Issue #475, opened 2013-05-07
Compilation errors (NSParameterAssert)


tonyarnold:
Hi guys,

I've just updated head from a checkout from a few days ago, and I can no longer get RAC to compile in my project, with the following errors from clang:

```
/Users/tonyarnold/Documents/Active Client Work/025.01.client-project/Pods/ReactiveCocoa/ReactiveCocoaFramework/ReactiveCocoa/NSObject+RACLifting.m:112:2: error: use of undeclared identifier '_cmd'; did you mean 'rcmd'?
        NSParameterAssert(block != nil);
        ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator6.1.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSException.h:139:38: note: expanded from macro 'NSParameterAssert'
#define NSParameterAssert(condition) NSAssert((condition), @"Invalid parameter not satisfying: %s", #condition)
                                     ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator6.1.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSException.h:100:65: note: expanded from macro 'NSAssert'
            [[NSAssertionHandler currentHandler] handleFailureInMethod:_cmd \
                                                                       ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator6.1.sdk/usr/include/unistd.h:709:6: note: 'rcmd' declared here
int      rcmd(char **, int, const char *, const char *, const char *, int *);
         ^
/Users/tonyarnold/Documents/Active Client Work/025.01.client-project/Pods/ReactiveCocoa/ReactiveCocoaFramework/ReactiveCocoa/NSObject+RACLifting.m:112:2: error: use of undeclared identifier 'self'
        NSParameterAssert(block != nil);
        ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator6.1.sdk/System/Library/Frameworks/Foundation.framework/Headers/NSException.h:139:38: note: expanded from macro 'NSParameterAssert'
#define NSParameterAssert(condition) NSAssert((condition), @"Invalid parameter not satisfying: %s", #condition)
                                     ^
[...]

PR #476, opened 2013-05-07
Change assert in RACLiftAndCallBlock to NSCParameterAssert

joshvera:
Fix for #475


sammoore:
I've been getting to know the RAC 2.x codebase, and out of curiosity, what triggered the change for all asserts to use the C versions? Wasn't this only necessary in certain situations and limits debugging information? Why is the reference to self avoided?

PR #1357, opened 2014-05-27
How to -retry: only matched error?

kkazuo:
How can I `-retry:` only some type of error?

With this patch, you can retry on time-out error only.

```
[signal retry:3 matchingError:^BOOL(NSError *error) {
    return [error.domain compare:NSURLErrorDomain] == 0 && error.code == NSURLErrorTimedOut;
}];
```

How do you think? Is there another recommended way without patch the core operators?

### DIFF
--- a/ReactiveCocoaFramework/ReactiveCocoa/RACSignal+Operations.h
+++ b/ReactiveCocoaFramework/ReactiveCocoa/RACSignal+Operations.h
@@ -764,6 +764,14 @@ typedef enum : NSUInteger {
 /// Resubscribes to the receiving signal if an error occurs.
 - (RACSignal *)retry;
 
+/// Resubscribes to the receiving signal if an error occurs, up until it has
+/// retried the given number of times.
+///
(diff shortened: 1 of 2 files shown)


krystianpaszek:
@jspahrsummers Shouldn't your solution look like this:

    __block NSUInteger retryCount = 0;

    RACSignal *errorHandlingSignal = [signal catch:^(NSError *error) {
        if ([error.domain compare:NSURLErrorDomain] == 0 && error.code == NSURLErrorTimedOut) {
            if (++retryCount < 3) {
                // Re-subscribe to `signal` again, which should be cold.
                return errorHandlingSignal;
            }
         }

        // We exceeded our retry count, or received an unexpected error, so let it
        // propagate.
        return [RACSignal error:error];
    }]

Notice the return after check for retryCount. Maybe I don't understand it properly, but for me it looks like after first error, your solution will return signal that won't catch any more errors.
If I'm right, then next question arises: how to implement it properly without having memory leaks?

PR #1470, opened 2014-08-18
[Legacy] Merge master in legacy

Coneko:
And we're back!

PR #1507, opened 2014-09-27
[Legacy] Change -map: to not create an extra RACLiveSubscriber

Coneko:
Depends on #1506 and #1470.

See #1377.


kastiglione:
@Coneko close?

kastiglione:
I asked @Coneko and he confirmed this can be closed.

PR #1579, opened 2014-11-21
Update to the latest LlamaKit

alanjrogers:
The latest LlamaKit .Error uses ErrorType protocol instead of straight NSError.

Should we have a precondition (or something?) that ensures the object is actually an NSError?


alanjrogers:
Closing due to staleness, and that it's been superseded by #1682

PR #1627, opened 2014-12-16
KVO via singleton proxy, redux

kastiglione:
Continuing @rspeyer's KVO proxy implementation from #1276 

Firstly, this pull request is to bring @rspeyer's changes up to date with `master`. This will mainly be updating the tests from Specta to Quick. Beyond that, perhaps `RACKVOProxy` and `RACKVOTrampoline` will be unified.

- [x] ~~Break out separate pull request to handle crash in (see https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1627#discussion_r22028875)~~
- [x] Refactor tests originally written using Specta's `AsyncBlock`
- [x] ~~Combine `RACKVOProxy` and `RACKVOTrampoline` (optional)~~

cc @epatey @modocache

### DIFF
--- a/ReactiveCocoaTests/RACKVOProxySpec.m
+++ b/ReactiveCocoaTests/RACKVOProxySpec.m
@@ -31,15 +31,15 @@ QuickSpecBegin(RACKVOProxySpec)
 
 qck_describe(@"racproxyobserve", ^{
 	__block TestObject *testObject;
-	
+
 	qck_beforeEach(^{
 		testObject = [[TestObject alloc] init];
 	});
-	
+
 	qck_afterEach(^{
 		testObject = nil;
 	});
-	
+
 	qck_describe(@"basic", ^{
 		qck_it(@"should handle multiple observations on the same value", ^{
 			__block int observedValue1 = 0;
@@ -50,20 +50,20 @@ qck_describe(@"racproxyobserve", ^{
 				subscribeNext:^(NSNumber *wrappedInt) {
 					observedValue1 = wrappedInt.intValue;
 				}];
-			
+
 			[[[RACObserve(testObject, testInt)
 				skip:1]
 				take:1]
 				subscribeNext:^(NSNumber *wrappedInt) {
 					observedValue2 = wrappedInt.intValue;
 				}];
-			
+
 			testObject.testInt = 2;
-			
+
 			expect(@(observedValue1)).toEventually(equal(@(testObject.testInt)));
 			expect(@(observedValue2)).toEventually(equal(@(testObject.testInt)));
 		});
-		
+
 		qck_it(@"can remove individual observation", ^{
 			__block int observedValue1 = 0;
 			__block int observedValue2 = 0;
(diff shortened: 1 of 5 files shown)


jspahrsummers:
@kastiglione Anything I can do to help?

kastiglione:
@jspahrsummers I just added a couple todos up top.

Although I spent some time investigating the first todo, it wasn't enough to understand where the dealloc race is coming from. That one is more investigative. The second todo might be more mechanical for you, depending on how obvious it is to you on how to convert the `AsyncBlock` tests to Quick.

kastiglione:
Also, the second todo may or may not be best done by using `waitUntil` from Quick/Nimble#72.

ReactiveCocoaTests/RACKVOProxySpec.m
```diff
@@ -0,0 +1,221 @@
...
+			dispatch_apply(numIterations, iterationQueue, ^(size_t index) {
+				RACDisposable *newDisposable = [RACObserve(testObject, testInt) subscribeCompleted:^{}];
+				[[disposable swapInDisposable:newDisposable] dispose];
+
+				dispatch_async(concurrentQueue, ^{
+					testObject.testInt = (int)index;
+				});
+			});
+
+			dispatch_barrier_sync(iterationQueue, ^{
```
kastiglione: Is this the crash? Should this be on `concurrentQueue`?

jspahrsummers: No, this isn't the crash. The scheduling is correct, because `concurrentQueue` is flushed (up top) before `testObject` is released by the spec.

kastiglione: `dispatch_apply` is documented to "waits for all iterations of the task block to complete before returning". At this point, the `iterationQueue` should be empty of any `dispatch_apply` blocks, but the `currentQueue` may still have blocks scheduled or in progress.

jspahrsummers: Ah, I misunderstood `dispatch_apply`. Nonetheless, this code is correct—this is the queue I intended to dispatch to, and it has the same semantics regardless of whether it's already empty or not.

`concurrentQueue` is properly drained at the top, before the test officially completes.

jspahrsummers:
I think this is ready to go. I hit all the to-do notes, and have seemingly fixed the [KVO crash](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1627#discussion_r22028875) (which I theorize to be from the automatic swizzling behavior).

joshaber:
:shipit:

kastiglione:
Since the test crash was fixed by using manual KVO notifying, have we concluded that automatic KVO is always subject to the crash, and should be avoided (at least with RAC)?

jspahrsummers:
I don't think we know enough about the implementation of automatic KVO to make a claim like that one way or another.

We only saw a problem in the tests because of highly concurrent observation and observer removal. I don't know if a real world application would actually hit it.

kastiglione:
Right. I guess "avoided" should be qualified by: concurrent changes and observer removal.

PR #1635, opened 2014-12-19
Replace MainScheduler with UIScheduler and QueueScheduler.mainQueueScheduler

### LINKED ISSUE #1619: Port -deliverOnMainThread to Swift
Since @jlawton's implementation in #1425 preserves ordering, it seems like this could be implemented as a custom `Scheduler`, despite the discussion in #1402 to the contrary.

That would avoid the need to implement it separately for each signal type, and might help in other ways too.

jspahrsummers:
This is _sort of_ a port of #1425/#1618 to Swift, except executed very differently.

I figured, if we're going to have immediate scheduling on the main thread as a feature, why not just make it the default? It seems cleaner and simpler to me than offering such a confusing choice to consumers.

Note that if you explicitly _want_ a scheduler hop, you can still use `delay(0)` to get that effect. Other than that, I don't _think_ there are any negative ramifications to this, but I'd love to hear thoughts from @jlawton @epatey @ReactiveCocoa/reactivecocoa.

Resolves #1619.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -228,7 +228,7 @@ public final class CocoaAction: NSObject {
 	private func startSignalOnMainThread<T: Equatable>(signal: ColdSignal<T>, next: T -> ()) {
 		let signalDisposable = signal
 			.skipRepeats(identity)
-			.deliverOn(MainScheduler())
+			.deliverOn(UIScheduler())
 			.start(next: next)
 
 		self.disposable.addDisposable(signalDisposable)
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -33,7 +33,7 @@ extension ImmediateScheduler {
 	}
 }
 
-extension MainScheduler {
+extension UIScheduler {
 	public func asRACScheduler() -> RACScheduler {
 		return RACScheduler.mainThreadScheduler()
 	}
--- a/ReactiveCocoa/Swift/Scheduler.swift
+++ b/ReactiveCocoa/Swift/Scheduler.swift
@@ -53,6 +53,7 @@ public final class ImmediateScheduler: Scheduler {
 /// A scheduler that performs all work on the main thread.
 public final class MainScheduler: DateScheduler {
 	private let innerScheduler = QueueScheduler(dispatch_get_main_queue())
+	private var queueLength: Int32 = 0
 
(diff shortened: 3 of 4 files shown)


jspahrsummers:
> I agree with @erikprice: users will notice eventually, probably after spending a lot of time debugging and figuring why something they expect to introduce a hop doesn't

Then is the problem setting appropriate expectations? Maybe there's a naming scheme we can figure out that clearly conveys the semantics.

> then to force the hop they'll deliver on a background scheduler and back to the main scheduler.

Do you have an example of when this might be important, and why `delay(0)` wouldn't be a better way to express the intention anyways?

Coneko:
> Do you have an example of when this might be important, and why `delay(0)` wouldn't be a better way to express the intention anyways?

I use it when I need to do something on the next runloop iteration.

`delay(0)` is better, I agree, but the only way one to be sure whether or not that would work is by reading the source. After all, if scheduling on the main thread scheduler from the main thread is reentrant, delaying by 0 might be as well.

It's also not as obvious to people who are more familiar with GCD queues than with `performSelector`.

I guess that might be the crux of my dislike of the reentrant scheduling. While schedulers are something more general than queues I've always reasoned about them as I would with queues, so any behaviour that doesn't match that bothers me.

jspahrsummers:
This discussion has only highlighted for me that the concept of one “main scheduler” is always going to confuse _someone_. Whether it introduces thread hops or not, the chosen behavior is going to be unintuitive to some subset of people.

So let's try to kill two birds with one stone. Here's a proposal:

 1. We add a `UIScheduler` that is _not_ a [`DateScheduler`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/677f713b4966f2dfc392fadfd65fa9243d30abc7/ReactiveCocoa/Swift/Scheduler.swift#L22). This has the reentrancy behavior we've been discussing, and hopefully avoids confusion by only allowing “as soon as possible” scheduling.
 1. We replace the `MainScheduler` type with a `QueueScheduler.mainQueueScheduler` property. This always follows GCD behavior, which is clear from its association with the `QueueScheduler` type.

Thoughts?

robrix:
:+1: I like that proposal in all of its specifics.

joshaber:
Yup, I like it :+1:

jlawton:
:+1:

erikprice:
:+1:

Coneko:
Sounds good.

jspahrsummers:
Updated. :zap:

joshaber:
:boom:

PR #1645, opened 2014-12-23
Syntactic sugar for Event<T>

### LINKED ISSUE #1643: Add syntactic sugar for Event constructors
All the boxing is really terrible, and will hopefully be able to go away in a future Swift release. Consequently, we should have custom constructor functions that don't require an explicit `Box`, like LlamaKit's `success()` and `failure()` functions.

jspahrsummers:
Resolves #1643.

### DIFF
--- a/ReactiveCocoa/Swift/ColdSignal.swift
+++ b/ReactiveCocoa/Swift/ColdSignal.swift
@@ -1282,6 +1282,28 @@ public enum Event<T> {
 		}
 	}
 
+	/// The value in this event, if it was a `Next`.
+	public var value: T? {
+		switch self {
+		case let .Next(value):
+			return value.unbox
+
+		default:
+			return nil
+		}
+	}
+
+	/// The error in this event, if it was an `Error`.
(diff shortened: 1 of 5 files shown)


ReactiveCocoa/Swift/ColdSignal.swift
```diff
@@ -1359,6 +1381,21 @@ extension Event: Printable {
 	}
 }
 
+/// Puts a `Next` event into the given sink.
+public func sendNext<T>(sink: SinkOf<Event<T>>, value: T) {
```
joshaber: Is there a reason not to make these funcs on `SinkOf`?

joshaber:
:confetti_ball:

ReactiveCocoa/Swift/ColdSignal.swift
jspahrsummers: They're restricted in the type of `SinkOf` accepted.

joshaber:
:sunglasses:

PR #1653, opened 2014-12-28
Add `Result` overload for `try`, allow default errors to be specified

jspahrsummers:
These are a couple of things that sometimes prevented me from using `try` and `tryMap` in Carthage, where they would otherwise make a lot of sense.

### DIFF
--- a/ReactiveCocoa/Swift/ColdSignal.swift
+++ b/ReactiveCocoa/Swift/ColdSignal.swift
@@ -581,6 +581,14 @@ extension ColdSignal {
 		return tryMap { (value, error) in f(value, error) ? value : nil }
 	}
 
+	/// Performs the given action upon each value in the receiver, bailing out
+	/// if the returned Result is an error.
+	public func try(f: T -> Result<()>) -> ColdSignal {
+		return tryMap { value -> Result<T> in
+			return f(value).map { _ in value }
+		}
+	}
+
 	/// Attempts to map each value in the receiver, bailing out with an error if
 	/// a given mapping is `nil`.
 	public func tryMap<U>(f: (T, NSErrorPointer) -> U?) -> ColdSignal<U> {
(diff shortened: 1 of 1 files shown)


joshaber:
If we're trying to keep the API small, I :doughnut: see the need for this.

jspahrsummers:
Is your concern the additional `try` operator, or the `defaultError` parameters as well?

joshaber:
Bolth.

jspahrsummers:
I guess I could go either way on `defaultError`. It seemed useful, but maybe that's just a smelly pattern that we've been using in Carthage.

`try()` over `Result` seems pretty important, if we're going to have a `try` operator at all. It parallels `tryMap` and encourages better error handling patterns than `NSError`.

joshaber:
The idea of a default error seems to encourage sloppiness.

`try` is alright, I suppose.

jspahrsummers:
@alanjrogers I think we're talking about different things. @joshaber was suggesting removing the ability to customize the error at all, so it would always be `RACError.Empty`.

alanjrogers:
o-rly-owls-13509350-282-282

joshaber:
Well I am cornfused. Why do we do that? Why don't we use the error we got: https://github.com/ReactiveCocoa/ReactiveCocoa/blob/9cfef279ff059180691381c3bcca45df97efb41a/ReactiveCocoa/Swift/ColdSignal.swift#L589

jspahrsummers:
We do, but there may not always be one.

joshaber:
I'd certainly put that on the caller, not on RAC then.

jspahrsummers:
Sooo… we _want_ the `defaultError` parameter now? I'm confused.

joshaber:
Yeah sorry, here's my full line of thought:

We should encourage the fallible `func` being wrapped in the `try...` to provide a meaningful error. Providing a default error encourages laziness and increases the size of the API.

joshaber:
I assume those are unrelated to this PR?

PR #1654, opened 2014-12-28
Add ColdSignal.try constructors for lazy operations

jspahrsummers:
Same motivation as #1653, but a logically distinct change that probably deserves some discussion on its own.


jspahrsummers:
Removed the `defaultError` stuff, per the discussion in #1653.

joshaber:
:+1:

Issue #1650, opened 2014-12-27
Thoughts on "The Great Swiftening" after a few hours playing with it


NachoSoto:
I recently played around with [RAC 3.0](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1382) and wanted to share some thoughts :) please keep in mind that this is only after a few hours of use, so most of what I bring up is probably due to me still being too used to `2.0`, and possibly to some lack of understanding on some of the underlying concepts here, so bear with me! :D

First some background: I used RAC 2.0 extensively for Elevate, in Objective-C, and for the past few months in Swift at Khan Academy. It’s obviously not great to have to do dynamic casts from `AnyObject!` all the time, but it’s better than nothing!
So far I’m really excited with what I’ve seen, but hopefully my thoughts can help make the transition to this new version easier for me and for other people :)

I’ve been trying to follow the discussions about the hot/cold separation for a while, and overall I think it’s a huge step forward.
These are some things I didn’t understand very well as I was trying to figure out the internals of each of the two types. I imagine most of these will be explained in the README though.

These are the docs on both:
```swift
/// A stream that will begin generating Events when a sink is attached, possibly
/// performing some side effects in the process. Events are pushed to the sink
/// as they are generated.
///
/// A corollary to this is that different sinks may see a different timing of
/// Events, or even a different version of events altogether.
///
/// Cold signals, once started, do not need to be retained in order to receive
/// events. See the documentation for startWithSink() or start() for more
/// information.
public struct ColdSignal<T> 
```
```swift
/// A push-driven stream that sends the same values to all observers.
///
/// Generally, hot signals do not need to be retained in order to observe the
/// values they send, but there are some exceptions. See the documentation for
/// observe() and HotSignal.weak() for more information.
public final class HotSignal<T>
```

The more I read those, the more confused I am :P Possibly the most important word in those two is `class` VS `struct`. I get that fundamental difference of `value type` ==> `different work per subscriber` VS `reference type` ==> `same to all subscribers`, but I can’t help to get confused by the rest of the docs and some other differences in the two types.

Like I said, it’s possible that I’m still too used to these two different concepts being only one (`RACSignal`), but I was surprised to see that not only they’re different conceptually, but also their APIs and functionality:
- Method to observe values: `start` VS `observe`.
- How values are “streamed”: `Event<T>` VS `<T>`.
- Corollary of the previous point: completion and possible errors VS only values.
[...]

PR #1655, opened 2014-12-28
Replace ColdSignal and HotSignal with SignalProducer and Signal

jspahrsummers:
After the long discussion in #1650, I invited @NachoSoto, @andymatuschak, and @kastiglione to talk more in-depth about the Swift APIs, and specifically where they need improvement.

We had a really, really great brainstorming session, and this is the result. The good news is that we all feel pretty good about this direction, and the bad news is that everything is getting rewritten again. :trollface:

The important bits are documented on the types themselves, but here’s the handy Quick Start guide of the proposed changes:

1. `HotSignal<T>` will be replaced by `Signal<T>`. Besides the name change, these signals can now send `Error` and `Completed` events. This avoids a lot of the lifecycle weirdness that has come up with a value-only `HotSignal`.
1. `ColdSignal<T>` will be replaced by `SignalProducer<T>`. As the name implies, a signal producer creates `Signal`s (via the `start` method). You can kind of think of the relationship as something like:
 * `SignalProducer : Signal :: RACSignal : subscription`
 * `SignalProducer : Signal :: class : instance`
1. Most of the basic operators are written in terms of `Signal`, but can be lifted to apply to `SignalProducer` instead.
1. Operators that benefit from “cold” semantics, like `retry` and `repeat`, are written in terms of `SignalProducer`.

The advantages here are:

* The relationship between “hot” and “cold” signals is now much more explicit and well-defined. A “cold” signal (a producer) creates “hot” signals, each of which represents a “subscription” (in RAC 2.x terms).
* Many operators only have to be written once, in terms of `Signal`, and can be lifted to the `SignalProducer` level.
* The semantics and/or difficulties of converting between “hot” and “cold” are more obvious, I think. You don’t need to swap between them on the fly, because they mean different things.
* Signal lifetime is now consistent, and completely defined by the event stream. Signal producers don’t do anything special—they really are just value types, and don’t encode any information about lifetime.

I’ll also make some notes in the changes below, to explain more specific things. Please leave feedback if you have it! :bow:


joshaber:
:sparkling_heart: This looks fantastic :sparkling_heart: 

A single signal type is awesome. I don't love the name of `SignalProducer` but I don't have any better ideas that haven't already been discussed.

--

Do we want to go ahead and merge this?

robb:
LGTM

jspahrsummers:
> Do we want to go ahead and merge this?

I'll go ahead and build the implementations here, unless there's a particular reason not to do that.

joshaber:
:+1:

kastiglione:
> I don't love the name of `SignalProducer` but I don't have any better ideas that haven't already been discussed.

It's a shame english cannot handle our scale.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -0,0 +1,474 @@
...
+	}
+}
+
+/// Combines the latest value of the receiver with the latest value from
+/// the given producer.
+///
+/// Signals started by the returned producer will not send a value until both
+/// inputs have sent at least one value each.
+public func combineLatestWith<T, U>(otherSignalProducer: SignalProducer<U>)(producer: SignalProducer<T>) -> SignalProducer<(T, U)> {
+	return producer.lift(combineLatestWith)(otherSignalProducer)
```
jspahrsummers: How freaking awesome is this?

jspahrsummers:
Trying to implement everything here was a mistake, since we're basically talking about the whole framework. So, instead, I've annotated outstanding TODOs and would like to merge this as-is (and create smaller PRs into `swift-development` after that).

:raised_hands:

jspahrsummers:
Gonna #yolo this so I can create dependencies that don't have crazy diffs. :trollface:

PR #1678, opened 2015-01-16
Backport RACQueueScheduler lldb/swift fixes

alanjrogers:
Right now, if you use ReactiveCocoa 2.4.4 in an app that mixes Swift and Objective-C, and you want to use LLDB on a swift class… you’re gonna have trouble.

We fixed this on the `swift-development` branch in #1573, so this is just a cherry-pick of the same fixes.


jspahrsummers:
:sunglasses:

PR #1681, opened 2015-01-16
++LlamaKit

eliperkins:
Now With Less Ducktyping™! Removes a bunch of parens, adds a few boxes, and parameterizes errors.

Probably closes #1579 FWIW too.


jspahrsummers:
Thanks! I had to fix the submodule URL, though, so I've opened https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1682 and added this commit to it.

And I apologize in advance for how all this code will be rewritten in #1655. :wink:

PR #1683, opened 2015-01-18
Parameterize signals by the errors they send

jspahrsummers:
Now that [`Result` is parameterized by its error type](https://github.com/LlamaKit/LlamaKit/releases/tag/v0.5.0), we should have some way of capturing that expressive power, and moving away from the use of `NSError` directly. This seems like the most natural step.

I was somewhat worried about the ability to compose signals with different error types, but I think it will actually turn out alright _as long as_ RAC avoids defining its own errors (because those will not automatically be included in whatever error type consumers end up using).

This also has some neat side benefits when paired with the empty `NoError` type, because we can now express that specific signals will _never_ send errors, or that certain signal operators (e.g., `takeUntil` or property binding) require that errors be handled by the consumer first.

/cc @andymatuschak

**To do:**

 - [x] Make `asRACSignal()` usable for producers and signals that don't already send `NSError`
 - [x] Remove `NSError`-based `try`?
 - [x] Document the error type parameters


alanjrogers:
👏

kastiglione:
> certain signal operators (e.g., `takeUntil` or property binding) require that errors be handled by the consumer first.

Fancy

andymatuschak:
I like that this really helps us capture some of the arrow-level semantics of signals. I am concerned that writing type signatures may now feel more like a chore. If only it were possible to create partially applied type aliases...

Not sure we'll be able to know until building something with it.

jspahrsummers:
> I am concerned that writing type signatures may now feel more like a chore.

Absolutely. I have the same concern.

The alternatives suck, though:

1. Use `NSError` everywhere. This is quite painful in Swift, when `enum`s are so great for this purpose.
1. Use a custom RAC error type that subsumes `NSError` and offers some customizability. To consumers, though, this isn't really any better than `NSError`, because they can't define their own error _type_.

jspahrsummers:
:interrobang:

joshaber:
:metal: `NoError` :metal:

corristo:
:cool:

PR #1688, opened 2015-01-20
Implemented a handful of `Signal` operators

andymatuschak:
Six down!

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -415,22 +415,37 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>) 
 /// is the current value. `initial` is supplied as the first member when `signal`
 /// sends its first value.
 public func combinePrevious<T, E>(initial: T)(signal: Signal<T, E>) -> Signal<(T, T), E> {
+	return signal |> reduce((initial, initial)) { previousCombinedValues, newValue in
+		return (previousCombinedValues.1, newValue)
+	}
+}
+
+/// Like `scan`, but sends only the final value and then immediately completes.
+public func reduce<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>) -> Signal<U, E> {
+	return signal
+		|> scan(initial, combine)
+		|> takeLast(1)
+}
+
+/// Aggregates `signal`'s values into a single combined value. When `signal` emits
+/// its first value, `combine` is invoked with `initial` as the first argument and
+/// that emitted value as the second argument. The result is emitted from the
+/// signal returned from `reduce`. That result is then passed to `combine` as the
+/// first argument when the next value is emitted, and so on.
+public func scan<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>) -> Signal<U, E> {
 	return Signal { observer in
-		let previousValueState = Atomic<T?>(nil)
+		let accumulatorState = Atomic(initial)
 		return signal.observe(next: { value in
-			previousValueState.modify { previousValue in
-				if let previousValue = previousValue {
-					sendNext(observer, (previousValue, value))
-				} else {
-					sendNext(observer, (initial, value))
-				}
-				return value
+			accumulatorState.modify { accumulator in
+				let newAccumulatorValue = combine(accumulator, value)
+				sendNext(observer, newAccumulatorValue)
+				return newAccumulatorValue
 			}
 			return
-		}, error: { error in
-			sendError(observer, error)
-		}, completed: {
-			sendCompleted(observer)
+			}, error: { error in
+				sendError(observer, error)
+			}, completed: {
+				sendCompleted(observer)
 		})
 	}
 }
@@ -438,8 +453,6 @@ public func combinePrevious<T, E>(initial: T)(signal: Signal<T, E>) -> Signal<(T
 /*
 TODO
 
-public func reduce<T, U>(initial: U, combine: (U, T) -> U)(signal: Signal<T>) -> Signal<U>
-public func scan<T, U>(initial: U, combine: (U, T) -> U)(signal: Signal<T>) -> Signal<U>
 public func skipRepeats<T: Equatable>(signal: Signal<T>) -> Signal<T>
 public func skipRepeats<T>(isRepeat: (T, T) -> Bool)(signal: Signal<T>) -> Signal<T>
 public func skipWhile<T>(predicate: T -> Bool)(signal: Signal<T>) -> Signal<T>
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -415,7 +415,7 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>) 
 /// is the current value. `initial` is supplied as the first member when `signal`
 /// sends its first value.
 public func combinePrevious<T, E>(initial: T)(signal: Signal<T, E>) -> Signal<(T, T), E> {
-	return signal |> reduce((initial, initial)) { previousCombinedValues, newValue in
+	return signal |> scan((initial, initial)) { previousCombinedValues, newValue in
 		return (previousCombinedValues.1, newValue)
 	}
 }
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -434,18 +434,12 @@ public func reduce<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, 
 /// first argument when the next value is emitted, and so on.
 public func scan<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>) -> Signal<U, E> {
 	return Signal { observer in
-		let accumulatorState = Atomic(initial)
-		return signal.observe(next: { value in
-			accumulatorState.modify { accumulator in
-				let newAccumulatorValue = combine(accumulator, value)
-				sendNext(observer, newAccumulatorValue)
-				return newAccumulatorValue
-			}
-			return
-		}, error: { error in
-			sendError(observer, error)
-		}, completed: {
-			sendCompleted(observer)
+		var accumulator = initial
+		return signal.observe(Signal.Observer { event in
+			observer.put(event.map { value in
+				accumulator = combine(accumulator, value)
+				return accumulator
+			})
 		})
 	}
 }
@@ -477,16 +471,12 @@ public func skipRepeats<T, E>(isRepeat: (T, T) -> Bool)(signal: Signal<T, E>) ->
 /// at which point the returned signal behaves exactly like `signal`.
 public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T, E> {
 	return Signal { observer in
-		let isSkipping = Atomic(true)
+		var shouldSkip = true
 		return signal.observe(next: { value in
-			isSkipping.modify { wasSkipping in
-				var shouldSkip = wasSkipping && predicate(value)
-				if !shouldSkip {
-					sendNext(observer, value)
-				}
-				return shouldSkip
+			shouldSkip = shouldSkip && predicate(value)
+			if !shouldSkip {
+				sendNext(observer, value)
 			}
-			return
 		}, error: { error in
 			sendError(observer, error)
 		}, completed: {
@@ -499,27 +489,19 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
 /// on the returned signal.
 public func takeLast<T,E>(count: Int)(signal: Signal<T,E>) -> Signal<T,E> {
 	return Signal { observer in
-		let buffer = Atomic<[T]>({
-			var bufferArray = [T]()
-			bufferArray.reserveCapacity(count)
-			return bufferArray
-		}())
-		
+		var buffer = [T]()
+		buffer.reserveCapacity(count)
+
 		return signal.observe(next: { value in
-			buffer.modify { oldBuffer in
-				var newBuffer = oldBuffer
-				newBuffer.append(value)
-				
-				while newBuffer.count > count {
-					newBuffer.removeAtIndex(0)
-				}
-				return newBuffer
+			buffer.append(value)
+
+			while buffer.count > count {
+				buffer.removeAtIndex(0)
 			}
-			return
 		}, error: { error in
 			sendError(observer, error)
 		}, completed: {
-			for bufferedValue in buffer.value {
+			for bufferedValue in buffer {
 				sendNext(observer, bufferedValue)
 			}
 			
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -421,8 +421,6 @@ public func combinePrevious<T, E>(initial: T)(signal: Signal<T, E>) -> Signal<(T
 }
 
 /// Like `scan`, but sends only the final value and then immediately completes.
-/// If `signal` completes before you observe the resulting signal, the resulting
-/// signal may send no values before completing.
 public func reduce<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>) -> Signal<U, E> {
 	// We need to handle the special case in which `signal` sends no values.
 	// We'll do that by sending `initial` on the output signal (before taking
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -442,10 +442,88 @@ public func scan<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>
 				return newAccumulatorValue
 			}
 			return
-			}, error: { error in
-				sendError(observer, error)
-			}, completed: {
-				sendCompleted(observer)
+		}, error: { error in
+			sendError(observer, error)
+		}, completed: {
+			sendCompleted(observer)
+		})
+	}
+}
+
+/// Forwards only those values from `signal` which are not duplicates of the
+/// immedately preceding value. The first value is always forwarded.
+public func skipRepeats<T: Equatable, E>(signal: Signal<T, E>) -> Signal<T, E> {
+	return signal |> skipRepeats { $0 == $1 }
+}
+
+/// Forwards only those values from `signal` which do not pass `isRepeat` with
+/// respect to the previous value. The first value is always forwarded.
+public func skipRepeats<T, E>(isRepeat: (T, T) -> Bool)(signal: Signal<T, E>) -> Signal<T, E> {
+	return signal
+		|> map { Optional($0) }
+		|> combinePrevious(nil)
+		|> filter {
+			switch $0 {
+			case let (.Some(a), .Some(b)) where isRepeat(a, b):
+				return false
+			default:
+				return true
+			}
+		}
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,127 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+	return Signal { observer in
+		let buffer = Atomic<[T]>({
+			var bufferArray = [T]()
+			bufferArray.reserveCapacity(count)
+			return bufferArray
+		}())
+		
+		return signal.observe(next: { value in
+			buffer.modify { oldBuffer in
+				var newBuffer = oldBuffer
```
kastiglione: Why the copy?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,127 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+}
+
+/// Aggregates `signal`'s values into a single combined value. When `signal` emits
+/// its first value, `combine` is invoked with `initial` as the first argument and
+/// that emitted value as the second argument. The result is emitted from the
+/// signal returned from `reduce`. That result is then passed to `combine` as the
+/// first argument when the next value is emitted, and so on.
+public func scan<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>) -> Signal<U, E> {
+	return Signal { observer in
+		let accumulatorState = Atomic(initial)
```
kastiglione: I believe the `Atomic` isn't necessary. Observers are called serially.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,127 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+			}
+		}
+		|> map { $0.1! }
+}
+
+/// Does not forward any values from `signal` until `predicate` returns false,
+/// at which point the returned signal behaves exactly like `signal`.
+public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T, E> {
+	return Signal { observer in
+		let isSkipping = Atomic(true)
```
kastiglione: This `Atomic` too.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,127 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+			sendCompleted(observer)
+		})
+	}
+}
+
+/// Waits until `signal` completes and then forwards the final `count` values
+/// on the returned signal.
+public func takeLast<T,E>(count: Int)(signal: Signal<T,E>) -> Signal<T,E> {
+	return Signal { observer in
+		let buffer = Atomic<[T]>({
```
kastiglione: And here.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: Yeah. It will only be one real copy because Swift is smart about avoiding spurious copies.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,127 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
 	}
 }
 
+/// Forwards events from `signal` with history: values of the returned signal
+/// are a tuple whose first member is the previous value and whose second member
+/// is the current value. `initial` is supplied as the first member when `signal`
+/// sends its first value.
+public func combinePrevious<T, E>(initial: T)(signal: Signal<T, E>) -> Signal<(T, T), E> {
+	return signal |> reduce((initial, initial)) { previousCombinedValues, newValue in
```
jspahrsummers: I think this should be `scan`.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: I was making this way, way harder on myself than it needed to be.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,127 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+		return (previousCombinedValues.1, newValue)
+	}
+}
+
+/// Like `scan`, but sends only the final value and then immediately completes.
+public func reduce<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>) -> Signal<U, E> {
+	return signal
+		|> scan(initial, combine)
```
jspahrsummers: There's an edge case here where the input signal is empty, so `scan` never sends a value. `reduce` should still send exactly one value (the initial value) in that case.

You could pehaps accomplish this by prepending a signal of `initial` before the `takeLast(1)`.

jspahrsummers:
Thanks for doing this! :sparkles:

If you feel like you won't have time to make fixes, I'd be happy to do so—just let me know.

andymatuschak:
Nope, I’m on it.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: Yeah. Worried about others making this mistake.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: Ah, right. Need SignalProducer.concat for that, which is still TODO. Will try to do this evening.

andymatuschak:
OK, please rereview.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -425,17 +425,127 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+					}
+				}
+				
+				if shouldSubscribe {
+					subscribeToSignalProducer!(innerSignalProducer)
+				}
+			}, error: { error in
+				sendError(observer, error)
+			}, completed: {
+				sendCompleted(observer)
```
kastiglione: Should this assign `selfCompleted` and call `completeIfAllowed()`?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,126 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+/// sends its first value.
+public func combinePrevious<T, E>(initial: T)(signal: Signal<T, E>) -> Signal<(T, T), E> {
+	return signal |> scan((initial, initial)) { previousCombinedValues, newValue in
+		return (previousCombinedValues.1, newValue)
+	}
+}
+
+/// Like `scan`, but sends only the final value and then immediately completes.
+/// If `signal` completes before you observe the resulting signal, the resulting
+/// signal may send no values before completing.
```
jspahrsummers: The resulting signal wouldn't send anything in that case, would it? This is a more general problem with `Signal`s, because you can miss events. (And so this documentation would apply to every operator.)

This is why [I don't want to support](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1689#issuecomment-70707397) operators like `concat` over hot signals.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,126 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+public func reduce<T, U, E>(initial: U, combine: (U, T) -> U)(signal: Signal<T, E>) -> Signal<U, E> {
+	// We need to handle the special case in which `signal` sends no values.
+	// We'll do that by sending `initial` on the resulting signal (before taking
+	// the last value).
+	let scannedSignal = signal |> scan(initial, combine)
+	let (liftingProducerForScannedSignal, liftingProducerSink: Signal<U, E>.Observer) = SignalProducer.buffer(0)
+	let liftingDisposable = scannedSignal.observe(liftingProducerSink)
```
jspahrsummers: This is a completely valid way to accomplish this, but it seems a little overcomplicated.

Could we use `Signal.pipe()` for the same purpose? That involves similar semantics and gives us similar levels of control, but avoids any need to lift.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,126 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+		})
+	}
+}
+
+/// Waits until `signal` completes and then forwards the final `count` values
+/// on the returned signal.
+public func takeLast<T,E>(count: Int)(signal: Signal<T,E>) -> Signal<T,E> {
+	return Signal { observer in
+		var buffer = [T]()
+		buffer.reserveCapacity(count)
```
jspahrsummers: TIL. :smile:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -410,16 +410,126 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>)
...
+
+		return signal.observe(next: { value in
+			buffer.append(value)
+
+			while buffer.count > count {
```
jspahrsummers: Doesn't this mean we've exceeded the reserved capacity? Not that that's an error—but it seems to defeat the potential optimization.

Should we remove and _then_ append instead?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -425,17 +425,127 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+		}
+	}
+}
+
+private struct ConcatState<T, E: ErrorType> {
+	/// Whether the signal-of-signals has completed yet.
+	var selfCompleted = false
+	
+	/// The signals waiting to be started.
+	var queuedSignalProducers = Array<SignalProducer<T, E>>()
```
jspahrsummers: We've generally favored separating the type annotation from the constructor for cases like this:

```swift
var queuedSignalProducers: [SignalProducer<T, E>] = []
```

TBH, I don't feel strongly, but this is the popular opinion. :space_invader:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -425,17 +425,127 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+	
+	/// Contains disposables for the currently active subscriptions.
+	var activeDisposables = Array<SerialDisposable>()
+}
+
+/// The Z combinator, which we use to make a recursive closure that we can
+/// nil out to avoid a retain cycle.
```
jspahrsummers: Not knowing much about combinatorial logic, how does this differ from the imperative approach? It seems like the same capturing semantics are necessary above. Is the main difference that the closure doesn't need to capture itself?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -425,17 +425,127 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+				// completion, preventing it from deallocating early.
+				subscribeToSignalProducer = nil
+			}
+		}
+		
+		subscribeToSignalProducer = Z { recur, signalProducer in
+			let serialDisposable = SerialDisposable()
+			let serialDisposableCompositeHandle = disposable.addDisposable(serialDisposable)
+			state.modify { oldState in
+				var newState = oldState
```
jspahrsummers: Can you annotate the argument as `(var oldState)` instead?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -425,17 +425,127 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+
+private struct ConcatState<T, E: ErrorType> {
+	/// Whether the signal-of-signals has completed yet.
+	var selfCompleted = false
+	
+	/// The signals waiting to be started.
+	var queuedSignalProducers = Array<SignalProducer<T, E>>()
+	
+	/// Contains disposables for the currently active subscriptions.
+	var activeDisposables = Array<SerialDisposable>()
```
jspahrsummers: Since concatenation is strictly serial, I think we'd only need a `latestCompleted` flag here. Combined with the properties above, that should be enough to determine whether the full signal-of-signals has been processed.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -425,17 +425,127 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+				}
+				
+				if let nextSignalProducer = nextSignalProducer {
+					recur(nextSignalProducer)
+				}
+			})
+		}
+		
+		producer.startWithSignal { signal, signalDisposable in
+			let outerSignalDisposable = signal.observe(next: { innerSignalProducer in
```
jspahrsummers: I apologize for not making this clear in the documentation, but you can use `signalDisposable` instead of this returned disposable for all intents and purposes. It's even preferable, because then the `Signal` can be disposed _before_ observation begins, if needed.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -425,17 +425,127 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+				var shouldSubscribe: Bool = true
```
jspahrsummers: There's an overload of `modify` that allows you to return a custom value as well. We could use that instead, with sufficient type annotation to help the compiler. :wink:

jspahrsummers:
Sorry for all the notes! I hope they're helpful, but lemme know if I can clarify anything.

ReactiveCocoa/Swift/Signal.swift
ka-interview: Yep, that's right. It would just complete immediately. 

Do you want me to remove the comment, since it applies to all these operators? Or do you want to support `reduce` on only `SignalProducer`s?

andymatuschak: Yep, that's right. It would just complete immediately. 

Do you want me to remove the comment, since it applies to all these operators? Or do you want to support `reduce` on only `SignalProducer`s?

jspahrsummers: > It would just complete immediately.

If the original signal is missed, every operator thereafter would result in an infinite signal (because the terminating event from the original would not be replayed).

To rephrase: if any signal operator misses _all_ the events from an input signal, the operator will probably not be able to do anything (and the result will therefore be infinite).

> Do you want me to remove the comment, since it applies to all these operators?

Yep.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: Oh, nice. Guess I didn't need to write `concat`. :)

ReactiveCocoa/Swift/Signal.swift
andymatuschak: > To rephrase: if any signal operator misses all the events from an input signal, the operator will probably not be able to do anything (and the result will therefore be infinite).

Whoa. That… sounds really hard to program against. Does it make sense to special-case observing a completed signal (i.e. send a completed event in that case)?

Will remove the comment.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: Yeah, thanks.

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: I find that a little less clear as a reader, but happy to make the change.

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: OK

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: I think that's going to require more effort from the reader than the present code.

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: Oh, jeez, yeah, what am I doing.

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: OK.

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: Yeah, this was copy pasta from `flatten:`.

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: I basically ported `-[RACSignal flatten:]`. That implementation captured a `weak` block to recur without creating a permanent cycle. Swift doesn't support weak closures, so this is my workaround.

jspahrsummers:
I want to pull [this conversation](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1688#discussion_r23490026) out to the top level, so it doesn't get hidden in a collapsed bit:

> > To rephrase: if any signal operator misses all the events from an input signal, the operator will probably not be able to do anything (and the result will therefore be infinite).
>
> Whoa. That… sounds really hard to program against. Does it make sense to special-case observing a completed signal (i.e. send a completed event in that case)?

I'm ambivalent, but let me play devil's advocate so we can fully tease out what the right choice is.

Reasons I think it could be a _bad_ idea to do that include:

 1. It would complicate the `Signal` contract. As the documentation describes, it's meant to be a thing that distributes the _same events_ to _all observers_ at _the same time_. If we special-cased termination, we'd be adding a pretty large asterisk to that description (i.e., complexity).
 1. What should happen with signals that errored out, instead of completing successfully? Should the error be replayed in the same way? It's hard for me to think of a cogent argument why or why not, but it certainly _feels_ weird.
 1. In both cases (error and completion), would you always want binary `Signal` operators to terminate because one of the inputs has already terminated? For example, should [`takeUntil`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/8c67470a34dde0b10e296a95a974c48b63990063/ReactiveCocoa/Swift/Signal.swift#L396) always terminate immediately if `trigger` has, or should it simply pass through `signal` in its entirety?

We came up with the [`SignalProducer.startWithSignal`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/8c67470a34dde0b10e296a95a974c48b63990063/ReactiveCocoa/Swift/SignalProducer.swift#L188) interface specifically to avoid any window where you might miss events. Why does termination warrant special treatment?

ReactiveCocoa/Swift/SignalProducer.swift
kastiglione: > Swift doesn't support weak closures

I didn't know this, good to know.

andymatuschak:
@jspahrsummers OK, you've persuaded me. I think we just need to really clearly communicate that observing `Signal`s—and in particular using signal operators—outside of a `SignalProducer.startWithSignal()` closure is dangerous and requires careful thought.

Just fixed up the last TODO. Let me know what else I can do.

jspahrsummers:
Looks great! The documentation is especially top-notch. :sparkles:

:shipit:

andymatuschak:
Thanks for the detailed review and for merging!

PR #1689, opened 2015-01-20
Add takeUntilReplacement to Signal

kastiglione:
Adds `takeUntilReplacement`, ~~and by necessity, `concat`~~. These are direct conversions of the objc versions.

~~Note that `concat` wasn’t listed in the `Signal`’s TODO. Was this just an omission, or is there another way that `takeUntilReplacement` should be implemented?~~

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -410,23 +410,6 @@ public func takeUntil<T, E>(trigger: Signal<(), NoError>)(signal: Signal<T, E>) 
 	}
 }
 
-/// Appends the events of `nextSignal` to the events of `signal`.
-public func concat<T, E>(nextSignal: Signal<T, E>)(signal: Signal<T, E>) -> Signal<T, E> {
-	return Signal { observer in
-		let serialDisposable = SerialDisposable()
-
-		serialDisposable.innerDisposable = signal.observe(next: { value in
-			sendNext(observer, value)
-		}, error: { error in
-			sendError(observer, error)
-		}, completed: {
-			serialDisposable.innerDisposable = nextSignal.observe(observer)
-		})
-
-		return serialDisposable
-	}
-}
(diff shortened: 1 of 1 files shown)


jspahrsummers:
I don't think `concat()` makes sense over `Signal`—it's too easy to use it incorrectly. More generally, I'd say we should _not_ offer any `Signal` operator that _waits to observe_ another signal, because missing events on the latter signal could cause unpredictable behavior.

In this particular case, the `concat()` could be avoided by simply using an `observe()` that ignores completion.

kastiglione:
> `concat()` could be avoided by simply using an `observe()` that ignores completion.

:+1:  That's how I had it at first, but I didn't like. In my infinite wisdom I though I'll just add `concat()` even though it's not in the TODO, because it's safe how _this_ is using it :trollface:

> _not_ offer any `Signal` operator that _waits to observe_ another signal, because missing events on the latter signal could cause unpredictable behavior.

If feel it might be ok to wait to observe for some kinds of signals, thinking specifically of `Signal`s that don't really have a start or an end. Some examples: push notifications, app life cycle, network reachability changes, OS notifications.

kastiglione:
Tests?

jspahrsummers:
> If feel it might be ok to wait to observe for some kinds of signals, thinking specifically of `Signal`s that don't really have a start or an end. Some examples: push notifications, app life cycle, network reachability changes, OS notifications.

It's not that there's _always_ a problem with doing this—I just feel like it'll be too easy for consumers to mess it up.

Worst case, you can create a `SignalProducer` that just attaches the observer to whatever plain `Signal` you have.

> Tests?

If you're offering. :smile: 

That can come in a separate PR, though. I'll merge this as-is for now. :shipit:

kastiglione:
> Worst case, you can create a SignalProducer that just attaches the observer to whatever plain Signal you have.

:+1:

PR #1382, opened 2014-06-19
The Great Swiftening (a.k.a. the new 3.0)

jspahrsummers:
Resolves #1365 and #1366. Per my comments in #1369, this is intended to be a long-running branch that will eventually become RAC 3.0, supplanting the [current `3.0-development` work](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/966).

To get us started, I basically just subtree merged [jspahrsummers/RxSwift](https://github.com/jspahrsummers/RxSwift) into here, then bridged several RAC classes back and forth.

PR #1696, opened 2015-01-24
New method for generating sequence iteratively

febeling:
I want to suggest to add a method

```objc
+ (instancetype)generateWithStart:(id)state iterate:(id (^)(id value))iterate;
```

to RACSequence. This method is modelled on IObservable.Generate, and also inspired by list comprehension in other languages.

C#.NET Rx has the method:

```
public static IObservable<TResult> Generate<TState, TResult>(
  TState initialState, 
  Func<TState, bool> condition, 
  Func<TState, TState> iterate, 
  Func<TState, TResult> resultSelector)
```

This offers the advantage that no constructor needs to be called. Here
is a C# usage example:

```
public static IObservable<int> Range(int start, int count)
{
  var max = start + count;
  return Observable.Generate(
    start, 
    value => value < max, 
    value => value + 1, 
    value => value);
}
```
(This example is from the [Rx online Book](http://www.introtorx.com/content/v1.0.10621.0/04_CreatingObservableSequences.html#ObservableGenerate) by Lee Campbell.)

Here the C# docs: [Observable.Generate](https://msdn.microsoft.com/en-us/library/system.reactive.linq.observable.generate.aspx)

It also feels intuitive when you know sequence generation based on function application in other functional languages. Arbitrary example from Clojure:

```clojure
user=> (take 5 (iterate inc 5))
(5 6 7 8 9)
```

See the docs of [clojure.core/iterate](https://clojuredocs.org/clojure.core/iterate) for details and examples.

The proposed change would make generating of a sequence based on a functional block more succinct. Consider this code from the test case:

```objc
RACSequence *sequence = [RACSequence generateWithStart:@1
    iterate: ^ (NSNumber *number) {
        return @(number.integerValue * 2);
    }];
```

Without this change, it's quite a bit more involved to create a sequence (and I hope I'm not overlooking something much simpler :). Complex recursive block usage and memory class headaches are easily all part of it.

This is the slimmed down version of the change, which I originally did more along the lines of the C# Rx interface. That includes a predicate for termination and a map transformation from sequence state to result values. The design guidelines of this project suggest to prefer operators for stream over extensions, though, and as `-map:`, `-take:` and `takeUntilBlock:` make these modifications straightforward I tend to leave them out, too. (Have a look at the febeling/ReactiveCocoa@sequence-generate branch for comparison in case anyone is curious.)

I wished for this method a number of times in real projects. Does it make sense?

### DIFF
--- a/ReactiveCocoa/RACSequence.h
+++ b/ReactiveCocoa/RACSequence.h
@@ -144,6 +144,52 @@
 /// tail. `headBlock` must not be nil.
 + (RACSequence *)sequenceWithHeadBlock:(id (^)(void))headBlock tailBlock:(RACSequence *(^)(void))tailBlock;
 
+/// Generate sequence from start value and iterate block.
+///
+/// This method creates a sequence from an initial state and an
+/// interate block, which calculate the next state value. The
+/// predicate block test the state value and terminates the sequence
+/// if true. The state value is transformed to a result value if a
+/// transform block is given.
+///
+/// The method corresponds to Rx `IObservable.Generate`.
+///
+/// start     – Initial state value.
+/// until     – Invoke on state to test if the sequence ends. Optional.
+/// iterate   – Invoked on state to calculate the next state.
+/// transform – Invoked to map state to result value. Optional.
+///
+/// Returns a lazy sequence.
++ (instancetype)generateWithStart:(id)state iterate:(id (^)(id value))iterate map:(id (^)(id value))transform until:(BOOL (^)(id x))predicate;
+
+/// Generate a sequence from start value and iterate block.
+///
+/// This method is equivalent to
+/// `+generateWithStart:iterate:map:until:`, but doesn't transform the
+/// state values.
+///
+/// state     – Initial state value.
+/// until     – Invoke on state to test if the sequence ends. Optional.
+/// iterate   – Invoked on state to calculate the next state.
+///
+/// Returns a lazy sequence.
++ (instancetype)generateWithStart:(id)state iterate:(id (^)(id value))iterate until:(BOOL (^)(id x))predicate;
+
+/// Generate a sequence from start value and iterate block.
+///
+/// This method is equivalent to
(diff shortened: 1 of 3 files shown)


jspahrsummers:
Thanks for the suggestion, and the very thorough pull request description! :sparkles:

Unfortunately, with [RAC 3.0](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1382) coming Soon™, we'd really like to avoid big additions to the RAC 2.x API. `RACSequence` itself is also something of a black sheep, because it's kind of a hack to work around the lack of basic sequence type in Objective-C. :confused: 

So, although I really appreciate the submission, I don't think we should incorporate this. You're more than welcome to keep a fork that provides this method, though!

febeling:
I see. :pensive: 

I'd be interested to understand why RACSequence is a hack or not a very satisfying sequence type? I thought it's simple and clean, and only a bit cumbersome to create, which is what this patch was trying to address. Is there a more fundamental problem that I'm overlooking?

jspahrsummers:
Well, it's a few things:

 * `RACStream` is super weird/annoying
 * `RACSignal` and `RACSequence` are pretty difficult to juggle between
 * `RACSequence` disallows `nil` (while `RACSignal` allows it)
 * `RACSequence.tail` is variously `nil` or an empty sequence when there are no more elements

I could go on. :stuck_out_tongue: 

It's _usable_, mostly, but these are reasons that I don't want to build more on top of it. Swift mostly obsoletes it too, with sequences part of the standard library.

febeling:
Thanks for the answer, I see your point, or points. I feel some of the annoyance with sequences might result from RACSequence being used so much less than signals. I imagine (it's easy when you try) that's got to do with the little straightforward ways to define them. That only relates to 1 and 2 of above, though. the other two are more serious (possibly mostly from an abstract API design point of view, though).

I was already under the impression that RAC 2.0 is in feature-freeze or maintenance mode when I was working on the suggestion.

PR #1698, opened 2015-01-24
Swift PropertySpec

JaviSoto:
Tests that rely on binding SignalProducers to Properties are commented out because the <~ is not enabled due to the Swift compiler crash.

### TODO
- [x] Fix test that verifies the Signal binding is teared down when disposed

### DIFF
--- a/ReactiveCocoaTests/Swift/PropertySpec.swift
+++ b/ReactiveCocoaTests/Swift/PropertySpec.swift
@@ -11,77 +11,62 @@ import Nimble
 import Quick
 import ReactiveCocoa
 
+private let initialPropertyValue = "InitialValue"
+private let subsequentPropertyValue = "SubsequentValue"
+
 class PropertySpec: QuickSpec {
 	override func spec() {
 		describe("ConstantProperty") {
 			it("should have the value given at initialization") {
-				let propertyValue = "StringValue"
-				let constantProperty = ConstantProperty(propertyValue)
+				let constantProperty = ConstantProperty(initialPropertyValue)
 
-				expect(constantProperty.value).to(equal(propertyValue))
+				expect(constantProperty.value).to(equal(initialPropertyValue))
 			}
 
 			it("should yield a producer that sends the current value then completes") {
-				let propertyValue = "StringValue"
-				let constantProperty = ConstantProperty(propertyValue)
+				let constantProperty = ConstantProperty(initialPropertyValue)
 
 				var sentValue: String?
 				var signalCompleted = false
 
 				constantProperty.producer.start(next: { value in
 					sentValue = value
-				},
-				completed: {
+				}, completed: {
 					signalCompleted = true
 				})
 
-				expect(sentValue).to(equal(propertyValue))
+				expect(sentValue).to(equal(initialPropertyValue))
 				expect(signalCompleted).to(beTruthy())
 			}
 		}
 
 		describe("MutableProperty") {
 			it("should have the value given at initialization") {
-				let propertyValue = "StringValue"
-				let mutableProperty = MutableProperty(propertyValue)
+				let mutableProperty = MutableProperty(initialPropertyValue)
 
-				expect(mutableProperty.value).to(equal(propertyValue))
+				expect(mutableProperty.value).to(equal(initialPropertyValue))
 			}
 
 			it("should yield a producer that sends the current value then all changes") {
-				let initialValue = "StringValue"
-				let subsequentValue = "NewStringValue"
-
-				let mutableProperty = MutableProperty(initialValue)
+				let mutableProperty = MutableProperty(initialPropertyValue)
 
 				var sentValue: String?
-				var signalCompleted = false
 
 				mutableProperty.producer.start(next: { value in
 					sentValue = value
-				},
-				completed: {
-					signalCompleted = true
 				})
 
-				expect(sentValue).to(equal(initialValue))
-				expect(signalCompleted).to(beFalsy())
-
-				mutableProperty.value = subsequentValue
-
-				expect(sentValue).to(equal(subsequentValue))
-				expect(signalCompleted).to(beFalsy())
+				expect(sentValue).to(equal(initialPropertyValue))
+				mutableProperty.value = subsequentPropertyValue
+				expect(sentValue).to(equal(subsequentPropertyValue))
 			}
 
 			it("should complete its producer when deallocated") {
-				let propertyValue = "StringValue"
-				var mutableProperty: MutableProperty? = MutableProperty(propertyValue)
+				var mutableProperty: MutableProperty? = MutableProperty(initialPropertyValue)
 
 				var signalCompleted = false
 
-				mutableProperty?.producer.start(next: { value in
-				},
-				completed: {
+				mutableProperty?.producer.start(completed: {
 					signalCompleted = true
 				})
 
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
+					let signal = Signal<Int, NoError>({ observer in
+						return signalDisposable
+					})
+
+					let mutableProperty = MutableProperty<Int>(0)
+
+					let bindingDisposable = mutableProperty <~ signal
+					bindingDisposable.dispose()
+
+					// This test is failing: disposing the binding isn't disposing the signal, or am I doing something wrong?
```
JaviSoto: @jspahrsummers do you see anything obviously wrong with this test?
Am I misinterpreting "tear down the binding"?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
+				let propertyValue = "StringValue"
+				let constantProperty = ConstantProperty(propertyValue)
+
+				var sentValue: String?
+				var signalCompleted = false
+
+				constantProperty.producer.start(next: { value in
+					sentValue = value
+				},
+				completed: {
```
jspahrsummers: Could you please join this line with the one above it?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
+				let mutableProperty = MutableProperty(initialValue)
+
+				var sentValue: String?
+				var signalCompleted = false
+
+				mutableProperty.producer.start(next: { value in
+					sentValue = value
+				},
+				completed: {
+					signalCompleted = true
```
jspahrsummers: I don't think this needs to be part of this example, since it's tested below.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
 
-			pending("should complete its producer when deallocated") {
+			it("should complete its producer when deallocated") {
+				let propertyValue = "StringValue"
+				var mutableProperty: MutableProperty? = MutableProperty(propertyValue)
+
+				var signalCompleted = false
+
+				mutableProperty?.producer.start(next: { value in
+				},
```
jspahrsummers: You can just omit the `next` closure.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
+			it("should have the value given at initialization") {
+				let propertyValue = "StringValue"
+				let mutableProperty = MutableProperty(propertyValue)
+
+				expect(mutableProperty.value).to(equal(propertyValue))
 			}
 
-			pending("should yield a producer that sends the current value then all changes") {
+			it("should yield a producer that sends the current value then all changes") {
+				let initialValue = "StringValue"
```
jspahrsummers: Could you lift this value out of all these blocks, since it's the same for all?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
 class PropertySpec: QuickSpec {
 	override func spec() {
 		describe("ConstantProperty") {
-			pending("should have the value given at initialization") {
+			it("should have the value given at initialization") {
+				let propertyValue = "StringValue"
```
jspahrsummers: Could you lift this value out of these examples and share it?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
+				let constantProperty = ConstantProperty(propertyValue)
+				let propertyOf = PropertyOf(constantProperty)
+
+				var sentValue: String?
+
+				propertyOf.producer.start(next: { value in
+					sentValue = value
+				})
+
+				expect(sentValue).to(equal(propertyValue))
```
jspahrsummers: It's probably worth testing completion here, since that's part of the `ConstantProperty` definition.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
-				pending("should update the property with values sent from the signal") {
+				it("should update the property with values sent from the signal") {
+					let finalPropertyValue = 815
+					var signalObserver: SinkOf<Event<Int, NoError>>?
+
+					let sendPropertyUpdate = {
+						sendNext(signalObserver!, finalPropertyValue)
+					}
+
+					let signal = Signal<Int, NoError>({ observer in
```
jspahrsummers: You probably want `Signal.pipe`. :wink:

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -14,66 +14,258 @@ import ReactiveCocoa
...
+						signalObserver = observer
+
+						return SimpleDisposable()
+					})
+
+					let mutableProperty = MutableProperty<Int>(0)
+
+					mutableProperty <~ signal
```
jspahrsummers: We should probably verify that the property has the same value immediately after this binding.

ReactiveCocoaTests/Swift/PropertySpec.swift
jspahrsummers: Yeah, this won't dispose the _signal_—what it means is that values should no longer be taken from the signal and assigned to the property.

jspahrsummers:
Thanks @JaviSoto! Hopefully those notes are helpful.

ReactiveCocoaTests/Swift/PropertySpec.swift
JaviSoto: Ahhh, got it!
(Also, I didn't know how to easily "push values to a signal". Thanks for pointing out `Signal.pipe`, I hadn't seen that API yet. I assume that's essentially the good ol' `RACSubject`? :))

ReactiveCocoaTests/Swift/PropertySpec.swift
JaviSoto: :+1: Didn't know about it. Thanks!

ReactiveCocoaTests/Swift/PropertySpec.swift
jspahrsummers: > I assume that's essentially the good ol' `RACSubject`?

Yup, basically.

ReactiveCocoaTests/Swift/PropertySpec.swift
JaviSoto: :+1:

ReactiveCocoaTests/Swift/PropertySpec.swift
JaviSoto: :+1:

ReactiveCocoaTests/Swift/PropertySpec.swift
JaviSoto: :+1: Changing that.

ReactiveCocoaTests/Swift/PropertySpec.swift
JaviSoto: :+1:

JaviSoto:
> Thanks @JaviSoto! Hopefully those notes are helpful.

Very :)

JaviSoto:
I think I fixed everything mentioned in the comments. Let me know if I missed anything.
Tests are now passing. @jspahrsummers, how do you feel about merging these?

PR #1699, opened 2015-01-25
SignalProducer tests

JaviSoto:
Putting this up already for discussion. Let me know if you prefer me to put this and #1698 in the same branch.

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalProducerNimbleMatchers.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerNimbleMatchers.swift
@@ -37,12 +37,7 @@ public func sendValues<T: Equatable, E: Equatable>(values: [T], sendError maybeS
 				return false
 			}
 
-			if let error = maybeSendError {
-				if error != sentError {
-					return false
-				}
-			}
-			else if sentError != nil {
+			if sentError != maybeSendError {
 				return false
 			}
 
--- a/ReactiveCocoaTests/Swift/SignalProducerNimbleMatchers.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerNimbleMatchers.swift
@@ -1,0 +1,55 @@
+//
+//  SignalProducerNimbleMatchers.swift
+//  ReactiveCocoa
+//
+//  Created by Javier Soto on 1/25/15.
+//  Copyright (c) 2015 GitHub. All rights reserved.
+//
+
+import Foundation
+
+import ReactiveCocoa
+import Nimble
+
+public func sendValue<T: Equatable, E: Equatable>(value: T?, #sendError: E?, #complete: Bool) -> NonNilMatcherFunc<SignalProducer<T, E>> {
+	return sendValues(value.map { [$0] } ?? [], sendError: sendError, complete: complete)
+}
+
+public func sendValues<T: Equatable, E: Equatable>(values: [T], sendError maybeSendError: E?, #complete: Bool) -> NonNilMatcherFunc<SignalProducer<T, E>> {
+	return NonNilMatcherFunc { actualExpression, failureMessage in
+		failureMessage.postfixMessage = "Send values \(values). Send error \(maybeSendError). Complete: \(complete)"
+		let maybeProducer = actualExpression.evaluate()
+
+		if let signalProducer = maybeProducer {
+			var sentValues: [T] = []
+			var sentError: E?
+			var signalCompleted = false
+
+			signalProducer.start(next: { value in
+				sentValues.append(value)
+			}, error: { error in
+				sentError = error
+			}, completed: {
+				signalCompleted = true
+			})
+
+			if sentValues != values {
+				return false
+			}
+
+			if let error = maybeSendError {
+				if error != sentError {
+					return false
+				}
+			}
+			else if sentError != nil {
+				return false
+			}
+
+			return signalCompleted == complete
+		}
+		else {
+			return false
+		}
+	}
+}
--- a/ReactiveCocoaTests/Swift/SignalProducerNimbleMatchers.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerNimbleMatchers.swift
@@ -17,6 +17,8 @@ public func sendValue<T: Equatable, E: Equatable>(value: T?, #sendError: E?, #co
 
 public func sendValues<T: Equatable, E: Equatable>(values: [T], sendError maybeSendError: E?, #complete: Bool) -> NonNilMatcherFunc<SignalProducer<T, E>> {
 	return NonNilMatcherFunc { actualExpression, failureMessage in
+		precondition(maybeSendError == nil || !complete, "Signals can't both send an error and complete")
+
 		failureMessage.postfixMessage = "Send values \(values). Send error \(maybeSendError). Complete: \(complete)"
 		let maybeProducer = actualExpression.evaluate()
 
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -6,53 +6,23 @@
 //  Copyright (c) 2015 GitHub. All rights reserved.
 //
 
+import Foundation
+
 import LlamaKit
 import Nimble
 import Quick
 import ReactiveCocoa
-import Foundation
-
-private func startSignalProducer<T: Equatable, E: Equatable>(signalProducer: SignalProducer<T, E>, #expectSentValue: T?, sentError expectSentError: E?, #complete: Bool) {
-	startSignalProducer(signalProducer, expectSentValues: expectSentValue.map { [$0] } ?? [], sentError: expectSentError, complete: complete)
-}
-
-private func startSignalProducer<T: Equatable, E: Equatable>(signalProducer: SignalProducer<T, E>, expectSentValues: [T] = [], sentError expectSentError: E?, #complete: Bool) {
-	var sentValues: [T] = []
-	var sentError: E?
-	var signalCompleted = false
-
-	signalProducer.start(next: { value in
-		sentValues.append(value)
-	},
-	error: { error in
-		sentError = error
-	},
-	completed: {
-		signalCompleted = true
-	})
-
-	expect(sentValues).to(equal(expectSentValues))
-
-	if let error = expectSentError {
-		expect(sentError).to(equal(expectSentError))
-	}
-	else {
-		expect(sentError).to(beNil())
-	}
-
-	expect(signalCompleted).to(equal(complete))
-}
 
 class SignalProducerSpec: QuickSpec {
 	override func spec() {
 		describe("init") {
 			it("should run the handler once per start()") {
 				var handlerCalledTimes = 0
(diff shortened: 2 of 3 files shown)


ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -72,13 +152,42 @@ class SignalProducerSpec: QuickSpec {
...
-			pending("should send the error") {
+			it("should send the error") {
+				let operationError = NSError(domain: "com.reactivecocoa.errordomain", code: 4815, userInfo: nil)
+				let operation: () -> Result<String, NSError> = {
+					return Result<String, NSError>.Failure(Box(operationError))
+				}
+
+				let signalProducer = SignalProducer.try(operation)
+
+				// This is not the current behavior, but should it?
```
JaviSoto: @jspahrsummers I realized that, contrary to what I was expecting, signals that just send an error they don't complete. What's the reasoning behind that?

JaviSoto: Oh, I just remembered: in RAC2, a signal either errors or completes, but never both. I guess that makes sense, but from an API usage perspective, it can be weird. For example, if you're just interested in the values, and want to know when it's *done sending values*, you would have to handle both the error and the completion. This can be useful (forcing the client to handle errors), but it would be easier to be able to rely on the completion event always being sent.

Considering this behavior has been this way in RAC for a very long time, it's probably for a very good reason :)

kastiglione: From [Event.swift](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/8c67470a34dde0b10e296a95a974c48b63990063/ReactiveCocoa/Swift/Event.swift#L17)

> `Next* (Error | Completed)?`

Yep, it's been like this for for a while. I've sometimes considered a unified `Completed` that has an optional error, but I strongly suspect that would be more complicated than it is now.

> if you're just interested in the values, and want to know when it's _done sending values_, you would have to handle both the error and the completion

I've found the existing operators provided good balance so that this has never been a pain point or a source of boilerplate noise, for me personally anyway. I also can't recall many issues/questions about this. The easy solution is to just use: `catchTo(SignalProducer.empty)`.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -10,11 +10,54 @@ import LlamaKit
...
+	expect(signalCompleted).to(equal(complete))
+}
 
 class SignalProducerSpec: QuickSpec {
 	override func spec() {
 		describe("init") {
-			pending("should run the handler once per start()") {
+			it("should run the handler once per start()") {
+				var handlerCalledTimes = 0
+				let signalProducer = SignalProducer<String, NSError>({ observer, disposable in
```
kastiglione: Use a trailing closure?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -72,13 +152,42 @@ class SignalProducerSpec: QuickSpec {
 		}
 
 		describe("SignalProducer.try") {
-			pending("should run the operation once per start()") {
+			it("should run the operation once per start()") {
+				var operationRunTimes = 0
+				let operation: () -> Result<String, NSError> = {
+					operationRunTimes++
+
+					return Result<String, NSError>.Success(Box("OperationValue"))
```
kastiglione: It's not necessary to redundantly type the `Result`. Can be just `.Success(…)`.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -72,13 +152,42 @@ class SignalProducerSpec: QuickSpec {
...
+				SignalProducer.try(operation).start()
+
+				expect(operationRunTimes).to(equal(2))
 			}
 
-			pending("should send the value then complete") {
+			it("should send the value then complete") {
+				let operationReturnValue = "OperationValue"
+				let operation: () -> Result<String, NSError> = {
+					return Result<String, NSError>.Success(Box(operationReturnValue))
```
kastiglione: Can be just `.Success(…)`.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -72,13 +152,42 @@ class SignalProducerSpec: QuickSpec {
...
+				let signalProducer = SignalProducer.try(operation)
+
+				startSignalProducer(signalProducer, expectSentValue: operationReturnValue, sentError: nil, complete: true)
 			}
 
-			pending("should send the error") {
+			it("should send the error") {
+				let operationError = NSError(domain: "com.reactivecocoa.errordomain", code: 4815, userInfo: nil)
+				let operation: () -> Result<String, NSError> = {
+					return Result<String, NSError>.Failure(Box(operationError))
```
kastiglione: Can be just `.Failure(…)`.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -10,11 +10,54 @@ import LlamaKit
 import Nimble
 import Quick
 import ReactiveCocoa
+import Foundation
+
+private func startSignalProducer<T: Equatable, E: Equatable>(signalProducer: SignalProducer<T, E>, #expectSentValue: T?, sentError expectSentError: E?, #complete: Bool) {
+	startSignalProducer(signalProducer, expectSentValues: expectSentValue.map { [$0] } ?? [], sentError: expectSentError, complete: complete)
+}
+
+private func startSignalProducer<T: Equatable, E: Equatable>(signalProducer: SignalProducer<T, E>, expectSentValues: [T] = [], sentError expectSentError: E?, #complete: Bool) {
```
kastiglione: In previous RAC tests, there's never been a helper like this. I believe, but could be wrong, the preference is to have tests directly use the API they're testing. I say this partly to ask aloud.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
almassapargali: or it could be just `success("OperationValue")`. LlamaKit states it's preferred because no need to deal with `Box()`.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
jspahrsummers: In addition to @kastiglione's points, I'll note that _separate_ completed/error events is “higher fidelity.”

In other words, it's always possible to treat an error event as completion, or vice-versa—but if completion were _always_ sent, it'd be much harder to detect a successful termination (because you could only do so through the absence of errors—which you would have to track statefully).

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
jspahrsummers: Yep, @almassapargali is right. We should use `success()` and `failure()` as much as possible.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
jspahrsummers: I would like to avoid a helper like this. If we _really_ want to make it more convenient, I'd favor writing one or more Nimble matchers instead.

jspahrsummers:
The separate branch is great, since they're completely independent PRs. Thanks again @JaviSoto! :sparkles:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
JaviSoto: > but if completion were always sent, it'd be much harder to detect a successful termination

I think that's the key :+1: Makes total sense. Let me fix my (incorrect) assumptions in these tests!

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
JaviSoto: :+1:

JaviSoto: :+1:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
JaviSoto: :+1: Didn't know about these. I'll change it everywhere.

JaviSoto: Mmmm what's the syntax to specify the "missing type" that it can't infer?

```swift
// Argument for generic type E can't be inferred
success("successValue")
```

```swift
// Cannot explicitly specialize a generic function
success<String, NoError>("successValue")
```

JaviSoto: Oh, I have to use `as`.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
JaviSoto: I initially implemented all of this in each method, and after a while I realized it was pretty WET. This method made the tests very concise and easy to read: they just describe what the expected output from the producer is.

That been said, I'm gonna look into implementing it as a matcher and see where that takes me :+1:

JaviSoto: I implemented it using a matcher (really neat API!). What do you guys think? https://github.com/JaviSoto/ReactiveCocoa/commit/c9dbd222e5bb5b380b1838f7ae897512ca737675

ReactiveCocoaTests/Swift/SignalProducerNimbleMatchers.swift
```diff
@@ -0,0 +1,55 @@
...
+
+			if sentValues != values {
+				return false
+			}
+
+			if let error = maybeSendError {
+				if error != sentError {
+					return false
+				}
+			}
```
JaviSoto: I *think* this could use some Applicative magic, but I haven't gotten there in my personal endeavor to FP-land :grinning: Any pointers?

kastiglione: It's not necessary to unwrap the optional for comparing. It can be just:

```
if sentError != maybeSendError {
    return false
}
```

JaviSoto: Oh, duh! what was I thinking?

JaviSoto:
Woot an unrelated test failed https://travis-ci.org/ReactiveCocoa/ReactiveCocoa/builds/48384428#L2966

jspahrsummers:
Is this ready for review? No worries if not—just checking on the status, since the title still says WIP.

JaviSoto:
@jspahrsummers Sorry about that, been super busy. I just merged swift-development back in. It was marked as WIP because there are still some more tests to write, but I'm out this weekend, so maybe it's better if we merge this and then I or somebody else writes the remaining tests?

almassapargali:
I'm planning to finish Signal's tests this weekend. Can take a look at these tests too.

jspahrsummers:
@Javi No problem, thanks! :sparkles:

PR #1702, opened 2015-01-25
Signal tests

almassapargali:
Here I started doing spec for Signal. Putting here for reviews. Please note that I'm new to ReactiveCocoa and I might totally misunderstood some things.

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -243,7 +243,27 @@ class SignalSpec: QuickSpec {
 		}
 
 		describe("map") {
-			pending("should transform the values of the signal") {
+			it("should transform the values of the signal") {
+				let numbers = [ 1, 2, 5 ]
+				var testScheduler = TestScheduler()
+				
+				let signal: Signal<Int, NoError> = Signal { observer in
+					testScheduler.schedule {
+						for number in numbers {
+							sendNext(observer, number)
+						}
+					}
+					return nil
+				}
+				
+				var afterMap: [Int] = []
+				
+				signal
+				|> map { $0 * 2 }
+				|> observe(next: { afterMap.append($0) })
+				
+				testScheduler.run()
+				expect(afterMap).to(equal([2, 4, 10]))
 			}
 		}
 
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -23,7 +23,7 @@ class SignalSpec: QuickSpec {
 			
 			it("should run the generator immediately") {
 				var didRunGenerator = false
-				Signal<Int, NoError> { observer in
+				Signal<AnyObject, NoError> { observer in
 					didRunGenerator = true
 					return nil
 				}
@@ -32,7 +32,7 @@ class SignalSpec: QuickSpec {
 			}
 
 			it("should keep signal alive if not terminated") {
-				weak var signal: Signal<Int, NoError>? = Signal { observer in
+				weak var signal: Signal<AnyObject, NoError>? = Signal { observer in
 					return nil
 				}
 				
@@ -40,7 +40,7 @@ class SignalSpec: QuickSpec {
 			}
 
 			it("should deallocate after erroring") {
-				weak var signal: Signal<Int, NSError>? = Signal { observer in
+				weak var signal: Signal<AnyObject, NSError>? = Signal { observer in
 					testScheduler.schedule {
 						sendError(observer, testError)
 					}
@@ -61,7 +61,7 @@ class SignalSpec: QuickSpec {
 			}
 
 			it("should deallocate after completing") {
-				weak var signal: Signal<Int, NSError>? = Signal { observer in
+				weak var signal: Signal<AnyObject, NSError>? = Signal { observer in
 					testScheduler.schedule {
 						sendCompleted(observer)
 					}
@@ -115,7 +115,7 @@ class SignalSpec: QuickSpec {
 			it("should dispose of returned disposable upon error") {
 				let disposable = SimpleDisposable()
 				
-				let signal: Signal<Int, NSError> = Signal { observer in
+				let signal: Signal<AnyObject, NSError> = Signal { observer in
 					testScheduler.schedule {
 						sendError(observer, testError)
 					}
@@ -138,7 +138,7 @@ class SignalSpec: QuickSpec {
 			it("should dispose of returned disposable upon completion") {
 				let disposable = SimpleDisposable()
 				
-				let signal: Signal<Int, NSError> = Signal { observer in
+				let signal: Signal<AnyObject, NSError> = Signal { observer in
 					testScheduler.schedule {
 						sendCompleted(observer)
 					}
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -198,7 +198,38 @@ class SignalSpec: QuickSpec {
 		}
 
 		describe("observe") {
-			pending("should stop forwarding events when disposed") {
+			var testScheduler: TestScheduler!
+			
+			beforeEach {
+				testScheduler = TestScheduler()
+			}
+			
+			it("should stop forwarding events when disposed") {
+				let disposable = SimpleDisposable()
+				
+				let signal: Signal<Int, NoError> = Signal { observer in
+					testScheduler.schedule {
+						for number in [1, 2] {
+							sendNext(observer, number)
+						}
+						sendCompleted(observer)
+						sendNext(observer, 4)
+					}
+					return disposable
+				}
+				
+				var fromSignal: [Int] = []
+				signal.observe(next: { number in
+					fromSignal.append(number)
+				})
+				
+				expect(disposable.disposed).to(beFalsy())
+				expect(fromSignal).to(beEmpty())
+				
+				testScheduler.run()
+				
+				expect(disposable.disposed).to(beTruthy())
+				expect(fromSignal).to(equal([1, 2]))
 			}
 
 			pending("should not trigger side effects") {
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -745,26 +745,6 @@ class SignalSpec: QuickSpec {
 
 		describe("takeWhile") {
 			pending("should take while the predicate is true") {
-//				let numbers = [ 1, 2, 4, 4, 5, 2 ]
-//				var testScheduler = TestScheduler()
-//				
-//				let signal: Signal<Int, NoError> = Signal { observer in
-//					testScheduler.schedule {
-//						for number in numbers {
-//							sendNext(observer, number)
-//						}
-//					}
-//					return nil
-//				}
-//				
-//				var result: [Int] = []
-//				
-//				signal
-//				|> takeWhile { $0 < 4 }
-//				|> observe(next: { result.append($0) })
-//				
-//				testScheduler.run()
-//				expect(result).to(equal([ 1, 2 ]))
 			}
 
 			pending("should complete if the predicate starts false") {
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -940,7 +940,34 @@ class SignalSpec: QuickSpec {
 		}
 
 		describe("materialize") {
-			pending("should reify events from the signal") {
+			it("should reify events from the signal") {
+				let (signal, observer) = Signal<Int, TestError>.pipe()
+				var latestEvent: Event<Int, TestError>?
+				signal
+				|> materialize
+				|> observe(next: { latestEvent = $0 })
+				
+				sendNext(observer, 2)
+				
+				expect(latestEvent).toNot(beNil())
+				if let latestEvent = latestEvent {
+					switch latestEvent {
+					case let .Next(box):
+						expect(box.unbox).to(equal(2))
+					default:
+						fail()
+					}
+				}
+				
+				sendError(observer, TestError.Default)
+				if let latestEvent = latestEvent {
+					switch latestEvent {
+					case .Error(_):
+						()
+					default:
+						fail()
+					}
+				}
 			}
 		}
 
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -232,7 +232,17 @@ class SignalSpec: QuickSpec {
 				expect(fromSignal).to(equal([1, 2]))
 			}
 
-			pending("should not trigger side effects") {
+			it("should not trigger side effects") {
+				var runCount = 0
+				let signal: Signal<(), NoError> = Signal { observer in
+					runCount += 1
+					return nil
+				}
+				
+				expect(runCount).to(equal(1))
+				
+				signal.observe()
+				expect(runCount).to(equal(1))
 			}
 
 			pending("should release observer after termination") {
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -14,7 +14,6 @@ import ReactiveCocoa
 class SignalSpec: QuickSpec {
 	override func spec() {
 		describe("init") {
-			let testError = NSError(domain: "SignalSpec", code: 1, userInfo: nil)
 			var testScheduler: TestScheduler!
 			
 			beforeEach {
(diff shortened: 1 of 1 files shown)


almassapargali:
Let me know if I should doing this 😆 (considering it's pretty bad)

almassapargali:
Also I'm not sure whether I understood "should deallocate..." test correctly. Obj-C tests checks that using disposables, so I did.

almassapargali:
Also I feel like I'm over checking testing variables like `errored`, `dealloced`, and `completed`. Should I check only final states, or leave as is?

almassapargali:
So when I came to `"should dispose of returned disposable upon error/completion"` I realised I misunderstood `"should deallocate after completing/erroring"` and actually that implementation belongs to `"should dispose..."` tests.

That said, can somebody show example of `"should deallocate..."` tests. I tried to define signal as variable and check if it's nil after erroring and it failed. Or does test really should check against nil and it's doing what it's supposed to do (show that Signal implementation has bugs).

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -26,13 +45,91 @@ class SignalSpec: QuickSpec {
...
+				}
+				
+				var fromSignal: [Int] = []
+				
+				signal.observe(
+					next: { number in
+						fromSignal.append(number)
+					},
+					error: { _ in },
+					completed: {}
```
kastiglione: The `error` and `completed` args have do-nothing defaults and can be left out.

almassapargali: @kastiglione yep seen that, though didn't know you can just left them out. Removed now.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -26,13 +45,91 @@ class SignalSpec: QuickSpec {
...
+				)
+				
+				expect(fromSignal).to(beEmpty())
+				
+				testScheduler.run()
+				
+				expect(fromSignal).toEventually(equal(numbers))
```
kastiglione: I don't think the needs to be eventually.

jspahrsummers:
Thanks for the pull request @almassapargali! At a glance, this looks like a fine start. I'll review it more in-depth when I have some more time. :smile:

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -14,10 +14,29 @@ import ReactiveCocoa
...
+			}
+			
+			it("should run the generator immediately") {
+				var didRunGenerator = false
+				Signal<Int, NoError> { observer in
+					didRunGenerator = true
+					return nil
+				}
+				
+				expect(didRunGenerator).to(beTrue())
```
jspahrsummers: Can you please use `beTruthy` and `beFalsy`, since those are more portable to other versions of Nimble?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -14,10 +14,29 @@ import ReactiveCocoa
 class SignalSpec: QuickSpec {
 	override func spec() {
 		describe("init") {
-			pending("should run the generator immediately") {
-			}
-
-			pending("should keep signal alive if not terminated") {
+			let testError = NSError(domain: "SignalSpec", code: 1, userInfo: nil)
```
jspahrsummers: Since signals can be parameterized by any `ErrorType`, I'd prefer if we could define a custom error `enum` for our unit tests.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -14,10 +14,29 @@ import ReactiveCocoa
...
+				
+				expect(didRunGenerator).to(beTrue())
+			}
+
+			it("should keep signal alive if not terminated") {
+				let signal: Signal<Int, NoError> = Signal { observer in
+					return nil
+				}
+				
+				expect(signal).toNot(beNil())
```
jspahrsummers: The `signal` property is strong, so the signal will be kept alive regardless. A better test would be to put the signal into a `weak var` and see if it disappears.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -26,13 +45,85 @@ class SignalSpec: QuickSpec {
...
-			pending("should dispose of returned disposable upon completion") {
+			it("should forward events to observers") {
+				let numbers = [ 1, 2, 5 ]
+				
+				let signal: Signal<Int, NoError> = Signal { observer in
+					testScheduler.schedule {
+						for number in numbers {
+							sendNext(observer, number)
+						}
+					}
```
jspahrsummers: Can you please test sending and handling completion too?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -26,13 +45,85 @@ class SignalSpec: QuickSpec {
...
+					return nil
+				}
+				
+				var fromSignal: [Int] = []
+				
+				signal.observe(
+					next: { number in
```
jspahrsummers: Can you write this more tightly inside the parentheses? For example:

```swift
signal.observe(next: { number in
    …
}, error: { error in
    …
}, completed: {
    …
})
```

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -26,13 +45,85 @@ class SignalSpec: QuickSpec {
...
+				
+				expect(fromSignal).to(beEmpty())
+				
+				testScheduler.run()
+				
+				expect(fromSignal).toEventually(equal(numbers))
+			}
+
+			it("should dispose of returned disposable upon error") {
+				var disposed = false
```
jspahrsummers: If you use a `SimpleDisposable`, you can get this behavior for free. :smile:

jspahrsummers:
I've left some notes on the work so far. I hope they're helpful! Please let me know if you have any questions.

ReactiveCocoaTests/Swift/SignalSpec.swift
almassapargali: Sure. Will change soon.

ReactiveCocoaTests/Swift/SignalSpec.swift
almassapargali: Ok, thanks. I think that's why I couldn't implement should deallocate tests. Will try again with weak variables.

ReactiveCocoaTests/Swift/SignalSpec.swift
almassapargali: Yep

ReactiveCocoaTests/Swift/SignalSpec.swift
almassapargali: Changed!

ReactiveCocoaTests/Swift/SignalSpec.swift
almassapargali: Please check out #1705 on this.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -14,25 +14,148 @@ import ReactiveCocoa
...
+				weak var signal: Signal<AnyObject, NSError>? = Signal { observer in
+					testScheduler.schedule {
+						sendError(observer, testError)
+					}
+					return nil
+				}
+				
+				var errored = false
+				
+				signal!.observe(error: { _ in errored = true })
```
almassapargali: I wonder if I should've used optional chain here, like `signal?.observe()` since `!` would theoretically crash tests if signal happens to be `nil`. Any thoughts?

jspahrsummers: Yep, I think we should use optional chaining wherever possible in our tests.

almassapargali:
Added some more tests, let me know if I'm doing something wrong.
Also skipped some pending tests, would be glad if you direct me on how to implement them.

jspahrsummers:
@almassapargali Which pending tests are you referring to?

almassapargali:
Between implemented ones. I'll write exact lines when get to Mac.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -14,30 +14,154 @@ import ReactiveCocoa
...
 		describe("Signal.pipe") {
-			pending("should keep signal alive if not terminated") {
+			
+			it("should keep signal alive if not terminated") {
+				weak var signal = Signal<(), NoError>.pipe().0
+				
+				expect(signal).toNot(beNil())
 			}
 
 			pending("should deallocate after erroring") {
```
almassapargali: Here I couldn't get weak variable to signal, while keeping pointer to observer. Tried this, but something still holds that variable:
screen shot 2015-01-29 at 11 37 01 am

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -46,12 +170,66 @@ class SignalSpec: QuickSpec {
...
+				expect(disposable.disposed).to(beFalsy())
+				expect(fromSignal).to(beEmpty())
+				
+				testScheduler.run()
+				
+				expect(disposable.disposed).to(beTruthy())
+				expect(fromSignal).to(equal([1, 2]))
 			}
 
 			pending("should not trigger side effects") {
```
almassapargali: Have no idea how to test this.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -65,25 +243,125 @@ class SignalSpec: QuickSpec {
 		}
```
almassapargali: How to get reference to observer on these tests?

almassapargali:
@jspahrsummers added side notes on skipped tests.

ReactiveCocoaTests/Swift/SignalSpec.swift
jspahrsummers: You may need to put some logic inside a closure, to put explicit bounds on the lifetime of some objects. See [the old tests](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-before-signal-producers/ReactiveCocoaTests/Swift/HotSignalSpec.swift#L18-L39) for some examples.

ReactiveCocoaTests/Swift/SignalSpec.swift
jspahrsummers: If you put some side effects (like incrementing a counter) in the generator passed to `init`, multiple `observe` invocations shouldn't affect how many times those side effects occur.

In other words, `Signal` effects should always occur _exactly_ once, and observers don't change that.

ReactiveCocoaTests/Swift/SignalSpec.swift
jspahrsummers: I think the easiest way to test this would be to capture some object in your observer callbacks, then make sure that object gets released (as a proxy for identifying when the observer itself is released).

almassapargali:
Argh, [this](https://travis-ci.org/ReactiveCocoa/ReactiveCocoa/builds/49026716#L1812) compiler fail is really bugging me. Can anybody else try running these tests please.

almassapargali:
I have couple of questions:

1. I have [test](https://github.com/almassapargali/ReactiveCocoa/commit/45efd2e7c3a90755e2755af7e39a0f8ece7528b6) which depends on #1710 to pass. Should I comment it out so CI tests passes, or leave as is?

2. I was limiting myself to use of plain Signals. Though @kastiglione used `Signal.pipe()` on [this](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1718/files#diff-9f6d020c3d0578997abeca0af04f0e16R140) tests. Now, I'm wondering should I prefer `Signal.pipe()` over plain signal, since it makes use of `TestScheduler` unnecessary and results in less code?

cc/ @jspahrsummers, @kastiglione

almassapargali:
@kastiglione couldn't reply on inline note, so writing here.

> Yep. I'm not sure which is better (if there is such thing as "better") wrapping in unsightly (), or breaking out an intermediate variable to call the observe() method on.

Agree that breaking out an intermediate variable looks more clear and readable (thus preferable). Though it'll introduce the need to declare signal as `var`. Don't now whether this important at all. So if you're ok I'll use intermediate variable.

kastiglione:
> 1\. I have test which depends on #1710 to pass. Should I comment it out so CI tests passes, or leave as is?

You can merge that branch (`producer-sync`) into the branch for this pull request. Then, the convention is to mark this pull request as depending on that pull request. This is typically noted in the first line of the pull request description, which is editable. See #1719 which depends on #1714.

> Though it'll introduce the need to declare signal as `var`.

Unless I'm missing something, this is not the case. For example, see [here]( https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1719/files#diff-a167eede91e4c815deb02376a37b6049R256).

almassapargali:
@kastiglione thanks on branch dependency convention. Will do that.

> Unless I'm missing something, this is not the case. For example, see here.

On this case signal declared already [here](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1702/files#diff-9f6d020c3d0578997abeca0af04f0e16R586).

almassapargali:
Depends on #1710 onwards.

almassapargali:
Did I mess up something when merging `producer-sync`?

jspahrsummers:
> I'm wondering should I prefer `Signal.pipe()` over plain signal, since it makes use of `TestScheduler` unnecessary and results in less code?

Unless an example is specifically testing a `Signal` constructor or `Signal.observe`, I think `Signal.pipe` is totally fine. Just use whichever is clearer/more appropriate to the test. :+1:

jspahrsummers:
Is this ready for review? No worries if not—just checking on the status, since the title still says WIP.

almassapargali:
@jspahrsummers not yet, will continue soon.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -14,50 +14,293 @@ import ReactiveCocoa
...
+				sendNext(sink, 1)
+				expect(testStr).to(equal("1"))
+				sendNext(sink, 2)
+				expect(testStr).to(equal("12"))
+
+				sendCompleted(sink)
+				expect(testStr).to(beNil())
 			}
 
 			pending("should release observer after disposal") {
```
almassapargali: @jspahrsummers what is difference between this and previous (`"should release observer after termination"`)  test?

almassapargali: I'm asking about implementation difference?

jspahrsummers: The intention for this one is that _cancellation_ should also remove all observers. It's a little weird given that we're in `SignalSpec`, but I think you'll need to use `SignalProducer.start…` here, and dispose of the returned `Disposable`.

almassapargali: Ok, thanks.

jspahrsummers: No problem!

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -14,121 +14,771 @@ import ReactiveCocoa
...
+						innerStr.appendString("\(value)")
+					})
+					testStr = innerStr
+				}
+				test()
+
+				expect(testStr).to(equal("12"))
+
+				disposable.dispose()
+//				expect(testStr).to(beNil())
```
almassapargali: Hey guys, any ideas why this doesn't pass?

jspahrsummers: Does it work if you use `toEventually` instead?

almassapargali: It fails too.

jspahrsummers: I dunno, then. Maybe you've found a bug?

If you can't figure out anything more, you can leave this test disabled for now.

almassapargali: Might (and should) removing reference to disposable change somethings?

almassapargali:
@jspahrsummers so far almost all tests implemented. Left ones are [this test](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1702/files#diff-9f6d020c3d0578997abeca0af04f0e16R306) and tests for `throttle`. I guess you're going to merge #1730, and if that's the case I can implement tests for it, adding #1730 as dependency.

ReactiveCocoaTests/Swift/SignalSpec.swift
jspahrsummers: I don't know without digging in myself, so it might be easier to disable this test for the moment if you don't find a workaround.

jspahrsummers:
No need to write tests for `throttle` in this PR. Thanks, though!

almassapargali:
Yeah, it was great to finally get to use ReactiveCocoa though :smile:

almassapargali:
:+1:

PR #1703, opened 2015-01-26
Fix race condition in Signal.takeUntilReplacement()

andymatuschak:
Fixes a subtle race condition that could occur if the replacement signal sent an event while the original signal observation was happening.


andymatuschak:
Oh, crap, uh, I set the wrong destination branch.

Issue #956, opened 2013-11-20
Operator that switches from one signal to another when the second signal sends a value


Coneko:
Basically it's `[[firstSignal takeUntil:secondSignal] concat:secondSignal]` except it needs to subscribe to `secondSignal` first and multicast it because of side effects.

I currently use it for signals that send values in response to events and need to send the current value on subscription in a thread-safe way, like `NSKeyValueObservingOptionInitial` does for KVO.

I can see it having use cases with time consuming work that can be easily approximated, but where the approximation is computed in parallel with the actual work. For example displaying a high resolution image with a low resolution fallback. The high resolution image might finish loading first if it was cached or if the low resolution image is unavailable.

Is this useful enough or tricky enough to implement to include it?

PR #1704, opened 2015-01-26
Fix race condition in Signal.takeUntilReplacement()

andymatuschak:
Fixes a subtle race condition that could occur if the replacement signal sent an event while the original signal observation was happening.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -571,47 +571,6 @@ public func takeLast<T,E>(count: Int)(signal: Signal<T,E>) -> Signal<T,E> {
 	}
 }
 
-public func takeUntilReplacement<T, E>(replacement: Signal<T, E>)(signal: Signal<T, E>) -> Signal<T, E> {
-	return Signal { observer in
-		/// When this disposable is disposed, we'll stop forwarding events from `signal` to
-		/// the observer. It's atomic because we need to defend against the case that the
-		/// replacement signal sends an event while we're starting to observe `signal`.
-		let signalDisposableAtomic = Atomic(SerialDisposable())
-
-		// When the replacement signal sends an event, make sure we're no longer forwarding
-		// events from `signal`, then forward that event.
-		let replacementDisposable = replacement.observe(SinkOf { event in
-			signalDisposableAtomic.modify { signalDisposable in
-				signalDisposable.dispose()
-				return signalDisposable
-			}
-			observer.put(event)
-		})
-
-		// If `replacement` hasn't already sent an event, start observing `signal`.
-		signalDisposableAtomic.modify { signalDisposable in
-			if !signalDisposable.disposed {
-				// Forard values and errors, but not completion events, to the observer.
-				signalDisposable.innerDisposable = signal.observe(next: { value in
-					sendNext(observer, value)
-				}, error: { error in
-					sendError(observer, error)
-				})
-			}
-			return signalDisposable
-		}
-
-		// Stop both observations when the observation on the output signal is disposed.
-		return ActionDisposable {
-			signalDisposableAtomic.modify { signalDisposable in
-				signalDisposable.dispose()
-				return signalDisposable
-			}
-			replacementDisposable.dispose()
-		}
-	}
-}
-
 /*
 TODO
 
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -513,34 +513,27 @@ public func takeUntilReplacement<T, E>(replacement: Signal<T, E>)(signal: Signal
 		// When the replacement signal sends an event, make sure we're no longer forwarding
 		// events from `signal`, then forward that event.
 		let replacementDisposable = replacement.observe(SinkOf { event in
-			signalDisposableAtomic.modify { signalDisposable in
+			signalDisposableAtomic.withValue { signalDisposable in
 				signalDisposable.dispose()
-				return signalDisposable
 			}
 			observer.put(event)
-			})
+		})
 
 		// If `replacement` hasn't already sent an event, start observing `signal`.
-		signalDisposableAtomic.modify { signalDisposable in
+		signalDisposableAtomic.withValue { signalDisposable -> Void in
 			if !signalDisposable.disposed {
 				// Forward values and errors, but not completion events, to the observer.
 				signalDisposable.innerDisposable = signal.observe(next: { value in
 					sendNext(observer, value)
-					}, error: { error in
-						sendError(observer, error)
+				}, error: { error in
+					sendError(observer, error)
 				})
 			}
-			return signalDisposable
 		}
 
 		// Stop both observations when the observation on the output signal is disposed.
-		return ActionDisposable {
-			signalDisposableAtomic.modify { signalDisposable in
-				signalDisposable.dispose()
-				return signalDisposable
-			}
-			replacementDisposable.dispose()
-		}
+		// No need to lock `signalDisposableAtomic` because the critical region has already run.
+		return CompositeDisposable([ signalDisposableAtomic.value, replacementDisposable ])
 	}
 }
 
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -505,42 +505,26 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
 /// instead, regardless of whether `signal` has sent events already.
 public func takeUntilReplacement<T, E>(replacement: Signal<T, E>)(signal: Signal<T, E>) -> Signal<T, E> {
 	return Signal { observer in
-		/// When this disposable is disposed, we'll stop forwarding events from `signal` to
-		/// the observer. It's atomic because we need to defend against the case that the
-		/// replacement signal sends an event while we're starting to observe `signal`.
-		let signalDisposableAtomic = Atomic(SerialDisposable())
+		let signalDisposable = SerialDisposable()
 
-		// When the replacement signal sends an event, make sure we're no longer forwarding
-		// events from `signal`, then forward that event.
-		let replacementDisposable = replacement.observe(SinkOf { event in
-			signalDisposableAtomic.modify { signalDisposable in
-				signalDisposable.dispose()
-				return signalDisposable
-			}
-			observer.put(event)
+		let replacementDisposable = replacement.observe(next: { value in
+			signalDisposable.dispose()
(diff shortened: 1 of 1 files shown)


kastiglione:
Nice :eyeglasses: . I guess we have a race in the [objc version](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/71f1dbfd1265538d30dae41152c7fd8929f5393c/ReactiveCocoa/RACSignal%2BOperations.m#L771-L775) too.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -505,26 +505,42 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+				signalDisposable.dispose()
+				return signalDisposable
+			}
+			observer.put(event)
 			})
+
+		// If `replacement` hasn't already sent an event, start observing `signal`.
+		signalDisposableAtomic.modify { signalDisposable in
+			if !signalDisposable.disposed {
+				// Forard values and errors, but not completion events, to the observer.
```
kastiglione: For**w**ard

andymatuschak: Thanks, fixed.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -505,26 +505,42 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
-				sendError(observer, error)
+		/// When this disposable is disposed, we'll stop forwarding events from `signal` to
+		/// the observer. It's atomic because we need to defend against the case that the
+		/// replacement signal sends an event while we're starting to observe `signal`.
+		let signalDisposableAtomic = Atomic(SerialDisposable())
+
+		// When the replacement signal sends an event, make sure we're no longer forwarding
+		// events from `signal`, then forward that event.
+		let replacementDisposable = replacement.observe(SinkOf { event in
+			signalDisposableAtomic.modify { signalDisposable in
```
kastiglione: I believe these `modify`s can all be `withValue`s.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -505,26 +505,42 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+				})
+			}
+			return signalDisposable
 		}
 
-		return CompositeDisposable([ signalDisposable, replacementDisposable ])
+		// Stop both observations when the observation on the output signal is disposed.
+		return ActionDisposable {
+			signalDisposableAtomic.modify { signalDisposable in
+				signalDisposable.dispose()
```
kastiglione: Is it necessary for this disposal to be wrapped by the atomic? At this point, the only parallel action could be disposal happening in the observation of `replacement`, and it's just calling `dispose`. Though not documented, `dispose`'s contract allows it to be called more than once with the extra calls being nops.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -505,26 +505,42 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+
+		// When the replacement signal sends an event, make sure we're no longer forwarding
+		// events from `signal`, then forward that event.
+		let replacementDisposable = replacement.observe(SinkOf { event in
+			signalDisposableAtomic.modify { signalDisposable in
+				signalDisposable.dispose()
+				return signalDisposable
+			}
+			observer.put(event)
 			})
```
kastiglione: This line is indented one too many.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -505,26 +505,42 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+
+		// If `replacement` hasn't already sent an event, start observing `signal`.
+		signalDisposableAtomic.modify { signalDisposable in
+			if !signalDisposable.disposed {
+				// Forward values and errors, but not completion events, to the observer.
+				signalDisposable.innerDisposable = signal.observe(next: { value in
+					sendNext(observer, value)
+					}, error: { error in
+						sendError(observer, error)
```
kastiglione: The above two lines are indented one too many.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -505,26 +505,42 @@ public func skipWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
-
-		if !signalDisposable.disposed {
-			signalDisposable.innerDisposable = signal.observe(next: { value in
-				sendNext(observer, value)
-			}, error: { error in
-				sendError(observer, error)
+		/// When this disposable is disposed, we'll stop forwarding events from `signal` to
+		/// the observer. It's atomic because we need to defend against the case that the
+		/// replacement signal sends an event while we're starting to observe `signal`.
+		let signalDisposableAtomic = Atomic(SerialDisposable())
```
kastiglione: Hmm… it's a strange composition. I feel like the `SerialDisposable` should be enough here, since it has an internal `Atomic` and is intended to make things related to disposal happen in serial. But the internal `Atomic` holds private implementation, so that's a dead end.

kastiglione:
Thanks for catching this @andymatuschak,

ReactiveCocoa/Swift/Signal.swift
andymatuschak: Ah, I hadn't seen `withValue`. Much nicer, thanks.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: Yeah, the issue is that it doesn't guarantee transactional access to `innerDisposable` and `isDisposed` as a pair.

ReactiveCocoa/Swift/Signal.swift
andymatuschak: No. You're right. I was just being consistent. I'll replace with the original and a comment.

andymatuschak:
Think I've addressed all your comments; thanks.

ReactiveCocoa/Swift/Signal.swift
kastiglione: If we see this pattern more, perhaps `Disposable`s should get something like `ifAlive(f: Void -> Void)`.

kastiglione:
:ship: 

@jspahrsummers would you like a look before merge?

jspahrsummers:
Yes, please. I'll try to get to it today.

jspahrsummers:
Maybe we could get rid of the `Atomic` entirely, and observe the replacement signal _after_ the original. Since we're dealing exclusively with hot signals (unlike [RAC 2.x](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/71f1dbfd1265538d30dae41152c7fd8929f5393c/ReactiveCocoa/RACSignal%2BOperations.m#L756)), the ordering of observation won't affect the behavior of arriving events, right?

andymatuschak:
Actually, yeah, that makes sense to me. So… okay, I think that means this pull request is moot—@kastiglione’s original code is fine. Dave, do you agree?

jspahrsummers:
I don't think the original code is fine, though, because there's a race condition in checking the status of the serial disposable. _If_, however, we reordered the observations, we shouldn't need to inspect the disposable's status.

kastiglione:
If both are truly hot and sending concurrently, then there's a race either
way. If they're both known to not send during function scope (thinking when
being setup by a `SignalProducer`?) then the order doesn't matter. True or
false?

jspahrsummers:
As I see it, the possible races are:

 1. If the observation order is **`replacement` then `signal`**, an event could arrive from `replacement` while we're starting to observe `signal`. Since it would get forwarded immediately, **we could send events from the original signal after those of the replacement**.
 1. If the observation order is **`signal` then `replacement`**, we could miss events from `replacement` while we're still setting up. However, since everything here is a hot signal, this _must_ be acceptable to any consumer—or else they need to guarantee that no events will be sent until the operator function returns.

I think the first race is unacceptable, but the second one is almost inevitable.

neilpa:
Should this be part of the core set of signal operators given the extra constraints required of the signals? There was a similar discussion about the need for this in #956 when it was added to RAC 2.

jspahrsummers:
@neilpa There are no _extra_ constraints required. The same limitations (at least for the second case I described above) apply to all hot signals.

kastiglione:
@jspahrsummers What conditions are you using to divide which operators are acceptable for `Signal` and which are not? Comparing this to `concat`, discussed in #1689, this one also "waits" to observe, though it is an unavoidable and very small wait. With `concat`, the wait is semantically obvious, with `takeUntilReplacement`, the name doesn't obviously indicate a wait (hence the standing race condition from the RAC 2 version). I would describe `takeUntilReplacement` as racey, in that timing can be surprising, where as `concat` perhaps isn't racey by the same "definition", since the timing semantics are explicitly sought. This isn't an argument for `concat`, I'm curious how you're thinking about types of `Signal` operators. I probably should re-read your comments, but I find myself wondering if some of the arguments made against `concat` are being set aside for `takeUntilReplacement`.

neilpa:
@jspahrsummers Fair point. I guess my underlying question is should there be default operators that coordinate events across multiple hot signals? The constraints are the same but what about a user's expectations?

jspahrsummers:
I did have reservations about this operator. Ultimately, however, I think it's in the same category as `sampleOn`.

The observation order within the invocation of _any_ binary operator over `Signal`s is unspecified, and the only way to _guarantee_ your desired ordering is to make sure that the `Signal`s do not send any events until all operator functions have been invoked. This is often easier than it sounds—for example, your UI controls are not going to send events before they're visible, so you can combine them at will.

My specific concern with `concat`, and other operators like it, is that it brings this inherent race condition _outside_ of the operator. If your first signal does not complete before the second signal sends a terminating event, you'll effectively be left with a “dangling” infinite signal. In a case like this, the absence of events while calling `concat` itself _is not a strong enough guarantee_ to avoid the race.

Does that make sense?

jspahrsummers:
Another way to state this: if you set up your signal graph before any events are sent, you won't run into these problems with `takeUntilReplacement`, `sampleOn`, and similar problems. However, combinators like `concat` can _still_ exhibit timing issues, because they are not related to the setup process itself.

andymatuschak:
Agreed with @jspahrsummers. Updated.

jspahrsummers:
Thanks @andymatuschak! :shipit:

kastiglione:
> Does that make sense?

Yep, exactly the kind of perspective I was seeking :pray:

PR #1705, opened 2015-01-26
added TestError

almassapargali:
@jspahrsummers you said [there](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1702#discussion_r23511903) (on #1702) you'd prefer custom error `enum` for tests. Did you mean something like this?

### DIFF
--- a/ReactiveCocoaTests/Swift/TestError.swift
+++ b/ReactiveCocoaTests/Swift/TestError.swift
@@ -15,6 +15,6 @@ enum TestError {
 
 extension TestError: ErrorType {
 	var nsError: NSError {
-		return NSError(domain: "", code: 0, userInfo: nil)
+		return NSError(domain: "org.reactivecocoa.ReactiveCocoa.Tests", code: 0, userInfo: nil)
 	}
 }
--- a/ReactiveCocoaTests/Swift/TestError.swift
+++ b/ReactiveCocoaTests/Swift/TestError.swift
@@ -1,0 +1,26 @@
+//
+//  TestError.swift
+//  ReactiveCocoa
+//
+//  Created by Almas Sapargali on 1/26/15.
+//  Copyright (c) 2015 GitHub. All rights reserved.
+//
+
+import Foundation
+import ReactiveCocoa
+
+enum TestError {
+	case Empty
+	case WithCode(Int)
+}
+
+extension TestError: ErrorType {
+	var nsError: NSError {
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/Swift/TestError.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Created by Almas Sapargali on 1/26/15.
+//  Copyright (c) 2015 GitHub. All rights reserved.
+//
+
+import Foundation
+import ReactiveCocoa
+
+enum TestError {
+	case Empty
+	case WithCode(Int)
```
jspahrsummers: We could probably just start with a `Default` error for now. We don't need multiple options—the main goal is just getting away from `NSError`.

ReactiveCocoaTests/Swift/TestError.swift
```diff
@@ -0,0 +1,26 @@
...
+}
+
+extension TestError: ErrorType {
+	var nsError: NSError {
+		switch self {
+		case .Empty:
+			return NSError(domain: "", code: 0, userInfo: nil)
```
jspahrsummers: Could we use a domain like `ReactiveCocoaTests` or something, so it's clear what happened if this shows up in logs?

jspahrsummers:
This is exactly the idea. Thanks!

ReactiveCocoaTests/Swift/TestError.swift
almassapargali: I though it may be useful to distinguish between many errors. Will remove now.

ReactiveCocoaTests/Swift/TestError.swift
almassapargali: Sure

ReactiveCocoaTests/Swift/TestError.swift
jspahrsummers: Maybe, but we can always add that in the future as needed.

almassapargali:
@jspahrsummers could you check this? I used `"org.reactivecocoa.ReactiveCocoa.Tests"` because this domain form was used all over code. Let me know if you want just `"ReactiveCocoaTests"`.

jspahrsummers:
:+1: That's fine.

almassapargali:
Cool! Thanks!

Issue #1691, opened 2015-01-23
Memory increase without bound in racRecursiveScheduler


JadianZheng:
In my case, I need to do some recurisve thing in the update thread. And I found the memory increse problem without idea to figure it.
Here is the example:
case 1:
```javascript
    [self.dataUpdateScheduler after:[NSDate date] repeatingEvery:1 withLeeway:1 schedule:^{
        NSMutableArray *abc = [NSMutableArray array];
        for (int i = 0; i < 222220; i++) {
            [abc addObject:@"hello"];
        }
    }];
```
case 2:
```javascript
    [self.dataUpdateScheduler scheduleRecursiveBlock:^(void (^reschedule)(void)) {
        NSMutableArray *abc = [NSMutableArray array];
        for (int i = 0; i < 222220; i++) {
            @autoreleasepool {
                [abc addObject:@"hello"];
            }
        }
        reschedule();
    }];
```

Case 1 without memory problem, and 2 has the memory increase problem.
Can anyone help me? Thank you.

jspahrsummers:
Background threads' autorelease pools are generally not drained until the thread goes away. Since `abc` exists in an autorelease pool, that means it may not be destroyed for as long as the thread exists.

If you put the `@autoreleasepool` around the array creation itself, do you still see an issue?

JadianZheng:
I change the example to the follow you say, and the problem still exist.

```
    [self.dataUpdateScheduler scheduleRecursiveBlock:^(void (^reschedule)(void)) {
        @autoreleasepool {
            NSMutableArray *abc = [NSMutableArray array];
            for (int i = 0; i < 222220; i++) {
                [abc addObject:@"hello"];
            }
        }
        
        reschedule();
    }];
```

jspahrsummers:
Do you have a full sample project, or an Instruments trace you can upload?

JadianZheng:
You can new a single view ios project. Replace the view Controller code to the follow. It's just a sample project.

```
#import "ViewController.h"

@interface ViewController ()

@property (nonatomic) RACScheduler *dataUpdateScheduler;

@end

@implementation ViewController

- (void)viewDidLoad {
    [super viewDidLoad];
    // Do any additional setup after loading the view, typically from a nib.
    
    self.dataUpdateScheduler = [RACScheduler schedulerWithPriority:RACSchedulerPriorityDefault name:@"com.RACScheduler.Memory-scheduler"];
    
[...]

PR #1707, opened 2015-01-27
fixing memory leak issue #1691

kkazuo:
add autorelease guard around a scheduled block.

I know that dispatch_async() will manage autorelease pool themselves.
But GCD has no guarantees as when those pools are drained.

https://developer.apple.com/library/ios/documentation/General/Conceptual/ConcurrencyProgrammingGuide/OperationQueues/OperationQueues.html


jspahrsummers:
:+1: Seems reasonable. Thanks!

PR #1710, opened 2015-01-29
Synchronous SignalProducer functions

kastiglione:
Depends on #1719 

This implements `single()`, `last()`, and `wait()` on `SignalProducer`, and the stubbed specs. Also supplies the specs for `first()`, which was already implemented.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -561,24 +561,7 @@ public func zipWith<T, U>(otherSignalProducer: SignalProducer<U>)(producer: Sign
 
 /// Starts the producer, then blocks, waiting for the first value.
 public func first<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
-	let semaphore = dispatch_semaphore_create(0)
-	var result: Result<T, E>?
-
-	producer
-		|> take(1)
-		|> start(next: { value in
-			result = success(value)
-			dispatch_semaphore_signal(semaphore)
-		}, error: { error in
-			result = failure(error)
-			dispatch_semaphore_signal(semaphore)
-		}, completed: {
-			dispatch_semaphore_signal(semaphore)
-			return
-		})
-
-	dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER)
-	return result
+	return producer |> take(1) |> single
 }
 
 /// Starts the producer, then blocks, waiting for events: Next and Completed.
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -558,7 +558,6 @@ public func takeUntilReplacement<T>(replacement: SignalProducer<T>)(producer: Si
 public func then<T, U>(replacement: SignalProducer<U>)(producer: SignalProducer<T>) -> SignalProducer<U>
 public func zipWith<T, U>(otherSignalProducer: SignalProducer<U>)(producer: SignalProducer<T>) -> SignalProducer<(T, U)>
 
-public func single<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>?
 public func last<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>?
 public func wait<T, E>(producer: SignalProducer<T, E>) -> Result<(), E>
 */
@@ -566,13 +565,43 @@ public func wait<T, E>(producer: SignalProducer<T, E>) -> Result<(), E>
 /// Starts the producer, then blocks, waiting for the first value.
 public func first<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
 	let semaphore = dispatch_semaphore_create(0)
-	var result: Result<T, E>? = nil
+	var result: Result<T, E>?
 
 	producer
 		|> take(1)
 		|> start(next: { value in
 			result = success(value)
 			dispatch_semaphore_signal(semaphore)
+		}, error: { error in
+			result = failure(error)
+			dispatch_semaphore_signal(semaphore)
+		}, completed: {
+			dispatch_semaphore_signal(semaphore)
+			return
+		})
+
+	dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER)
+	return result
+}
+
+/// Starts the producer, then blocks, waiting for events: Next and Completed.
+/// When a single value or error is sent, the returned `Result` will represent
+/// those cases. However, when no values are sent, or when more than one value
+/// is sent, `nil` will be returned.
+public func single<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
+	let semaphore = dispatch_semaphore_create(0)
+	var result: Result<T, E>?
+
+	producer
+		|> take(2)
+		|> start(next: { value in
+			if result != nil {
+				// Move into failure state after recieving another value.
+				result = nil
+				return
+			}
+
+			result = success(value)
 		}, error: { error in
 			result = failure(error)
 			dispatch_semaphore_signal(semaphore)
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -596,7 +596,7 @@ public func single<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
 
 /// Starts the producer, then blocks, waiting for the last value.
 public func last<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
-	return producer |> takeLast(1) |> first
+	return producer |> takeLast(1) |> single
 }
 
 /// Starts the producer, then blocks, waiting for completion.
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -618,19 +618,8 @@ public func last<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
 
 /// Starts the producer, then blocks, waiting for completion.
 public func wait<T, E>(producer: SignalProducer<T, E>) -> Result<(), E> {
-	let semaphore = dispatch_semaphore_create(0)
-	var result: Result<(), E>!
-
-	producer.start(error: { error in
-		result = failure(error)
-		dispatch_semaphore_signal(semaphore)
-	}, completed: {
-		result = success(())
-		dispatch_semaphore_signal(semaphore)
-	})
-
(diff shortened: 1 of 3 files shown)


ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,46 +233,129 @@ class SignalProducerSpec: QuickSpec {
...
+
+				var result: Result<(), NoError>?
+				dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)) {
+					result = producer |> wait
+				}
+				expect(sink).toEventuallyNot(beNil())
+				expect(result).to(beNil())
+
+				sendCompleted(sink)
+				expect(result?.value).toNot(beNil())
```
kastiglione: I couldn't do `equal()` on `()`, so this checks non-`nil`. Unless there's a way to do a matcher for `()`?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -585,6 +581,75 @@ public func first<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
...
+}
+
+/// Starts the producer, then blocks, waiting for the last value.
+public func last<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
+	let semaphore = dispatch_semaphore_create(0)
+	var result: Result<T, E>?
+
+	producer
+		|> takeLast(1)
+		|> start(next: { value in
```
jspahrsummers: This can just use `first()`.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -585,6 +581,75 @@ public func first<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
...
+	dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER)
+	return result
+}
+
+/// Starts the producer, then blocks, waiting for completion.
+public func wait<T, E>(producer: SignalProducer<T, E>) -> Result<(), E> {
+	let semaphore = dispatch_semaphore_create(0)
+	var result: Result<(), E>!
+
+	producer.start(error: { error in
```
jspahrsummers: This could probably be implemented in terms of one of the others, if you're interested in trying it.

jspahrsummers:
:dancer:

ReactiveCocoa/Swift/SignalProducer.swift
kastiglione: :ok_hand:

kastiglione:
Still aiming to implement `first()` in terms of `single()` using:

```swift
return producer |> take(1) |> single
```

But that causes `first()`'s tests to fail. Will continue to look into it.

kastiglione:
:lollipop:  

Turns out `take()` was completing after `count+1` nexts. So that's fixed. However, `take(0)` has a strange behavior where it won't complete until a value is sent, even though it should be able to complete right away. My mind is still thinking of RAC 2 signals, so I tried to fix this by introducing a `Signal.empty` and returning that when given a `count` of zero, but of course that's not gonna work. I've left that issue alone.

kastiglione:
Apologies for the commit thrash.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -585,6 +594,17 @@ public func first<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
...
 }
 
+/// Starts the producer, then blocks, waiting for the last value.
+public func last<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
+	return producer |> takeLast(1) |> single
+}
+
+/// Starts the producer, then blocks, waiting for completion.
+public func wait<T, E>(producer: SignalProducer<T, E>) -> Result<(), E> {
+	let result = producer |> map { _ in () } |> last
```
jspahrsummers: You could use a `concat` with a producer of one value `()` instead of the `map` here. That would also avoid the need for `??` below, I think.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,46 +233,129 @@ class SignalProducerSpec: QuickSpec {
 		}
 
 		describe("first") {
-			pending("should start a signal then block on the first value") {
+			it("should start a signal then block on the first value") {
+				var sink: Signal<Int, NoError>.Observer!
+				let producer = SignalProducer<Int, NoError> { observer, _ in
```
jspahrsummers: Can we use `SignalProducer.buffer` instead, to avoid this weird saving of the observer?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,46 +233,129 @@ class SignalProducerSpec: QuickSpec {
...
+					sink = observer
+				}
+				expect(sink).to(beNil())
+
+				var result: Result<Int, NoError>?
+				dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)) {
```
jspahrsummers: This can be removed with the [use of `buffer`](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1710/files#r24038113).

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,46 +233,129 @@ class SignalProducerSpec: QuickSpec {
...
 
-			pending("should return a nil result if no values are sent before completion") {
+			it("should return a nil result if no values are sent before completion") {
+				let result = SignalProducer<Int, NoError>.empty |> single
+				expect(result).to(beNil())
 			}
 
-			pending("should return a nil result if too many values are sent before completion") {
+			it("should return a nil result if more than one value is sent before completion") {
+				let result = SignalProducer<Int, NoError>(values: [1, 2]) |> single
```
jspahrsummers: Can you put spaces inside array literals?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,46 +233,129 @@ class SignalProducerSpec: QuickSpec {
...
+			it("should return a nil result if no values are sent before completion") {
+				let result = SignalProducer<Int, NoError>.empty |> first
+				expect(result).to(beNil())
 			}
 
-			pending("should return an error if one occurs before the first value") {
+			it("should return an error if one occurs before the first value") {
+				let result = SignalProducer<Int, TestError>(error: .Default) |> first
+				expect(result?.error).to(equal(TestError.Default))
 			}
```
jspahrsummers: Can you also test `first` over multiple values?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,46 +233,129 @@ class SignalProducerSpec: QuickSpec {
...
+			it("should return a nil result if no values are sent before completion") {
+				let result = SignalProducer<Int, NoError>.empty |> last
+				expect(result).to(beNil())
 			}
 
-			pending("should return an error if one occurs") {
+			it("should return an error if one occurs") {
+				let result = SignalProducer<Int, TestError>(error: .Default) |> last
+				expect(result?.error).to(equal(TestError.Default))
 			}
```
jspahrsummers: Can you also test `last` over multiple values?

jspahrsummers:
:house_with_garden:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
kastiglione: By pre-filling the buffer before calling `first` etc? If so, the test won't really be testing the "then block" part of the spec description.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
kastiglione: Addressed in 4911bad.

ReactiveCocoa/Swift/SignalProducer.swift
kastiglione: :+1:  (May have to depend on #1714 for the use of `concat`)

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,46 +233,119 @@ class SignalProducerSpec: QuickSpec {
 		}
 
 		describe("first") {
-			pending("should start a signal then block on the first value") {
+			it("should start a signal then block on the first value") {
+				let (producer, sink) = SignalProducer<Int, NoError>.buffer()
+
+				var result: Result<Int, NoError>?
+				dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)) {
```
kastiglione: @jspahrsummers [You requested this be removed](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1710#discussion_r24038166), but since the comment was hidden with updates I'm re-commenting on this line.

My question back:

> By pre-filling the buffer before calling `first` etc? If so, the test won't really be testing the "then block " part of the spec description.

jspahrsummers: Yeah, that's fair. Can we make this more deterministic, though? Maybe by using a dispatch group that we wait upon?

ReactiveCocoa/Swift/SignalProducer.swift
kastiglione: > That would also avoid the need for `??` below, I think.

Don't think so, `concat`ing a static value doesn't make the type system statically know there is a value, so the `last` operator still returns an optional.

kastiglione: I wouldn't be able to use `concat`, but could use `then`. If the `??` is still needed, is swapping `map` for `then` much of an improvement?

kastiglione: I could get rid of the `??` if we want to force unwrap the result, since *we* know the signal will produce a value, even if the *compiler* doesn't.

jspahrsummers: I think `then` plus force-unwrapping would be ideal, since those most closely reflect our intended semantics.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
kastiglione: Addressed in 4911bad.

kastiglione:
:arrows_counterclockwise:

jspahrsummers:
:twisted_rightwards_arrows:

PR #1714, opened 2015-01-31
Fix, test, and refactor concat()

kastiglione:
This fixes a crash in a `concat()` in cases where a signals completes before following signals are added to the queue. The crash happens while attempting to dequeue from the empty queue. This arose when doing a binary `concat` with the first signal being `empty`, which completes synchronously before the second signal is handled on the outer signal.

EDIT: Turned into a bunch of refactoring. See https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1714#issuecomment-73197857 for description.

cc @andymatuschak

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -530,7 +530,7 @@ private struct ConcatState<T, E: ErrorType> {
 	/// Indicates whether the most recently processed inner signal has completed yet.
 	var latestSignalCompleted = true
 
-	/// Determination `concat` completion in its entirety.
+	/// Determines `concat` completion in its entirety.
 	var isComplete: Bool {
 		return selfCompleted && latestSignalCompleted
 	}
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -463,22 +463,12 @@ private final class ConcatState<T, E: ErrorType> {
 	/// The top level disposable of a started `concat` producer.
 	let disposable: CompositeDisposable
 
-	var spinlock = OS_SPINLOCK_INIT
+	/// The signals waiting to be started.
+	var queuedSignalProducers: [SignalProducer<T, E>] = []
 
 	init(observer: Signal<T, E>.Observer, disposable: CompositeDisposable) {
 		self.observer = observer
 		self.disposable = disposable
-	}
-
-	/// The signals waiting to be started.
-	var queuedSignalProducers: [SignalProducer<T, E>] = []
-
-	func lock() {
-		withUnsafeMutablePointer(&spinlock, OSSpinLockLock)
-	}
-
-	func unlock() {
-		withUnsafeMutablePointer(&spinlock, OSSpinLockUnlock)
 	}
 
 	func enqueueSignalProducer(producer: SignalProducer<T, E>) {
@@ -522,6 +512,10 @@ private final class ConcatState<T, E: ErrorType> {
 			}
 		})
 	}
+
+	var spinlock = OS_SPINLOCK_INIT
+	func lock() { withUnsafeMutablePointer(&spinlock, OSSpinLockLock) }
+	func unlock() { withUnsafeMutablePointer(&spinlock, OSSpinLockUnlock) }
 }
 
 /// `concat`s `next` onto `producer`.
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -438,22 +438,17 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 	return SignalProducer { observer, disposable in
 		let state = ConcatState(observer: observer, disposable: disposable)
 
+		let completion: SignalProducer<T, E> = .empty |> on(completed: {
+			sendCompleted(observer)
+		})
+
 		producer.startWithSignal { signal, signalDisposable in
 			signal.observe(next: { innerSignalProducer in
 				state.enqueueSignalProducer(innerSignalProducer)
 			}, error: { error in
 				sendError(observer, error)
 			}, completed: {
-				var isComplete = false
-
-				state.atomic.modify {
-					state.selfCompleted = true
-					isComplete = state.isComplete
-				}
-
-				if isComplete {
-					sendCompleted(observer)
-				}
+				state.enqueueSignalProducer(completion)
 			})
 				
 			disposable.addDisposable(signalDisposable)
@@ -475,22 +470,17 @@ private func startNextSignalProducer<T, E>(signalProducer: SignalProducer<T, E>,
 		sendError(state.observer, error)
 	}, completed: {
 		var nextSignalProducer: SignalProducer<T, E>?
-		var isComplete = false
 
 		serialDisposableCompositeHandle.remove()
 		state.atomic.modify {
 			state.latestSignalCompleted = true
-			if state.queuedSignalProducers.isEmpty {
-				isComplete = state.isComplete
-			} else {
+			if !state.queuedSignalProducers.isEmpty {
 				nextSignalProducer = state.queuedSignalProducers.removeAtIndex(0)
 			}
 		}
 		
 		if let nextSignalProducer = nextSignalProducer {
 			startNextSignalProducer(nextSignalProducer, state)
-		} else if isComplete {
-			sendCompleted(state.observer)
 		}
 	})
 }
@@ -509,16 +499,8 @@ private final class ConcatState<T, E: ErrorType> {
 
 	let atomic = Atomic()
 
-	/// Whether the signal-of-signals has completed yet.
-	var selfCompleted = false
-	
 	/// Indicates whether the most recently processed inner signal has completed yet.
 	var latestSignalCompleted = true
-
-	/// Determines `concat` completion in its entirety.
-	var isComplete: Bool {
-		return selfCompleted && latestSignalCompleted
-	}
 
 	/// The signals waiting to be started.
 	var queuedSignalProducers: [SignalProducer<T, E>] = []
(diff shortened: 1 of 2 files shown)


kastiglione:
Found and fixed a logic bug where the producer returned from `concat()` would complete even when producers are still in the queue.

kastiglione:
I'm going to fill out the test stubs too…

kastiglione:
Ready for review.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -475,10 +475,10 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 				serialDisposableCompositeHandle.remove()
 				state.modify { (var state) in
 					state.latestSignalCompleted = true
-					let outerSignalIsComplete = completeIfAllowed(state)
-					if !outerSignalIsComplete {
-						nextSignalProducer = state.queuedSignalProducers[0]
-						state.queuedSignalProducers.removeAtIndex(0)
+					if state.queuedSignalProducers.isEmpty {
+						completeIfAllowed(state)
```
jspahrsummers: I know this wasn't from your changes, but can you refactor this operator so that all invocations of `completeIfAllowed` occur _outside_ of the `modify` block? Otherwise, we'll be holding the spinlock for way longer than intended, and could even deadlock.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -158,16 +158,126 @@ class SignalProducerSpec: QuickSpec {
 		}
 
 		describe("concat") {
-			pending("should start subsequent inner signals upon completion") {
+			context("sequencing") {
```
jspahrsummers: Can you please use `describe` instead of `context`, for consistency?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -158,16 +158,126 @@ class SignalProducerSpec: QuickSpec {
...
+				var previousSink: Signal<Int, NoError>.Observer!
+				var previousProducer: SignalProducer<Int, NoError>!
+
+				var subsequentSink: Signal<Int, NoError>.Observer!
+				var subsequentProducer: SignalProducer<Int, NoError>!
+				var subsequentStarted = false
+
+				beforeEach {
+					outerProducer = SignalProducer { observer, _ in
+						outerSink = observer
```
jspahrsummers: Can this use `SignalProducer.buffer` or something instead?

jspahrsummers:
Made a few notes which might significantly change the code, so I'll pause here.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
kastiglione: Just for this use of `context`, or the other use I had too?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
kastiglione: Aaaaaah… I wondered if there was something analogous to `pipe` for `SignalProducer`, yet I've skipped over `buffer` dozens of times on the name alone :see_no_evil:

ReactiveCocoa/Swift/SignalProducer.swift
kastiglione: > could even deadlock.

Is that for cycles? Or other ways too?

kastiglione:
:100:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -443,18 +443,12 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
...
-				
-				// A strong reference is held to `subscribeToSignalProducer` until
-				// completion, preventing it from deallocating early.
-				subscribeToSignalProducer = nil
-
-				return true
-			} else {
-				return false
-			}
+		let complete💯: Void -> Void = {
```
jspahrsummers: nopenopenope

jspahrsummers:
:angry:

ReactiveCocoa/Swift/SignalProducer.swift
jspahrsummers: Criss-crossing signals as well.

kastiglione:
> :angry: 

Sorry, eh :disappointed: 

I saw an email with:

> Criss-crossing signals as well.

Was that related to an older revision?

kastiglione:
Did a big 'ol bunch of refactoring. Kinda like how it's turned out so far, but eager to hear feedback.

Table of Contents, with more info below.

1. Wanted to get rid of the recursive closure and the retain cycle uncertainty.
1. Wanted to make the atomic operations on `ConcatState` its own responsibility.
1. Got rid of the boolean pair `selfCompleted` and `latestSignalCompleted`.

Re 1: With the `Z` combinator, I wasn't entirely sure the `completeIfAllowed` closure needed to strongly capture the `subscribeToSignalProducer` block, nor that it needed to `nil` it out. I _think_ the recursion and curried nature of `Z()` took care of, but instead of proving this, I figured it would be simpler to make `subscribeToSignalProducer` a free function that had no free variables. Then it could refer to itself no problem. To simplify calling `subscribeToSignalProducer`, I added some immutable state to `ConcatState`, specifically the concat observer and disposable.

Re 2: Not much to say on this, check out locking inside `ConcatState`.

Re 3: I quite liked how this played out. In order to get rid of the `selfCompleted` flag, I made it so that upon completion of the outer producer, one more inner producer is added to the queue. This last producer is just `.empty` with an `on(:completed)` side effect of calling `sendCompleted(observer)`. Next, to get rid of the `latestSignalCompleted` flag, I changed the semantics of the queue such that the active producer stays in the queue until it is completed. Thus, if the queue is empty, then the last producer *has* completed, and if it is not empty, then the last producer *has not* completed.

The result is that this implementation moved logic from being mostly inline in the `SignalProducer` definitions, to being responsibility of `ConcatState`.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -521,21 +458,71 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
...
+		self.disposable = disposable
+	}
+
+	func enqueueSignalProducer(producer: SignalProducer<T, E>) {
+		var shouldStart = true
+
+		lock()
+		// An empty queue means the concat is idle, ready & waiting to start the
+		// next producer given.
+		shouldStart = self.queuedSignalProducers.isEmpty
```
jspahrsummers: Can you please remove references to `self` that are unnecessary?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -521,21 +458,71 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
...
+
+			if let nextSignalProducer = self.dequeueSignalProducer() {
+				self.startNextSignalProducer(nextSignalProducer)
+			}
+		})
+	}
+
+	var spinlock = OS_SPINLOCK_INIT
+	func lock() { withUnsafeMutablePointer(&spinlock, OSSpinLockLock) }
+	func unlock() { withUnsafeMutablePointer(&spinlock, OSSpinLockUnlock) }
```
jspahrsummers: Can we wrap the array in an `Atomic` instead of doing this?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -521,21 +458,71 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
...
+
+		return nextSignalProducer
+	}
+
+	/// Subscribes to the given signal producer.
+	func startNextSignalProducer(signalProducer: SignalProducer<T, E>) {
+		let serialDisposable = SerialDisposable()
+		let serialDisposableCompositeHandle = disposable.addDisposable(serialDisposable)
+
+		serialDisposable.innerDisposable = signalProducer.start(next: { value in
```
jspahrsummers: If you use `startWithSignal`, the serial disposable should be unnecessary.

jspahrsummers:
Very nice! Just a few more notes.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -521,21 +458,68 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
...
+			return queue
+		}
+
+		return nextSignalProducer
+	}
+
+	/// Subscribes to the given signal producer.
+	func startNextSignalProducer(signalProducer: SignalProducer<T, E>) {
+		signalProducer.startWithSignal { signal, disposable in
+			self.disposable.addDisposable(disposable)
```
kastiglione: @jspahrsummers Does this do it?

kastiglione:
:ferris_wheel:

ReactiveCocoa/Swift/SignalProducer.swift
jspahrsummers: Yep!

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -521,21 +458,68 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
...
 
-/// The Z combinator, which we use to make a recursive closure that we can
-/// nil out to avoid a retain cycle.
-private func Z<T, U>(f: (T -> U, T) -> U)(x: T) -> U {
-	return f(Z(f), x)
+	/// The top level disposable of a started `concat` producer.
+	let disposable: CompositeDisposable
+
+	/// The active producer, if any, and the producers waiting to be started.
+	var queuedSignalProducers: Atomic<[SignalProducer<T, E>]> = Atomic([])
```
jspahrsummers: Can you make this `let`, since it's an immutable reference to mutable data?

jspahrsummers:
Just one more note, and then feel free to merge this yourself. :roller_coaster:

PR #1716, opened 2015-01-31
Reimplement throttle out of map, delay, switchToLatest

adly-holler:
Hi folks, I really love your work! Hope you can use this.

### DIFF
--- a/ReactiveCocoa/RACSignal+Operations.m
+++ b/ReactiveCocoa/RACSignal+Operations.m
@@ -141,14 +141,14 @@ static RACDisposable *subscribeForever (RACSignal *signal, void (^next)(id), voi
 	NSCParameterAssert(predicate != nil);
 
 	return [[[self map:^(id value) {
-        if (predicate(value)) {
-            return [[RACSignal return:value] delay:interval];
-        } else {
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/RACSignal+Operations.m
```diff
@@ -140,58 +140,15 @@ - (RACSignal *)throttle:(NSTimeInterval)interval valuesPassingTest:(BOOL (^)(id
...
-				hasNextValue = YES;
-				nextDisposable.disposable = [delayScheduler afterDelay:interval schedule:^{
-					flushNext(YES);
-				}];
-			}
-		} error:^(NSError *error) {
-			[compoundDisposable dispose];
-			[subscriber sendError:error];
-		} completed:^{
-			flushNext(YES);
```
kastiglione: The current implementation flushes the last throttled value on completion. The approach below does not, it will wait the full delay.

kastiglione:
Thanks for the pull request. I've had a bit of a look and made one comment. Composing out of other operators is nice, if it can be made to work.

adly-holler:
Hmm I think implementing that behavior would probably cost too much concision to be worth it. =(

PR #1717, opened 2015-01-31
implemented takeWhile()

almassapargali:


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -551,10 +551,27 @@ public func takeLast<T,E>(count: Int)(signal: Signal<T,E>) -> Signal<T,E> {
 	}
 }
 
+/// Forwards any values from `signal` until `predicate` returns false,
+/// at which point the returned signal will complete.
+public func takeWhile<T,E>(predicate: T -> Bool)(signal: Signal<T,E>) -> Signal<T,E> {
```
kastiglione: I've seen both `<T,E>` and `<T, E>`. It seems we're using the latter but nothing is documented in the [style guide](https://github.com/github/swift-style-guide).

kastiglione:
Since there are not tests for `takeWhile` in #1702, what do you think about adding them here?

ReactiveCocoa/Swift/Signal.swift
almassapargali: Thanks, didn't know about that style guide.

almassapargali:
I added them, then commented out ([this commit](https://github.com/almassapargali/ReactiveCocoa/commit/637282222f659169e745134e343fe5cb3b4ee8f6)). Wanted to test, continuously getting Segmentation fault: 11. Can't figure out what is causing it. Error logs points to clearly correct code.

ReactiveCocoa/Swift/Signal.swift
kastiglione: To be clear, I'm saying it doesn't state one way or the other. But currently, mostly there's a space between generic type parameters.

almassapargali: Oh, I just looked at functions around whether to leave space, and it turns out only `takeLast()` which is preceding didn't have it. 😁

almassapargali: @kastiglione it's ok to add spaces on other functions too on this branch?

kastiglione: I think it's fine.

kastiglione:
I just tried, not sure *why* the problem exists, but in my experimenting, I isolated the problem to using the free function version of `observe()`. Replacing it with the `observe()` method allows it to compile. I put up a pull request to add those tests, it depends on this pull request.

jspahrsummers:

PR #1718, opened 2015-01-31
Implement stubbed takeWhile() tests

kastiglione:
Depends on #1717 

As mentioned in #1717, these tests use `observe()` as a method, because the free function `observe()` crashes swiftc.

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -152,9 +152,11 @@ class SignalSpec: QuickSpec {
 					completed = true
 				})
 
-				sendNext(observer, 4)
-				expect(latestValue).to(equal(4))
-				expect(completed).to(beFalse())
+				for value in -1...4 {
+					sendNext(observer, value)
+					expect(latestValue).to(equal(value))
+					expect(completed).to(beFalse())
+				}
 
 				sendNext(observer, 5)
 				expect(latestValue).to(equal(4))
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -133,10 +133,43 @@ class SignalSpec: QuickSpec {
...
+
+				signal.observe(next: { value in
+					latestValue = value
+				}, completed: {
+					completed = true
+				})
+
+				sendNext(observer, 4)
+				expect(latestValue).to(equal(4))
+				expect(completed).to(beFalse())
```
jspahrsummers: Can you please test at least one more value that fails the predicate?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -133,10 +133,43 @@ class SignalSpec: QuickSpec {
...
 
-			pending("should complete if the predicate starts false") {
+			it("should complete if the predicate starts false") {
+				var completed = false
+
+				signal.observe(completed: {
+					completed = true
+				})
+
+				sendNext(observer, 5)
```
jspahrsummers: Can we verify that the value is not passed through in this case?

jspahrsummers:
:sparkler:

ReactiveCocoaTests/Swift/SignalSpec.swift
kastiglione: :+1:

ReactiveCocoaTests/Swift/SignalSpec.swift
kastiglione: Fails, or passes? Just not sure since this comment is on the section for a value that passes, and I also don't know what other kind of failing value you tested here,

jspahrsummers: Sorry, I meant another value which will get passed through.

kastiglione:
:radio:

jspahrsummers:
:gem:

PR #1719, opened 2015-02-01
Add then() SignalProducer

kastiglione:
Depends on #1714 ([Compare view](https://github.com/ReactiveCocoa/ReactiveCocoa/compare/concat-nada...then))

This adds `then()` for `SignalProducer`.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -475,8 +475,9 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 				serialDisposableCompositeHandle.remove()
 				state.modify { (var state) in
 					state.latestSignalCompleted = true
-					let outerSignalIsComplete = completeIfAllowed(state)
-					if !outerSignalIsComplete && !state.queuedSignalProducers.isEmpty {
+					if state.queuedSignalProducers.isEmpty {
+						completeIfAllowed(state)
+					} else {
 						nextSignalProducer = state.queuedSignalProducers.removeAtIndex(0)
 					}
 					return state
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -567,14 +567,16 @@ public func wait<T, E>(producer: SignalProducer<T, E>) -> Result<(), E>
 /// which case `replacement` will not be started, and none of its events will be
 /// be forwarded. All values sent from `producer` are ignored.
 public func then<T, U, E>(replacement: SignalProducer<U, E>)(producer: SignalProducer<T, E>) -> SignalProducer<U, E> {
-	let relay = SignalProducer<U, E> { observer, disposable in
-		let producerDisposable = producer.start(error: { error in
-			sendError(observer, error)
-		}, completed: {
-			sendCompleted(observer)
-		})
+	let relay = SignalProducer<U, E> { observer, observerDisposable in
+		producer.startWithSignal { signal, signalDisposable in
+			observerDisposable.addDisposable(signalDisposable)
 
-		disposable.addDisposable(producerDisposable)
+			signal.observe(error: { error in
+				sendError(observer, error)
+			}, completed: {
+				sendCompleted(observer)
+			})
+		}
 	}
 
 	return relay |> concat(replacement)
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -476,9 +476,8 @@ public func concat<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 				state.modify { (var state) in
 					state.latestSignalCompleted = true
 					let outerSignalIsComplete = completeIfAllowed(state)
-					if !outerSignalIsComplete {
-						nextSignalProducer = state.queuedSignalProducers[0]
-						state.queuedSignalProducers.removeAtIndex(0)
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -218,17 +218,86 @@ class SignalProducerSpec: QuickSpec {
 			}
 		}
 
+		describe("ignoreValues") {
+			it("should not forward values from original producer") {
+				let producer: SignalProducer<Int, NoError> = SignalProducer<Int, NoError>(values: [ 1, 2 ]) |> ignoreValues
```
kastiglione: The typing here is unfortunate. Is there a way to have it to use the obvious default?

almassapargali: What if you change function declaration [here](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1719/files#diff-48c01e61977d5d46be16565ecfb017e2R566) from:
```
public func ignoreValues<T, U, E>(producer: SignalProducer<T, E>) -> SignalProducer<U, E>
```
to:
```
public func ignoreValues<T, E>(producer: SignalProducer<T, E>) -> SignalProducer<T, E>
```

kastiglione: That would probably do the trick for these lines in the test, but would break `then()`, which has the takes a `SignalProducer<T, E>` and returns a `SignalProducer<U, E>`.

almassapargali: Yep, nevermind. I was just in the scope of current test. Btw, I still think changing it may be good idea, since it doesn't really introduces new type.

jspahrsummers:
I don't like having `ignoreValues`. Isn't it just `filter { _ in false }`?

kastiglione:
> I don't like having `ignoreValues`.

We got rid of `then` in RAC 2, what have you learned since then? I think of `ignoreValues` as having at least slightly more utility than `then`, it can be used with, say, `takeUntil`.

> Isn't it just `filter { _ in false }`?

Depends how it's being composed. In this case, I first tried to implement `then` with that catch all `filter`, but that returns a producer of type `T` but the replacement has type `U`. Since I was going to need to "[reconstitute](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1719/files#diff-48c01e61977d5d46be16565ecfb017e2R567)" a producer of type `U`, I figured this type filtering+retyping was just annoying enough to validate the need for `ignoreValues`. Is there a way to use `filter` and then cast to a producer of type `U`? Pretty sure I tried `as`.

jspahrsummers:
> We got rid of `then` (but kept `ignoreValues`) in RAC 2, what have you learned since?

We need `then` because there's no easy way to join operations of different types. It doesn't matter in Objective-C (where you can use `ignoreValues` plus `concat`) because everything is `id`. :trollface:

With response to the rest of your question, you're right that `ignoreValues` isn't just `filter`—however, it is still easy:

```swift
producerToIgnore |> then(.empty)
```

kastiglione:
> ```swift
> producerToIgnore |> then(.empty)
> ```

I find this to be super not obvious at *first glance*, but I'll leave out `ignoreValues` anyway because if it is a problem, someone will come asking.

ps. I didn't realize you could do `.empty` like that, cool.

kastiglione:
:recycle:

kastiglione:
Note at the bottom, this depends on #1714, which is not yet resolved.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -563,6 +562,24 @@ public func last<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>?
 public func wait<T, E>(producer: SignalProducer<T, E>) -> Result<(), E>
 */
 
+/// Waits for completion of `producer`, *then* forwards all events from
+/// `replacement`. Any error sent from `producer` is forwarded immediately, in
+/// which case `replacement` will not be started, and none of its events will be
+/// be forwarded. All values sent from `producer` are ignored.
+public func then<T, U, E>(replacement: SignalProducer<U, E>)(producer: SignalProducer<T, E>) -> SignalProducer<U, E> {
+	let relay = SignalProducer<U, E> { observer, disposable in
+		let producerDisposable = producer.start(error: { error in
```
jspahrsummers: Can this use `startWithSignal`?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -219,16 +219,54 @@ class SignalProducerSpec: QuickSpec {
...
+				let (original, sink) = SignalProducer<Int, NoError>.buffer()
+
+				var subsequentStarted = false
+				let subsequent = SignalProducer<Int, NoError> { observer, _ in
+					subsequentStarted = true
+				}
+
+				let producer = original |> then(subsequent)
+				producer.start()
+				expect(subsequentStarted).to(beFalse())
```
jspahrsummers: Can you please use `beFalsy` and `beTruthy` for greater backwards- and cross-compatibility?

jspahrsummers:
:telephone_receiver:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
kastiglione: Where do I look to learn about the compatibility issues?

kastiglione:
:banana:

jspahrsummers:
Looks good! Please feel free to self-merge after #1714 is in.

PR #1723, opened 2015-02-02
RFC: Overload operators for NoError

kastiglione:
Depends on #1710 ([Compare](https://github.com/ReactiveCocoa/ReactiveCocoa/compare/producer-sync...noerror-sync-funcs))

What do we think about operators being overloaded for `NoError`? Is the convenience “easy” at the cost of “simple”?

This pull request has some examples for the blocking methods in #1710 `{first,single,last,wait}`, but there may be other cases where overloading for `NoError` will make for calling code with less type system appeasing.

For example:

```swift
func first<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>?
// and
func first<T, E: ErrorType>(producer: SignalProducer<T, NoError>) -> T?
```

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -174,10 +174,10 @@ public func take<T, E>(count: Int)(signal: Signal<T, E>) -> Signal<T, E> {
 			if taken < count {
 				taken++
 				sendNext(observer, value)
+			}
 
-				if taken == count {
-					sendCompleted(observer)
-				}
+			if taken == count {
+				sendCompleted(observer)
 			}
 		}, error: { error in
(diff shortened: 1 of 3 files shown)


jspahrsummers:
Overloaded return types are quite painful with Swift today. They require a lot of manual type annotations IME.

I also don't think users should have to change their code to change a `NoError` signal into one that can error (though the reverse makes sense). For example, I don't want to add and remove `.value` after `first()` when that parameter changes.

kastiglione:
:ok:

jspahrsummers:
I do really appreciate you opening this pull request, and the thinking & motivation behind it. Thanks! :heart:

kastiglione:
:gem:

PR #1728, opened 2015-02-03
implementations of <~ for producers and poperties

kaybutter:
This is my attempt to fix the <~ Operator. As you can see I simply reuse the signal version of the operator. This implementation doesn't crash the compiler. 

I hope I handled disposal right. :)

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -132,16 +132,11 @@ public func <~ <T>(property: MutableProperty<T>, signal: Signal<T, NoError>) -> 
 /// The binding will automatically terminate when the property is deinitialized,
 /// or when the created signal sends a `Completed` event.
 public func <~ <T>(property: MutableProperty<T>, producer: SignalProducer<T, NoError>) -> Disposable {
-	let disposable = CompositeDisposable()
-	let propertyDisposable = property.producer.start(completed: {
-		disposable.dispose()
-	})
-	
-	disposable.addDisposable(propertyDisposable)
+	var disposable: Disposable!
 	
 	producer.startWithSignal { signal, signalDisposable in
-		disposable.addDisposable(signalDisposable)
 		property <~ signal
+		disposable = signalDisposable
 	}
 	
 	return disposable
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -137,6 +137,10 @@ public func <~ <T>(property: MutableProperty<T>, producer: SignalProducer<T, NoE
 	producer.startWithSignal { signal, signalDisposable in
 		property <~ signal
 		disposable = signalDisposable
+		
+		property.producer.start(completed: {
+			signalDisposable.dispose()
+		})
 	}
 	
 	return disposable
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -124,8 +124,6 @@ public func <~ <T>(property: MutableProperty<T>, signal: Signal<T, NoError>) -> 
 	return disposable
 }
 
-/*
-FIXME: DISABLED DUE TO COMPILER BUG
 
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -138,46 +136,22 @@ public func <~ <T>(property: MutableProperty<T>, producer: SignalProducer<T, NoE
 	let propertyDisposable = property.producer.start(completed: {
```
jspahrsummers: I think you could move this into the `startWithSignal` closure below, then simply dispose of `signalDisposable` upon completion. That would avoid the need to allocate an additional `CompositeDisposable`.

jspahrsummers:
I think this will work. Thanks!

Could you please also try enabling @JaviSoto's commented-out tests for `<~`?

ReactiveCocoa/Swift/Property.swift
kaybutter: I can move that inside the closure, but which `Disposable` should the method return if I don't create a `CompositeDisposable`? The `startWithSignal` method doesn't return one, or should I pass the one from the closure into a mutable variable outside of it? Doesn't look that clean either.

jspahrsummers: Ah, yeah, just save the disposable given to `startWithSignal`. I know it's a little unclean, but it's totally valid—the only reason that method doesn't itself return a `Disposable` is because I thought that'd confuse its purpose too much.

kaybutter:
Apparently most of that disposable magic was redundant. Please review again :)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -134,50 +132,21 @@ FIXME: DISABLED DUE TO COMPILER BUG
...
 	producer.startWithSignal { signal, signalDisposable in
-		disposable.addDisposable(signalDisposable)
-
-		signal.observe(next: { [weak property] value in
-			property?.value = value
-			return
-		}, completed: {
-			disposable.dispose()
-		})
+		property <~ signal
```
jspahrsummers: Although this implementation is much simpler (kudos!), there's unfortunately a bit of a problem right here.

If the property deinitializes, and the returned disposable is _not_ disposed, the producer's work may not be cancelled. In order for that to happen, we need to make sure that `signalDisposable` is disposed when `property.producer` completes.

The important bit is that it doesn't need to be put into any kind of serial or composite disposable, though.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -136,98 +136,98 @@ class PropertySpec: QuickSpec {
 
 			describe("from a SignalProducer") {
 				pending("should start a signal and update the property with its values") {
```
jspahrsummers: Can you please change these from `pending` to `it`? :smile:

jspahrsummers:
This is looking really good! Just a few more notes.

ReactiveCocoa/Swift/Property.swift
kaybutter: But if the property deinitializes, the inner <~ will take care of disposing the Signal. Shouldn't that be enough?

kaybutter: Okay, now that the tests are actually enabled they seem to agree with you.

ReactiveCocoaTests/Swift/PropertySpec.swift
kaybutter: Oh. How did I miss that 😳

jspahrsummers:
Seems good to me. Is this ready to be merged, or were you still working on anything?

ReactiveCocoa/Swift/Property.swift
jspahrsummers: The reason is, disposing an _observation_ to a `Signal` is not the same as disposing the work of the `SignalProducer` that created it. Signals can have N observers, so the disposal of any of them shouldn't affect delivery to the others.

kaybutter:
Nope. Ready to go 👍

Issue #1396, opened 2014-07-01
Eager throttle


olegam:
I need to operate on a signal with the following behavior:
1. If a `next` is received and it did not send a `next` for the last `interval` seconds it should be forwarded immediately.
2. If a `next` is received and another next was just sent it will be queued and sent `interval` seconds after the previous one unless a new `next` is received before that.
3. If another `next` is received while one is waiting the first one is discarded and the latest will be sent `interval` seconds after the last `next` was sent.

The signature could be like:

```
- (RACSignal *)eagerThrottle:(NSTimeInterval)interval;
```

Anyone who has done something like this? I have a lot of use cases where I prefer this behavior over `throttle:`.

PR #1729, opened 2015-02-03
Implement throttle for Signal

adly-holler:
Man, Swift really makes ReactiveCocoa come to life! What fun!

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -569,10 +569,80 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
 	}
 }
 
+/// Throttles values for which `predicate` returns true.
+///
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Thank you so much for submitting this! However, I'd really like for RAC 3.0's version of `throttle` to fix the [unnecessary delay](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1396) it often adds. In short, we should pass through values as quickly as possible _while preserving_ the interval between them (i.e., not always having to wait for the full interval).

This was implemented in the previous iteration of the Swift API, but only for `HotSignal`, which did not have a concept of error or completion. Still, [the implementation](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-before-signal-producers/ReactiveCocoa/Swift/HotSignal.swift#L404-L433) might be of use, if you want to try porting it.

adly-holler:
Feast your eyes on #1730

PR #1730, opened 2015-02-03
Swift throttle redux

adly-holler:
Here's another go

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -569,20 +569,12 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
 	}
 }
 
-/// Throttle values sent by the receiver for which `predicate` returns true,
-/// so that at least `interval` seconds pass between each, then forwards
-/// them on the given scheduler.
-/// 
-/// Values that do not pass `predicate`, Errors, and Completed events
-/// are forwarded immediately. If an unthrottled value is forwarded while
-/// a throttled value is pending, the throttled value is discarded.
-///
-/// If the receiver completes before the last throttled value is sent,
-/// it will be sent immediately.
+/// Throttle values sent by the receiver, so that at least `interval`
+/// seconds pass between each, then forwards them on the given scheduler.
 ///
 /// If multiple values are received before the interval has elapsed, the
 /// latest value is the one that will be passed on.
-public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType, predicate: T -> Bool = { _ in true })(signal: Signal<T, E>) -> Signal<T, E> {
+public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
 	precondition(interval >= 0)
 
 	return Signal { observer in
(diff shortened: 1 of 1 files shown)


adly-holler:
Oh gotta use predicate, hold up

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -569,6 +569,88 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType, predicate: T -> Bool = { _ in true })(signal: Signal<T, E>) -> Signal<T, E> {
+	precondition(interval >= 0)
+
+	return Signal { observer in
+		let state = Atomic(ThrottleState<T>())
+		let flush = {
+			state.modify { state in
+				if let value = state.pendingValue {
+					let now = scheduler.currentDate
+					sendNext(observer, value)
```
neilpa: Forwarding events may be dangerous while holding the lock. Should try to hoist this outside of the modify block

neilpa: @jspahrsummers had similar [feedback](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1714#discussion-diff-24039204) on another PR.

adly-holler:
You're absolutely right, and there's no need to. On it.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -569,6 +569,88 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+							flush()
+						}
+					}
+					compositeDisposable.addDisposable(disposableFromScheduler)
+					newState.scheduleDisposable = ScopedDisposable(compositeDisposable)
+					return newState
+				}
+			} else {
+				state.modify { (var state) in
+					sendNext(observer, value)
```
neilpa: Same here about the locking

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -569,6 +569,88 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+					if prev == nil || now.timeIntervalSinceDate(prev!) >= interval {
+						scheduleDate = now
+					} else {
+						scheduleDate = prev!.dateByAddingTimeInterval(interval)
+					}
+					
+					let compositeDisposable = CompositeDisposable()
+					let disposableFromScheduler = scheduler.scheduleAfter(scheduleDate) {
+						if !compositeDisposable.disposed {
+							flush()
```
neilpa: Same here about the locking

jspahrsummers:
Er, why the use of the predicate? That doesn't match the signature of [the `throttle` that I referenced](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-before-signal-producers/ReactiveCocoa/Swift/HotSignal.swift#L404-L433).

adly-holler:
It was a really handy additional feature in RAC 2 (say you're implementing an autocomplete field and you want to throttle the text signal only if it wasn't cleared) and since we have defaults in Swift it provides no clutter if it isn't used. Want me to remove it?

jspahrsummers:
One of my main concerns with the Swift API is keeping it simple enough that it's not overwhelming. This was a big problem with RAC 2.x, and it'll only be greater with generics and a new language. :wink:

Since we can always add APIs more easily than we can remove or change them, I'd like to leave out the predicate behavior for now. Thanks for thinking of it, though, @adly-holler!

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -569,6 +569,71 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+		}, completed: {
+			flush()
+			sendCompleted(observer)
+		})
+	}
+}
+
+// TODO: Move this inside function and rename when compiler stops segfaulting
+private struct ThrottleState<T> {
+	var previousDate: NSDate? = nil
```
kastiglione: This is never assigned to.

jspahrsummers:
Is this ready for re-review? (We usually leave an emoji or comment to indicate handoff, since commits don't notify on their own.)

adly-holler:
Oh! Yeah, a review would be great. ⚡

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: _(Addressed.)_

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -569,6 +569,72 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+	var pendingValue: T? = nil
+	var scheduleDisposable: ScopedDisposable? = nil
```
jspahrsummers: Why is this a `ScopedDisposable`? TBH, I'm not even sure how that will work inside a value type.

I think it should just be a normal disposable unless there's a reason we can't implement the scoping ourselves.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -569,6 +569,72 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+///
+/// If multiple values are received before the interval has elapsed, the
+/// latest value is the one that will be passed on.
+public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
+	precondition(interval >= 0)
+
+	return Signal { observer in
+		let state = Atomic(ThrottleState<T>())
+		let flush: () -> Void = {
+			let (_, doSend: (() -> Void)?) = state.modify { (var state) in
```
jspahrsummers: Could this maybe yield `T?` instead of the closure, to keep the signature simpler and easier to understand?

jspahrsummers:
Actually, scratch those comments. I'm gonna branch off this PR and make some changes directly, because I've spotted some larger architectural stuff that needs to be reworked.

Since it'll be easier for me to make the changes and then explain them, I'll open a new PR and CC everyone there.

Sorry about the trouble!

adly-holler:
Yo sorry for taking up your time reviewing. I'm still learning this functional-valuetype-Swifty stuff – next PR will be better.

jspahrsummers:
@adly-holler No, it's okay! Sorry, that sounded way more negative than I intended. There's a lot going on here, and most of it is non-obvious (in part because we have zero high-level documentation right now).

I'm going to try to thoroughly explain the changes I make, so hopefully that clarifies any reworking.

PR #1734, opened 2015-02-08
Preserve overall order in -groupBy:transform:

bencochran:
Given the following code:

```objective-c
RACSignal *numbers = [[[@"1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30" componentsSeparatedByString:@" "].rac_sequence
    map:^(NSString *string) {
        return @([string integerValue]);
    }]
    signal];

[[[numbers
    groupBy:^(NSNumber *number) {
        return @(number.integerValue % 3);
    }]
    flattenMap:^(RACSignal *signal) {
        return [signal collect];
    }]
    subscribeNext:^(NSArray *group) {
        NSLog(@"%@", group.firstObject);
    }];
```
I would expect to see “1, 2, 3” as my output. But instead I get “3, 1, 2”

Removing the `collect` and just watching all the signals makes the values all come through in the correct order. It’s the signal completion that’s happening out of order. This line’s the culprit:

```objective-c
[groups.allValues makeObjectsPerformSelector:@selector(sendCompleted)];
```

Which sends completions based on `NSArray`’s internal ordering.

This resolves that issue by also keeping an ordered list of the `RACGroupedSignal`s and sending completion based on that.

### DIFF
--- a/ReactiveCocoa/RACSignal+Operations.m
+++ b/ReactiveCocoa/RACSignal+Operations.m
@@ -1074,6 +1074,7 @@ static RACDisposable *subscribeForever (RACSignal *signal, void (^next)(id), voi
 
 	return [[RACSignal createSignal:^(id<RACSubscriber> subscriber) {
 		NSMutableDictionary *groups = [NSMutableDictionary dictionary];
+		NSMutableArray *orderedGroups = [NSMutableArray array];
 
 		return [self subscribeNext:^(id x) {
 			id<NSCopying> key = keyBlock(x);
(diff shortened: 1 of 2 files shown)


jspahrsummers:
I think you might be the only person in the world using this method. :wink:

This looks good, though! Could you please add a regression test for this issue?

bencochran:
I think I may be :smiley: 

(It’s specialized enough, I could see it being left out of future versions…)

iheart2code:
I can't believe you've left @jspahrsummers hanging for 23 days!

Issue #1700, opened 2015-01-25
Consider replacing Signal.Type.pipe with Signal.Type.disposablePipe


andymatuschak:
Clients will often have to add to the returned signal's disposable, e.g.:

```swift
(signal, observer) = Signal.pipe()
someOtherSignal.observe(observer)
return signal
```

The canonical API should certainly allow—but also should probably encourage—developers to think about the correct memory management.

jspahrsummers:
Related to/dependent upon our resolution to #1701.

Issue #1701, opened 2015-01-25
Swift signal disposable model can lead to leaks


andymatuschak:
I think we still haven't quite gotten the `Disposable` model quite right. Consider:

```swift
let disposable = producer.startWithSignal { signal, disposable in
  disposable.add(signal |> map { $0 } |> observe(next: { println($0) }))
}
disposable.dispose()
```

The `map` `Signal` is leaked.

Perhaps disposal should flow downstream? Not sure.

jspahrsummers:
Thanks for catching this! I've been thinking about this since you pointed it out.

The primary question we need to answer is: _why_ is disposal important for `Signal`s, when they're meant to be producer controlled? In other words, why can't disposal be a concern that's external to the class entirely?

My conclusion: it's important because **`Signal` disposal breaks the retain cycle between a signal and its observers**. This is why infinite signals require explicitly disposal, while finite signals generally do not (because terminating events will result in the same behavior).

As mentioned in chat, I did—in the previous iteration of the Swift APIs—try to make `HotSignal` (the precursor to `Signal`) obey “normal” memory management semantics, but this is way harder to use, because it forces a choice between two obnoxious designs:

 1. `observe` keeps a strong reference to the `Signal`, in which case _all_ observations must be explicitly disposed to avoid leaks.
 1. `observe` keeps a weak reference to the `Signal`, in which case you need to guarantee that a strong reference exists to the `Signal` in some other way.

So, if we agree that we want more convenient memory management than that, we effectively have to create our own retain cycles that can be broken through disposal.

**The primary use case for explicit disposal is cancellation**—for example, a consumer decides to cancel `SignalProducer.startWithSignal` and the associated work. In that scenario, it makes sense that everything attached to the produced `Signal` should get thrown away. It would also be inappropriate to send a completed or error event, because then the act of cancellation itself can result in work by downstream consumers.

Since `Signal` is meant to be producer-controlled, it seems reasonable that the producer should be the one to cancel it. To that end, **I think we should do the following**:

 1. [Return a `Disposable` from `Signal.pipe`](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1700). Disposing it will tear down the signal and its dependents.
 1. Pass a `Disposable` (perhaps a `CompositeDisposable`) into `Signal.init`. Disposing it will tear down the signal and its dependents.
[...]

PR #1735, opened 2015-02-09
Add an Interrupted event

jspahrsummers:
One possible resolution for #1700 and #1701.

By sending an `Interrupted` event when `SignalProducer.start` is disposed of, we allow each `Signal` in the chain to perform teardown tasks. Downstream `Signal`s would otherwise leak, because disposal fundamentally only goes _upstream_ toward the producer. Events go _downstream_ toward the consumer, allowing cancellation to propagate through the chain.

This also happens to resolve @andymatuschak's concern in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1688#issuecomment-71301825 that `Signal`s which have _already_ terminated can still be observed but simply will never generate events. Now, we can immediately generate an `Interrupted` event and detach the new observer.

I've done a lot more thinking about this than I could hope to write in the body of a PR, so please ask any questions you have and I'll dump my thoughts into an answer.

**To do:**

 - [x] Update tests as necessary
 - [x] Add new lifetime tests
 - [x] Merge from `swift-development`
 - [x] Fix unstated compiler error (lol)

### DIFF
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -81,12 +81,16 @@ private func optionalize<T, E>(signal: Signal<T, E>) -> Signal<T?, E> {
 
 /// Creates a RACSignal that will start() the producer once for each
 /// subscription.
+///
+/// Any `Interrupted` events will be silently discarded.
 public func asRACSignal<T: AnyObject, E>(producer: SignalProducer<T, E>) -> RACSignal {
 	return asRACSignal(producer |> optionalize)
 }
 
 /// Creates a RACSignal that will start() the producer once for each
 /// subscription.
+///
+/// Any `Interrupted` events will be silently discarded.
 public func asRACSignal<T: AnyObject, E>(producer: SignalProducer<T?, E>) -> RACSignal {
 	return RACSignal.createSignal { subscriber in
 		let selfDisposable = producer.start(next: { value in
@@ -104,11 +108,15 @@ public func asRACSignal<T: AnyObject, E>(producer: SignalProducer<T?, E>) -> RAC
 }
 
 /// Creates a RACSignal that will observe the given signal.
+///
+/// Any `Interrupted` event will be silently discarded.
 public func asRACSignal<T: AnyObject, E>(signal: Signal<T, E>) -> RACSignal {
 	return asRACSignal(signal |> optionalize)
 }
 
 /// Creates a RACSignal that will observe the given signal.
+///
+/// Any `Interrupted` event will be silently discarded.
 public func asRACSignal<T: AnyObject, E>(signal: Signal<T?, E>) -> RACSignal {
 	return RACSignal.createSignal { subscriber in
 		let selfDisposable = signal.observe(next: { value in
@@ -120,7 +128,8 @@ public func asRACSignal<T: AnyObject, E>(signal: Signal<T?, E>) -> RACSignal {
 		})
 
 		return RACDisposable {
-			selfDisposable.dispose()
+			selfDisposable?.dispose()
+			return
 		}
 	}
 }
@@ -140,7 +149,12 @@ extension RACCommand {
 			// FIXME: Workaround for <~ being disabled on SignalProducers.
 			|> startWithSignal { signal, disposable in
 				let bindDisposable = enabledProperty <~ signal
-				disposable.addDisposable(bindDisposable)
+				
+				signal.observe(SinkOf { event in
+					if event.isTerminating {
+						bindDisposable.dispose()
+					}
+				})
 			}
 
 		return Action(enabledIf: enabledProperty) { (input: AnyObject?) -> SignalProducer<AnyObject?, NSError> in
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -143,19 +143,9 @@ extension RACCommand {
 	public func asAction(file: String = __FILE__, line: Int = __LINE__) -> Action<AnyObject?, AnyObject?, NSError> {
 		let enabledProperty = MutableProperty(true)
 
-		self.enabled.asSignalProducer()
+		enabledProperty <~ self.enabled.asSignalProducer()
 			|> map { $0 as Bool }
 			|> catch { _ in SignalProducer<Bool, NoError>(value: false) }
-			// FIXME: Workaround for <~ being disabled on SignalProducers.
-			|> startWithSignal { signal, disposable in
-				let bindDisposable = enabledProperty <~ signal
-				
-				signal.observe(SinkOf { event in
-					if event.isTerminating {
-						bindDisposable.dispose()
-					}
-				})
-			}
 
 		return Action(enabledIf: enabledProperty) { (input: AnyObject?) -> SignalProducer<AnyObject?, NSError> in
 			let executionSignal = RACSignal.defer {
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -141,14 +141,16 @@ public func map<T, U, E>(transform: T -> U)(signal: Signal<T, E>) -> Signal<U, E
 /// Preserves only the values of the signal that pass the given predicate.
 public func filter<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T, E> {
 	return Signal { observer in
-		return signal.observe(next: { value in
-			if predicate(value) {
-				sendNext(observer, value)
+		return signal.observe(Signal.Observer { event in
+			switch event {
+			case let .Next(value):
+				if predicate(value.unbox) {
+					sendNext(observer, value.unbox)
+				}
+
+			default:
+				observer.put(event)
 			}
-		}, error: { error in
-			sendError(observer, error)
-		}, completed: {
-			sendCompleted(observer)
 		})
 	}
 }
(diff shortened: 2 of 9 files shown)


NachoSoto:
> Fix unstated compiler error (lol)

Now I understand your [tweet](https://twitter.com/jspahrsummers/status/564592866089906176). Maybe you already know this, but here's a little trick to find out exactly why `swiftc` is crashing. If you copy the command executed in the compile phase (`CompileSwiftSources normal arm64 com.apple.xcode.tools.swift.compiler`) and run it from the command line you can see exactly what's going on. My guess is that these errors are output through `stderr` and `Xcode` doesn't show them. This is the error in this branch (coming from `Event.swift`):

```json
{
  "kind": "finished",
  "name": "compile",
  "pid": 10610,
  "output": "Basic Block in function '_TFC13ReactiveCocoa19CompositeDisposablecfMS0_USs12SequenceType_USs13GeneratorType__FQ_S0_' does not have terminator!\nlabel %entry3\nLLVM ERROR: Broken function found, compilation aborted!\n",
  "exit-status": 1
}
```

The idea behind this PR sounds completely sound to me, although `disposables` is probably my weakest spot in RAC, so what do I know :)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -168,25 +156,32 @@ public func take<T, E>(count: Int)(signal: Signal<T, E>) -> Signal<T, E> {
...
-				taken++
-				sendNext(observer, value)
-			} else {
-				sendCompleted(observer)
+		return signal.observe(Signal.Observer { event in
+			switch event {
+			case let .Next(value):
+				if taken < count {
+					taken++
+					fallthrough
```
kastiglione: no shit!

jspahrsummers:
Thanks @NachoSoto!

ReactiveCocoa/Swift/Signal.swift
neilpa: Definitely confusing at first glance. I also find it a bit disturbing in this context since it's essentially `goto default` from an inner scope.

jspahrsummers: In this context, I think it makes sense (although maybe I should be restructuring the code to make it clearer).

The idea is, “if we've not exhausted our count, use the default behavior of passing everything through.” The inverse applies to `skip` too.

neilpa: A restructuring might be worthwhile given this isn't how `fallthrough` is generally used. Maybe inverting the conditional and adding an early return?

```swift
case let .Next(value):
    if taken == count {
        sendCompleted(observer, value)
        return
    }
    taken++
    fallthrough
default:
    observer.put(event)
```

Not sure this is ideal but at least it's more obvious what's going on.

neilpa:
I noticed you changed `never` to send the interrupted event in 7e79659 but then reverted it in c30ccdd. That seems like the right behavior since this is supposed to be an infinite empty signal. I'm however curious if there were scenarios you had in mind where it makes sense to have an operator that immediately interrupts?

ReactiveCocoa/Swift/Signal.swift
kastiglione: Maybe just call `sendCompleted` explicitly? It's to the point. Unless I'm missing subtleties, the tradeoff of micro-reuse at the cost of the indirect control flow doesn't seem worth it.

kastiglione:
> I've done a lot more thinking about this than I could hope to write in the body of a PR, so please ask any questions you have and I'll dump my thoughts into an answer.

My one question: can you further elaborate on your thinking on this subject? :trollface:

kastiglione:
I'm half serious with the above comment. After looking at this PR, I don't know what to think. I liked the idea, but when I see this code I'm uncertain.

These are observations/thoughts that I'm not sure what to think of:

* Strange that an externally triggered event is now *handled* in the same class of events that producers internally send. Is that "complecting"?
* Could an operator (or any composition point) accidentally leave out the `interrupted:` handler and now lose interruption? Seems like interruption should be opt-out, not opt-in.
* `Event` now has three cases that all terminate a signal, and, at least now, that *feels* messier.
* Interesting that disposal always was a kind of an event. Essentially a "phantom" event, i.e. it was an unnamed, unobservable.

jspahrsummers:
@kastiglione Thanks for asking some great questions! I'll take my usual devil's advocate position just so we can find out what makes sense and what doesn't, but I'll note that I am still very ambivalent myself.

> Strange that an externally triggered event is now *handled* in the same class of events that producers internally send. Is that "complecting"?

What distinguishes an “external” event from an “internal” one? I'm not being punchy—this is something I struggled with too—but _all_ signal events are in some sense “externally” driven. The only question is whether they're determined ahead of time, or at some unforeseen cancellation point.

In any case, disposal _is_ an event that has to be handled whether we reify it or not, because we'd be subject to leaks if we didn't do so.

> Could an operator (or any composition point) accidentally leave out the `interrupted:` handler and now lose interruption?

Yes.

> Seems like interruption should be opt-out, not opt-in.

I agree that it _should_, but I honestly just couldn't find a sane way to make it work.

This comes back to what I mentioned above—when a `Signal` needs to be disposed, we're fundamentally trying to send information _downstream_ so that operators along the way can stop what they're doing.

This is very different from RAC 2.x, where disposal was actually a behavior that traveled _upstream_.

> `Event` now has three cases that all terminate a signal, and, at least now, that *feels* messier.

I thought about pulling out the terminating event types into their own `enum`, but I ultimately decided that it wouldn't add any value. My thinking was: they all have to be handled one way or another, so why introduce a level of indirection?

> Interesting that disposal always was a kind of an event. Essentially a "phantom" event, i.e. it was an unnamed, unobservable.

Right, exactly! This is the key insight that I think we've arrived at here, even if we don't ultimately take this approach.

By treating disposal as an event in its own right—consistent with all other signal events—I think we actually _simplify_ the API, because it suddenly follows a much more familiar pattern.

jspahrsummers:
@ReactiveCocoa/reactivecocoa Did anyone else have further thoughts here? Is there something I can do to explain or demonstrate it further?

neilpa:
Just a few comments

>> Interesting that disposal always was a kind of an event. Essentially a "phantom" event, i.e. it was an unnamed, unobservable.

> Right, exactly! This is the key insight that I think we've arrived at here, even if we don't ultimately take this approach.

This last point is the most salient to me. Proper disposal can be hard and this simplifies it by offering a clear hook in. We may need to iterate more but I'm all for this as a first aproach. 

> This also happens to resolve @andymatuschak's concern in #1688 (comment) that Signals which have already terminated can still be observed but simply will never generate events. Now, we can immediately generate an Interrupted event and detach the new observer.

This implies there's actually two types of `Interrupted` events

1. The signal was prematurely disposed before `Completed` or `Error`
2. The signal already terminated do to an earlier `Completed`, `Error` or `Interrupted`

I don't foresee any problem with this in practice but wanted to call this out in case others had thoughts.

jspahrsummers:
@joshaber

kastiglione:
Can someone photoshop this gif to remove the beard? How am I supposed to believe this is @jspahrsummers?

alanjrogers:
> Can someone photoshop this gif to remove the beard? How am I supposed to believe this is @jspahrsummers?

It looks more like @joshaber

kastiglione:
That was my original comment that I didn't go with :P

jspahrsummers:
@andymatuschak Could I bother you for a glance here too, since you found the `Signal` leaks originally? :bow:

jspahrsummers:
I'm feeling more and more strongly that this is the proper way to resolve #1700 and #1701.

Because we have no other candidate solutions, I think we should merge this and see how it feels. We can choose a different approach before shipping if this seems unwieldy or painful.

Ready for review and merge.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -140,7 +149,12 @@ extension RACCommand {
 			// FIXME: Workaround for <~ being disabled on SignalProducers.
 			|> startWithSignal { signal, disposable in
 				let bindDisposable = enabledProperty <~ signal
-				disposable.addDisposable(bindDisposable)
+
```
NachoSoto: This is repeated in a few places. How do you feel about something like `bindDisposable.disposeAfterSignalTermination(signal)`?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -13,52 +13,51 @@ import LlamaKit
...
 
+		let generatorDisposable = SerialDisposable()
 		let sink = Observer { event in
 			self.lock.lock()
 
 			if let observers = self.observers {
+				if event.isTerminating {
+					// Disallow any further events (e.g., any triggered
+					// recursively).
+					self.observers = nil
```
NachoSoto: Nice

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -78,46 +77,40 @@ public final class Signal<T, E: ErrorType> {
...
-		return ActionDisposable {
-			if let token = token {
+		if token == nil {
+			sink.put(.Interrupted)
+			return nil
+		} else {
+			return ActionDisposable {
 				self.lock.lock()
-				self.observers?.removeValueForToken(token)
+				self.observers?.removeValueForToken(token!)
```
NachoSoto: Nit: reverse the `if` to avoid the unwrapping here:
```swift
if let token = token {
} else {
}
```

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -334,10 +340,20 @@ public func materialize<T, E>(signal: Signal<T, E>) -> Signal<Event<T, E>, NoErr
...
-			observer.put(event)
-		}, completed: {
-			sendCompleted(observer)
+		return signal.observe(SinkOf { event in
+			switch event {
+			case let .Next(innerEvent):
+				observer.put(innerEvent.unbox)
+
+			case .Error:
+				assert(false)
```
NachoSoto: I think `fatalError` would be more appropriate here, since the condition is always false.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
jspahrsummers: This code can change to just use `<~` now. I don't think we should add anything like that.

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: This condition is literally impossible per the types—the compiler just requires it.

NachoSoto: Right, I think `fatalError("This can't happen")` can be clearer :)

jspahrsummers:
@NachoSoto I went ahead and pushed, because it seemed like you were done reviewing. For future, we usually use an emoji or text sign-off for that too. :smile:

NachoSoto:
I was :D your changes look good! :)

PR #1736, opened 2015-02-09
[WIP] Automatically upload releases via Travis, recommend binary installation

jspahrsummers:
I don’t know if this is something we actually want to do, but I thought I’d put it out there. It does make our instructions a lot simpler. :wink:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,12 @@ notifications:
   email: false
   slack:
     secure: C9QTry5wUG9CfeH3rm3Z19R5rDWqDO7EhHAqHDXBxT6CpGRkTPFliJexpjBYB4sroJ8CiY5ZgTI2sjRBiAdGoE5ZQkfnwSoKQhWXkwo19TnbSnufr3cKO2SZkUhBqOlZcA+mgfjZ7rm2wm7RhpCR/4z8oBXDN4/xv0U5R2fLCLE=
+deploy:
+  provider: releases
+  api-key:
+    secure: fuanFoD8mM8N30xCXyDg1GYjhMCS4eQU0hfZWV9j42tW5cQ+7cZvhDI3HrjGhkszLVQgTyxMzysniLXa7s/O0xq4DGCF7UkJUPDJm5WjO1nj9Tl248O/aIasa49EUcb6lYeDNQ2CxLQv/LPXaxc7Km31GJKMfbVgnz7Vp1HRosc=
+  file: FIXME
+  skip_cleanup: true
+  on:
+    tags: false
+    all_branches: false
--- a/README.md
+++ b/README.md
@@ -486,11 +486,13 @@ ReactiveCocoa supports OS X 10.8+ and iOS 8.0+.
 
 ## Importing ReactiveCocoa
 
-We recommend using [Carthage](https://github.com/Carthage/Carthage) to add ReactiveCocoa to your project. Simply install Carthage, then add the following to your `Cartfile`:
+We recommend using [Carthage](https://github.com/Carthage/Carthage) to add ReactiveCocoa to your project. Simply install Carthage, add the following to your `Cartfile`:
 
 ```
 github "ReactiveCocoa/ReactiveCocoa"
 ```
+
+… then run `carthage update`.
 
 If you don’t want to use Carthage, you can also add ReactiveCocoa as a Git submodule. If you do this, make sure to use `git submodule update --init --recursive` to pull in ReactiveCocoa’s dependencies as well.
 
--- a/README.md
+++ b/README.md
@@ -486,34 +486,18 @@ ReactiveCocoa supports OS X 10.8+ and iOS 8.0+.
 
 ## Importing ReactiveCocoa
 
-To add RAC to your application:
+We recommend using [Carthage](https://github.com/Carthage/Carthage) to add ReactiveCocoa to your project. Simply install Carthage, then add the following to your `Cartfile`:
 
- 1. Add the ReactiveCocoa repository as a submodule of your application's
-    repository.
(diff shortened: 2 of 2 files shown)


mdiep:
I think we should keep the normal instructions in case someone doesn't want to use Carthage or CocoaPods. But I'd support plugging Carthage before those instructions.

kastiglione:
:+1: to keeping the basic instructions. If not inline, at least linked to.

jspahrsummers:
It's hard for me to recommend the subproject setup with the myriad ways that exist to mess it up. And once you want to use another library that depends on RAC, it's pretty unclear what to do.

In my mind, we built Carthage for a reason—to _replace_ the submodule and subproject workflow with something that solves transitivity and makes setup easier. If it's not the easiest and _first recommended_ workflow for installing one of our frameworks, I kinda feel like we've failed.

None of this is to say we should break subproject setups. Users that know how can still do that without any issues—but anyone that doesn't know how would probably be better off with some tooling assistance.

mdiep:
> If it's not the easiest and first recommended workflow for installing one of our frameworks, I kinda feel like we've failed.

I'm not disagreeing with that. Carthage should be the easiest and the first recommended workflow. But I think it would benefit everyone to keep the other instructions available.

If nothing else, the other instructions show how much you'd gain by using Carthage. :stuck_out_tongue_closed_eyes:

jspahrsummers:
That's fair. The duplication (across our frameworks) is pretty annoying, though—maybe we should centralize those instructions somehow?

swizzlr:
Shouldn't the first way be "download the latest framework from releases and drag it into the project"?

Carthage being a slight enhancement on that?

jspahrsummers:
> Shouldn't the first way be "download the latest framework from releases and drag it into the project"?

We're not currently very good about uploading binaries to our tagged releases, but this is a good point.

@ReactiveCocoa/reactivecocoa How would others feel about doing this for each tag?

kastiglione:
I don't know how much it would be used, but still :+1:

robrix:
> @ReactiveCocoa/reactivecocoa How would others feel about doing this for each tag?

It’s just another thing to forget. Unless it’s automated, I’d leave it out.

ashfurrow:
I understand the desire to make using ReactiveCocoa as simple as possible, and I've long understood the reluctance to include instructions for installing RAC with CocoaPods. However, I feel that if ReactiveCocoa were to support one dependency manager over another that's commonly used, people _will_ perceive this as an explicit endorsement bred out of the developer overlap between ReactiveCocoa and Carthage. 

I believe that this decision should be made with *only* the best interests of the users of ReactiveCocoa at heart, and should ignore the interests of Carthage (and CocoaPods) because those are separate projects from RAC.

neilpa:
>> Shouldn't the first way be "download the latest framework from releases and drag it into the project"?

> We're not currently very good about uploading binaries to our tagged releases, but this is a good point.

Won't you want this anyways since the [next version of Carthage](https://github.com/Carthage/Carthage/pull/257) will be able to pull from relelases instead of having to rebuild the frameworks.

Coneko:
> That's fair. The duplication (across our frameworks) is pretty annoying, though—maybe we should centralize those instructions somehow? Any ideas?

That's unavoidable: duplication between installation instructions across different projects is a good thing, it means the installation process is fairly standardised.

We can recommend installing via Carthage. It's definitely an "explicit endorsement bred out of the developer overlap between ReactiveCocoa and Carthage", but there is a very practical difference between Carthage and CocoaPods as far as ReactiveCocoa is concerned: there being so much overlap between the people working on both of them means the two projects will actively try to support each other.

I'm ok with it as long as the vanilla build process remains on CI.

jspahrsummers:
Hmm, @neilpa is right, though. Binary frameworks will also be the easiest thing for users.

Therefore, I'd like to recommend binaries as the preferred setup mechanism, and mention subprojects, Carthage, and CocoaPods in asides.

We can automate the uploading of binaries via Travis, but (against @Coneko's recommendation) this would basically require building with `carthage`, since we would want to upload fat binaries for iOS, and doing that by hand is super annoying.

To summarize, how does everyone feel about:

1. Building with Carthage on CI
1. Automatically uploading binary frameworks via Travis when a release is tagged
1. Recommending binaries as the preferred installation mechanism
1. Pointing out that subprojects, Carthage, and CocoaPods are also possible (in that order)

ashfurrow:
:+1:

robrix:
That seems reasonable to me, but I’m curious what @Coneko thinks re: CI.

alanjrogers:
> Building with Carthage on CI

If we're talking travis, and we want to start using Xcode 6.3/Swift 1.2, we're going to be waiting a while.

jspahrsummers:
> If we're talking travis, and we want to start using Xcode 6.3/Swift 1.2, we're going to be waiting a while.

Well, we already use Travis, so /shrug.

Besides, this whole discussion is in reference to `master`. :wink:

jspahrsummers:
Blocked (by choice) on https://github.com/Carthage/Carthage/issues/258.

Coneko:
I'm just worried about issues that might pop up in the vanilla build and not in the Carthage one, but given that Carthage uses xcodebuild I'm not sure what that might be.

I know there are issues that can make the Xcode build fail even when xcodebuild succeeds, but the current CI without Carthage uses xcodebuild anyway right?

jspahrsummers:
@Coneko Yeah, `xcodebuild` is basically unavoidable for CLI builds. There is `xctool`, but that eventually uses `xcodebuild` as well.

However, I'm going to temporarily close this out, because https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1775 will make it impossible to use Travis until the next stable version of Xcode.

PR #1737, opened 2015-02-09
change iOS code signing identity for release builds to allow embedding the framework in your project

ben9m:


jspahrsummers:
This change shouldn't be necessary. It makes it harder for people to build RAC during development (because they now need a distribution certificate), and adding the framework to the “Embedded Frameworks & Libraries” section in your target settings—with “Code Sign on Copy” checked—should re-sign it automatically with the correct certificate.

ben9m:
This change only affects distribution (App Store and Ad Hoc) builds. Without it, Xcode continues to sign the embedded framework with the iOS Developer certificate rather than the iOS Distribution. I'm not sure if it causes App Store problems but it does with Ad Hoc.

ben9m:
It should only affect Release builds, not Debug.


—

jspahrsummers:
I understand. Sometimes it's useful to build for the Release configuration during development too.

My point is that this doesn't affect the ability to bundle RAC for Distribution (because of the Code Sign on Copy feature), but it _does_ require that users now have a Distribution certificate to play around with the Release configuration.

PR #1739, opened 2015-02-10
Swift zipWith() implementation

almassapargali:


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -571,6 +571,45 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+			}
+		}
+		
+		let onError = { sendError(observer, $0) }
+		let onCompleted = { sendCompleted(observer) }
+		
+		let signalDisposable = signal.observe(next: { value in
+			lock.lock()
+			signalTeeth.append(value)
+			zipUp()
```
neilpa: I would structure this differently to avoid sending events under lock. Instead of sending the event inside `zipUp`, that method should return an optional tuple to send if there are values available from both signals. That loop is unnecessary too since at most one tuple will be sent per event on either of the zipped signals.

neilpa:
I spent more time thinking about and tinkering with different approaches to implement this one. Here's [the gist](https://gist.github.com/neilpa/ab7eba6ea1e6b57dfcea) of where I landed. Feel free to take as much or as little from that as you like.

kastiglione:
FWIW, I find the pair of arrays to be clearer than an array of `Either`s.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -571,6 +571,45 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
 	}
 }
 
+/// Zips elements of two signals into pairs. The elements of any Nth pair
+/// are the Nth elements of the two input signals.
+public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) -> Signal<(T, U), E> {
+	return Signal { observer in
+		let lock = NSRecursiveLock()
```
kastiglione: This should probably use the `Atomic` class.

neilpa: Won't that result in a copying of the array everytime you mutate it via `modify`?

almassapargali:
@neilpa, @kastiglione 
As for sending values inside lock, current implementation of `combineLatestWith` does it too [here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-development/ReactiveCocoa/Swift/Signal.swift#L211-L227). It's added by @jspahrsummers in [this commit](https://github.com/ReactiveCocoa/ReactiveCocoa/commit/e0f90ca364099bc1a5c5e5ce86b5af3f5b7c070d).

As for implementation design, I was consulting [this](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-before-signal-producers/ReactiveCocoa/Swift/HotSignal.swift#L498-L536) old implementation.

Btw, what do you think about [this gist](https://gist.github.com/almassapargali/068aeccce27b205f527a). I used `Either` enum, to have all logic on one place, and moved `sendNext()` outside of lock?

ReactiveCocoa/Swift/Signal.swift
kastiglione: I dunno. Hopefully the compiler can optimize them structs. Guess we'd have to inspect IL or assembly. Either way, premature optimization, no?

almassapargali: I was taking `combineLatestWith` [implementation](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-development/ReactiveCocoa/Swift/Signal.swift#L209-L257) as design sample and it used `NSRecursiveLock`.
Can you elaborate on why it should use `Atomic` please?

kastiglione: I didn't see that implementation. @jspahrsummers will have to answer that, but there's no doubt a reason I'm missing.

jspahrsummers: `combineLatestWith` uses a recursive lock because it has to send values under synchronization, or else run into ordering violations.

jspahrsummers: I can't tell whether `Atomic` could potentially introduce ordering issues here, but that inability to tell means we should probably keep using a recursive lock.

neilpa: I don't think that `zipWith` has the same potential for ordering problems as `combineWithLatest`. As an observer of the zipped signal, it doesn't matter which of the inner signals triggers the zipped value to be sent.

jspahrsummers: :+1:

samwgoldman:
Would it be possible to generalize zipWith to something like this:

```
public func zipWith<T,U,X,E>(f: (U, T) -> X)(otherSignal: Signal<U,E>)(signal: Signal<T,E>) -> Signal<X, E> 
```

And provide an implementation where X = (U, T) instead?

jspahrsummers:
@samwgoldman You can achieve the same result with composition (`zipWith` plus `map`), so I'm not sure that would add anything but implementation complexity.

samwgoldman:
@jspahrsummers Fair enough. I thought it might be painful to destructure the tuple, but I tried it out and realized that you can pass tuples as function arguments without destructuring, so I'm happy to use `zipWith` and `map` as combinators as you suggest. Thanks. :)

jspahrsummers:
@almassapargali I needed `zipWith`, so I went ahead finished up this PR in #1746. Thanks for getting it started!

almassapargali:
👍
I wasn't sure how to implement it with atomic.

PR #1740, opened 2015-02-10
Implement try and tryMap for Signal

neilpa:


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -571,13 +573,40 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+	}
+}
+
+/// Applies `operation` to values from `signal` with `Success`ful results mapped
+/// on the returned signal and `Failure`s sent as `Error` events.
+public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>) -> Signal<U, E> {
+	return Signal { observer in
+		signal.observe(next: { value in
+			switch operation(value) {
+			case .Success(let val):
```
jspahrsummers: Could you please lift these `let`s out to the `case` level?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -257,18 +257,71 @@ class SignalSpec: QuickSpec {
...
-			}
-
-			pending("should error if a mapping fails") {
+			it("should forward mapped values upon success") {
+				let (baseSignal, sink) = Signal<Int, TestError>.pipe()
+				var signal = baseSignal |> tryMap { num -> Result<Bool, TestError> in
+					return success(num % 2 == 0)
+				}
+				
+				var even: Bool!
```
jspahrsummers: Could you please use an explicit optional? I know it “shouldn't” happen here, but crashes in the test runner are very hard to debug.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -257,18 +257,71 @@ class SignalSpec: QuickSpec {
...
+				signal.observe(next: { value in
+					even = value
+				})
+				
+				sendNext(sink, 1)
+				expect(even).to(beFalse())
```
jspahrsummers: Can you please use `beFalsy` and `beTruthy`, for better cross-compatibility? :smile:

jspahrsummers:
Looking good! Sorry about the delay in reviewing this.

ReactiveCocoaTests/Swift/SignalSpec.swift
neilpa: After converting `even` to an optional `beFalsy` will incorrectly match on `nil` values if it's never set. I can change the truth-y case if you like but I think consistency here would be better.

jspahrsummers: Maybe we could just use `equal()` in that case?

neilpa:
No problem. Fixed the first two, thoughts on the third?

jspahrsummers:
Already replied. :wink:

ReactiveCocoaTests/Swift/SignalSpec.swift
neilpa: :ok:

jspahrsummers:
Thanks @neilpa!

PR #1741, opened 2015-02-11
[UPD] we really don't want to call description without debug

kirillyakimovich:
rac_description is a debug method and its result is ignored when RAC_DEBUG_SIGNAL_NAMES is not defined.

But for tuples description is calculated anyway and then just ignored. If tuple contains arrays it may be a problem

And now implementation really do not call description if not asked.

### DIFF
--- a/ReactiveCocoa/NSObject+RACDescription.m
+++ b/ReactiveCocoa/NSObject+RACDescription.m
@@ -24,7 +24,11 @@
 @implementation NSValue (RACDescription)
 
 - (NSString *)rac_description {
-	return self.description;
+    if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL) {
+        return self.description;
+    } else {
+        return @"(description skipped)";
+    }
 }
 
 @end
@@ -32,7 +36,11 @@
 @implementation NSString (RACDescription)
 
 - (NSString *)rac_description {
-	return self.description;
+    if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL) {
+        return self.description;
+    } else {
+        return @"(description skipped)";
+    }
 }
 
 @end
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/NSObject+RACDescription.m
```diff
@@ -24,7 +24,11 @@ - (NSString *)rac_description {
 @implementation NSValue (RACDescription)
 
 - (NSString *)rac_description {
-	return self.description;
+    if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL) {
```
jspahrsummers: I don't think we really need the check for this case, since `NSValue`s are usually just primitives.

ReactiveCocoa/NSObject+RACDescription.m
```diff
@@ -32,7 +36,11 @@ - (NSString *)rac_description {
 @implementation NSString (RACDescription)
 
 - (NSString *)rac_description {
-	return self.description;
+    if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL) {
```
jspahrsummers: I don't think we really need the check for this case, since an `NSString` description is just the string itself.

ReactiveCocoa/NSObject+RACDescription.m
```diff
@@ -40,7 +48,11 @@ - (NSString *)rac_description {
 @implementation RACTuple (RACDescription)
 
 - (NSString *)rac_description {
-	return self.allObjects.description;
+    if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL) {
```
jspahrsummers: Can you please use tabs instead of spaces here?

jspahrsummers:
Thanks for catching this! Just a few notes. :sparkles:

kirillyakimovich:
Updated.
Agreed with primitives. I made it only for "consistency".

Happy to participate in this amazing project :sunny:

PR #1742, opened 2015-02-12
Fix the issue:

JadianZheng:
### function (after:repeatingEvery:withLeeway:schedule:) not work


jspahrsummers:
Thanks for the pull request!

However, it sounds like the returned `RACDisposable` just has to be retained. That's okay (though an accidental part of the design), because it's rare to want a truly _infinite_ timer.

We should just document that requirement, instead of implementing a workaround like this.

PR #1744, opened 2015-02-13
Use DISPATCH_TIME_NOW-based timer for reschedule

apalancat:
The current method using the system wall clock stops sending when the system clock is moved backwards to a time before the initial date.
I see that you are using walltime to avoid skew when the computer goes to sleep, I don't know if this method would have the same problem, I'm not sure how to test it.

I also just started using ReactiveCocoa, so I don't know how to add tests for this, or if this issue has already been discussed.

### DIFF
--- a/ReactiveCocoa/RACQueueScheduler.m
+++ b/ReactiveCocoa/RACQueueScheduler.m
@@ -95,7 +95,7 @@
 	uint64_t intervalInNanoSecs = (uint64_t)(interval * NSEC_PER_SEC);
 	uint64_t leewayInNanoSecs = (uint64_t)(leeway * NSEC_PER_SEC);
 
-	dispatch_time_t startDateDelay = (date.timeIntervalSince1970 - NSDate.date.timeIntervalSince1970) * NSEC_PER_SEC;
+	int64_t startDateDelay = (int64_t)((date.timeIntervalSince1970 - NSDate.date.timeIntervalSince1970) * NSEC_PER_SEC);
 	dispatch_source_t timer = dispatch_source_create(DISPATCH_SOURCE_TYPE_TIMER, 0, 0, self.queue);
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Thanks for the pull request! This is definitely tricky.

We basically have to choose between two alternatives:

 1. Behaving weirdly if the system clock changes (when using wall time)
 1. Unintentionally increasing our timer interval across sleep/wake (when not using wall time), because the clock only counts up while awake

The latter sucks for longer intervals, while the former is just generally sucky. I don't particularly like either, but I guess the second option would be preferable to me.

@ReactiveCocoa/reactivecocoa Thoughts?

kastiglione:
Number 2 gets my vote.

andymatuschak:
I prefer number 2 also, but I've encountered scenarios where that would be
bad (eg do something once a day). Perhaps for those cases there should be a
date based signal?

jspahrsummers:
`NSTimer` [isn't walltime based](http://stackoverflow.com/a/16597866/515514), and that's the closest thing to our behavior, so we should probably match it (thereby following option 2 [above](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1744#issuecomment-74403154)).

Unfortunately, this is a breaking change. I don't know how significant the ramifications would be, but silently changing timer behavior seems pretty insidious. :worried: 

Maybe we can just make this change to the [RAC 3.0 Swift APIs](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1382) instead?

> Perhaps for those cases there should be a date based signal?

@andymatuschak That may make sense, but it can come separately, in any case.

joshaber:
>Maybe we can just make this change to the RAC 3.0 Swift APIs instead?

:+1:

jspahrsummers:
Gonna close this out, and we can discuss it further in a 3.0-targeted PR or issue. :smile:

PR #1745, opened 2015-02-14
Add a DynamicProperty for KVC/KVO

### LINKED ISSUE #1738: Add a PropertyType instance supporting KVC/KVO?
I've waffled on this a few times myself, so I'd like to open it up to the larger conversation.

I was originally of the mindset that we should focus on “the Swift way, the right way” for RAC 3.0, and eschew dynamic Objective-C features like KVC and KVO. The problem, of course, is that those features are _really damn useful_.

Even besides their appearances in data binding—for which `MutableProperty` should be a suitable replacement—there are some APIs that require the use of KVO (`NSOperation` and Core Data come to mind). And isn't part of The Promise of RAC™ that we'll unify all these disparate paradigms into one, all-consuming abstraction?

Anyways, I'm highly ambivalent. What does @ReactiveCocoa/reactivecocoa think?

jspahrsummers:
Resolves #1738.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -95,6 +95,57 @@ extension MutableProperty: SinkType {
 	}
 }
 
+/// Wraps a `dynamic` property, or one defined in Objective-C, using Key-Value
+/// Coding and Key-Value Observing.
+///
+/// Use this class only as a last resort! `MutableProperty` is generally better
+/// unless KVC/KVO is required by the API you're using (for example,
+/// `NSOperation`).
+@objc public final class DynamicProperty: NSObject, PropertyType {
+	public typealias Value = AnyObject?
+
+	private weak var object: NSObject?
+	private let keyPath: String
+
+	/// The current value of the property, as read and written using Key-Value
+	/// Coding.
+	public var value: AnyObject? {
+		get {
+			return object?.valueForKeyPath(keyPath)
+		}
+
+		set(newValue) {
+			object?.setValue(newValue, forKeyPath: keyPath)
+		}
+	}
+
+	/// A producer that will create a Key-Value Observer for the given object,
+	/// send its initial value then all changes over time, and then complete
+	/// when the observed object has deallocated.
+	///
+	/// By definition, this only works if the object given to init() is
+	/// KVO-compliant. Most UI controls are not!
+	public var producer: SignalProducer<AnyObject?, NoError> {
+		if let object = object {
+			return object.rac_valuesForKeyPath(keyPath, observer: nil).asSignalProducer()
+				// Errors aren't possible, but the compiler doesn't know that.
+				|> catch { error in
+					fatalError("Received unexpected error from KVO signal: \(error)")
+					return .empty
+				}
+		} else {
+			return .empty
+		}
+	}
+
+	/// Initializes a property that will observe and set the given key path of
+	/// the given object. `object` must support weak references!
+	public init(object: NSObject, keyPath: String) {
+		self.object = object
+		self.keyPath = keyPath
+	}
+}
+
 infix operator <~ {
 	associativity right
 	precedence 90
@@ -133,16 +184,16 @@ public func <~ <T>(property: MutableProperty<T>, signal: Signal<T, NoError>) -> 
 /// or when the created signal sends a `Completed` event.
 public func <~ <T>(property: MutableProperty<T>, producer: SignalProducer<T, NoError>) -> Disposable {
 	var disposable: Disposable!
-	
+
 	producer.startWithSignal { signal, signalDisposable in
 		property <~ signal
 		disposable = signalDisposable
-		
+
 		property.producer.start(completed: {
 			signalDisposable.dispose()
 		})
 	}
-	
+
 	return disposable
 }
 
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -140,7 +140,7 @@ extension MutableProperty: SinkType {
 
 	/// Initializes a property that will observe and set the given key path of
 	/// the given object. `object` must support weak references!
-	public init(object: NSObject, keyPath: String) {
+	public init(object: NSObject?, keyPath: String) {
 		self.object = object
 		self.keyPath = keyPath
 	}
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -95,6 +95,57 @@ extension MutableProperty: SinkType {
...
 }
 
+/// Wraps a `dynamic` property, or one defined in Objective-C, using Key-Value
+/// Coding and Key-Value Observing.
+///
+/// Use this class only as a last resort! `MutableProperty` is generally better
+/// unless KVC/KVO is required by the API you're using (for example,
+/// `NSOperation`).
+@objc public final class DynamicProperty: NSObject, PropertyType {
+	public typealias Value = AnyObject?
```
JaviSoto: Could we make DynamicProperty generic over this type? (With the constraint of it being an AnyObject)

jspahrsummers: That would prevent it from being exposed to Objective-C (which I thought might be handy).

And, ultimately, we're going to have to cast somewhere, because KVC and KVO just use `AnyObject?` for everything.

JaviSoto: Oh, I didnt even consider Obj-C compatibility. We could always have a subclass of this just for swift perhaps (`DynamicPropertyOf<T>` or something). We would cast, but that would be an implementation detail, this API would be nicer to use  (when working with the property's producer)

jspahrsummers: My biggest question is, what should happen if the cast fails? You get `nil`? That might make sense, but could also be surprising.

This keeps it simple and pushes that complexity to the consumer.

JaviSoto: I think using `as!` would be reasonable: If the KVO observation yields results of a type you didn't expect, something is very wrong.

jspahrsummers: A library crashing its host is almost always wrong. :stuck_out_tongue_winking_eye:

JaviSoto: I agree, but if it doesn't crash in RAC, I would make my app crash anyway, by using `as!` when I get a value from the signal. Put another way: how is an app supposed to handle gracefully the KVO subscription sending values of a type you didn't expect? I think most Obj-C apps would crash with "unrecognized selector" due to the implicit cast, cause we rarely do isKindOfClass in those cases...

jspahrsummers: My point is simply that I think it's appropriate to pass this responsibility to the application, to decide whether it wants to take the risk of crashing or not. RAC doesn't have the information to know which kind of cast is appropriate here.

JaviSoto: Yeah, that's fair :)

I know that whenever I use this, I'll definitely implement in my app the generic version with the forced cast though :D

jspahrsummers: Also, you can think of this as another disincentive toward using KVC/KVO. :stuck_out_tongue:

JaviSoto: s/KVO/Obj-C :P

ReactiveCocoa/Swift/Property.swift
```diff
@@ -95,6 +95,57 @@ extension MutableProperty: SinkType {
...
+		} else {
+			return .empty
+		}
+	}
+
+	/// Initializes a property that will observe and set the given key path of
+	/// the given object. `object` must support weak references!
+	public init(object: NSObject?, keyPath: String) {
+		self.object = object
+		self.keyPath = keyPath
```
kastiglione: Do mutable strings get copied? Wondering how Swift code should handle Strings that might come from Objc.

jspahrsummers: `String` is always an immutable value type, so yes.

jspahrsummers:
Could I please get a review and merge on this?

ReactiveCocoa/Swift/Property.swift
NachoSoto: I'm with @jspahrsummers here, it's trivial to `map` the `producer` to cast.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -95,6 +95,57 @@ extension MutableProperty: SinkType {
...
+	///
+	/// By definition, this only works if the object given to init() is
+	/// KVO-compliant. Most UI controls are not!
+	public var producer: SignalProducer<AnyObject?, NoError> {
+		if let object = object {
+			return object.rac_valuesForKeyPath(keyPath, observer: nil).asSignalProducer()
+				// Errors aren't possible, but the compiler doesn't know that.
+				|> catch { error in
+					fatalError("Received unexpected error from KVO signal: \(error)")
+					return .empty
```
NachoSoto: `fatalError` is defined as `@noreturn`, so this return shouldn't be needed?

jspahrsummers: You'd think!

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -94,6 +94,80 @@ class PropertySpec: QuickSpec {
...
 
+		describe("DynamicProperty") {
+			var object: ObservableObject!
+			var property: DynamicProperty!
+
+			let propertyValue: () -> Int? = {
+				if let value: AnyObject = property?.value {
+					return value as? Int
+				} else {
+					return -1
```
NachoSoto: Why `-1` instead of `nil`?

NachoSoto:
So excited about this! :tada:

ReactiveCocoa/Swift/Property.swift
NachoSoto: I guess it's because it fails to infer the return type, which you could do with `{ (() -> SignalProducer<(), NoError>), error in` but meh, not worth it.

ReactiveCocoaTests/Swift/PropertySpec.swift
jspahrsummers: Uhhh… I don't know/remember. I'll change it.

jspahrsummers:
:roller_coaster:

NachoSoto:
:ship:

jspahrsummers:
@NachoSoto Feel free to merge at your discretion. :smile:

NachoSoto:
:boom:

PR #1756, opened 2015-02-14
timeoutWithError() swift implementation

### LINKED ISSUE #1752: Implement timeoutWithError
https://github.com/ReactiveCocoa/ReactiveCocoa/blob/15c2c9a473f5bf09b79061f603cb122a257162e1/ReactiveCocoa/Swift/Signal.swift#L687

Previous implementations in [`swift-before-signal-producers`](https://github.com/ReactiveCocoa/ReactiveCocoa/releases/tag/swift-before-signal-producers) may be of use.

almassapargali:
Here is my (first) take on it. Fixes #1752.

I think we could let disposables do the work, and let cancel sending error event, without introducing `completed` variable. But since `scheduleAfter()` returns optional `Disposable` I kept that variable, and condition.

Maybe make `scheduleAfter()` return `Disposable`, not `Disposable?`?

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -693,9 +693,9 @@ public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInter
 
 		let signalDisposable = signal.observe(observer)
 
-		let dis = CompositeDisposable([ signalDisposable ])
-		dis.addDisposable(schedulerDisposable)
-		return dis
+		let disposable = CompositeDisposable([ signalDisposable ])
+		disposable.addDisposable(schedulerDisposable)
+		return disposable
 	}
 }
 
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -682,8 +682,7 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>) 
 
 /// Forwards events from `signal` until `interval`. Then if signal isn't completed yet,
 /// errors with `error` on `scheduler`.
-public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E>
-{
+public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
 	precondition(interval > 0)
 
 	var completed = false
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -685,19 +685,13 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>) 
 public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
 	precondition(interval >= 0)
 
-	var completed = false
 	return Signal { observer in
 		let date = scheduler.currentDate.dateByAddingTimeInterval(interval)
 		let schedulerDisposable = scheduler.scheduleAfter(date) {
-			if !completed {
-				sendError(observer, error)
-			}
+			sendError(observer, error)
 		}
 
 		let signalDisposable = signal.observe(Signal.Observer { event in
-			if event.isTerminating {
-				completed = true
-			}
 			observer.put(event)
 		})
 
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -683,7 +683,7 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>) 
 /// Forwards events from `signal` until `interval`. Then if signal isn't completed yet,
 /// errors with `error` on `scheduler`.
 public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
-	precondition(interval > 0)
+	precondition(interval >= 0)
 
 	var completed = false
 	return Signal { observer in
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -691,9 +691,7 @@ public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInter
 			sendError(observer, error)
 		}
 
-		let signalDisposable = signal.observe(Signal.Observer { event in
-			observer.put(event)
-		})
+		let signalDisposable = signal.observe(observer)
 
 		let dis = CompositeDisposable([ signalDisposable ])
 		dis.addDisposable(schedulerDisposable)
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -680,11 +680,38 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>) 
 	}
 }
 
+/// Forwards events from `signal` until `interval`. Then if signal isn't completed yet,
+/// errors with `error` on `scheduler`.
+public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E>
+{
+	precondition(interval > 0)
+
+	var completed = false
+	return Signal { observer in
+		let date = scheduler.currentDate.dateByAddingTimeInterval(interval)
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -680,11 +680,38 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
 	}
 }
 
+/// Forwards events from `signal` until `interval`. Then if signal isn't completed yet,
+/// errors with `error` on `scheduler`.
+public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E>
+{
```
kastiglione: Style: Open parens are placed on the end of line, not beginning of new line.

almassapargali: Sure, I usually do so. Not sure why missed this one.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -680,11 +680,38 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
...
+
+		let signalDisposable = signal.observe(Signal.Observer { event in
+			if event.isTerminating {
+				completed = true
+			}
+			observer.put(event)
+		})
+
+		let dis = CompositeDisposable([ signalDisposable ])
+		dis.addDisposable(schedulerDisposable)
```
kastiglione: Perhaps `CompositeDisposable` could accept `[Disposable?]`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -680,11 +680,37 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
...
+/// Forwards events from `signal` until `interval`. Then if signal isn't completed yet,
+/// errors with `error` on `scheduler`.
+public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
+	precondition(interval >= 0)
+
+	var completed = false
+	return Signal { observer in
+		let date = scheduler.currentDate.dateByAddingTimeInterval(interval)
+		let schedulerDisposable = scheduler.scheduleAfter(date) {
+			if !completed {
```
kastiglione: I wonder if we need to track completion state and make this conditional. We could just blindly call `sendError()` which will do nothing if signal has already completed.

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: Swift has a hard time with that overload. :confused:

almassapargali: Yeah, compiler insisted on this.

ReactiveCocoa/Swift/Signal.swift
almassapargali: @kastiglione yep, I'd prefer that too, though look for PR description.

almassapargali:
Thanks! Btw, what is your thoughts on this?

> Maybe make `scheduleAfter()` return `Disposable`, not `Disposable?`?

I think it'd be good contract, like scheduling action returns `Disposable` which will cancel that action, if disposed?

cc\ @jspahrsummers @kastiglione

ReactiveCocoa/Swift/Signal.swift
kastiglione: I wondered about that.

almassapargali:
Right now I needed to track completion, because lack of cancellation scheduled action.

jspahrsummers:
@almassapargali That was mostly for consistency with `schedule`. In general, though, it's conceivable that some scheduler implementation may not actually be able to cancel operations—I'd rather represent that by returning `nil` than returning a disposable that doesn't do anything.

Why does it matter if it might return `nil`? It should be just as easy to use, either way.

almassapargali:
I wanted to have some way to cancel scheduled action. But yes, I was missing this:

> In general, though, it's conceivable that some scheduler implementation may not actually be able to cancel operations.

Is PR ok otherwise?

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: It's still okay to blindly `sendError`, even if this can't be disposed. If the signal has already completed, it can't error as well.

jspahrsummers:
> Is PR ok otherwise?

There's still [this note above](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1756#discussion_r24716197) from @kastiglione.

ReactiveCocoa/Swift/Signal.swift
almassapargali: Oh sorry guys, I misread "which will do nothing if signal has already completed.". I was thinking he's assuming disposal would cancel it. Sure, will change now.

almassapargali:
updated

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -680,11 +680,31 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
...
+	precondition(interval >= 0)
+
+	return Signal { observer in
+		let date = scheduler.currentDate.dateByAddingTimeInterval(interval)
+		let schedulerDisposable = scheduler.scheduleAfter(date) {
+			sendError(observer, error)
+		}
+
+		let signalDisposable = signal.observe(Signal.Observer { event in
+			observer.put(event)
```
kastiglione: I don't think it's necessary to wrap `observer`. This could be `signal.observe(observer)`.

almassapargali: What are you talking about?? :p (sorry, it's just shameful really)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -680,11 +680,29 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
 	}
 }
 
+/// Forwards events from `signal` until `interval`. Then if signal isn't completed yet,
+/// errors with `error` on `scheduler`.
+public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
+	precondition(interval >= 0)
```
kastiglione: Why do we allow `timeoutWithError()` to have an `interval` is exactly `0`? From the outside, it would almost seem to _guarantee_ that this signal sends the error. But knowing the implementation details, it's more like: "send error if the given signal doesn't complete synchronously". In either case, both are kind of suspicious. Can we make remove ambiguity by saying the interval has to be strictly `> 0`?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -680,11 +680,29 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
...
+
+	return Signal { observer in
+		let date = scheduler.currentDate.dateByAddingTimeInterval(interval)
+		let schedulerDisposable = scheduler.scheduleAfter(date) {
+			sendError(observer, error)
+		}
+
+		let signalDisposable = signal.observe(observer)
+
+		let dis = CompositeDisposable([ signalDisposable ])
```
kastiglione: Apologies for the nitpicky request, but for consistency can we name this `disposable`? Thanks in advance :love_letter:

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: > But knowing the implementation details, it's more like: "send error if the given signal doesn't complete synchronously".

I kind of like this use of `timeout`. What if we documented it instead?

kastiglione: > What if we documented it instead?

I was going to suggest this too. I got a bit carried away with that sentence about this being suspicious.

ReactiveCocoa/Swift/Signal.swift
almassapargali: It's totally ok :ok_hand:

ReactiveCocoa/Swift/Signal.swift
almassapargali: What would you suggest? Something like:

"If given `interval` is 0, signal errors instantly, unless it completes synchronously."?

jspahrsummers: How about something like:

> If the `interval` is 0, the timeout will be scheduled immediately. The signal must complete synchronously (or on a faster scheduler) to avoid the timeout.

kastiglione: > …errors if given signal fails to complete synchronously

almassapargali: I'm not a native here. So, you choose.

kastiglione: I like @jspahrsummers's.

almassapargali: Ok. Will commit that soon. 👍

almassapargali:
ping

jspahrsummers:
Looks good, thanks! :shipit:

almassapargali:
:+1:

Issue #1751, opened 2015-02-14
Implement repeat and retry


jspahrsummers:
https://github.com/ReactiveCocoa/ReactiveCocoa/blob/15c2c9a473f5bf09b79061f603cb122a257162e1/ReactiveCocoa/Swift/SignalProducer.swift#L539-L540

Previous implementations in [`swift-before-signal-producers`](https://github.com/ReactiveCocoa/ReactiveCocoa/releases/tag/swift-before-signal-producers) may be of use.

NachoSoto:
I'll try to do this tomorrow! They're the operators I'm most excited about.

jspahrsummers:
Thanks! I've assigned it to you, but feel free to remove yourself if you don't work on it after all.

NachoSoto:
This is finished with #1760 and #1761 :tada: 
That was fun!

PR #1758, opened 2015-02-14
allValues method for SignalProducer

NachoSoto:
I found myself needing this in order to easily write tests for #1751.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -611,6 +611,19 @@ public func single<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
 	return result
 }
 
+/// Starts the producer, then blocks, waiting for all values.
+public func allValues<T, E>(producer: SignalProducer<T, E>) -> Result<[T], E> {
+	let resultOrNil = producer
+		|> reduce([T]()) { (var previous, value) in
+			previous.append(value)
+
+			return previous
+		}
+		|> single
+
+	return resultOrNil ?? success([T]())
+}
+
 /// Starts the producer, then blocks, waiting for the last value.
 public func last<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
 	return producer |> takeLast(1) |> single
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -438,6 +438,40 @@ class SignalProducerSpec: QuickSpec {
 			}
 		}
 
+		describe("allValues") {
+			it("should start a signal then block until completion") {
+				let (producer, sink) = SignalProducer<Int, NoError>.buffer()
+
+				var result: Result<[Int], NoError>?
+
+				let group = dispatch_group_create()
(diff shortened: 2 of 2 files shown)


JaviSoto:
I literally wrote this last night also as a helper for tests. It's almost as if we were twins. 

What do you think about having an overload of this function for the case `E: NoError` where we can return `[T]`?

kastiglione:
@JaviSoto I brought that up here: #1723

NachoSoto:
Sounds good :+1:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -611,6 +611,19 @@ public func single<T, E>(producer: SignalProducer<T, E>) -> Result<T, E>? {
 	return result
 }
 
+/// Starts the producer, then blocks, waiting for all values.
+public func allValues<T, E>(producer: SignalProducer<T, E>) -> Result<[T], E> {
+	let resultOrNil = producer
+		|> reduce([T]()) { (var previous, value) in
```
NachoSoto: Maybe I should extract this into `collect` like in `RACSignal`? Then make `allValues` simply add `single`.

jspahrsummers:
One of the big problems with RAC 2.x has been its huge API surface area, because it intimidates newcomers (“What do all these things mean?! I just want to do XYZ!”) and introduces many different ways to do any given thing.

Since this can be trivially implemented in terms of `reduce`:

```swift
|> reduce([]) { $0 + [ $1 ] }
|> single
```

I think I'd rather omit this. Thanks for the suggestion, though!

NachoSoto:
Sounds completely reasonable :)

kastiglione:
> One of the big problems with RAC 2.x has been its huge API surface area, because it intimidates newcomers (“What do all these things mean?! I just want to do XYZ!”) and introduces many different ways to do any given thing

Opinion dump, please imagine "IMO"'s sprinkled all over! :trollface: 

* I think part of the problem with RAC 2 was all the classes level concepts: streams, sequences, subjects, multicast connections, etc. I know we've introduced new concepts in RAC.swift, but I have the sense it will be a much smaller API overall.
* Having said that, I also think operators are _by far_ the most valuable part of RAC, I hate to see them scrimped on, especially if indeed the API surface is already smaller as suggested above.
* Having to use `reduce([]) { $0 + [ $1 ] }` for `collect()` is a big :-1: IMO. I don't think conciseness of implementation should be a bar for whether a named operator exists. Giving common operators a semantically meaningful and readable name is a good thing.
* Optimizing for less named operators is optimizing for the new learner, but not the code.
* Is there something like a number we can aim for, a dunbar number of sorts, for the core operators? It seems there's an optimization being applied here without a way of measuring where it puts us.
* The question of "does this operator belong" came up in RAC 1, RAC 2, RAC 3, I keep thinking we need an ecosystem so some of these operators can exist and be readily available and commonly used, even if we don't want them in RAC core. I say this because many of these operators are common operators in other places, Haskell, Clojure, JS libraries, etc, so it seems that people would want them, expect them.

kastiglione:
So sorry.

jspahrsummers:
@kastiglione Those are pretty compelling points. I would like to discuss it further, but I think we should probably take it to a new thread.

TLDR for this, though: I'd be _more_ okay with `collect` than a blocking method a la `single`.

NachoSoto:
I agree with both: I think `collect` is a valuable method (even if it's redundant with `reduce`), but `allValues` is much more easily implemented by composing existing operators.

Issue #1750, opened 2015-02-14
Implement switch and switchMap


jspahrsummers:
https://github.com/ReactiveCocoa/ReactiveCocoa/blob/15c2c9a473f5bf09b79061f603cb122a257162e1/ReactiveCocoa/Swift/SignalProducer.swift#L536-L537

Previous implementations in [`swift-before-signal-producers`](https://github.com/ReactiveCocoa/ReactiveCocoa/releases/tag/swift-before-signal-producers) may be of use.

brow:
I don't think `switch` is a legal identifier in Swift. How about `switchToLatest`?

kastiglione:
`switcheroo`

brow:
You can assign to me since I have #1759 going and plan to finish today.

PR #1759, opened 2015-02-14
Implement switchToLatest and switchMap

brow:
Towards #1750.

Tested:
- [x] should forward values from the latest inner signal
- [x] should forward an error from an inner signal
- [x] should forward an error from the outer signal
- [x] should complete when the original and latest signals have completed

TODO:
- [x] Refactor specs with `SignalProducer.buffer`
- [x] Implement switchMap

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -589,7 +589,11 @@ public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 							case .Completed:
 								updateState { $0.completeInnerSignal(innerSignal) }
 							default:
-								sink.put(event)
+								state.withValue { value -> () in
+									if value.isIncompleteLatestInnerSignal(innerSignal) {
+										sink.put(event)
+									}
+								}
 							}
 						})
 					}
@@ -629,13 +633,21 @@ private struct LatestState<T, E: ErrorType> {
 	}
 	
 	func completeInnerSignal(signal: Signal<T, E>) -> LatestState<T, E> {
-		switch latestInnerSignal {
-		case .Incomplete(let latestSignal) where signal === latestSignal:
+		if isIncompleteLatestInnerSignal(signal) {
 			return LatestState(
 				outerSignalComplete: outerSignalComplete,
 				latestInnerSignal: .Complete)
-		case .Incomplete, .Complete:
+		} else {
 			return self
+		}
+	}
+	
+	func isIncompleteLatestInnerSignal(signal: Signal<T, E>) -> Bool {
+		switch latestInnerSignal {
+		case .Incomplete(let latestSignal) where signal === latestSignal:
+			return true
+		case .Incomplete, .Complete:
+			return false
 		}
 	}
 	
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -570,7 +570,7 @@ public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 			let latestInnerDisposable = SerialDisposable()
 			disposable.addDisposable(latestInnerDisposable)
 			
-			let state = Atomic(LatestState<T, E>.initial)
+			let state = Atomic(LatestState<T, E>())
 			let updateState = { (action: LatestState<T, E> -> LatestState<T, E>) -> () in
 				state.modify(action)
 				if state.value.isComplete {
@@ -607,14 +607,8 @@ public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 }
 
 private struct LatestState<T, E: ErrorType> {
-	let outerSignalComplete: Bool
-	let latestInnerSignal: LatestStateInnerSignal<T, E>
-
-	static var initial: LatestState {
-		return LatestState(
-			outerSignalComplete: false,
-			latestInnerSignal: .Complete)
-	}
+	private let outerSignalComplete = false
+	private let latestInnerSignal = LatestStateInnerSignal<T, E>.Complete
 	
 	func completeOuterSignal() -> LatestState<T, E> {
 		return LatestState(
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -617,24 +617,20 @@ private struct LatestState<T, E: ErrorType> {
 	}
 	
 	func addInnerSignal(signal: Signal<T, E>) -> LatestState<T, E> {
-		if isComplete {
-			return self
-		} else {
-			return LatestState(
+		return isComplete
+			? self
+			: LatestState(
 				outerSignalComplete: outerSignalComplete,
 				latestInnerSignal: .Incomplete(signal))
-		}
 	}
 	
 	func completeInnerSignal(signal: Signal<T, E>) -> LatestState<T, E> {
-		if isIncompleteLatestInnerSignal(signal) {
-			return LatestState(
+		return isIncompleteLatestInnerSignal(signal)
+			? LatestState(
 				outerSignalComplete: outerSignalComplete,
 				latestInnerSignal: .Complete)
-		} else {
-			return self
-		}
-	}
+			: self
+}
 	
 	func isIncompleteLatestInnerSignal(signal: Signal<T, E>) -> Bool {
 		switch latestInnerSignal {
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -564,32 +564,32 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
 /// have both completed.
 public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> SignalProducer<T, E> {
 	return SignalProducer<T, E> { sink, disposable in
-		let producerCompleted = Atomic(false)
-		let latestCompleted = Atomic(false)
+		let outerSignalCompleted = Atomic(false)
+		let latestInnerSignalCompleted = Atomic(false)
 		
 		let completeIfNecessary: () -> () = {
-			if producerCompleted.value && latestCompleted.value {
+			if outerSignalCompleted.value && latestInnerSignalCompleted.value {
 				sendCompleted(sink)
 			}
 		}
 
-		let latestDisposable = SerialDisposable()
-		disposable.addDisposable(latestDisposable)
+		let latestInnerDisposable = SerialDisposable()
+		disposable.addDisposable(latestInnerDisposable)
 
-		producer.startWithSignal { signal, producerDisposable in
-			disposable.addDisposable(producerDisposable)
+		producer.startWithSignal { outerSignal, outerDisposable in
+			disposable.addDisposable(outerDisposable)
 			
-			signal.observe(
-				next: { signal in
-					latestCompleted.value = false
+			outerSignal.observe(
+				next: { innerProducer in
+					latestInnerSignalCompleted.value = false
 					
-					signal.startWithSignal { signal, signalDisposable in
-						latestDisposable.innerDisposable = signalDisposable
+					innerProducer.startWithSignal { innerSignal, innerDisposable in
+						latestInnerDisposable.innerDisposable = innerDisposable
 						
-						signal.observe(SinkOf { event in
+						innerSignal.observe(SinkOf { event in
 							switch event {
 							case .Completed:
-								latestCompleted.value = true
+								latestInnerSignalCompleted.value = true
 								completeIfNecessary()
 								
 							default:
@@ -602,7 +602,7 @@ public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 				}, error: { error in
 					sendError(sink, error)
 				}, completed: {
-					producerCompleted.value = true
+					outerSignalCompleted.value = true
 					completeIfNecessary()
 				})
 		}
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -554,14 +554,14 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
     return producer |> map(transform) |> concat
 }
 
-/// Returns a signal that forwards values from the latest signal sent on
-/// `producer`, ignoring values sent on previous inner signals.
+/// Returns a producer that forwards values from the latest producer sent on
+/// `producer`, ignoring values sent on previous inner producers.
 ///
-/// An error sent on `producer` or the latest inner signal will be sent on the
-/// returned signal.
+/// An error sent on `producer` or the latest inner producer will be sent on the
+/// returned producer.
 ///
-/// The returned signal completes when `producer` and the latest inner signal
-/// have both completed.
+/// The returned producer completes when `producer` and the latest inner 
+/// producer have both completed.
 public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> SignalProducer<T, E> {
 	return SignalProducer<T, E> { sink, disposable in
 		producer.startWithSignal { outerSignal, outerDisposable in
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -608,12 +608,12 @@ public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> S
 
 private struct LatestState<T, E: ErrorType> {
 	private let outerSignalComplete = false
-	private let latestInnerSignal = LatestStateInnerSignal<T, E>.Complete
+	private let latestIncompleteSignal: Signal<T, E>? = nil
 	
 	func completeOuterSignal() -> LatestState<T, E> {
 		return LatestState(
 			outerSignalComplete: true,
-			latestInnerSignal: latestInnerSignal)
+			latestIncompleteSignal: latestIncompleteSignal)
 	}
 	
 	func addInnerSignal(signal: Signal<T, E>) -> LatestState<T, E> {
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -527,13 +527,62 @@ public func concat<T, E>(next: SignalProducer<T, E>)(producer: SignalProducer<T,
...
+			
+			signal.observe(
+				next: { signal in
+					latestDisposable.innerDisposable = nil
+					latestCompleted.value = false
+					
+					signal.startWithSignal { signal, signalDisposable in
+						latestDisposable.innerDisposable = signalDisposable
+						
+						materialize(signal).observe { event in
```
kastiglione: `observe()` is overridden to take a `Signal<T, E>.Observer`, aka `SinkOf<Event<T, E>>`. I'm not sure if the `materialize()` necessary, could it just take an observer/sink?

brow: Is this what you're envisioning?
```swift
signal.observe(SinkOf { event in
    switch event {
        // ...
    }
})
```

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -527,13 +527,62 @@ public func concat<T, E>(next: SignalProducer<T, E>)(producer: SignalProducer<T,
...
+								sink.put(event)
+							}
+						}
+					}
+					
+					return
+				},
+				error: { error in
+					sendError(sink, error)
+				},
```
jspahrsummers: I prefer to join these with the next line, a la the style of `-subscribeNext:error:completed:` used in RAC 2.x.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -527,13 +527,62 @@ public func concat<T, E>(next: SignalProducer<T, E>)(producer: SignalProducer<T,
...
+
+		let latestDisposable = SerialDisposable()
+		disposable.addDisposable(latestDisposable)
+
+		producer.startWithSignal { signal, producerDisposable in
+			disposable.addDisposable(producerDisposable)
+			
+			signal.observe(
+				next: { signal in
+					latestDisposable.innerDisposable = nil
```
jspahrsummers: It should be fine to wait until the assignment below (as long as `latestCompleted.value` is moved too), because that closure will be invoked before anything starts sending.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -527,13 +527,62 @@ public func concat<T, E>(next: SignalProducer<T, E>)(producer: SignalProducer<T,
 	return SignalProducer(values: [producer, next]) |> concat
 }
 
+public func switchToLatest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> SignalProducer<T, E> {
```
jspahrsummers: Could you please document this? :pray:

jspahrsummers:
Can you please also add `switchMap` here?

While we're on the subject, I'd love to talk about the names. It would be great for switch-map to have the same naming convention as `mergeMap` and `concatMap`. As you noted, though, `switch` isn't a valid identifier for our function.

@ReactiveCocoa/reactivecocoa Is there a better name for `switch[toLatest]` that still describes the semantics accurately?

ReactiveCocoa/Swift/SignalProducer.swift
kastiglione: Yeah. Although I'm not sure if we prefer that to be `SinkOf` or `Signal.Observer`.

JaviSoto:
I'm curious: `switchMap` is the `bind` variant that in RAC2 was called `flattenMap`, right? But I guess that's not a great name that expresses the different semantics between the 3 methods. Perhaps something like mapLatest...?

kastiglione:
> `switchMap` is the bind variant that in RAC2 was called `flattenMap`, right?

That was `mergeMap()`.

brow:
`select`/`selectMap`?

brow:
Or perhaps `multiplex` or `mux`? http://en.wikipedia.org/wiki/Multiplexer

ReactiveCocoa/Swift/SignalProducer.swift
brow: Signal.swift contains `signal.observe(Signal.Observer {...` twice and `signal.observe(SinkOf {...` four times. I'll go with the latter for now.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -527,13 +527,59 @@ public func concat<T, E>(next: SignalProducer<T, E>)(producer: SignalProducer<T,
...
+	return SignalProducer<T, E> { sink, disposable in
+		let producerCompleted = Atomic(false)
+		let latestCompleted = Atomic(false)
```
kastiglione: I'm not sure if this would make the code more reasonable, but this could be done with a single boolean, say `canComplete`, I think.

jspahrsummers:
> `select`/`selectMap`?

Unfortunately, `select` often itself means `map`—see Smalltalk or Rx.NET.

> Or perhaps `multiplex` or `mux`? http://en.wikipedia.org/wiki/Multiplexer

This one occurred to me too, but I'm not sure if it implies the “latest” part of the behavior. /shrug

brow:
`graft`? http://en.wikipedia.org/wiki/Climatic_Research_Unit_documents :trollface:

mdiep:
`latest`/`latestMap`?

robrix:
`select` means `filter` in Smalltalk, but that has the same problem, really. `latest`/`latestMap` sounds good to me.

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: I like multiplex.

andymatuschak:
Like latest.

robb:
:+1: `latest`/`latestMap`

kastiglione:
:heavy_plus_sign::one:  `latest`

brow:
Being an adjective, `latest` kinda obscures the parallel with `concat[enate]` and `merge` and implies the same type as `last`. But I suppose that's what type signatures are for, and `latest` does best capture how the function differs from others with the same type sig.

I'm renaming to `latest` due to overwhelming support, and I think all other comments have been addressed as well.

brow:
Oops, need to resolve conflicts and document `latestMap`.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -549,6 +549,67 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+		}
+
+		let latestDisposable = SerialDisposable()
+		disposable.addDisposable(latestDisposable)
+
+		producer.startWithSignal { signal, producerDisposable in
+			disposable.addDisposable(producerDisposable)
+			
+			signal.observe(
+				next: { signal in
```
neilpa: Shadowing `signal` here and below makes this code a bit harder to reason over. Especially since this one is a `SignalProducer` and not a `Signal`. Maybe `innerProducer` or `latestProducer` here and similarly with its produced signal below

brow: Thanks for bringing attention to this, @neilpa. I was using [swift-before-signal-producers](https://github.com/ReactiveCocoa/ReactiveCocoa/releases/tag/swift-before-signal-producers) as a guide, and we use shadowing liberally there.

The argument I can imagine in favor of shadowing is that it ensures the shadowed referent isn't used inside the scope where it is shadowed, which makes the code easier to reason over in a way.

Re: referring to `SignalProducer`s as "signals", there's a lot of precedent all over the post-`ColdSignal` codebase, but I'm not sure if it's deliberate. 

I'd love to get more opinions from @ReactiveCocoa/reactivecocoa on both of these practices.

brow:
```
    ✗ -[RACSchedulerSpec _scheduler__should_cancel_future_blocks_when_disposed] (38 ms) (0)
```
Works on my machine? :hankey:

ReactiveCocoa/Swift/SignalProducer.swift
andymatuschak: I find it easier to reason about code which doesn't shadow (except in the case of `let x: Int?; if let x = x…`).

I also think we should not refer to `SignalProducer`s as "signals"—I suspect the other examples are unintentional.

neilpa: The other operators on producers of producers (`merge` and `concat`) are explicit with naming and avoid shadowing.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -549,6 +549,67 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+					latestCompleted.value = false
+					
+					signal.startWithSignal { signal, signalDisposable in
+						latestDisposable.innerDisposable = signalDisposable
+						
+						signal.observe(SinkOf { event in
+							switch event {
+							case .Completed:
+								latestCompleted.value = true
```
kastiglione: Is there a race here? Consider when the outer producer and inner producers are sending on mutually separate threads. The outer producer could send a new producer and set `latestCompleted` to false. Then, before the current active inner producer is disposed, the inner producer sends completed, which sets `latestCompleted` to true, which is not correct. If the outer producer then completes before this inner producer has completed, the inner producer will be prematurely cut off.

neilpa: Yea, @jspahrsummers [mentioned this before](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1759#discussion_r24719355). Setting `lastCompleted` needs to happen after disposal of the previous signal.

kastiglione: Thanks. I saw that, but didn't pay close enough attention to see what it was saying.

> Setting `lastCompleted` needs to happen after disposal of the previous signal.

Is that sufficient to prevent a theoretical race? Even though the previous signal has been disposed, disposal doesn't interrupt any active observation callback. There's no _guarantee_ that the 1. completion event starts being handled, 2. it gets disposed, 3. `lastCompleted` set to false, 4. completion event sets `lastCompleted` to true.

neilpa: Ugh, you're right. Checking that `signalDisposable.disposed` is false before setting `latestCompleted` should remedy that. However, that introduces a third synchronization object making this code a bit hairy. 

Alternatively, we could package the two completed bools and the latest signal in a `LatestState` struct. Wrap that in an `Atomic` so that we only have one lock to consider. Then in the inner completion handler check if it's still the latest signal and only set `latestCompleted` if so.

kastiglione:
I can't remember if we ever tried this in RAC.m, but I feel like I remember an implementation of `switch()` on top of `merge()`. I haven't given the corner cases any thought, but what about this:

```swift
public func latest<T, E>(producer: SignalProducer<SignalProducer<T, E>, E>) -> SignalProducer<T, E> {
	var previousObserver = Signal<T, E>.Observer { _ in }

	return producer |> mergeMap { innerProducer in
		SignalProducer { latestObserver, latestDisposable in
			sendCompleted(previousObserver)
			previousObserver = latestObserver

			innerProducer.startWithSignal { signal, signalDisposable in
				latestDisposable.addDisposable(signalDisposable)
				signal.observe(latestObserver)
			}
		}
	}
}
```

I used this with the tests from this pull request, and they passed, for whatever that's worth.

jspahrsummers:
@kastiglione That won't work if the returned producer is started multiple times separately. Signal producer operators shouldn't close over any external state.

neilpa:
@jspahrsummers There should probably be a test for this case as well

jspahrsummers:
Agreed!

kastiglione:
@jspahrsummers duh, thanks.

brow:
Moving back to WIP for:
- [ ] Test case and fix for [race in `latest`](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1759#discussion_r24783661) 
- [ ] Don't refer to `SignalProducer`s as "signals" in identifiers and comments
- [ ] Avoid gratuitous shadowing

ReactiveCocoa/Swift/SignalProducer.swift
brow: Oh, also, `latestCompleted` never gets set if the outer producer completes before sending any producers :see_no_evil:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -554,6 +554,124 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+						latestInnerDisposable.innerDisposable = innerDisposable
+						state.value = state.value.addInnerSignal(innerSignal)
+						
+						innerSignal.observe(SinkOf { event in
+							switch event {
+							case .Completed:
+								updateState { $0.completeInnerSignal(innerSignal) }
+							default:
+								state.withValue { value -> () in
+									if value.isIncompleteLatestInnerSignal(innerSignal) {
```
brow: @kastiglione @neilpa Based on https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1759#discussion_r24790861, I think this check is necessary to guarantee an event is never forwarded from a non-latest inner signal. Agree?

brow:
Ping. Especially @kastiglione and @neilpa to see if I covered the race conditions they had in mind.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -554,6 +554,124 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+}
+
+private struct LatestState<T, E: ErrorType> {
+	let outerSignalComplete: Bool
+	let latestInnerSignal: LatestStateInnerSignal<T, E>
+
+	static var initial: LatestState {
+		return LatestState(
+			outerSignalComplete: false,
+			latestInnerSignal: .Complete)
```
neilpa: Set these at the default values and remove this

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -554,6 +554,124 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+		producer.startWithSignal { outerSignal, outerDisposable in
+			disposable.addDisposable(outerDisposable)
+
+			let latestInnerDisposable = SerialDisposable()
+			disposable.addDisposable(latestInnerDisposable)
+			
+			let state = Atomic(LatestState<T, E>.initial)
+			let updateState = { (action: LatestState<T, E> -> LatestState<T, E>) -> () in
+				state.modify(action)
+				if state.value.isComplete {
```
neilpa: There's an overload of `modify` that allows for returning user data. You should use that to return the completion state.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -554,6 +554,124 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+				}, completed: {
+					updateState { $0.completeOuterSignal() }
+				})
+		}
+	}
+}
+
+private struct LatestState<T, E: ErrorType> {
+	let outerSignalComplete: Bool
+	let latestInnerSignal: LatestStateInnerSignal<T, E>
```
neilpa: I would shorten this and related names `latestSignal`, `inner` is redundant.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -554,6 +554,124 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+										sink.put(event)
+									}
+								}
+							}
+						})
+					}
+				}, error: { error in
+					sendError(sink, error)
+				}, completed: {
+					updateState { $0.completeOuterSignal() }
```
neilpa: Rather than a method for this can you call the initializer directly

ReactiveCocoa/Swift/SignalProducer.swift
neilpa: Yea, although I would forgo the method and do the check explicitly here.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -554,6 +554,124 @@ public func concatMap<T, U, E>(transform: T -> SignalProducer<U, E>)(producer: S
...
+		switch latestInnerSignal {
+		case .Complete:
+			return outerSignalComplete
+		case .Incomplete:
+			return false
+		}
+	}
+}
+
+private enum LatestStateInnerSignal<T, E: ErrorType> {
```
neilpa: In this context I would use `latestCompleted` bool instead. You're not gaining anything from the enum and it would simplify the above cases where your forced into pattern matching.

ReactiveCocoa/Swift/SignalProducer.swift
brow: I'm not seeing a way to use that version of `modify` here that's less verbose or more clear than what I've got. How would you do it?

ReactiveCocoa/Swift/SignalProducer.swift
brow: I didn't intent the memberwise initializer to be public, and in forthcoming commits it isn't. The idea  is to use methods on `LatestState` expose only valid state transitions.

ReactiveCocoa/Swift/SignalProducer.swift
brow: The enum, compared to separate `latestCompleted` and `latest[Inner]Signal` vars, does prove that we don't use a reference to an inner signal after it completes and avoids holding that reference.

Also, it may be a matter of taste, but I like the explictness of these `switch` statements :smile:

ReactiveCocoa/Swift/SignalProducer.swift
neilpa: I don't see the need for the extra level of indirection here. This is already a private struct just for this method. Hiding the transitions behind the methods made the code harder to grok. I had to keep bouncing back and forth while reviewing.

ReactiveCocoa/Swift/SignalProducer.swift
neilpa: Alternatively you could use an optional latest signal to capture the same semantics

kastiglione:
@brow Apologies but I won't be able to look in time before it will presumably be merged.

jspahrsummers:
I agree that the state transitions on the `struct` are a bit confusing and obfuscating, at least to my :eyes:. It also differs from how other composite operators are implemented.

Can you move the state transitions back into the operator implementation itself? Leaving introspective capabilities on the state structure makes sense, though.

ReactiveCocoa/Swift/SignalProducer.swift
brow: Oh, I like that. I'll try it.

brow:
I think we're done here

PR #1760, opened 2015-02-14
Implemented times for SignalProducer

NachoSoto:
Completes part of #1751.


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,21 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
 public func switchMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalProducer<T>) -> SignalProducer<U>
 
-public func repeat<T>(count: Int)(producer: SignalProducer<T>) -> SignalProducer<T>
 public func retry<T>(count: Int)(producer: SignalProducer<T>) -> SignalProducer<T>
 */
 
+/// Repeats `producer` `count` times.
+/// Repeating `0` times results in a equivalent signal producer.
+public func repeat<T, E>(count: Int)(producer: SignalProducer<T, E>) -> SignalProducer<T, E> {
+	precondition(count >= 0)
```
NachoSoto: Alternatively I can do
```swift
switch count {
case 0:
	return producer
case let count where count > 0:
	return producer |> concat(producer |> repeat(count - 1))
default:
	fatalError("Invalid count: \(count)")
}
```

Let me know what you prefer :)

jspahrsummers: A `precondition` like this is fine.

This actually raises an interesting question, though. My mental model of `count` was “the number of times you want to see the signal events,” but you've interpreted it as “the number of _additional_ times.”

The problem is that I don't think either interpretation is actually wrong. Maybe we should call this operator `times()` or something instead, so that it's clear which semantics it has?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,21 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
+
+	if count == 0 {
+		return producer
+	} else {
+		return producer |> concat(producer |> repeat(count - 1))
```
jspahrsummers: This is clever, but for a huge count (e.g., `Int.max`) will assumably blow the stack, since `repeat()` recursively invokes itself immediately.

Can you refactor this to use iteration instead, and add a test that verifies that an overflow doesn't occur with an `Int.max` repeat?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -299,10 +299,52 @@ class SignalProducerSpec: QuickSpec {
 		}
 
 		describe("repeat") {
-			pending("should start a signal N times upon completion") {
+			it("should start a signal N times upon completion") {
+				let original = SignalProducer<Int, NoError>(values: [1, 2, 3])
```
jspahrsummers: Can you please add spaces inside these array literals?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -299,10 +299,52 @@ class SignalProducerSpec: QuickSpec {
...
+			it("should not repeat upon error") {
+				let results: [Result<Int, TestError>] = [
+					success(1),
+					success(2),
+					failure(.Default)
+				]
+
+				var operationIndex = 0
+				let operation: () -> Result<Int, TestError> = {
+					assert(operationIndex < results.count, "Operation started too many times")
```
jspahrsummers: Can this be an `expect()` instead? An assertion will probably bring down the test runner.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -299,10 +299,52 @@ class SignalProducerSpec: QuickSpec {
...
+				let expectedEvents: [Event<Int, TestError>] = [
+					.Next(Box(1)),
+					.Next(Box(2)),
+					.Error(Box(.Default))
+				]
+
+				// Can't test for equality because Array<T> is not Equatable,
+				// and neither is Event<T, E>.
+				expect(result.count).to(equal(expectedEvents.count))
+				expect(result[0] == expectedEvents[0]).to(beTruthy())
```
jspahrsummers: Why can't this use `to(equal())`?

jspahrsummers: Could we just do this over the whole arrays instead?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -299,10 +299,52 @@ class SignalProducerSpec: QuickSpec {
...
+				let operation: () -> Result<Int, TestError> = {
+					assert(operationIndex < results.count, "Operation started too many times")
+
+					return results[operationIndex++]
+				}
+
+				let original = SignalProducer.try(operation)
+				let producer = original |> repeat(2)
+
+				let result = (producer |> materialize |> allValues).value!
```
jspahrsummers: We should avoid force-unwrapping in tests, because any failure will bring down the rest runner.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: Notice my comment, it's not possible :/

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: Oh good call

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: Is there an equivalent to `except` that actually marks the test case as a failure and stops? If I just replace this with `except` the next line would be an out of bounds runtime crash.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: :+1:

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: That's a good point, I think `times` is a good name for it :)

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: Never mind, I thought of a way to write this more concisely.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
jspahrsummers: Hrm, I don't know. I guess I would just change this into a conditional, then, and do `expect(false)` or something silly in the alternate case.

NachoSoto: :+1:

almassapargali: There is fail() afaik

NachoSoto:
This is ready for review again :recycle:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -298,14 +298,58 @@ class SignalProducerSpec: QuickSpec {
...
+		describe("times") {
+			it("should start a signal N times upon completion") {
+				let original = SignalProducer<Int, NoError>(values: [ 1, 2, 3 ])
+				let producer = original |> times(3)
+
+				let result = producer |> reduce([]) { $0 + [$1] } |> single
+				expect(result?.value).to(equal([ 1, 2, 3, 1, 2, 3, 1, 2, 3 ]))
+			}
+
+			it("should produce an equivalent signal producer if count is 0") {
```
kastiglione: "count is _1_"

neilpa:
Continuing my comment from [here](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1762#issuecomment-74437536). Instead of `concat`, `times` should simply wrap the producer and restart it on completed events until `count` is exhausted.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: Good catch, changed!

jspahrsummers:
I agree with @neilpa. Could you please make that change, @NachoSoto?

NachoSoto:
I don't mind at all, but it's unfortunate that we need to implement it from scratch instead of relying on existing operators as an optimization. Or is there another reason I'm not seeing?

jspahrsummers:
I don't think “not crashing from stack overflow” can be classified as merely an optimization. :stuck_out_tongue:

NachoSoto:
I was thinking that it might be more appropriate to make concat not overflow, but now that I think about it, it's not common to concat many producers, but it might be very common to make a producer repeat infinitely, and that shouldn't be O(n).

Thanks guys! I'll make the change today :)

neilpa:
I actually did this yesterday to make sure what I was suggesting worked - neilpa/ReactiveCocoa@81c2cf6d2da0e8c09dc77a082dd1a353b00a1ffc.  Go ahead and take that if you want.

NachoSoto:
@neilpa haha, I had just finished writing it :) my implementation is slightly different, let me know if it's correct as well.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,41 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
 
-public func repeat<T>(count: Int)(producer: SignalProducer<T>) -> SignalProducer<T>
 public func retry<T>(count: Int)(producer: SignalProducer<T>) -> SignalProducer<T>
 */
 
+/// Repeats `producer` a total of `count` times.
+/// Repeating `1` times results in a equivalent signal producer.
+public func times<T, E>(count: Int)(producer: SignalProducer<T, E>) -> SignalProducer<T, E> {
+	precondition(count >= 1)
+
```
neilpa: Could short circuit on `1` and return `producer`

NachoSoto: True! changed

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,41 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
+			signal.observe(next: { value in
+				sendNext(observer, value)
+			}, error: { error in
+				sendError(observer, error)
+			}, completed: {
+				if count == 1 {
+					sendCompleted(observer)
+				} else {
+					let nextSignal = producer |> times(count - 1)
```
neilpa: `nextProducer` since this isn't a `Signal`

neilpa: Actually you don't need this at all. There's a `startWithSignal` free function so you can chain it with `|>`.

neilpa:
I'm not 100% on the `Disposable`s management but I think it's correct. I also like the way you handled the inner repeat over my approach of recursively calling the original `startWithSignal` closure.

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Oh neat!

NachoSoto:
Addressed @neilpa's comments. The implementation is much cleaner now :)

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,39 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
+
+		producer.startWithSignal { signal, signalDisposable in
+			serialDisposable.innerDisposable = signalDisposable
+
+			signal.observe(next: { value in
+				sendNext(observer, value)
+			}, error: { error in
+				sendError(observer, error)
+			}, completed: {
+				producer |> times(count - 1) |> startWithSignal { signal, signalDisposable in
```
neilpa: Current style is to split these across multiple lines ([see single](https://github.com/NachoSoto/ReactiveCocoa/blob/signal-producer-repeat/ReactiveCocoa/Swift/SignalProducer.swift#L621-L637)) unless they're one-liners.

neilpa:
One last note

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,41 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
 public func switchMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalProducer<T>) -> SignalProducer<U>
 
-public func repeat<T>(count: Int)(producer: SignalProducer<T>) -> SignalProducer<T>
 public func retry<T>(count: Int)(producer: SignalProducer<T>) -> SignalProducer<T>
 */
 
+/// Repeats `producer` a total of `count` times.
+/// Repeating `1` times results in a equivalent signal producer.
+public func times<T, E>(count: Int)(producer: SignalProducer<T, E>) -> SignalProducer<T, E> {
+	precondition(count >= 1)
```
jspahrsummers: I kinda think 0 should be equivalent to returning `empty`. Does that seem reasonable?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,41 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
+				sendNext(observer, value)
+			}, error: { error in
+				sendError(observer, error)
+			}, completed: {
+				producer
+					|> times(count - 1)
+					|> startWithSignal { signal, signalDisposable in
+						serialDisposable.innerDisposable = signalDisposable
+						signal.observe(observer)
+				}
```
jspahrsummers: Can you please fix the indentation here?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,41 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
+		producer.startWithSignal { signal, signalDisposable in
+			serialDisposable.innerDisposable = signalDisposable
+
+			signal.observe(next: { value in
+				sendNext(observer, value)
+			}, error: { error in
+				sendError(observer, error)
+			}, completed: {
+				producer
+					|> times(count - 1)
```
jspahrsummers: It still doesn't seem like we need to be recursing here, especially because each recursion adds more memory footprint.

Can this be rewritten to use a `var` which tracks how many iterations are left, and simply invoke the outer `startWithSignal` again when there's more to do?

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Yeah makes sense!

ReactiveCocoaTests/Swift/SignalProducerSpecHelpers.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//  Created by Nacho Soto on 2/14/15.
+//  Copyright (c) 2015 GitHub. All rights reserved.
+//
+
+import ReactiveCocoa
+import LlamaKit
+import Nimble
+
+extension SignalProducer {
```
jspahrsummers: Can this please go into the actual spec file?

ReactiveCocoaTests/Swift/SignalProducerSpecHelpers.swift
```diff
@@ -0,0 +1,34 @@
...
+	/// Creates a producer that can be started as many times as elements in `results`.
+	/// Each signal will immediately send either a value or an error.
+	static func tryWithResults<C: CollectionType where C.Generator.Element == Result<T, E>, C.Index.Distance == Int>(results: C) -> SignalProducer<T, E> {
```
jspahrsummers: With the above change, can you please make this `private`?

jspahrsummers:
Sorry, I found a few more things. :stuck_out_tongue_closed_eyes:

ReactiveCocoaTests/Swift/SignalProducerSpecHelpers.swift
NachoSoto: Absolutely! I wasn't sure what the best place for this was

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Done!

NachoSoto:
I think I addressed all the comments :8ball:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -552,10 +552,51 @@ public func mergeMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalPr
...
+			producer.startWithSignal { signal, signalDisposable in
+				serialDisposable.innerDisposable = signalDisposable
+
+				signal.observe(next: { value in
+					sendNext(observer, value)
+				}, error: { error in
+					sendError(observer, error)
+				}, completed: {
+					if --remainingTimes > 0 {
+						iterate()
```
jspahrsummers: I think this will strongly capture the block within itself, creating a leak.

You might want @andymatuschak's [implementation of the Z combinator](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/63f55928b5cc613433bfe252cc66ee5ce69df8bd/ReactiveCocoa/Swift/SignalProducer.swift#L535-L539) instead.

jspahrsummers:
:soon:

ReactiveCocoa/Swift/SignalProducer.swift
neilpa: Or @robrix's [`fix`](https://github.com/robrix/Prelude/blob/master/Prelude/Fix.swift) which is what I had used previously. Either also has the added benefit of avoiding the two step initialization dance.

robrix: `Z` and `fix` are exactly equivalent, I think.

NachoSoto:
Does [this](https://github.com/NachoSoto/ReactiveCocoa/commit/7e9063081082dbf8167167c02ce821ded758d446?w=0) look good?

jspahrsummers:
Yup, that should do it! Just waiting on CI and then I'll merge this.

NachoSoto:
Thank you!
Let me know if you think we should move the `fix` function somewhere else, happy to change that!

NachoSoto:
:roller_coaster:? :D

jspahrsummers:
Derp, sorry!

PR #1761, opened 2015-02-15
Implemented retry for SignalProducer

NachoSoto:
Finishes #1751.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -612,7 +612,6 @@ TODO
 public func switch<T>(producer: SignalProducer<SignalProducer<T>>) -> SignalProducer<T>
 public func switchMap<T, U>(transform: T -> SignalProducer<U>)(producer: SignalProducer<T>) -> SignalProducer<U>
 
-public func retry<T>(count: Int)(producer: SignalProducer<T>) -> SignalProducer<T>
 */
 
 /// Repeats `producer` a total of `count` times.
@@ -653,6 +652,19 @@ public func times<T, E>(count: Int)(producer: SignalProducer<T, E>) -> SignalPro
 		}
 
 		iterate()
+	}
+}
+
+/// Ignores errors up to `count` times.
+public func retry<T, E>(count: Int)(producer: SignalProducer<T, E>) -> SignalProducer<T, E> {
+	precondition(count >= 0)
+
+	if count == 0 {
+		return producer
+	} else {
+		return producer |> catch { _ in
+			producer |> retry(count - 1)
+		}
 	}
 }
 
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -433,13 +433,48 @@ class SignalProducerSpec: QuickSpec {
 		}
 		
 		describe("retry") {
-			pending("should start a signal N times upon error") {
+			it("should start a signal N times upon error") {
+				let results: [Result<Int, TestError>] = [
+					failure(.Error1),
+					failure(.Error2),
+					success(1)
+				]
+
+				let original = SignalProducer.tryWithResults(results)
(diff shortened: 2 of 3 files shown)


jspahrsummers:
I'll let this sit for now, as some of the notes for #1760 apply here too.

NachoSoto:
Thanks! Working on ~~`repeat`~~ `times` now, I'll rebase this when done.

NachoSoto:
Rebased and ready!

jspahrsummers:
This needs the same improvements made to `times`. :pray:

NachoSoto:
Rebasing now! :)

jspahrsummers:
I mean that corresponding changes should be made to `retry` too. :wink:

Also, please don't rebase! Merge instead.

NachoSoto:
I'll merge from now on, but I have to rebase in this case because this has commits that I removed in the other branch.

Is the implementation itself not correct in this case?

jspahrsummers:
This can suffer the same O(n) growth that we discussed for `times` in #1760. It should use a counter instead.

NachoSoto:
```swift
fit("does not overflow") {
	let original: SignalProducer<Int, TestError> = SignalProducer.try {
		return failure(.Default)
	}

	let producer = original |> retry(Int.max)
	let result = producer |> single

	expect(result?.error).to(equal(TestError.Default))
}
```

I see now :) because catch calls the closure synchronously.
Aren't you worried about not trying to reuse existing operators instead of implementing them all from scratch?

I'll fix this tonight, in any case!

jspahrsummers:
> Aren't you worried about not trying to reuse existing operators instead of implementing them all from scratch?

If you can find a way to reuse `catch` without creating O(n) allocations, I would love to reuse it. :wink:

NachoSoto:
Looking at `catch` implementation now I realize it's not as trivial as I was thinking :P

neilpa:
Instead, you may be able to abstract away the recursive bits from `times` to create a more general way for restarting producers of producers. A factory of sorts for the `startWithHandler` closure.

NachoSoto:
@neilpa you gave me a good idea to reuse the implementation between `retry` and `times`, since they're essentially the same except error VS completed.

I tried implementing it again to check if the test passed, but unfortunately an implementation like this (with the test I wrote above) still overflows the stack if everything is synchronous:
```swift
/// Ignores errors up to `count` times.
public func retry<T, E>(count: Int)(producer: SignalProducer<T, E>) -> SignalProducer<T, E> {
	precondition(count >= 0)

	if count == 0 {
		return producer
	} else {
			let serialDisposable = SerialDisposable()
			disposable.addDisposable(serialDisposable)

			var remainingTimes = count

			let iterate = fix { recur in
				{
						serialDisposable.innerDisposable = signalDisposable

						signal.observe(next: { value in
							sendNext(observer, value)
						}, error: { error in
							if remainingTimes-- > 0 {
								recur()
							} else {
								sendError(observer, error)
							}
						}, completed: {
							sendCompleted(observer)
						})
					}
				}
			}

			iterate()
		}
	}
}
```

I can't think of an alternative implementation right now :/

neilpa:
This is unavoidable without an async hop via a scheduler to force the stack to unwind. RAC 2 had [`-scheduleRecursiveBlock:`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/Documentation/DesignGuidelines.md#avoid-stack-overflow-from-deep-recursion) for this problem and maybe that's the answer here. However, that precludes synchronous event delivery. It's always possible to turn a `Signal` async with `observeOn` but you cant go the other way once you've lost it.

jspahrsummers:
I think it's fine if an infinite, synchronous signal would overflow the stack. The difference from RAC 2 is that we have the capability to interrupt a synchronous signal.

My main concern is preventing O(n) _allocations_.

NachoSoto:
In that case, the current implementation meets that criterion :) Memory remains constant in each iteration.

jspahrsummers:
Hm, okay.

PR #1763, opened 2015-02-15
Add `combineLatest` and `zip` over 2–10 signals or producers

### LINKED ISSUE #1747: Re-add overloaded combineLatest() and zip() over multiple arguments
It's useful to be able to `combineLatest()` and `zip()` with [an arbitrary number of signals](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-before-signal-producers/ReactiveCocoa/Swift/ColdSignal.swift#L1023-L1129), rather than join (and unpack) them one-by-one.

jspahrsummers:
_Alternate title:_ “Why We Need Metaprogramming”

Resolves #1747.


neilpa:
What do you think about separating these sorts of overloads into another file? This is a decent amount of boilerplate that might be worth splitting out so that `Signal(Producer).swift` are more focused.

jspahrsummers:
Ideally our interfaces should be viewable in a format that doesn't include implementations at all, whether that's the built-in Swift “header” generator, a documentation generator like [jazzy](https://github.com/realm/jazzy), or something else entirely.

With that in mind, I don't think we should structure source files for easy API lookup. They should be more about conceptually related APIs.

neilpa:
Makes sense. You'll probably need to dupe you're documentation comments on all the overloads though.

PR #1767, opened 2015-02-20
UITextField -rac_textSignal and -rac_newTextChannel send values on all e...

infimache:
...diting events

Sending values on <code>UIControlEventAllEditingEvents</code> as opposed to <code>UIControlEventEditingDidBegin</code> and <code>UIControlEventEditingDidBegin</code> would resolve following issue:

1. edit text field
2. press return key, while auto completion changes text
3. resignFirstResponder
```objc
- (BOOL) textFieldShouldReturn:(UITextField *)textField {
    [textField resignFirstResponder];
    return YES;
}
```
then new value is not send via <code>-rac_textSignal</code>.

I am aware, that proposed solution has disadvantage of sending (in most cases) last value twice, but IMHO it is irrelevant in most cases, and applying some kind of filtering would unnecessarily impact performance.

### DIFF
--- a/ReactiveCocoa/UITextField+RACSignalSupport.h
+++ b/ReactiveCocoa/UITextField+RACSignalSupport.h
@@ -14,15 +14,15 @@
 @interface UITextField (RACSignalSupport)
 
 /// Creates and returns a signal for the text of the field. It always starts with
-/// the current text. The signal sends next when the UIControlEventEditingChanged
-/// or UIControlEventEditingDidBegin control event is fired on the control.
+/// the current text. The signal sends next when the UIControlEventAllEditingEvents
+/// control event is fired on the control.
 - (RACSignal *)rac_textSignal;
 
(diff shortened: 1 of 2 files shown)


jspahrsummers:
Thanks for the pull request! It almost seems like we should be watching for `ValueChanged` instead. Does that get triggered from edit events?

infimache:
Unfortunatelly, <code>ValueChanged</code> does not get triggered.

jspahrsummers:
Alright, well, this fix makes sense to me. :+1:

I don't think we'd break existing consumers, since the frequency of firing shouldn't really be relied upon, so I think this would be okay to incorporate in the next minor or patch release.

Does anyone in @ReactiveCocoa/reactivecocoa feel differently?

kastiglione:
> Does anyone feel differently?

nope

jspahrsummers:
:cool:

Thanks @infimache! :shipit:

PR #1772, opened 2015-02-22
Replace merge/concat/latest with one join() operator

### LINKED ISSUE #1768: Combine merge/latest/concat into one operator with a “strategy” parameter?
I've been thinking about combining `merge`, `concat`, and `latest` (and their corresponding mapping variants) into one operator—perhaps called `join`—that accepts a choice of joining strategy.

Although it's generally frowned upon to use parameters to _completely_ change the behavior of a function, I think it could be beneficial in this case as a way to force consumers to think about which strategy is most appropriate (vs. RAC 2.x, where almost everyone uses `flattenMap` without thinking about it).

Thoughts?

jspahrsummers:
Resolves #1768.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -631,7 +631,7 @@ extension JoinStrategy: Printable {
 /// Joins together the inner producers sent upon `producer` according to the
 /// semantics of the given strategy.
 ///
-/// If `producer` or any of the inner producers emit an error, the returned
+/// If `producer` or an active inner producer emits an error, the returned
 /// producer will forward that error immediately.
 public func join<T, E>(strategy: JoinStrategy)(producer: SignalProducer<SignalProducer<T, E>, E>) -> SignalProducer<T, E> {
 	switch strategy {
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -581,6 +581,53 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
 	}
 }
 
+/// Describes how multiple signals or producers should be joined together.
+public enum JoinStrategy: Equatable {
+	/// The signals should be merged, so that any value received on any of the
+	/// input signals will be forwarded immediately to the output signal.
+	///
+	/// The resulting signal will complete only when all inputs have completed.
+	case Merge
+
+	/// The signals should be concatenated, so that their values are sent in the
+	/// order of the signals themselves.
+	///
+	/// The resulting signal will complete only when all inputs have completed.
+	case Concat
+
+	/// Only the events from the latest input signal should be considered for
+	/// the output. Any signals received before that point will be disposed of.
+	///
+	/// The resulting signal will complete only when the signal-of-signals and
+	/// the latest signal has completed.
+	case SwitchToLatest
+}
+
+public func == (lhs: JoinStrategy, rhs: JoinStrategy) -> Bool {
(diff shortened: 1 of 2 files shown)


NachoSoto:
Looks great!

JaviSoto:
I like it!

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -581,6 +581,80 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+	///
+	/// The resulting signal will complete only when all inputs have completed.
+	case Concat
+
+	/// Only the events from the latest input signal should be considered for
+	/// the output. Any signals received before that point will be disposed of.
+	///
+	/// The resulting signal will complete only when the signal-of-signals and
+	/// the latest signal has completed.
+	case SwitchToLatest
```
mdiep: Could we still call this just _Latest_?

neilpa: :+1: like the succinctness and symmetry to the other values

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -581,6 +581,80 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
 	}
 }
 
+/// Describes how multiple signals or producers should be joined together.
+public enum JoinStrategy: Equatable {
+	/// The signals should be merged, so that any value received on any of the
+	/// input signals will be forwarded immediately to the output signal.
+	///
+	/// The resulting signal will complete only when all inputs have completed.
+	case Merge
```
joshaber: In RAC 2, `-flatten:` takes a max number of signals to flatten concurrently (https://github.com/ReactiveCocoa/ReactiveCocoa/blob/54098f8b457d5fa6b2a285d790c2c07c099b98f5/ReactiveCocoa/RACSignal%2BOperations.h#L221). Would we want to parameterize `.Merge` with that?

mdiep: I think this would fit better as a separate case. _Limit_, maybe?

neilpa: Or `.Concurrent`

jspahrsummers: I agree that a separate case seems more appropriate.

Also, I wasn't planning on reimplementing that operator for the first Swift release. :flushed:

jspahrsummers:
Ready for review and merge.

NachoSoto:
:shipit:

kastiglione:
I haven't reviewed but a :+1: to a cursor look at the use in the tests :smiley:.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -581,6 +581,80 @@ public func catch<T, E, F>(handler: E -> SignalProducer<T, F>)(producer: SignalP
...
+			return "latest"
+		}
+	}
+}
+
+/// Joins together the inner producers sent upon `producer` according to the
+/// semantics of the given strategy.
+///
+/// If `producer` or any of the inner producers emit an error, the returned
+/// producer will forward that error immediately.
```
brow: This isn't quite true. Under the "latest" strategy, an error emitted by an inner producer that is not the latest inner producer is not forwarded.

jspahrsummers:
@brow Thanks for [catching that](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1772#discussion_r25560851)!

Since this got some :+1:s, I'm gonna self-merge it.

PR #1775, opened 2015-02-23
Swift 1.2b3

### LINKED ISSUE #1753: Upgrade to Swift 1.2
Depends upon https://github.com/LlamaKit/LlamaKit/pull/27.

robrix:
- [x] What on earth is going on with the `value`/`setValue:` “overrides” in `DynamicProperty`?
- [x] Depends on https://github.com/jspahrsummers/xcconfigs/pull/33
- [x] Runtime crash at `ActionSpec.swift:148`? _What_ crash at `Actio*********ENDSIGNAL`
- [x] Runtime crash at `SignalSpec.swift:925` in `delay`.
- [x] Runtime crash in `-[ReactiveCocoa_MacTests.SignalSpec timeoutWithError__should_error_if_not_completed_before_the_interval_has_elapsed]`

Fixes #1753.


NachoSoto:
Did you try adding `@objc` to the property?

jspahrsummers:
Can you delete `.travis.yml` on this branch, so Travis stops trying to build it?

robrix:
@NachoSoto Yes, it’s a no-go.

I still have no idea where the mystery `value` property/method is coming from.

robrix:
Narrowed it down a little: it still occurs if I delete all of the dependencies, all of the tests, and all of the Objective-C sources.

robrix:
Or not. Deleting _everything_ but `PropertyType` and `DynamicProperty` still triggers this build error, except not in a new file. Something in the build settings…?

robrix:
It’s Cocoa. Attempting to compile this:

```swift
import Cocoa
class C: NSObject {
	var value: Int { return 0 }
}
```

produces this error:

```
value.swift:3:17: error: getter for 'value' overrides Objective-C method 'value' from superclass 'NSObject'
        var value: Int { return 0 }
                       ^
<unknown>:0: note: Objective-C method 'value' defined by method 'value()' here
```

robrix:
Evidently this is [not a bug](https://twitter.com/UINT_MIN/status/570052388796522497), although I’m extremely skeptical of arbitrary informal protocols breaking `NSObject` subclasses, especially when the methods (NB: _not_ property! Hence why `override` fails, I suppose?) in question have the same type.

I don’t anticipate a workaround at this point (short of splitting all the AppKit stuff into a separate framework—and even then who knows if it would work?) so we should consider renaming.

robrix:
@jspahrsummers Do you have any thoughts on the `value` name problem? The need to pick a name which does not now, nor will ever conflict with any informal protocol in Cocoa scares the bejeezus out of me, but I’ve no reason to believe the Swift team will roll this back.

alanjrogers:
@robrix would good ol' `rac_value/rac_setValue` work?

jspahrsummers:
`value` is part of the `PropertyType` protocol, which I'd rather not change just for this lulz.

What if we:

 1. Created an Objective-C superclass for `DynamicProperty`
 1. Implemented `value` of the superclass in terms of another _unimplemented_ property
 1. Implemented the latter in `DynamicProperty`, adding conformance to `PropertyType` at the same time

robrix:
@jspahrsummers Unfortunately, that fails at step 3: `PropertyType` has to be conformed to in its entirety by the class declaring conformance (according to the rules of Swift 1.2b2 today).

I have [filed a radar](http://www.openradar.me/19961299) about this debacle. In the meantime it appears we have no choice but to a) rename, or b) sit on our hands and hope.

robrix:
A good moment, I think, for a gif which hubot found today:

robrix:
https://twitter.com/jckarter/status/570736798264983552

robrix:
…and that doesn’t actually work.

robrix:
…but it does if you set it on the `get`/`set`. We’re good.

robrix:
Correction: we’re not good, or at least not yet:

```
<unknown>:0: error: -[ReactiveCocoa_MacTests.PropertySpec DynamicProperty__should_write_the_underlying_object] : failed: caught "NSUnknownKeyException", "[<_TtC22ReactiveCocoa_MacTestsP33_EDA0DD2D130CD37D8B7288BCB98F94FA16ObservableObject 0x10b2d55c0> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key value."
```

Okay, so we do still need an ObjC superclass.

jspahrsummers:
How can I help here?

robrix:
There’s crashes in the tests that need workarounds. Would be a good idea to file a radar about having to go from curried to uncurried functions to work around crashes too :disappointed:

jspahrsummers:
Ugh, what a shitshow. Maybe we should just wait for the next beta.

robrix:
Yeah, maybe.

jspahrsummers:
Nimble (and therefore our tests) currently fails to build on beta 3, and LlamaKit will generate a warning until https://github.com/LlamaKit/LlamaKit/pull/38 is merged.

However, the main project now builds on 1.2b3. :tada:

jspahrsummers:
@ReactiveCocoa/reactivecocoa I made some changes here, so can someone else please review and merge this? :pray:

NachoSoto:
Woot! Does it build with -O and -whole-module-optimizations too?

neilpa:
When merging we should open an issue to add back `travis.yml` once they support Swift 1.2.

jspahrsummers:
> Woot! Does it build with -O and -whole-module-optimizations too?

@NachoSoto It does build, but tests start failing with `-O`. :cry: 

> When merging we should open an issue to add back `travis.yml` once they support Swift 1.2.

@neilpa :ok_hand:

neilpa:
@jspahrsummers Your updates look good although the currying change is unfortunate.

neilpa:
>> Woot! Does it build with -O and -whole-module-optimizations too?

> @NachoSoto It does build, but tests start failing with -O. :cry:

Do we still want to wait on this then?

jspahrsummers:
Nope, `-O` doesn't work on Swift 1.1 either.

neilpa:
Swift 1.2 here we come then.

NachoSoto:
@jspahrsummers do tests fail or crash?
Maybe we can ping someone at Apple about it, KA works fine with -O now, I was hoping they had fixed ~~all~~ most compiler optimization bugs by now.

jspahrsummers:
@NachoSoto They crash in RAC framework code. I didn't bother to look into how crazy a fix would be, but I'm fairly confident that it would be the result of compiler or optimizer bugs.

NachoSoto:
I filed a radar: http://www.openradar.me/20167826

jspahrsummers:
@NachoSoto :zap:

PR #1790, opened 2015-02-28
Throttle (redux (redux))

jspahrsummers:
Builds upon @adly-holler's implementation in #1730, with these additional changes:

- Completion no longer flushes the latest throttled value, since the contract of the operator is that _at least_ `interval` seconds pass between each value.
- Error and completed events are now scheduled, ensuring that all events get delivered on the given scheduler.
- Simplifies the logic for determining when to send values.

/cc @neilpa @kastiglione 

**To do:**

- [x] Add tests
- [x] Fix [race condition](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1790#discussion_r25550454)
- [x] Merge `swift-development`

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -569,20 +569,12 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
 	}
 }
 
-/// Throttle values sent by the receiver for which `predicate` returns true,
-/// so that at least `interval` seconds pass between each, then forwards
-/// them on the given scheduler.
-/// 
-/// Values that do not pass `predicate`, Errors, and Completed events
-/// are forwarded immediately. If an unthrottled value is forwarded while
-/// a throttled value is pending, the throttled value is discarded.
-///
-/// If the receiver completes before the last throttled value is sent,
-/// it will be sent immediately.
+/// Throttle values sent by the receiver, so that at least `interval`
+/// seconds pass between each, then forwards them on the given scheduler.
 ///
 /// If multiple values are received before the interval has elapsed, the
 /// latest value is the one that will be passed on.
-public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType, predicate: T -> Bool = { _ in true })(signal: Signal<T, E>) -> Signal<T, E> {
+public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
 	precondition(interval >= 0)
 
 	return Signal { observer in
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -569,6 +569,50 @@ public func takeWhile<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signa
...
+		let signalDisposable = signal.observe(next: { value in
+			let scheduleDate = previousDate.value?.dateByAddingTimeInterval(interval) ?? scheduler.currentDate
+
+			schedulerDisposable.innerDisposable = scheduler.scheduleAfter(scheduleDate) {
+				previousDate.value = scheduler.currentDate
+
+				// There's a small race where a new disposable could be set
+				// before we've updated the `previousDate`, which would schedule
+				// a value too early. Disallow that.
+				schedulerDisposable.innerDisposable = nil
```
jspahrsummers: I was more than a bit distracted when I wrote this, and I'm realizing now that it could result in values being lost if they're sent in that race condition window. I'll fix this up.

jspahrsummers:
Ready for review.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -716,6 +716,67 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
...
+/// If multiple values are received before the interval has elapsed, the
+/// latest value is the one that will be passed on.
+///
+/// If the input signal completes while a value is being throttled, that value
+/// will be discarded and the returned signal will complete immediately.
+public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType)(signal: Signal<T, E>) -> Signal<T, E> {
+	precondition(interval >= 0)
+
+	return Signal { observer in
+		let state: Atomic<ThrottleState<T>> = Atomic(ThrottleState())
```
neilpa: Could shorten this to `let state = Atomic(ThrottleState<T>())`.

jspahrsummers: This is the style that the other GitHub folks want to follow, keeping the type clarification separate from the initialization.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -716,6 +716,67 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
...
+		let schedulerDisposable = SerialDisposable()
+
+		let disposable = CompositeDisposable()
+		disposable.addDisposable(schedulerDisposable)
```
neilpa: This is a pretty common pattern. Can we add a convenience initializer that takes a disposable.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -716,6 +716,67 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
 	}
 }
 
+/// Throttle values sent by the receiver, so that at least `interval`
+/// seconds pass between each, then forwards them on the given scheduler.
+///
+/// If multiple values are received before the interval has elapsed, the
+/// latest value is the one that will be passed on.
+///
+/// If the input signal completes while a value is being throttled, that value
```
neilpa: nit: completes => terminates

ReactiveCocoa/Swift/Signal.swift
neilpa: Fair enough, hadn't realized.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -716,6 +716,67 @@ public func tryMap<T, U, E>(operation: T -> Result<U, E>)(signal: Signal<T, E>)
...
+					let scheduleDate = state.previousDate?.dateByAddingTimeInterval(interval) ?? scheduler.currentDate
+					return (state, scheduleDate)
+				}
+
+				schedulerDisposable.innerDisposable = scheduler.scheduleAfter(scheduleDate) {
+					let (_, pendingValue) = state.modify { (var state) -> (ThrottleState<T>, T?) in
+						let value = state.pendingValue
+						if value != nil {
+							state.pendingValue = nil
+							state.previousDate = scheduleDate
```
neilpa: Should this use `scheduler.currentDate` instead?

neilpa:
Few comments but looks good

ReactiveCocoa/Swift/Signal.swift
neilpa: Nevermind, just noticed your last commit and explanation the PR.

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: There is one that accepts a list. I'll use that.

jspahrsummers:
Thanks for the review, @neilpa! I think I've addressed your comments.

If it all looks good, feel free to merge at will. :smile:

jspahrsummers:
Looks like b1956ba is untenable, because Swift 1.1 is bad at type resolution. :wink:

neilpa:
Unfortunate, but not a big deal.

PR #1792, opened 2015-02-28
Add promoteErrors operator to promote NoError signals/producers

### LINKED ISSUE #1786: Add an operator for promoting a NoError signal/producer
Signals and producers that are parameterized over `NoError` (i.e., those that cannot send errors) cannot automatically be combined with signals and producers that might error. Instead, you have to lift them using something like:

```swift
producer |> catch { _ in SignalProducer<T, RealError>.empty }
```

… which would not make any sense to most readers (and which does not work on plain `Signal`s). We should add an operator to encapsulate this use case.

/cc @robrix

jspahrsummers:
Resolves #1786.


robrix:
:+1: LGTM!

NachoSoto:
:tada:

robrix:
:clap:

PR #1802, opened 2015-03-04
Remaining producer tests

### LINKED ISSUE #1794: Remaining tests for SignalProducer
There are still some `pending` tests in the `SignalProducerSpec`.

neilpa:
Will resolve #1794

* [x] `init`
  * [x] ~~`observer` lifetime~~
  * [x] `disposable` lifetime
* [x] `buffer`
* [x] `startWithSignal`
* [x] `start`
* [x] `lift`
  * [x] unary operators
  * [x] binary operators
* [x] `timer`
* [x] `on`
* [x] `startOn`
* [x] `catch`
* [x] `observer` lifetimes (is this even possible?)

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -284,10 +284,33 @@ class SignalProducerSpec: QuickSpec {
 		}
 
 		describe("start") {
-			pending("should immediately begin sending events") {
+			it("should immediately begin sending events") {
+				let producer = SignalProducer<Int, NoError>(values: [1, 2])
+
+				var values: [Int] = []
+				var completed = false
+				producer.start(next: {
+					values.append($0)
+				}, completed: {
+					completed = true
+				})
+
+				expect(values).to(equal([1, 2]))
+				expect(completed).to(beTruthy())
 			}
 
-			pending("should send interrupted if disposed") {
+			it("should send interrupted if disposed") {
+				let producer = SignalProducer<(), NoError>.never
+
+				var interrupted = false
+				let disposable = producer.start(interrupted: {
+					interrupted = true
+				})
+
+				expect(interrupted).to(beFalsy())
+
+				disposable.dispose()
+				expect(interrupted).to(beTruthy())
 			}
 
 			pending("should release sink when disposed") {
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -34,7 +34,7 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("should dispose of added disposables upon completion") {
-				let addedDisposable = SerialDisposable()
+				let addedDisposable = SimpleDisposable()
 				var test: () -> () = { }
 
 				let producer = SignalProducer<(), NoError>() { observer, disposable in
@@ -50,7 +50,7 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("should dispose of added disposables upon error") {
-				let addedDisposable = SerialDisposable()
+				let addedDisposable = SimpleDisposable()
 				var test: () -> () = { }
 
 				let producer = SignalProducer<(), TestError>() { observer, disposable in
@@ -66,7 +66,7 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("should dispose of added disposables upon interruption") {
-				let addedDisposable = SerialDisposable()
+				let addedDisposable = SimpleDisposable()
 				var test: () -> () = { }
 
 				let producer = SignalProducer<(), NoError>() { observer, disposable in
@@ -82,7 +82,7 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("should dispose of added disposables upon start() disposal") {
-				let addedDisposable = SerialDisposable()
+				let addedDisposable = SimpleDisposable()
 
 				let producer = SignalProducer<(), TestError>() { _, disposable in
 					disposable.addDisposable(addedDisposable)
@@ -278,7 +278,7 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("should dispose of added disposables if disposed") {
-				let addedDisposable = SerialDisposable()
+				let addedDisposable = SimpleDisposable()
 				var test: () -> () = { }
 
 				let producer = SignalProducer<Int, NoError>() { _, disposable in
@@ -360,7 +360,7 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("should dispose of added disposables upon completion") {
-				let addedDisposable = SerialDisposable()
+				let addedDisposable = SimpleDisposable()
 				var test: () -> () = { }
 
 				let producer = SignalProducer<Int, TestError>() { observer, disposable in
@@ -376,7 +376,7 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			it("should dispose of added disposables upon error") {
-				let addedDisposable = SerialDisposable()
+				let addedDisposable = SimpleDisposable()
 				var test: () -> () = { }
 
 				let producer = SignalProducer<Int, TestError>() { observer, disposable in
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -392,7 +392,32 @@ class SignalProducerSpec: QuickSpec {
 		}
 
 		describe("catch") {
-			pending("should invoke the handler and start new producer for an error") {
+			it("should invoke the handler and start new producer for an error") {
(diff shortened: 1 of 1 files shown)


neilpa:
This is ready for review. The only remaining tests were for observer/sink lifetimes. Was that really the intention for those or were they supposed to be `Signal`s? I doubt it'll be possible to test the observer case since you can't weakly capture non-class types.

jspahrsummers:
@neilpa We want to test it with `SignalProducer`s _and_ `Signal`s. I was thinking we could test it by creating a `SinkType` which we pass in to `start()`.

The easiest way to do that is probably `SinkOf` capturing some object that we can watch the lifetime of.

neilpa:
As always, another layer of indirection to the rescue

neilpa:
@jspahrsummers I ended up creating a `TestSink` class for monitoring directly instead of `SinkOf` which made things a bit simpler.

This did expose a bug in `start` leaking the observer. However, I've yet to figure out why given how simple `start`'s implementation is it doesn't make sense. Maybe another pair of eyes will help but for now I've left that test marked pending.

jspahrsummers:
Hmm, I don't know why that test would fail. Can you try making it more similar to the [`Signal` observer lifetime test](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/0bb278e8c94e65c896cc62749b623327887d3b89/ReactiveCocoaTests/Swift/SignalSpec.swift#L368-L391) and see what happens?

neilpa:
I tried something like it and a few other permutations to try and force ARC to kick in all with no luck. One thing I discovered is putting an exact copy of `start` in an extension method inside the tests works fine. Something peculiar is going on here since it only fails when crossing the module boundary.

neilpa:
I'm going to remove WIP here but will look into the `start` issue more this evening. If needed we can split that out into a separate issue.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -30,19 +30,86 @@ class SignalProducerSpec: QuickSpec {
 				expect(handlerCalledTimes).to(equal(2))
 			}
 
-			pending("should release signal observers when given disposable is disposed") {
+			it("should release signal observers when given disposable is disposed") {
+				var dispose: () -> () = {}
```
jspahrsummers: Is there any reason not to retain the `Disposable` itself?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -30,19 +30,86 @@ class SignalProducerSpec: QuickSpec {
...
+				expect(testSink).toNot(beNil())
+
+				dispose()
+				expect(testSink).to(beNil())
 			}
 
-			pending("should dispose of added disposables upon completion") {
+			it("should dispose of added disposables upon completion") {
+				let addedDisposable = SimpleDisposable()
+				var test: () -> () = { }
```
jspahrsummers: Can we save the sink/observer instead?

jspahrsummers: This applies to the below tests too.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -154,83 +274,427 @@ class SignalProducerSpec: QuickSpec {
...
 			}
 
-			pending("should send interrupted if disposed") {
+			it("should send interrupted if disposed") {
+				var interrupted = false
+				var scheduler = TestScheduler()
+
+				SignalProducer<Int, NoError>(value: 42)
+					|> startOn(scheduler)
+					|> startWithSignal { signal, disposable in
```
jspahrsummers: I think this case could be simpler by using `start(interrupted:)`, then disposing of the returned `Disposable` when ready.

jspahrsummers: Ah, sorry—this is specifically in a group for `startWithSignal`. My bad. :flushed:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -154,83 +274,427 @@ class SignalProducerSpec: QuickSpec {
...
+						signal.observe(interrupted: {
+							interrupted = true
+						})
+
+						scheduler.schedule {
+							disposable.dispose()
```
jspahrsummers: Could we just save the `Disposable` and do this outside of the block, to avoid any need for the `TestScheduler`?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -154,83 +274,427 @@ class SignalProducerSpec: QuickSpec {
...
+					interrupted = true
+				})
+
+				expect(interrupted).to(beFalsy())
+
+				disposable.dispose()
+				expect(interrupted).to(beTruthy())
 			}
 
 			pending("should release sink when disposed") {
```
jspahrsummers: Can you add a `// FIXME` comment to this test, near the point of failure, so it's easy to find and remember what's wrong later on?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -154,83 +274,427 @@ class SignalProducerSpec: QuickSpec {
...
 
 		describe("lift") {
 			describe("over unary operators") {
-				pending("should invoke transformation once per started signal") {
+				it("should invoke transformation once per started signal") {
+					let baseProducer = SignalProducer<Int, NoError>(values: [1, 2])
+
+					var counter = 0
+					let transform = { (signal: Signal<Int, NoError>) -> Signal<Int, NoError> in
+						counter += 1
```
jspahrsummers: This can be `++`. :smile:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -154,83 +274,427 @@ class SignalProducerSpec: QuickSpec {
...
 				}
 
-				pending("should not miss any events") {
+				it("should not miss any events") {
+					let baseProducer = SignalProducer<Int, NoError>(values: [1, 2, 3])
+					let otherProducer = SignalProducer<Int, NoError>(values: [4, 5, 6])
+
+					let transform = { (signal: Signal<Int, NoError>) -> Signal<Int, NoError> -> Signal<Int, NoError> in
+						return { otherSignal in
+							return zip(signal, otherSignal) |> map { return $0.0 + $0.1 }
```
jspahrsummers: Can you split the tuple using arguments to `map`, like so:

```swift
map { first, second in first + second }
```

jspahrsummers:
Thanks @neilpa! This is looking really great. :sparkles:

Some of my notes apply in multiple places, but I wanted to avoid leaving duplicate comments. It's generally pattern-level things, like saving a closure of a `Disposable` instead of the `Disposable` itself.

Hope the review is helpful.

jspahrsummers:
Ah, and about `start`: I think we can leave that test disabled for now, merge this, and deal with it later.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
neilpa: IIRC `++` breaks type inference even harder on these closures and requires an explicit `return` to "fix" it.

jspahrsummers: Seems like `+=` would do that too, but I may be misremembering.

neilpa: Double checked the stdlib, `+=` has no return value so doesn't confuse the compiler here.

jspahrsummers: :ok_hand:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -154,83 +274,428 @@ class SignalProducerSpec: QuickSpec {
...
+				expect(addedDisposable.disposed).to(beTruthy())
 			}
 
-			pending("should send interrupted if disposed") {
+			it("should send interrupted if disposed") {
+				var interrupted = false
+				var disposable: Disposable!
+
+				SignalProducer<Int, NoError>(value: 42)
+					|> startOn(TestScheduler())
```
neilpa: Only looked briefly but without this `startOn` the signal observer doesn't receive the interrupted event.

neilpa:
@jspahrsummers This should fix all instances of the issues you noted

neilpa:
There are some more broken curried functions on 1.2b3 that these tests expose. Was going to add them here but I'll open a new PR.

jspahrsummers:
I fixed any that were crashing during the test run. Are you seeing more?

neilpa:
Missed your `startOn` fix since I hadn't merged yet. I'm hitting another in `Signal.timeWithError`.

jspahrsummers:
@neilpa I briefly saw that too, but then a completely clean build worked.

neilpa:
Welp, that "fixed" it for me to

PR #1813, opened 2015-03-07
tests for SignalProducer (only observer lifetime)

335g:
Since most tests for SignalProducer have completed (#1802), I create pull request only observer lifetime tests.

This is ready for review. 
But observer cannot be released only at ```start()``` test. 

What's wrong?
Please advice when you get a chance? :bow:


neilpa:
Thanks for the PR, however, I just pushed implementations for these as well. I'd like to stick with my version since they test the observers more directly.

You're also right that there seems to be a bug in `start` not releasing its observers.

Thanks again and sorry about the mixup :cry:

335g:
It is certainly direct :+1: 

I will also PR, If there is something to help.

Issue #1815, opened 2015-03-09
Re-add constructor like +[RACSignal defer:] to SignalProducer


jspahrsummers:
It's sometimes hard to express imperative things with `SignalProducer`s without an operator like this. Maybe it could be called `run {}` or something action-y like that.

NachoSoto:
How would this ~~defer~~ differ from `SignalProducer`'s `startHandler` constructor?

jspahrsummers:
This function would accept a closure of type `() -> SignalProducer`.

iCodesign:
Is it something like this:

```
public static func defer(producerHandler: ()-> SignalProducer) -> SignalProducer {
	return self { observer, compositeDisposable in
		let diposable = producerHandler().start(observer)
		compositeDisposable.addDisposable(diposable)
	}
}
```

jspahrsummers:
@iCodesign Sort of. That particular sample would require some changes (maybe a different name, using `startWithSignal` instead). Would you be interested in opening a pull request where we can discuss it further?

jspahrsummers:
As I see it, `+[RACSignal defer:]` has three use cases:

1. Performing some work before the stream begins. In RAC 3, you can use `on(started:)` for this.
1. Performing an operation which may succeed or fail. In RAC 3, you can use `SignalProducer.self.try()` for this.
1. Performing some work before _creating_ the stream.

The last case is where I thought we might still want `defer`, but it's actually possible to use `try` for this too. Just return a `Result`, then use `joinMap` to tack on an additional producer based on that result if need be.

So, after thinking about it a bit, I've decided that `defer` is actually not worth including right now.

PR #1822, opened 2015-03-12
add defer to SignalProducer

iCodesign:
This is a prototype. Probably need more discussion. #1815


jspahrsummers:
Hey @iCodesign, thanks for opening this pull request! :sparkles:

After I stared at it for a while, though, I had a hard time justifying its inclusion. I think we should [leave it out for now](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1815#issuecomment-79248393), and maybe revisit it at some point in the future.

Thank you anyways, though!

PR #1832, opened 2015-03-18
add DependentType

### LINKED ISSUE #1828: The correct way to bind DynamicProperty to a signal
I can bind `MutableProprty` to a signal using `<~` operator. However there is no way to do the same for `DynamicProperty`, even if nothing (technically) prevents from it, since `DynamicProperty` can also be modified. 

Do I miss something?

335g:
Resolves #1828 

But I'm worried about 2 points.

1) This code allows only ```Signal<AnyObject?, _>```
2) Sense for name ```DependentType```

What do you think?

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -47,10 +47,16 @@ public struct ConstantProperty<T>: PropertyType {
 	}
 }
 
+/// Represents a property that's value is depended on Signal/Producer.
+/// i.e., Operator (<~) binds a signal to a property.
+public protocol DependentType: class, PropertyType {
+	var value: Value { get set }
+}
+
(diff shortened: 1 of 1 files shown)


jspahrsummers:

Can we call the protocol `MutablePropertyType`?

> This code allows only ```Signal<AnyObject?, _>```

That's fine—callers can `map` their signal values before they bind.

335g:
OK.

Fixed. :+1:

jspahrsummers:
Thanks for fixing that up!

I also made a couple of tweaks to some additional stuff I noticed in 05c9f0ee811795a03c19de4e016648faf3898f96 and c4a22f2b7ab5918e298e9854d85316186f98f853. Sorry for not noticing those before.

335g:
I committed without worry too much. :sweat: 
Thanks for your modification.

PR #1833, opened 2015-03-18
Added versions of <~ to bind values of a Signal or SignalProducer into any T -> () function

JaviSoto:
Not sure how useful this is, and whether we would prefer to keep it outside of the 1.0 API to limit the API surface, but I found myself doing:

```swift
someProducer.start(next: { value in
    self.doSomethingWithValue(value: value)
})
```

(Which I later realized I could just write like:)

```swift
someProducer.start(next: self.doSomethingWithValue)
```

And it reminded me of the `-[RACSignal setKeyPath:onObject:]` from RAC2. I think it would make sense to be able to use the `<~` operator also with `T -> ()` functions, like so:

```swift
self.doSomethingWithValue <~ someProducer
```

Does this make sense?

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -205,3 +205,29 @@ public func <~ <T>(property: MutableProperty<T>, producer: SignalProducer<T, NoE
 public func <~ <T, P: PropertyType where P.Value == T>(destinationProperty: MutableProperty<T>, sourceProperty: P) -> Disposable {
 	return destinationProperty <~ sourceProperty.producer
 }
+
+
+/// Invokes the provided function with the values sent by `producer`
+///
+/// The binding will automatically terminate when the producer sends a `Completed` event,
+/// or when the returned disposable is disposed of.
+public func <~ <T>(f: T -> (), producer: SignalProducer<T, NoError>) -> Disposable {
+	return producer.start(next: f)
+}
+
+/// Invokes the provided function with the values sent by `signal`
+///
+/// The binding will automatically terminate when the signal sends a `Completed` event,
+/// or when the returned disposable is disposed of.
+public func <~<T>(f: T -> (), signal: Signal<T, NoError>) -> Disposable {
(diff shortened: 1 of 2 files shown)


JaviSoto:
This may actually not be a good idea :-/ I just realized something, that may be obvious to you guys, but it wasn't obvious at all to me:

```swift
self.doSomethingWithValue <~ someProducer
```

or

```swift
someProducer.start(next: self.doSomethingWithValue)
```

Cause the producer to retain `self` (which can create a retain cycle if the producer is for example, `self.property.producer`). What you want is:

```swift
self.property.producer.start(next: { [unowned self] value in
    self.doSomethingWithValue(value)
}
```

So while this is not technically this API's fault, it's definitely a dangerous tool because you can create a retain cycle in a very non-obvious way :-/

brow:
I :+1: the sentiment. You can save a few characters using trailing closure syntax, but yeah, my views all have bindings like this:

```swift
model.saveButtonEnabled.producer.start { [weak self] enabled in
  self?.saveButton?.enabled = enabled
  return // Swift 1.1 :(
}
```

jspahrsummers:
IMO we shouldn't do this. The partial application capturing behavior is potentially surprising, as you've suggested, and I feel like this pattern discourages thinking about `error` and `completed` events (which are important for understanding lifetimes).

@brow FWIW, the fact that trailing closure syntax works for the `next` block is completely an accident. I don't want to recommend it, for the same reason as above. You could also use `DynamicProperty` for bindings like the one you have there.

JaviSoto:
Totally agreed @jspahrsummers :+1: Closing this!

brow:
@jspahrsummers I don't see how a `DynamicProperty` can help here. Did you mean you mean `MutableProperty`? 

I could define `enabled: MutableProperty` on `UIButton` and bind that like `saveButton.enabled <~ model.saveButtonEnabled`. Nice. Thanks. :+1:

brow:
Ah, I see now that it's intended for `<~` to be defined over `DynamicProperty`. #1828

PR #1834, opened 2015-03-18
Fixing throttle bug

MichaelMcGuire:
This fixes a bug in the throttle operator.  There was an error in the logic for scheduling the next signal that would cause a drift backwards.  This meant that given enough time passing, throttle would have no effect.  The fix is to make sure that the scheduled date is never in the past.  The change in the test demonstrates the issue.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -763,7 +763,8 @@ public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: Date
 					let (_, scheduleDate) = state.modify { (var state) -> (ThrottleState<T>, NSDate) in
 						state.pendingValue = value.unbox
 
-						let scheduleDate = state.previousDate?.dateByAddingTimeInterval(interval) ?? scheduler.currentDate
+						let proposedScheduleDate = state.previousDate?.dateByAddingTimeInterval(interval) ?? scheduler.currentDate
+						let scheduleDate = proposedScheduleDate.laterDate(scheduler.currentDate)
 						return (state, scheduleDate)
 					}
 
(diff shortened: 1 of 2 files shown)


jspahrsummers:
Thanks for the pull request! However, I'm not sure I understand where the bug lies.

The implementation is trying to avoid forward skew from the scheduler—if an event was scheduled for the right time, but ends up running 5 minutes later, IMO `throttle` should use the _scheduled_ date rather than the _actual_ date.

It seems like this PR would change it to the latter behavior. Is that correct?

MichaelMcGuire:
The bug isn't fixing a skew from the scheduler.  I noticed this bug after playing around with applying throttle of half a second to a button tapped signal (I'm pretty new to FRP and was playing around with performing the stuff in [this tutorial](https://gist.github.com/staltz/868e7e9bc2a7b8c1f754).  When I would tap the button a bunch of times, I would get one signal, then the interval and then the signal again (as expected). However, if I would tap the button, wait a few seconds) and then start tapping the button a bunch, I would get a signal per tap.  Digging into the implementation, it looks like this is what was occurring:

1. On first tap, the the event gets scheduled for scheduler's current date (time=0)
2. On the next run loop, the event gets sent.
3. After waiting a few seconds (say time is now 5), tapping will schedule the next event using the previous schedule date plus the interval (time=0.5).
4. On the next run loop, the event gets sent because it was scheduled in the past.
5. Tapping again will schedule the next event using the previous schedule date plus the interval (time = 1.0).
6. On the next run loop, the event gets sent because it was scheduled in the past.

The best way I could get to simulate this in the test was to increase the amount of time that passed after sending the `2` value.  If you go back to the previous code, but keep my change to the test, you'll see that `5` gets immediately scheduled after the `scheduler.advance()` call, which fails the test.

I hope that explanation makes sense.

jspahrsummers:
Ah, yep, it does. Thank you for detailing that out, and for the fix! :sparkles::shipit:

PR #1837, opened 2015-03-21
Don’t crash the compiler so much with default args

robrix:
`nil` turns out to be 99% less crashy than `doNothing`, so :fire:


robrix:
Wrong base branch! :skull:

Issue #1829, opened 2015-03-16
Swift compiler crash using observe free function


MichaelMcGuire:
I'm fairly new to FRP and decided to try working through some of [this linked intro to FRP](https://gist.github.com/staltz/868e7e9bc2a7b8c1f754) using the new 3.0 version.  Fairly quickly I came across an issue where the Swift compiler crashes with a segmentation fault 11 error.  I'm guessing it is a bug in the compiler, but wanted to make sure I wasn't doing something stupid.  To recreate (using Xcode 6.3β3 w/Swift 1.2):

1. Create a Single View Application (iOS)
2. Use Carthage to add latest from swift-development branch
    * Cartfile has single line: `github "ReactiveCocoa/ReactiveCocoa" "swift-development"`
    * Run `carthage update`
    * Drag built frameworks into Embedded Binaries section of app target
3. Add `import ReactiveCocoa` to import section of `ViewController.swift` file.
4. Add following code to `viewDidLoad()` function of included `ViewController.swift` file:

````swift
override func viewDidLoad() {
    super.viewDidLoad()
    
    let (signal, sink) = Signal<String, NoError>.pipe()
    
    signal.observe(next: { value in println("\(value)") })
    signal |> observe(next: { value in println("\(value)") })

    sendNext(sink, "Hello, world");
}
```
This compiles and run fine: printing out "Hello, World" to the output twice.  However, if you comment the first observe line, you get a SourceKit crash and the app no longer compiles.  No matter what I have tried, I can't use the free version of observe to work unless I first use the method off `Signal`.  It seems so strange that simply using the instance function first gets the free version to work.

Programmer error, Swift compiler bug or ReactiveCocoa issue?

(and thanks to everyone for their hard work on 3.0, I'm really loving what I see!)

neilpa:
Compiler bug. Might be another problem with curried functions that we've seen elsewhere (although those were at runtime).

Here's the full stack

```
0  swift                    0x00000001047b8388 llvm::sys::PrintStackTrace(__sFILE*) + 40
1  swift                    0x00000001047b8864 SignalHandler(int) + 452
2  libsystem_platform.dylib 0x00007fff8e94bf1a _sigtramp + 26
3  libsystem_platform.dylib 000000000000000000 _sigtramp + 1902854400
[...]

PR #1838, opened 2015-03-21
Try to crash less

robrix:
`nil` turns out to be 99% less crashy than `doNothing`, so :fire:.

Fixes a nasty crasher: https://github.com/Carthage/Carthage/pull/368


neilpa:
:+1: 

I wonder if this also works around #1829

robrix:
I bet it does.

jspahrsummers:
Thanks for digging into this, @robrix! :zap:

robrix:
You’re welcome! :smile:

PR #1844, opened 2015-03-22
Signal<T, E:ErrorType> binds to DynamicProperty

335g:
I want to bind `Signal<T, E>` to `DynamicProperty`.

Specifically, I want `DynamicProperty <~ Signal(UIKit)` through `RACSignal.toSignalProducer()`.
But it's not easy, because `RACSignal.toSignalProducer()` returns `SignalProducer<AnyObject?, NSError>`.

I think that operator (<~) doesn't need `Sigal<T, NoError>`. 
Operator (<~) doesn't do anything for error.

How about?

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -164,7 +164,7 @@ infix operator <~ {
 ///
 /// The binding will automatically terminate when the property is deinitialized,
 /// or when the signal sends a `Completed` event.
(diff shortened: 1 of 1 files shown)


kaybutter:
Why not `catch` the error first before binding to the property? Seems unsafe to have the operator just discard any errors...

335g:
Oh, I've been overlooked `catch` function. :open_mouth: 
Use `catch` is safety though it's not bad for binding, because I don't know what the error is coming.

Thanks :hand:

335g:
By the way, How would do I do for `Signal` ?

PR #1847, opened 2015-03-23
[WIP] Work around swift_getTupleTypeMetadata crash

jspahrsummers:
See:
https://twitter.com/jckarter/status/579854984147181568
https://twitter.com/jckarter/status/579855419092267009

### DIFF
--- a/ReactiveCocoaTests/Swift/PropertySpec.swift
+++ b/ReactiveCocoaTests/Swift/PropertySpec.swift
@@ -73,6 +73,12 @@ class PropertySpec: QuickSpec {
 				mutableProperty = nil
 				expect(signalCompleted).to(beTruthy())
 			}
+
+			it("should not crash in a tuple") {
+				let tuple = (MutableProperty(0), MutableProperty(false))
+				expect(tuple.0.value).to(equal(0))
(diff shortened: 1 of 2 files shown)


jspahrsummers:
This doesn't yet fix the crash in https://github.com/Carthage/Carthage/pull/368#issuecomment-84451730. :cry:

jspahrsummers:
That fix didn't work either. :weary:

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -73,6 +73,12 @@ class PropertySpec: QuickSpec {
 				mutableProperty = nil
 				expect(signalCompleted).to(beTruthy())
 			}
+
+			it("should not crash in a tuple") {
```
jspahrsummers: This test failed at the start of the PR, then passed with both “fixes,” so it's not a good indicator of whether the crash is gone.

Instead, I've been building https://github.com/Carthage/Carthage/pull/368 against this branch and testing a complex `carthage bootstrap` with it.

jspahrsummers:
None of these attempted fixes have worked. :cry:

PR #1850, opened 2015-03-24
Add ignoreNil for Signal

neilpa:
Essentially filter+map without the need to force unwrap values in the latter if you implement it in terms of the two.

~~I went with the name `flatMap` to be consistent with the stdlib. I also considered `filterMap` and am open to other suggestions~~. ~~Lets go with `filterMap`~~. Take 3, `ignoreNil`.

@ReactiveCocoa/reactivecocoa thoughts?

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -177,12 +177,12 @@ public func filter<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T
 	}
 }
 
-/// Applies `transform` to values from `signal` with non-`nil` results unwrapped and
-/// forwared on the returned signal.
-public func filterMap<T, U, E>(transform: T -> U?)(signal: Signal<T, E>) -> Signal<U, E> {
+/// Unwraps non-`nil` values from `signal` and forwards them on the returned
+/// signal, `nil` values are dropped.
+public func ignoreNil<T, E>(signal: Signal<T?, E>) -> Signal<T, E> {
 	return Signal { observer in
 		signal.observe(next: { value in
-			if let val = transform(value) {
+			if let val = value {
 				sendNext(observer, val)
 			}
 		}, error: { error in
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -180,20 +180,7 @@ public func filter<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T
 /// Unwraps non-`nil` values from `signal` and forwards them on the returned
 /// signal, `nil` values are dropped.
 public func ignoreNil<T, E>(signal: Signal<T?, E>) -> Signal<T, E> {
-	// This is equivalent to `signal |> filter { $0 != nil } |> map { $0! }`
-	return Signal { observer in
-		signal.observe(next: { value in
-			if let val = value {
-				sendNext(observer, val)
-			}
-		}, error: { error in
-			sendError(observer, error)
-		}, completed: {
-			sendCompleted(observer)
-		}, interrupted: {
-			sendInterrupted(observer)
-		})
-	}
+	return signal |> filter { $0 != nil } |> map { $0! }
 }
 
 /// Returns a signal that will yield the first `count` values from the
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -177,6 +177,24 @@ public func filter<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T
 	}
 }
 
+/// Applies `transform` to values from `signal` with non-`nil` results unwrapped and
+/// forwared on the returned signal.
+public func flatMap<T, U, E>(transform: T -> U?)(signal: Signal<T, E>) -> Signal<U, E> {
+	return Signal { observer in
+		signal.observe(next: { value in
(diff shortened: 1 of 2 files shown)


robrix:
`flatMap`’s type would be `(A -> Signal<B>) -> Signal<A> -> Signal<B>`. This is `filterMap`, definitely.

NachoSoto:
I think this operator is certainly useful, like you said implementing this as filter+map is really annoying.

However, I don't think the name correctly describes the operator.
> I went with the name flatMap to be consistent with the stdlib

The `flatMap` you're referring to is `Optional.flatMap`. That name is accurate because it belongs to `Optional`. However, a `Signal.flatMap` operator would have the signature:
```swift
func flatMap<T, U, E>(transform: T -> Signal<U>)(signal: Signal<T, E>) -> Signal<U, E>
```

I'm also worried that the operator is doing 2 different things: mapping from `T` to `U`, and filtering optionals. Since `map` is already available, how about
```swift
func filterNils<T, E>(signal: Signal<T?, E>) -> Signal<T, E>
```

neilpa:
@NachoSoto I experimented `filterNils` and it would be enough to avoid the forced unwrapping. However, I've found the combined operator a bit nicer than having to append the each `map` with `filterNils`. Also, if `filterNils` is all you need than that's just `filterMap(id)`.

robrix:
I’ve considered writing this for `SequenceType` several times, so I’m :+1:

brow:
I've needed this enough to [write it](https://github.com/brow/Sukhasana/blob/master/Sukhasana/Other/SignalOperators.swift), too. I called it `mapOptional` after Haskell's `mapMaybe`.

```swift
func mapOptional<T, U, E>(transform: T -> U?)(signal: Signal<T, E>) -> Signal<U, E> {
  return signal |> map(transform) |> filter { $0 != nil } |> map { $0! }
}
```

I imagine your implementation performs better?

mdiep:
I've already done something like this as well.

jspahrsummers:
I'd prefer an operator that just did filtering. I don't want to provide a ton of convenience operators—it bloats the API surface and can turn off newcomers, especially if it's not immediately obvious how to apply them.

The previous Swift implementation(s) called a filtering operator `ignoreNil`. I think `ignore` is better than `filter`, which implies the inverse: that it gives you _only_ those things.

neilpa:
Although I prefer the combined operator I understand wanting to stick with a minimal set of primitives. An `ignoreNil` still eliminates the need to `filter` then force unwrap via `map` which was my main issue.

Also, going to throw out `drop` as possible alternative to `ignore`. There's at least some precedent in the stdlib with `dropFirst` and `dropLast`.

robrix:
Notably, `dropFirst` doesn’t change the type of the elements. I think `ignore` is probably better for that reason.

neilpa:
I've replaced `filterMap` with `ignoreNil` if someone wants to review.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -177,6 +177,24 @@ public func filter<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T
...
+				sendNext(observer, val)
+			}
+		}, error: { error in
+			sendError(observer, error)
+		}, completed: {
+			sendCompleted(observer)
+		}, interrupted: {
+			sendInterrupted(observer)
+		})
+	}
```
JaviSoto: Perhaps it might be easier to understand by just composing existing operators like:

```swift
return signal 
           |> filter { $0 != nil }
           |> map { $0! }
```

neilpa: Maybe my aversion is unfounded, but I was explicitly trying to avoid that pattern with the forced unwrapping. I could add a comment that this is functionally equivalent to that though.

JaviSoto: Yeah, you're right your implementation is a bit safer for that reason, it just seems so noisy having to explicitly forward the rest of the events.

neilpa: We could push some of this into `Event` but I'm not sure it's worth it.

NachoSoto: > but I was explicitly trying to avoid that pattern with the forced unwrapping.

I don't see any risk, specially since
- It's very well tested.
- The implementation is so simple that we can reason about it.

I would actually opt for @JaviSoto's approach: by reusing existing operators we have less code to maintain.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -177,6 +177,25 @@ public func filter<T, E>(predicate: T -> Bool)(signal: Signal<T, E>) -> Signal<T
 	}
 }
 
+/// Unwraps non-`nil` values from `signal` and forwards them on the returned
+/// signal, `nil` values are dropped.
+public func ignoreNil<T, E>(signal: Signal<T?, E>) -> Signal<T, E> {
```
NachoSoto: I like it :100:

jspahrsummers:
Looks good! :sparkles:

I agree this should be written in terms of `filter` and `map`, though. Do you mind making that change? :pray:

neilpa:
Done, but I still have my doubts about not implementing this one explicitly. Although, that's mostly residual scars from RAC 2 and some of the perf issues we hit with everything being implemented in terms of `bind`. At least now it's a lot easier to sub in hand-rolled operators if it becomes an issue.

jspahrsummers:
Thanks! :bow:

PR #1857, opened 2015-03-26
DynamicProperty Lifetime

### LINKED ISSUE #1855: DynamicProperty binding lifetime
It seems reasonable that DynamicPropertys should be kept alive at least long enough for the use case 

```swift
DynamicProperty(cell, "imageView.image") <~ myImageSignal
```

to be possible without storing a strong reference to the property. As it is today (`v3.0-alpha3`), that code completely fails – the binding only sticks if you keep the property alive somewhere. Maybe keep it alive until the object will deallocate? Or keep it alive until the run loop ends without a signal bound to it?

335g:
Resolve #1855

- [x] property lifetime with binding Signal
- [x] property lifetime with binding SignalProducer

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -152,6 +152,8 @@ extension MutableProperty: SinkType {
 		self.object = object
 		self.keyPath = keyPath
 		
+		/// DynamicProperty stay alive as long as object is alive.
+		/// This is made possible by strong reference cycles.
 		super.init()
 		object?.rac_willDeallocSignal()?.toSignalProducer().start(completed: { self })
 	}
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -114,7 +114,7 @@ extension MutableProperty: SinkType {
 
 	private weak var object: NSObject?
 	private let keyPath: String
-	
+
 	/// The current value of the property, as read and written using Key-Value
 	/// Coding.
 	public var value: AnyObject? {
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -114,7 +114,7 @@ extension MutableProperty: SinkType {
 
 	private weak var object: NSObject?
 	private let keyPath: String
-
+	
 	/// The current value of the property, as read and written using Key-Value
 	/// Coding.
 	public var value: AnyObject? {
@@ -151,6 +151,9 @@ extension MutableProperty: SinkType {
 	public init(object: NSObject?, keyPath: String) {
 		self.object = object
 		self.keyPath = keyPath
+		
+		super.init()
+		object?.rac_willDeallocSignal()?.toSignalProducer().start(completed: { self })
 	}
 }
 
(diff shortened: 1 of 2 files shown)


335g:
This is ready for review.

I’m worried about whether or not `MutablePropertyType` needs class object. It is not always necessary class, because property isn't weak reference in `Event.next`. Is there other reason to be a class?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -195,15 +195,60 @@ class PropertySpec: QuickSpec {
 					sendNext(observer, subsequentPropertyValue)
 					expect(mutableProperty.value).to(equal(initialPropertyValue))
 				}
-
-				it("should tear down the binding when the property deallocates") {
```
jspahrsummers: This test is still valid over `MutableProperty`. Can you please restore it?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -135,12 +138,21 @@ extension MutableProperty: SinkType {
...
+				|> map{ $0 as! () }
+			
 			return object.rac_valuesForKeyPath(keyPath, observer: nil).toSignalProducer()
 				// Errors aren't possible, but the compiler doesn't know that.
 				|> catch { error in
 					assert(false, "Received unexpected error from KVO signal: \(error)")
 					return .empty
 				}
+				|> takeUntil(willDeallocProducer)
+				|> takeUntil(objectWillDeallocProducer)
```
jspahrsummers: `-rac_valuesForKeyPath:` already terminates automatically when the object deallocates.

jspahrsummers:
I think I would rather flip this around a bit, and make the lifetime of `DynamicProperty` _itself_ depend on the lifetime of the object being bound.

We could do this by watching `rac_willDeallocSignal()` inside the `init` method, and capturing `self` in that observation, ensuring that the reference to the `DynamicProperty` is only released when the `NSObject` deallocates.

That might be a bit unintuitive, so we should also document it. Ultimately, I think it's what people will _expect_, even if it's inconsistent with how our other property types behave.

335g:
I thought that `MutableProperty` should be retained too when it is bound by `Signal/Producer`, because of consistency for `MutablePropertyType`. 

Normally, people will use `MutableProperty` with retain, so I think your thought is right too. :+1:

I will edit as above pointed out. :runner:

335g:
Fixed. :+1: 

I might have misunderstood.
Is this right?

※ 2nd, 3rd tests are not directly related this issue. If dividing better, I will divide for other PR. :bow:

ReactiveCocoa/Swift/Property.swift
```diff
@@ -151,6 +151,9 @@ extension MutableProperty: SinkType {
 	public init(object: NSObject?, keyPath: String) {
 		self.object = object
 		self.keyPath = keyPath
+		
+		super.init()
+		object?.rac_willDeallocSignal()?.toSignalProducer().start(completed: { self })
```
jspahrsummers: Can you add a comment clarifying why we reference `self` in the block?

jspahrsummers:
The implementation looks good, thank you! Just one more comment. :sparkles:

335g:
I added comment.

I'm not good at English.
Please correct it, if strange wording. :bow:

jspahrsummers:
Thanks! :shipit:

PR #1860, opened 2015-03-27
Ignore unhelpful warnings from -Wreceiver-is-weak in RACObserve

numist:
https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1595 changed `RACObserve` to message a weakified target, but this emits a compiler warning when `-Wreceiver-is-weak` is enabled.

This is fine for `RACObserve` since the lifetime of the signal is tied to the lifetime of the target, so it's safe to fix with a `clang diagnostic ignored`.


jspahrsummers:
Thanks! :heart:

PR #1866, opened 2015-03-31
MKAnnotationView prepare for reuse signal

kazmasaurus:
Adds the same `rac_prepareForReuseSignal` that's in all the other reusable views to MKAnnotationView.


jspahrsummers:
Awesome, thanks! :sparkles:

PR #1885, opened 2015-04-11
Overloaded skipRepeats for Array<T: Equatable>

NachoSoto:
I'm not sure if this is something we want to include, but I've needed this myself.

This is necessary because `Array` is not `Equatable` *yet*. Hopefully one day Swift will allow partial specialization of types (is that the right term?) so this won't be necessary.


jspahrsummers:
You can use `skipRepeats(==)` for this, I think, or `skipRepeats { $0 == $1 }` at the very least.

I'd rather not bring this in, because the `Equatable` restriction applies to _every_ collection, and I don't think we should be offering an overload for everything under the sun.

NachoSoto:
Makes sense :) thanks!

PR #1886, opened 2015-04-11
Rename join/joinMap to flatten/flatMap

### LINKED ISSUE #1842: Rename joinMap to flatMap for stdlib consistency?
`joinMap` predates `flatMap` in the stdlib, but maybe it's worth renaming it to be consistent now.

We would still have our `strategy` parameter, which makes it a little different, but it would at least highlight the parallels.

Thoughts?

jspahrsummers:
Resolves #1842.

Unfortunately, as you can see, this breaks the symmetry we had with `join{,Map}`, but I think the consistency with the standard library is worth it.

This is hopefully the last _major_ breaking change in RAC 3—if we even decide to accept it. :open_hands:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -203,11 +203,12 @@ Because `-flattenMap:` is the easiest to use, it was often
 incorrectly chosen even when concatenation or switching semantics are more
 appropriate.
 
-**RAC 3 distills these concepts down into just two operators, named `join` and `joinMap`.**
-Both accept a “strategy” which determines how the producer-of-producers should
-be integrated, which can be one of:
+**RAC 3 distills these concepts down into just two operators, `flatten` and `flatMap`.**
+Note that these do _not_ have the same behavior as `-flatten` and `-flattenMap:`
+from RAC 2. Instead, both accept a “strategy” which determines how the
+producer-of-producers should be integrated, which can be one of:
 
- * `.Merge`, which is equivalent to `-flatten` or `+merge:`
+ * `.Merge`, which is equivalent to RAC 2’s `-flatten` or `+merge:`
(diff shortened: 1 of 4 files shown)


kastiglione:
Too bad we can't open a pull request to bring "to flat" into english as a verb.

neilpa:
:+1: for consistency with the stdlib here.

jspahrsummers:
I'm still ambivalent, so I'd like @joshaber to make the final call here.

joshaber:
I go back and forth. I think there's probably only 4.39 people in the world who understand how this `flatMap` would relate to Swift's stdlib `flatMap` so I'm not sure that's a compelling reason. And I think it's confusing for those used to `flatMap` from RAC 2.

joshaber:
Unless anyone has a compelling argument for changing it, :-1:

jspahrsummers:
I'm not strongly in favor of `flatMap` (mostly because of the inconsistency in `flatten`), but I do think it's a mildly better name.

`join` doesn't build any intuition or convey what it means to perform the operation. At best, it sounds like “joining a thread,” which isn't really an accurate mental model. But _flattening_ at least implies that you're taking a complicated structure and “flattening it out,” even if it's not exactly clear how that happens.

`-flattenMap:` was a poor name for the operation in RAC 2 that is better known as “merge,” and I don't think we should be constrained by our past mistakes—especially with the strategy argument it has (that can't be omitted), which will make it clear that only _one_ version matches the previous behavior.

joshaber:
>join doesn't build any intuition or convey what it means to perform the operation. At best, it sounds like “joining a thread,” which isn't really an accurate mental model.

Legit :+1:

Issue #1887, opened 2015-04-12
Update Result type


neilpa:
Should we migrate our `Result` type before 3.0 given that [LlamaKit](https://github.com/LlamaKit/LlamaKit/blob/master/README.md) probably isn't going to be maintained anymore.

jspahrsummers:
Yeah, I suppose we should use [antitypical/Result](https://github.com/antitypical/Result) here.

/cc @robrix

robrix:
:+1:

ikesyo:
[antitypical/Result](https://github.com/antitypical/Result)'s `Box` is internal, but RAC is using the `Box` type in `Event.swift`. What should we do for it?

robrix:
Just add an internal `Box` type.

iCodesign:
:+1: `Box` is just an internal type to compromise with Swift limitations.

PR #1891, opened 2015-04-12
Setting APPLICATION_EXTENSION_API_ONLY to YES in iOS and Mac frameworks

NachoSoto:
Per the documentation:

```
When enabled, this causes the compiler and linker to
disallow use of APIs that are not available
to app extensions and to disallow linking to frameworks
that have not been built with this setting enabled. 
```

Since we're not using any of these APIs, we can safely turn this setting on.

This allows `ReactiveCocoa` to be used in application extensions without receiving this warning:
screen shot 2015-04-12 at 11 00 23

The only problem is that compiling now prompts this same error with `LlamaKit`, but I imagine we can enable this in https://github.com/antitypical/Result after #1887.

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,3 +1,3 @@
-github "jspahrsummers/xcconfigs" >= 0.7.1
+github "jspahrsummers/xcconfigs" >= 0.8
 github "Quick/Quick" "master"
 github "Quick/Nimble" "master"
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
 github "LlamaKit/LlamaKit" "v0.6.0"
 github "Quick/Nimble" "6a679cbf20da3fe570c26296b3029dd168b72c00"
 github "Quick/Quick" "155f858bce082373fc60d0c0a63415bf336c676b"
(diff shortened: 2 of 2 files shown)


jspahrsummers:
What sorts of things does this restrict?

If it's not a crazy list, we should consider adding it to [the inherited xcconfig](https://github.com/jspahrsummers/xcconfigs/blob/2e77204b59c3d97c24e5dd34966fb32c231194f0/Base/Targets/Framework.xcconfig) instead.

NachoSoto:
This is the official [list](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/ExtensibilityPG/ExtensionOverview.html#//apple_ref/doc/uid/TP40014214-CH2-SW6).

I don't think there's a way to disable each individually, but I agree I should move it to the `xcconfig` :)

jspahrsummers:
:+1: Let's do that, and then we can replace this change with a bump to the xcconfigs.

NachoSoto:
https://github.com/jspahrsummers/xcconfigs/pull/36

NachoSoto:
Updated! :)

jspahrsummers:
:boom: Thanks!

PR #1894, opened 2015-04-14
> Unannotated single-expression closures with non-Void return types can ...

ikesyo:
...now be used in Void contexts

The cases of `observeOn` and `sampleOn` in Signal.swift cannot be omitted due to compilation error.


JaviSoto:
:+1:

jspahrsummers:
Looks good, thanks! :sparkles:

PR #1895, opened 2015-04-14
Reduce boxing and unboxing

ikesyo:
Is it worth doing?


jspahrsummers:
Since this is primarily an optimization, I'd like to leave it out for now. It's more important for the RAC 3 API and basic implementation to stabilize than it is to start making performance improvements.

I also get nervous any time we add more overloading, because we'll probably run into new toolchain issues. :scream: 

Thank you, though! :sparkles:

furyu-ikeda-sho:
Sure, it makes sense. Thank you too!

PR #1896, opened 2015-04-14
Workarounds for Swift runtime memory corruption

jspahrsummers:
“Fixes” crashes [like these](https://gist.github.com/jspahrsummers/b77aeb5ff0eabe67feb7), as reported in https://github.com/Carthage/Carthage/pull/368#issuecomment-87943571.

I basically just addressed these with the brute force approach: run until we crash, and then find another way to express the code there. :confused: 

/cc @robrix @jckarter

### DIFF
--- a/ReactiveCocoa/Swift/Bag.swift
+++ b/ReactiveCocoa/Swift/Bag.swift
@@ -48,6 +48,16 @@ internal struct Bag<T>: SequenceType {
 	}
 	
 	internal func generate() -> GeneratorOf<T> {
-		return GeneratorOf(elements.values.generate())
+		var values: [T] = []
+
+		var index = next + 1
+		while index > 0 {
+			--index
+			if let value = elements[index] {
+				values.append(value)
+			}
+		}
+
+		return GeneratorOf(values.generate())
 	}
 }
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -52,21 +52,12 @@ internal final class Atomic<T> {
 	///
 	/// Returns the old value.
 	func modify(action: T -> T) -> T {
-		let (oldValue, _) = modify { oldValue in (action(oldValue), 0) }
-		return oldValue
-	}
-	
-	/// Atomically modifies the variable.
-	///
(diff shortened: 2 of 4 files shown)


jspahrsummers:
Ugh, now the tests are crashing.

robrix:
I’ve found curried function definitions are prone to runtime crashes. Are we using them fairly widely…?

jspahrsummers:
@robrix Yeah, they're essential to the functionality of `|>`.

jspahrsummers:
No matter what I fix here, the crash always shifts around. Definitely some sort of corruption issue, and I'm not sure how to work around it. :weary:

robrix:
@jspahrsummers Right, but what I mean is, sometimes this crashes:

```swift
func foo(x: Int)(_ y: Int) -> Int
```

where this does not:

```swift
func foo(x: Int) -> Int -> Int
```

This seems particularly common when the function has type constraints.

jspahrsummers:
Merging from `swift-development` fixes test crashes, but now [the Carthage crash](https://github.com/Carthage/Carthage/pull/368) is back. :finnadie:

robrix:
Curried function syntax is said to be fixed in Xcode 6.4.

jspahrsummers:
Finally had a chance to update this branch and re-test it on Xcode 6.3.1.

It seems to be :sparkles: now, but that's probably more due to @ikesyo's work on other crashes than anything I changed here. :smile:

robrix:
:tada:

ReactiveCocoa/Swift/Bag.swift
```diff
@@ -48,6 +48,16 @@ internal struct Bag<T>: SequenceType {
...
+
+		var index = next + 1
+		while index > 0 {
+			--index
+			if let value = elements[index] {
+				values.append(value)
+			}
+		}
+
+		return GeneratorOf(values.generate())
```
robrix: This makes me so sad!

robrix: To be clear: this seems like a very good solution. Just a heartbreaking necessity.

robrix:
I’m :+1: on all of this.

PR #1900, opened 2015-04-16
Make "next" the trailing closure for Signal.observe() and SignalProducer.start()

sharplet:
First, a little story:

When I first decided to test out the great new Swift API (and it really is lovely!), I had to work out how to get values out of a signal. The very first time I observed a signal, this is more or less what I wrote:

```swift
producer.start { next in
  // do something with next value
}
```

However, this didn't compile!

```
/path/to/Sadface.swift: error: cannot invoke 'start' with an argument list of type '((_) -> _)'
                                        producer.start { next in
                                                 ^
```

After I little bit of digging I realised that because the `interrupted:` parameter is the last in the parameter list, my trailing closure was not the closure I expected it to be.

*****

So I'm floating this breaking API change so it can be considered before the 3.0 release. My argument is this:

- Trailing closure syntax is a thing, so we may as well make the trailing closure useful
- The values sent on a signal are the valuable bit, so making `next:` the trailing closure makes sense to me

*****

So basically this just reorders some parameter lists to make `next:` the trailing closure, and adds some tests to make the semantics explicit. If this is a desirable change, there's probably a bunch of other things to do like updating documentation! Also I may well have missed other parts of the API where this would also make sense.

Thanks for your time!

### DIFF
--- a/ReactiveCocoa/Swift/Event.swift
+++ b/ReactiveCocoa/Swift/Event.swift
@@ -64,7 +64,7 @@ public enum Event<T, E: ErrorType> {
 
 	/// Creates a sink that can receive events of this type, then invoke the
 	/// given handlers based on the kind of event received.
-	public static func sink(next: (T -> ())? = nil, error: (E -> ())? = nil, completed: (() -> ())? = nil, interrupted: (() -> ())? = nil) -> SinkOf<Event> {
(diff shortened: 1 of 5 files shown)


jspahrsummers:
Thanks for the PR! :sparkles:

I'm worried about confusing users into thinking that values are the only thing you'll receive upon a signal/producer, and about the loss of explicitness here, **but** since Swift allows a trailing closure here regardless, I agree with your point that it should be the most useful one.

So, I think this makes sense, although I'll avoid using the syntax myself, and I don't think any official docs should make reference to it. :wink:

sharplet:
Makes sense to me, and thanks! :tada:

PR #1901, opened 2015-04-16
Use for-in loop for a sequence, not use the generator directly

ikesyo:
It seems to me that this is neater way.


jspahrsummers:
This is definitely desired! It just used to crash the compiler. :wink:

ikesyo:
:confetti_ball:

PR #1904, opened 2015-04-16
Add override of map for transforming signals with optional values

MichaelMcGuire:
Adding an override for map that allows for a transformation of `Signal<T?,E>` to `Signal<U?,E>` but using a transform function that maps from `T` to `U`.  Nil values are passed through.  I have found this to be helpful simplifying my transform functions where I want to allow `nil` values to continue to be sent.

I'm fairly new to both FP and FRP and had originally used a different name for the transformation.  However, in searching on this transformation, it seemed that `map` was the correct name, at least when viewing the use of it in the Swift standard library ([map](http://swiftdoc.org/func/map/)).  Sorry if there is A) a better name, or B) an existing operator that does this. :smile:

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -142,6 +142,25 @@ public func map<T, U, E>(transform: T -> U)(signal: Signal<T, E>) -> Signal<U, E
 	}
 }
 
+// Maps each non-nil value in a signal to a new value.  Nils are passed through
+// to the new signal.
+public func map<T, U, E>(transform: T -> U)(signal: Signal<T?, E>) -> Signal<U?, E> {
+	return Signal<U?, E> { observer in
+		return signal.observe(Signal.Observer { event in
+			switch event {
+			case let .Next(value):
+				sendNext(observer, value.unbox.map(transform))
+			case let .Error(error):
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -142,6 +142,25 @@ public func map<T, U, E>(transform: T -> U)(signal: Signal<T, E>) -> Signal<U, E
 	}
 }
 
+// Maps each non-nil value in a signal to a new value.  Nils are passed through
+// to the new signal.
+public func map<T, U, E>(transform: T -> U)(signal: Signal<T?, E>) -> Signal<U?, E> {
+	return Signal<U?, E> { observer in
```
NachoSoto: This could be implemented very easily as 
```swift
let originalSignal: Signal<T?, E>
let newSignal: Signal<U?, E> = originalSignal.map {
  $0.map(transform)
}
```

So I'm not sure it's worth adding to `Signal` itself :)

MichaelMcGuire: Ha! So it is.  Yes, may not be worth adding then.  Althou

MichaelMcGuire:
Since @NachoSoto's solution is trivial, I'll go ahead and close this for now.

PR #1902, opened 2015-04-16
Uncurry combinePrevious

ikesyo:
To avoid runtime crash in Swift 1.2. Fixes #1890.

PR #1905, opened 2015-04-17
Specify Release configuration for test build, fix some optimization issues

ikesyo:
Depends on #1902. To detect issues caused by the compiler optimization.

Now unfortunately, [this test](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/091d09960df11c3538b9e59f732d1e3ad8bb2e97/ReactiveCocoaTests/Objective-C/RACKVOWrapperSpec.m#L623) is failing :cry:

### DIFF
--- a/ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa iOS.xcscheme
+++ b/ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa iOS.xcscheme
@@ -40,7 +40,7 @@
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       shouldUseLaunchSchemeArgsEnv = "YES"
-      buildConfiguration = "Debug">
+      buildConfiguration = "Release">
       <Testables>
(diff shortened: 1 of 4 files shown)


ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa iOS.xcscheme
```diff
@@ -40,7 +40,7 @@
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       shouldUseLaunchSchemeArgsEnv = "YES"
-      buildConfiguration = "Debug">
+      buildConfiguration = "Release">
```
NachoSoto: :+1:

jspahrsummers:
Thanks @ikesyo! This looks great. :sparkles:

That one test is failing because it's assigning directly into a `__weak` variable, so it will disappear almost immediately with optimization. I think we could either make it strong, or else create a temporarily strong reference to it just within the `@autoreleasepool`.

ikesyo:
@jspahrsummers I fixed that with the latter way. :boom:

jspahrsummers:
Looks good! :shipit:

Issue #1911, opened 2015-04-17
Bi-directional MutableProperty binding


gavrix:
Currently, `MutableProperty` implements `Signal` (via `producer` property) and `SinkType` effectively allowing cross bind two mutable properties to each other. Within current implementation that leads to a loop.

In RAC2 we had `RACChannel`concept and ability to bind two KVO observings through `RACChannelTo` macro. 

Would it make more sense to make `MutableProperty` act same way as RACChannelTerminal? Meaning, whenever value came from a signal bound to this property do not pass it into own producer, and only pass those values set directly through calling 
```
property.value = <value>
```

BTW, within current implementation, even though `MutableProperty` conforms to `SinkType` (has `put` method) values coming from signal set to the property thorough it's `value` property. ([here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/v3.0-beta.1/ReactiveCocoa/Swift/Property.swift#L180-L184))

gavrix:
After having a second though I tried modelling the use case and ended up with `PropertyTermianl` #1915  which sort of fixed my problem of mutual property bindings

jspahrsummers:
Thanks for the issue and the pull request! :sparkles:

The reason channels and terminals haven't been ported to RAC 2.x is because **two-way bindings are hard**, and I don't think we've gotten it right yet. I'd even go further and argue that two-way bindings are _usually_ a smell, and should be reformulated to explicitly give precedence to one side or the other.

Although I am interested in revisiting two-way bindings at some point in the future, I don't think we should include anything in the first release of RAC 3.0—because then we're required to support it for all future updates in the 3.x cycle.

/cc @Coneko @andymatuschak who also care a lot about this problem

gavrix:
Ok I got that.

Just for the note, a couple of words about reasoning behind that mutual properties binding.
I was looking first time how I can replace `RAC`/`RACObserve` in my particular uses, for example `UIButton`'s `enabled` property. I have to use `DynamicProperty` but this property is of type `AnyObject?` and ideally I would implement that property in ViewModel as `Bool`. So I tried to hide that `AnyObject?` property with something which could bind to `Bool`-parameterized property.

ended up with this:

```swift
class PropertiesContainer<P1 : MutablePropertyType, P2 : MutablePropertyType> {
    let proxyProperty : P1
    let originalProperty : P2
    init (proxy : P1, original : P2) {
        self.proxyProperty = proxy
[...]

PR #1915, opened 2015-04-20
`PropertyTerminal` implementation

gavrix:
This follows `RACChannelTerminal` concept and provides a way to support mutual property bindings through wrapping them in PropertyTerminal

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -159,6 +159,35 @@ extension MutableProperty: SinkType {
 	}
 }
 
+public class PropertyTerminal<P : MutablePropertyType> : MutablePropertyType {
+	typealias Value = P.Value
+	
+	public let property : P
+	private var _producerLocked : Bool
+	
+	public var value : Value {
+		get {
+			return self.property.value
+		}
+		set (x) {
+			self._producerLocked = true
+			self.property.value = x
+			self._producerLocked = false
+		}
+	}
+	
+	public var producer: SignalProducer<Value, NoError> {
+		get {
+			return self.property.producer |> filter { _ in !self._producerLocked }
+		}
+	}
+	
+	public init (property : P) {
+		self.property = property
+		self._producerLocked = false
+	}
+}
+
 infix operator <~ {
 	associativity right
 	precedence 90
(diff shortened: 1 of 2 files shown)


jspahrsummers:
Per https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1911#issuecomment-94597651, I don't think we're ready to tie ourselves to this in RAC 3.0, but we may revisit in the future.

Thank you for the pull request, though! :heart:

gavrix:
Ok I got that.

Just for the note, a couple of words about reasoning behind that mutual properties binding.
I was looking first time how I can replace `RAC`/`RACObserve` in my particular uses, for example `UIButton`'s `enabled` property. I have to use `DynamicProperty` but this property is of type `AnyObject?` and ideally I would implement that property in ViewModel as `Bool`. So I tried to hide that `AnyObject?` property with something which could bind to `Bool`-parameterized property.

ended up with this:

```swift
class PropertiesContainer<P1 : MutablePropertyType, P2 : MutablePropertyType> {
    let proxyProperty : P1
    let originalProperty : P2
    init (proxy : P1, original : P2) {
        self.proxyProperty = proxy
        self.originalProperty = original;
    }
}

extension UIControl {


    private struct AssociatedKeys {
        static var EnabledProperty = 0
    }

    
    public var enabledProperty : MutableProperty<Bool> {
        get {
            typealias EnabledPropertyProxyCotnainerType = PropertiesContainer<PropertyTerminal<MutableProperty<Bool>>, PropertyTerminal<DynamicProperty>>
            var propertiesPair : EnabledPropertyProxyCotnainerType?

            propertiesPair = objc_getAssociatedObject(self, &AssociatedKeys.EnabledProperty) as? EnabledPropertyProxyCotnainerType
            if propertiesPair == nil {
                let proxyProeprty = PropertyTerminal(property: MutableProperty(self.enabled))
                let originProperty = PropertyTerminal(property: DynamicProperty(object: self, keyPath: "enabled"));

                originProperty <~ proxyProeprty.producer |> map { $0 as AnyObject }
                proxyProeprty <~ originProperty.producer |> filter {$0 is Bool } |> map { $0 as! Bool }

                propertiesPair = PropertiesContainer(proxy: proxyProeprty, original: originProperty)
                objc_setAssociatedObject(self, &AssociatedKeys.EnabledProperty, propertiesPair, UInt(OBJC_ASSOCIATION_RETAIN))
            }

            return propertiesPair!.proxyProperty.property
        }
    }

}

``` 

So that was my uses case of mutual binding to provide type cast for Dynamic properties.

What's your thoughts on this problem? I mean typecasting `DynamicProperty` ?

Issue #1890, opened 2015-04-12
"Invalid pointer dequeued from free list" crash when using `skipRepeats` 


kaybutter:
I have been experiencing weird crashes since updating to Swift 1.2. They seem to be connected to `skipRepeats` or its call to `combinePrevious`. 

This simple example will reproduce the crash for me:

``` Swift
SignalProducer<String, NoError>(values: ["a", "b", "c", "a", "a", "c"])
  |> skipRepeats
  |> reduce("", +)
  |> start(next: { string in
    NSLog("\(string)")
  })
```

It crashes with a `SIGABRT` in `combinePrevious` that is used by `skipRepeats` and prints the following:

```
malloc: *** error for object 0x610000261f00: Invalid pointer dequeued from free list
*** set a breakpoint in malloc_error_break to debug
```

I have set up an example project that demonstrates the issue:
https://github.com/kaybutter/rac-crash

codeflows:
Tried the example project and it's crashing for me as well: https://gist.github.com/codeflows/c11a661251d0b86074fb

NachoSoto:
I just hit this as well in my own app :/ I'm using `skipRepeats` in several places, but only one of them reproduced this issue.

jspahrsummers:
As with https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1888#issuecomment-92211972, any kind of memory corruption sounds like a Swift toolchain issue.

We should try finding a workaround, but I would suggest filing a radar so the underlying problem gets fixed.

JaviSoto:
I'm getting a different crash here now after https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1902 :-/

![screen shot 2015-04-18 at 3 57 48 pm](https://cloud.githubusercontent.com/assets/666807/7217512/b6d953ee-e5e3-11e4-8ca9-49635db9e771.png)

JaviSoto:
@NachoSoto gave me a great idea for a workaround. This doesn't crash:

```swift
public func skipRepeatsWorkAround<T, E>(isRepeat: (T, T) -> Bool)(signal: Signal<T, E>) -> Signal<T, E> {
	return signal
		|> map { Box($0) }
		|> skipRepeats { isRepeat($0.unbox, $1.unbox) }
		|> map { $0.unbox }
}
```

NachoSoto:
[...]

PR #1917, opened 2015-04-20
Use multiple if-let unwrapping in skipRepeats

ikesyo:
Hopefully fixes https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1890#issuecomment-94210667.


JaviSoto:
Clever! 
Did you reproduce the crash before the fix, and stopped reproducing it after?

jspahrsummers:
tumblr_m2qjcz5pzy1qe2c25

ikesyo:
@JaviSoto I'm sorry, didn't confirmed that. :sweat:

PR #1918, opened 2015-04-20
Buffer behaviour fixed according to #1916

### LINKED ISSUE #1916: Buffer capacity should describe number of values, not events
Right now, `buffer(0)` is a buffer of zero events, and `buffer(1)` is a buffer of exactly one event.

I don't think this makes much sense, because it means that a 0-buffer will not replay anything, and a 1-buffer can never replay a value after it's terminated.

`RACReplaySubject`, on the other hand, treats the capacity as the number of _values_ that should be kept, and always replays termination. I think we should match that behavior with `buffer()`.

gavrix:
Fixes #1916.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -111,12 +111,16 @@ public struct SignalProducer<T, E: ErrorType> {
 		lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.buffer"
 
 		var events: [Event<T, E>] = []
+		var terminationEvent : Event<T,E>?
 		var observers: Bag<Signal<T, E>.Observer>? = Bag()
 
 		let producer = self { observer, disposable in
 			lock.lock()
 			for event in events {
 				observer.put(event)
+			}
+			if let termEvent = terminationEvent {
+				observer.put(termEvent)
 			}
 
 			let token = observers?.insert(observer)
@@ -137,12 +141,13 @@ public struct SignalProducer<T, E: ErrorType> {
 			// If not disposed…
 			if let liveObservers = observers {
 				if event.isTerminating {
+					terminationEvent = event
 					observers = nil
-				}
-
-				events.append(event)
-				while events.count > capacity {
-					events.removeAtIndex(0)
+				} else {
+					events.append(event)
+					while events.count > capacity {
+						events.removeAtIndex(0)
+					}
 				}
 
 				for observer in liveObservers {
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -118,6 +119,9 @@ public struct SignalProducer<T, E: ErrorType> {
 			for event in events {
 				observer.put(event)
 			}
+			if let termEvent = terminationEvent {
```
jspahrsummers: You can use `let terminationEvent` here, to shadow the variable with the unwrapped version.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -111,6 +111,7 @@ public struct SignalProducer<T, E: ErrorType> {
 		lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.buffer"
 
 		var events: [Event<T, E>] = []
+		var terminationEvent : Event<T,E>?
```
jspahrsummers: Can you please remove the space to the left of the `:`, to match the surrounding code?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,6 +233,29 @@ class SignalProducerSpec: QuickSpec {
...
 				expect(error).to(equal(TestError.Default))
 			}
+			
+			it("should always replay termination event") {
+				let (producer, sink) = SignalProducer<Int, TestError>.buffer(0)
+				var completed = false
+				sendCompleted(sink)
+				
+				producer .start(
+					completed: {
```
jspahrsummers: Can you please reformat this to match the other code in these tests?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,6 +233,29 @@ class SignalProducerSpec: QuickSpec {
...
+			it("should replay values after being terminated") {
+				let (producer, sink) = SignalProducer<Int, TestError>.buffer(1)
+				var value : Int?
+				
+				sendNext(sink, 123)
+				sendCompleted(sink)
+				producer.start(next: {val in
+					value = val
+				})
+				expect(value).to(equal(123))
```
jspahrsummers: Can this also verify that `.Completed` was sent after the value?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -233,6 +233,29 @@ class SignalProducerSpec: QuickSpec {
...
+				let (producer, sink) = SignalProducer<Int, TestError>.buffer(0)
+				var completed = false
+				sendCompleted(sink)
+				
+				producer .start(
+					completed: {
+					completed = true
+				})
+				expect(completed).to(beTruthy())
+			}
```
jspahrsummers: Can you please add a blank line in between these examples?

jspahrsummers:
This looks great, thanks! :sparkles:

I've left just a few minor notes above.

gavrix:
@jspahrsummers I made some adjustements according to comments, let me know if there if there's something yet to be improved.

jspahrsummers:
Looks good! :gem:

PR #1922, opened 2015-04-20
Update Signal.pipe example

richeterre:
This fixes the changelog's example code for `Signal.pipe`, which currently doesn't compile.


NachoSoto:
Thanks!

PR #1925, opened 2015-04-21
Use class name, not dynamicType

ikesyo:
Because this is final class.


JaviSoto:
Yeah, this makes sense to me. Any reason why it may have used `dynamicType` in the first place?

neilpa:
@JaviSoto I suspect leftovers from the earlier Swift betas.

JaviSoto:
That would make sense. I will defer this to somebody else to merge, I want to make sure I'm not missing something here :)

jspahrsummers:
This was basically just to avoid duplicating the class name unnecessarily, but it seems too obfuscated and not that big a deal. :+1: on merging.

PR #1926, opened 2015-04-21
Optimize collect() performance and memory usage

ikesyo:
Use a class type for the accumulator, avoid copying array.

It might not be acceptable because of premature optimization, but the improvements are not so bad.

Using the following test case and Debug Session's Memory Report in Xcode for Mac tests with 5 times:

```swift
fit("test collect performance") {
	let producer = SignalProducer<Int, NoError> { observer, _ in
		for i in 1...100_000 {
			sendNext(observer, i)
		}
		sendCompleted(observer)
	}

	let startTime = CACurrentMediaTime()

	producer |> collect |> start(completed: { println("Completed") })

	let endTime = CACurrentMediaTime()
	println("elapsed time: \(endTime - startTime)")
}
```

- old impl: 20.02MB memory, 4.511005624 seconds elapsed in average
- new impl: 7.58MB memory, 1.84319988 seconds elapsed in average

How do you think guys?

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -215,9 +215,18 @@ public func take<T, E>(count: Int)(signal: Signal<T, E>) -> Signal<T, E> {
 	}
 }
 
+private final class CollectState<T> {
+	var values: [T] = []
+
+	func append(value: T) -> Self {
+		values.append(value)
+		return self
+	}
+}
+
 /// Returns a signal that will yield an array of values when `signal` completes.
 public func collect<T, E>(signal: Signal<T, E>) -> Signal<[T], E> {
-	return signal |> reduce([]) { $0 + [ $1 ] }
+	return signal |> reduce(CollectState()) { $0.append($1) } |> map { $0.values }
 }
 
 /// Forwards all events onto the given scheduler, instead of whichever
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -215,6 +215,8 @@ public func take<T, E>(count: Int)(signal: Signal<T, E>) -> Signal<T, E> {
 	}
 }
 
+/// A reference type which wraps an array to avoid copying it for performance and
+/// memory usage optimization.
 private final class CollectState<T> {
 	var values: [T] = []
 


NachoSoto:
Nice :) the implementation is still fairly simple.

I wonder though, can you compare them in release? (with -O). I imagine this will still be better, but I wonder by how much.

ikesyo:
The test are running in release already thanks to #1905. :wink: 

The results with optimization off are:

- old impl: 19.08MB memory, 4.820028233 seconds elapsed in average
- new impl: 7.58MB memory, 2.038176889 seconds elapsed in average

NachoSoto:
@ikesyo oh of course, I forgot that was already merged.

This is definitely a substantial improvement, and given that the implementation is just as simple, I'm sold.
Does anyone have anything against this *premature* optimization?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -215,9 +215,18 @@ public func take<T, E>(count: Int)(signal: Signal<T, E>) -> Signal<T, E> {
...
+	func append(value: T) -> Self {
+		values.append(value)
+		return self
+	}
+}
+
 /// Returns a signal that will yield an array of values when `signal` completes.
 public func collect<T, E>(signal: Signal<T, E>) -> Signal<[T], E> {
-	return signal |> reduce([]) { $0 + [ $1 ] }
+	return signal |> reduce(CollectState()) { $0.append($1) } |> map { $0.values }
```
kastiglione: How do the numbers compare, if at all, to:

```swift
return signal |> reduce([]) { (var array, item) in
    array.append(item)
    return array
}
```

ikesyo: I don't have the data now, but I tested the code once. The result was better for memory usage ,but that required more seconds than the original implementation.

ikesyo: Oh I'm sorry, my memory was broken :sweat: 

This is the results for `var array` version with optimization on:

- 10.88MB memory, 4.155653999 seconds elapsed in average

jspahrsummers:
Generally :+1: on this idea, but can you please add an explanatory comment to the class, since its purpose isn't exactly obvious? :pray:

ikesyo:
I added a documentation comment. :smiley:

jspahrsummers:
Thanks! :shipit:

Issue #1868, opened 2015-03-31
The use of rac_description on the input to RACCommand execute: causes problems


Rupert-RR:
I haven't done a pull request because I'm not sure whether there is perhaps a good reason for this.
Anyway, I am coming across a crash using RAC with Swift when I call ``execute`` on a command passing in an object which doesn't inherit from ``NSObject``. The problem is that the final ``return`` of the method tries to set a name on the connection signal, using the ``rac_description`` of the input. Given that the input is an ``id`` I would have thought that this wouldn't be correct even in pure Objective C if we were to pass in an object not inheriting from ``NSObject``, but this certainly is the case for a pure Swift object.
Is there a reason we are doing this? Wouldn't the description of the command be more pertinent to the connection than its input object?
Thanks in advance for any explanation.
(If this is just an oversight, I'm happy to submit a pull request for it)

Rupert-RR:
A quick browse through the source code has found other instances too (other instances exist but are called on ``self`` which is always an object inheriting from ``NSObject``):
* ``RACEagerSequence return:``
* ``RACSignal+Operations aggregateWithStart::`` (both methods)
* ``RACStream mapReplace:``
* ``RACStream combinePreviousWithStart:``
* ``RACStream ignore:``
* ``RACStream startWith:``
* ``RACStream scanWithStart::`` (both methods)
* ``RACUnarySequence return:``

jspahrsummers:
Hmm, this is a good catch. We haven't really had to worry about objects that don't inherit from `NSObject`, because even proxies end up proxying an `NSObject` in typical Objective-C code.

Swift, of course, changes things. Maybe we could create a (private) global function that checks whether `rac_description` is implemented before calling it?

Rupert-RR:
Excuse my ignorance, but how would that work (especially in Swift)? methods like ``respondsToSelector:`` belong again to ``NSObject`` no? Or are you talking about some kind of global function that the developer can override to do nothing if he is using non-``NSObject`` objects?
Sorry I can't bring any suggestion to the table - I just don't see any reliable way to test for object inheritance in this case.

jspahrsummers:
`-respondsToSelector:` is part of the `<NSObject>` protocol, which Swift objects (and basically any object that wants to play nicely with Objective-C) implement.

Rupert-RR:
Ok, I hadn't realised that. Sounds like a plan then!

PR #1928, opened 2015-04-21
Issue #1868 RACDescription on non-NSObject objects

Rupert-RR:
This addresses issue #1868 by adding a global command to be used anywhere where we try to use ``rac_description`` on objects which are not known to inherit from ``NSObject``.
This is my first PR for this project, so be gentle!

### DIFF
--- a/ReactiveCocoa/NSObject+RACDescription.h
+++ b/ReactiveCocoa/NSObject+RACDescription.h
@@ -19,3 +19,6 @@
 - (NSString *)rac_description;
 
 @end
+
+// Global function to be called on any object not known to inherit from NSObject
+NSString *rac_description(id object);
--- a/ReactiveCocoa/NSObject+RACDescription.h
+++ b/ReactiveCocoa/NSObject+RACDescription.h
@@ -21,4 +21,4 @@
 @end
 
 // Global function to be called on any object not known to inherit from NSObject
-NSString *rac_description(id object);
+NSString *rac_description(id object);NSString *RACDescription(id object);
--- a/ReactiveCocoa/NSObject+RACDescription.h
+++ b/ReactiveCocoa/NSObject+RACDescription.h
@@ -8,17 +8,9 @@
 
 #import <Foundation/Foundation.h>
 
-// A private category providing a terser but faster alternative to -description.
-@interface NSObject (RACDescription)
-
-// A simplified description of the receiver, which does not invoke -description
+// A simplified description of the object, which does not invoke -description
 // (and thus should be much faster in many cases).
 //
 // This is for debugging purposes only, and will return a constant string
 // unless the RAC_DEBUG_SIGNAL_NAMES environment variable is set.
-- (NSString *)rac_description;
-
-@end
-
-// Global function to be called on any object not known to inherit from NSObject
-NSString *rac_description(id object);NSString *RACDescription(id object);
+NSString *RACDescription(id object);
--- a/ReactiveCocoa/NSObject+RACDescription.m
+++ b/ReactiveCocoa/NSObject+RACDescription.m
@@ -49,7 +49,7 @@
 
 @end
 
-NSString *rac_description(id object) {
+NSString *RACDescription(id object) {
 	if ([object respondsToSelector:@selector(rac_description)]) {
 		return [object rac_description];
 	} else {
--- a/ReactiveCocoa/NSObject+RACDescription.m
+++ b/ReactiveCocoa/NSObject+RACDescription.m
@@ -48,3 +48,15 @@
 }
 
 @end
+
+NSString *rac_description(id object) {
+	if ([object respondsToSelector:@selector(rac_description)]) {
+		return [object rac_description];
+	} else {
+		if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL && [object respondsToSelector:@selector(description)]) {
+			return [object description];
+		} else {
+			return @"(description skipped)";
+		}
+	}
+}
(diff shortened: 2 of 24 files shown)


ReactiveCocoa/NSObject+RACDescription.h
```diff
@@ -19,3 +19,6 @@
 - (NSString *)rac_description;
 
 @end
+
+// Global function to be called on any object not known to inherit from NSObject
+NSString *rac_description(id object);
```
jspahrsummers: Can you please call this `RACDescription`, to align with Apple's conventions for C functions?

ReactiveCocoa/NSObject+RACDescription.m
```diff
@@ -48,3 +48,15 @@ - (NSString *)rac_description {
 }
 
 @end
+
+NSString *rac_description(id object) {
+	if ([object respondsToSelector:@selector(rac_description)]) {
+		return [object rac_description];
+	} else {
+		if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL && [object respondsToSelector:@selector(description)]) {
```
jspahrsummers: Everything should respond to `-description`, I think.

ReactiveCocoa/NSObject+RACDescription.m
```diff
@@ -48,3 +48,15 @@ - (NSString *)rac_description {
 }
 
 @end
+
+NSString *rac_description(id object) {
+	if ([object respondsToSelector:@selector(rac_description)]) {
```
jspahrsummers: `-respondsToSelector:` is kind of expensive (especially when called frequently), so maybe we could default to the behavior of `-[NSObject rac_description]` here, and remove that method entirely.

Then, if `RAC_DEBUG_SIGNAL_NAMES` is set _and_ `-rac_description` is implemented (by a subclass only, since we'd have removed it from `NSObject`), we could use that instead—allowing us to preserve the `NSValue`, `NSString`, and `RACTuple` implementations.

Does that make sense?

jspahrsummers:
@Rupert-RR Thanks for the PR! Sorry for the delay in reviewing this (I was on vacation).

I've left some comments on the code above. Hope they're helpful! :sparkles:

Rupert-RR:
Hi! Likewise, sorry for not having replied before - I've been snowed under.

ReactiveCocoa/NSObject+RACDescription.h
Rupert-RR: Ok

ReactiveCocoa/NSObject+RACDescription.m
Rupert-RR: Are you sure about this? As far as I can see, ``-description`` is a property implemented on ``NSObject`` (and also by some other base classes) but there is no guarantee that it would be implemented for an ``id`` object.

ReactiveCocoa/NSObject+RACDescription.m
Rupert-RR: Sounds like a sensible idea, but I'm concerned that the ``-class`` method (used in the ``rac_description`` implementation) is not necessarily implemented by an ``id`` object.

jspahrsummers: Everything `id` has to obey the `<NSObject>` protocol—which includes a `-class` method—as a matter of practicality. AFAIK Swift objects are included in that regard.

ReactiveCocoa/NSObject+RACDescription.m
jspahrsummers: `-description` is part of [`<NSObject>`](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1928#discussion_r30458419), which is different from the `NSObject` root class.

ReactiveCocoa/NSObject+RACDescription.m
Rupert-RR: Just for my own instruction, do you have any sources for that?

ReactiveCocoa/NSObject+RACDescription.m
```diff
@@ -48,3 +36,15 @@ - (NSString *)rac_description {
 }
 
 @end
+
+NSString *RACDescription(id object) {
+	if (getenv("RAC_DEBUG_SIGNAL_NAMES") != NULL) {
+		if ([object respondsToSelector:@selector(rac_description)]) {
+			return [object rac_description];
+		} else {
+			return [[NSString alloc] initWithFormat:@"<%@: %p>", object.class, object];
```
Rupert-RR: This causes a compiler error "Property '``class``' not found on object of type '``__string id``'.
Do ``id`` objects definitely guarantee to implement this?

jspahrsummers: Yes, you just can't use dot syntax with `id`—use `[object class]` instead.

Rupert-RR: Oops - hadn't thought of that! I never really used to use dot syntax, but I suppose that now I've spent too much time with Swift!

Rupert-RR:
Ok, I think this should be good (like a naughty iOS developer I forgot to check the OSX builds!).
All tests are now passing.
I should probably add in tests for non-``NSObject`` objects in some of these signals to actually cover what I've done, but I will come back to that later.

jspahrsummers:
Looks good, thank you! :shipit:

Rupert-RR:
Thanks for your help and comments!

PR #1933, opened 2015-04-23
Upgrade to the new Result type

### LINKED ISSUE #1887: Update Result type

ikesyo:
Resolves #1887.


ReactiveCocoa/Swift/Event.swift
```diff
@@ -6,7 +6,15 @@
 //  Copyright (c) 2015 GitHub. All rights reserved.
 //
 
-import LlamaKit
+import Result
+
+public final class Box<T> {
```
NachoSoto: Is there a reason to not keep this `internal`?

I would also move it to its own file.

ReactiveCocoa/Swift/Event.swift
```diff
@@ -49,7 +57,7 @@ public enum Event<T, E: ErrorType> {
 	public func map<U>(f: T -> U) -> Event<U, E> {
 		switch self {
 		case let .Next(value):
-			return Event<U, E>.Next(Box(f(value.unbox)))
+			return Event<U, E>.Next(Box(f(value.value)))
```
NachoSoto: Maybe keep this as `unbox` to minimize changes?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -51,12 +51,10 @@ public struct SignalProducer<T, E: ErrorType> {
 	/// then complete, or immediately send an error, depending on the given
 	/// Result.
 	public init(result: Result<T, E>) {
-		switch result {
```
NachoSoto: Why this change? I think it's more clear as a `switch`.

NachoSoto:
Looking good! :star2:

ReactiveCocoa/Swift/SignalProducer.swift
ikesyo: New `Result` is not a enum type, so `switch` can't be used. And also the following could't be compiled:

```swift
public init(result: Result<T, E>) {
	result.analysis(ifSuccess: { value in
		self.init(value: value)
	}), ifFailure: { error in
		self.init(error: error)
	}
}
```

ReactiveCocoa/Swift/Event.swift
ikesyo: Because `Event` is `public`, `Box` also must be `public`.

ReactiveCocoa/Swift/Event.swift
ikesyo: Yeah, I had some thoughts about it. The changes are mainly for users who are using/would like to use this library for iOS 7 target (I am one of them) by importing the sources to an app target directly.

In the case, `Result`'s `Box` is visible in the entire app target (because that is included in the target), including RAC sources. To avoid conflict between `Result`'s `Box` and RAC's `Box`, the user must comment out either of them. If `box.value` and `box.unbox` are both exists, either A or B becomes uncompilable with the commenting out (so the user must fix the getters of which one of the two).

This is why I unified the unboxing getter.

ikesyo:
I noticed the existence of a class type `ManagedBuffer<Value, Element>` in standard lib, which can be used like:

```swift
extension ManagedBuffer {
	static func create(value: Value) -> ManagedBuffer {
		return self.create(1, initialValue: { _ in value })
	}
}

public enum Event<T, E: ErrorType> {
	case Next(ManagedBuffer<T, ()>)
	case Error(ManagedBuffer<E, ()>)
	case Completed
	case Interrupted

	public func map<U>(f: T -> U) -> Event<U, E> {
		switch self {
		case let .Next(value):
			// `ManagedBuffer` has `final var value: Value`
			return Event<U, E>.Next(ManagedBuffer.create(f(value.value)))
		case let .Error(error):
			return Event<U, E>.Error(error)
		case .Completed:
			return .Completed
		case .Interrupted:
			return .Interrupted
		}
	}
}

public func sendNext<T, E>(sink: SinkOf<Event<T, E>>, value: T) {
	sink.put(Event<T, E>.Next(ManagedBuffer.create(value)))
}
```

Could it be an alternative to `Box`?

neilpa:
> Could it be an alternative to Box?

No, that's not what `ManagedBuffer` is for.

Can we just add a dependency on @robrix's [`Box`](https://github.com/robrix/box)? We need a `Box` type here and that seems like the right choice.

ReactiveCocoa/Swift/Event.swift
robrix: `Result`’s `Box` is purely internal to `Result`.

robrix:
`Result` moved off Box.framework to use an internal `Box` type. It managed to keep its `Box` purely internal by hiding its use in an internal `enum`, `State`.

I suggest we follow suit here and ship a tiny `Box` type in RAC, because:

- dependencies are apparently viewed with skepticism because tooling is insufficient
- it’s doubtful that we’d ever use `BoxType` or `MutableBox` or anything else provided by Box.framework

I further suggest that it be public, because `switch`ing over `Event` is much more useful than over `Result`: there are several more `case`s, and therefore hiding `Box` may have a relatively high cost for many valid uses of `Box`.

If we find that the analysis methods suffice, we can always hide `Box` away later using the same technique as `Result`.

neilpa:
I understand wanting to make `Result` itself standalone, but I think that's the wrong choice for RAC.

If we ship a public `Box` it sets the wrong precedent and is going to be problematic for consumers when multiple frameworks do this. Now instead of simply using `Box` I have use `ReactiveCocoa.Box` or `SomeOtherFramework.Box` everywhere to avoid the naming ambiguity.

mdiep:
If `Box` is going to be public, then I think it makes sense to use Box.framework.

ReactiveCocoa/Swift/Event.swift
ikesyo: Sorry for the misunderstanding. 😓 For iOS 7, we totally can't use dynamic framework. So `Result`'s source files also should be included in an app target, `internal` means within the app target for the case.

ikesyo:
> No, that's not what ManagedBuffer is for.

OK, I got it. I'm sorry. 🙇

ReactiveCocoa/Swift/SignalProducer.swift
robrix: Hm. This is a problem I hadn’t foreseen. Maybe `Result` should be just a simple `enum` and depend on Box.framework too?

ReactiveCocoa/Swift/Event.swift
robrix: I was under the impression that RAC was iOS8+, but I guess that’s my misunderstanding!

neilpa: It is iOS8+, but can be made to work on iOS7

ReactiveCocoa/Swift/SignalProducer.swift
neilpa: > Maybe Result should be just a simple enum and depend on Box.framework too?
:+1:

iCodesign: I love the `enum` `result`, much easier to use.

robrix: Alright, [filed an issue on `Result` for this](https://github.com/antitypical/Result/issues/36). May have a chance to look at this today.

ikesyo:
Updated to the `Result` with `Box`. :boom:

robrix:
:tada:

mdiep:
Ideally, Result would be set to a release, not just pinned to a SHA.

robrix:
To which end I’ve just [tagged 0.4](https://github.com/antitypical/Result/releases/tag/0.4).

ikesyo:
OK, I did it. :wink:

robrix:
Looks pretty good to me. @neilpa, @NachoSoto, any further comments?

neilpa:
:+1:

ReactiveCocoa.xcodeproj/project.pbxproj
```diff
@@ -17,6 +17,14 @@
 		B696FB821A7640C00075236D /* TestError.swift in Sources */ = {isa = PBXBuildFile; fileRef = B696FB801A7640C00075236D /* TestError.swift */; };
 		BFA6B94D1A7604D400C846D1 /* SignalProducerNimbleMatchers.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFA6B94A1A76044800C846D1 /* SignalProducerNimbleMatchers.swift */; };
 		BFA6B94E1A7604D500C846D1 /* SignalProducerNimbleMatchers.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFA6B94A1A76044800C846D1 /* SignalProducerNimbleMatchers.swift */; };
+		CD183E821AED1E0E00848F5D /* Box.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = CD183E811AED1E0E00848F5D /* Box.framework */; };
```
NachoSoto: How come these are repeated 4 times?

NachoSoto:
Other than that, it looks good! :ship:

ReactiveCocoa.xcodeproj/project.pbxproj
robrix: My guess: Per target. RAC Mac, RAC iOS, RAC tests Mac, RAC tests iOS.

ikesyo: > My guess: Per target. RAC Mac, RAC iOS, RAC tests Mac, RAC tests iOS.

Exactly. :ok_hand:

robrix:
Alright, :shipit:!

PR #1942, opened 2015-04-27
Guarantee to update refresh control on main thread

### LINKED ISSUE #1940: [UIRefreshControl+RACCommandSupport] RefreshControl does not guarantee to call -endRefresh on main thread
I noticed that if I assign a RACComand, whose `signalBlock` is working on the background thread to the `rac_command` of a `UIRefreshControl`, the refresh control won't call `-endRefreshing` on main thread, so it won't end the animation immediately after the signal sends completed.

liyihe96:
Resolves #1940.

### DIFF
--- a/ReactiveCocoa/Objective-C/UIRefreshControl+RACCommandSupport.m
+++ b/ReactiveCocoa/Objective-C/UIRefreshControl+RACCommandSupport.m
@@ -36,7 +36,7 @@ static void *UIRefreshControlDisposableKey = &UIRefreshControlDisposableKey;
 	// Like RAC(self, enabled) = command.enabled; but with access to disposable.
 	RACDisposable *enabledDisposable = [command.enabled setKeyPath:@keypath(self.enabled) onObject:self];
 
-	RACDisposable *executionDisposable = [[[[self
+	RACDisposable *executionDisposable = [[[[[self
 		rac_signalForControlEvents:UIControlEventValueChanged]
 		map:^(UIRefreshControl *x) {
 			return [[[command
@@ -46,7 +46,7 @@ static void *UIRefreshControlDisposableKey = &UIRefreshControlDisposableKey;
 					return [RACSignal return:x];
 				}];
 		}]
-		concat]
+		concat] deliverOnMainThread]
 		subscribeNext:^(UIRefreshControl *x) {
 			[x endRefreshing];
 		}];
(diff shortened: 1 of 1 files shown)


neilpa:
Thanks for the PR!

However I'm not sure we want to make this change. IIRC, signals associated with commands are required to deliver events on the main thread. That means the signal returned from the generator block should be getting the `deliverOnMainThread` treatment and not here.

liyihe96:
Oh! I didn't know that the signal returned from the generator block is required to deliver on main thread, thanks!

kastiglione:
Actually, `RACCommand` takes care of delivering on the main thread. This is briefly mentioned in the [docs](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/5f1c27c04b142693793140ebf7d286254dacd6ff/ReactiveCocoa/RACCommand.h#L37). The relevant line of code is [here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/5f1c27c04b142693793140ebf7d286254dacd6ff/ReactiveCocoa/RACCommand.m#L155).

liyihe96:
Isn't that line of code used to deliver `executionSignals` on main thread but not the return value of `-execute:`? Sorry, I am not that familiar with ReactiveCocoa yet, if I am wrong. But when I checked if the subscription block is invoked on main thread using `[NSThread isMainThread]`, it gives me NO.

    RACDisposable *executionDisposable = [[[[self
		rac_signalForControlEvents:UIControlEventValueChanged]
		map:^(UIRefreshControl *x) {
			return [[[command
				execute:x]
				catchTo:[RACSignal empty]]
				then:^{
					return [RACSignal return:x];
				}];
		}]
		concat]
		subscribeNext:^(UIRefreshControl *x) {
            NSLog(@"Is on main thread: %d",[NSThread isMainThread]);
			[x endRefreshing];
		}];
   
    // output
    // Is on main thread: 0

kastiglione:
Ah yes, it's the execution signals, not their values, that are being delivered on main.

Is that the exact code you're using? I'm not sure what part of that would deliver from another thread/queue. Can you set a breakpoint in the subscribe block?

liyihe96:
Yeah, my PR is to apply `-deliverOnMainThread` before the `-subscribeNext:`, but I think apply it in the generator block will make more sense. Thanks!

jspahrsummers:
I'm :+1: on this change. Can you please put `deliverOnMainThread` on its own line, though? :pray:

liyihe96:
Just did, glad I could help. 😁

jspahrsummers:
Great, thank you! :sparkles:

PR #1949, opened 2015-04-30
Minor refactorings in Disposable for more immutability

ikesyo:


jspahrsummers:
I'm :+1: on the changes to `Atomic` variables, but the updates to the `State` structure seem unnecessary, since [value types are immutable in Swift](https://github.com/jspahrsummers/enemy-of-the-state/blob/52d1d5e4b399261b0745778bfc8aa8760c1b5c9b/Enemy%20of%20the%20State.md#value-types-are) anyways.

ikesyo:
> but the updates to the State structure seem unnecessary

I got it. :ok: I reverted the changes.

jspahrsummers:
:gem:

PR #1952, opened 2015-05-01
Fix comment

siemensikkema:
Changed reference to MainScheduler to UIScheduler which is probably what is meant here


jspahrsummers:
Agreed, thanks! :sparkles:

PR #1953, opened 2015-05-01
@noescape for Signal/SignalProducer, fix memory corruption in times()

ikesyo:
Running tests at 99b9995 leads to the result like following captures:


JaviSoto:
Nice! I think this looks good :) I'll let another pair of eyes look at it though :shipit:

neilpa:
Agreed

PR #1954, opened 2015-05-02
Minor improvements for 3.0 codebase

ikesyo:
Please see each commit for details. :wink:

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -191,13 +191,7 @@ public final class CocoaAction: NSObject {
 	/// object given to execute(), if it can be downcast successfully, or nil
 	/// otherwise.
 	public convenience init<Input: AnyObject, Output, Error>(_ action: Action<Input?, Output, Error>) {
-		self.init(action, { input in
-			if let input: AnyObject = input {
-				return input as? Input
-			} else {
-				return nil
-			}
-		})
+		self.init(action, { $0 as? Input })
 	}
 
 	deinit {
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -79,7 +79,7 @@ extension RACSignal {
 
 /// Turns each value into an Optional.
 private func optionalize<T, E>(signal: Signal<T, E>) -> Signal<T?, E> {
-	return signal |> map { Optional.Some($0) }
+	return signal |> map { Optional($0) }
 }
 
 /// Creates a RACSignal that will start() the producer once for each
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -169,7 +169,7 @@ infix operator <~ {
 ///
 /// The binding will automatically terminate when the property is deinitialized,
 /// or when the signal sends a `Completed` event.
-public func <~ <T, P: MutablePropertyType where P.Value == T>(property: P, signal: Signal<T, NoError>) -> Disposable {
+public func <~ <P: MutablePropertyType>(property: P, signal: Signal<P.Value, NoError>) -> Disposable {
 	let disposable = CompositeDisposable()
 	let propertyDisposable = property.producer.start(completed: {
 		disposable.dispose()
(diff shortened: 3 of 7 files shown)


ReactiveCocoa/Swift/Action.swift
```diff
@@ -191,13 +191,7 @@ public final class CocoaAction: NSObject {
...
 	/// otherwise.
 	public convenience init<Input: AnyObject, Output, Error>(_ action: Action<Input?, Output, Error>) {
-		self.init(action, { input in
-			if let input: AnyObject = input {
-				return input as? Input
-			} else {
-				return nil
-			}
-		})
+		self.init(action, { $0 as? Input })
```
NachoSoto: Hmm I'm not 100% sure about this, but shouldn't this be `as? Input?`?

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -79,7 +79,7 @@ extension RACSignal {
 
 /// Turns each value into an Optional.
 private func optionalize<T, E>(signal: Signal<T, E>) -> Signal<T?, E> {
-	return signal |> map { Optional.Some($0) }
+	return signal |> map { Optional($0) }
```
NachoSoto: Hmm I personally think the previos version is more accurate.

NachoSoto:
Other than those 2, the rest of improvements are great!

ReactiveCocoa/Swift/ObjectiveCBridging.swift
ikesyo: [here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/f7611e4747816e9828ab8b48674a1e531c64c4d4/ReactiveCocoa/Swift/Bag.swift#L32) and [here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/f7611e4747816e9828ab8b48674a1e531c64c4d4/ReactiveCocoa/Swift/Signal.swift#L531) use `Optional(value)`, so I unified with it. Should change the two instead?

ReactiveCocoa/Swift/Action.swift
ikesyo: `as?` operator returns `nil` if the value is not the casting type, so the return type would be `Input?`, I think it's OK.

NachoSoto: But `input` is an optional, and you're trying to cast it to a non optional. I guess if tests pass then this "just works", but I don't realy understand why.

ikesyo: When `$0 as? Input?` is used, the return type is `Input??`. To specify `Input?` as the casting type, we should use `as!` operator: `$0 as! Input?`

I think `$0 as? Input` is safer and more understandable than `$0 as! Input?`.

jspahrsummers: I think what @NachoSoto is asking is why `AnyObject? as? Input` would work (since that's what this statement seems to imply). I have to admit that I don't understand why it works either.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
jspahrsummers: This is fine.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -169,7 +169,7 @@ infix operator <~ {
 ///
 /// The binding will automatically terminate when the property is deinitialized,
 /// or when the signal sends a `Completed` event.
-public func <~ <T, P: MutablePropertyType where P.Value == T>(property: P, signal: Signal<T, NoError>) -> Disposable {
+public func <~ <P: MutablePropertyType>(property: P, signal: Signal<P.Value, NoError>) -> Disposable {
```
jspahrsummers: Ah, neat. Didn't know you could do this.

jspahrsummers:
:+1: Thanks!

ReactiveCocoa/Swift/Action.swift
ikesyo: I'v read [the type casting document](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/TypeCasting.html), but there was no description about taking optional values. :disappointed: 

But like the case in [ObjectiveCBridging.swift](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/f7611e4747816e9828ab8b48674a1e531c64c4d4/ReactiveCocoa/Swift/ObjectiveCBridging.swift#L149), casting operators can handle optional values as its input anyway. Thank you for the discussion @NachoSoto @jspahrsummers :bow:

ReactiveCocoa/Swift/Property.swift
NachoSoto: I don't think this used to compile.

PR #1961, opened 2015-05-04
Add zip and combineLatest array operators

larcus94:
So I've tried to use RAC as a Futures & Promises library in an app I'm working on and realized that the `zip` and `combineLatest` operators would only return tuples which is rather inconvenient when the number of signals or signal producers is variable.

So I've added the array operators to make code as follows possible:
```swift
zip(repositories.map { github.getRepository($0.fullName) }).start(error: nil, completed: nil, interrupted: nil, next:  {}
```

This is my first pull request for RAC so excuse me if I forgot about something very obvious.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -292,7 +292,7 @@ public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal
 	}
 }
 
-private func combineLatestWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
+internal func combineLatestWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
 	return signal |> combineLatestWith(otherSignal) |> map { $0.0 + [$0.1] }
 }
 
@@ -720,7 +720,7 @@ public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) ->
 	}
 }
 
-private func zipWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
+internal func zipWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
 	return signal |> zipWith(otherSignal) |> map { $0.0 + [$0.1] }
 }
 
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -263,7 +263,13 @@ private func observeWithStates<T, U, E>(signalState: CombineLatestState<T>, othe
 	}, interrupted: onInterrupted)
 }
 
-private func combineLatestWith<T, U, V, E>(otherSignal: Signal<U, E>, transform: ((T, U) -> V))(signal: Signal<T, E>) -> Signal<V, E> {
+/// Combines the latest value of the receiver with the latest value from
+/// the given signal.
+///
+/// The returned signal will not send a value until both inputs have sent
+/// at least one value each. If either signal is interrupted, the returned signal
+/// will also be interrupted.
+public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) -> Signal<(T, U), E> {
 	return Signal { observer in
 		let lock = NSRecursiveLock()
 		lock.name = "org.reactivecocoa.ReactiveCocoa.combineLatestWith"
@@ -272,7 +278,7 @@ private func combineLatestWith<T, U, V, E>(otherSignal: Signal<U, E>, transform:
 		let otherState = CombineLatestState<U>()
 		
 		let onBothNext = { () -> () in
-			sendNext(observer, transform(signalState.latestValue!, otherState.latestValue!))
+			sendNext(observer, (signalState.latestValue!, otherState.latestValue!))
 		}
 		
 		let onError = { sendError(observer, $0) }
@@ -286,18 +292,8 @@ private func combineLatestWith<T, U, V, E>(otherSignal: Signal<U, E>, transform:
 	}
 }
 
-/// Combines the latest value of the receiver with the latest value from
-/// the given signal.
-///
-/// The returned signal will not send a value until both inputs have sent
-/// at least one value each. If either signal is interrupted, the returned signal
-/// will also be interrupted.
-public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) -> Signal<(T, U), E> {
-	return signal |> combineLatestWith(otherSignal) { ($0, $1) }
-}
-
 public func combineLatestWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
-	return signal |> combineLatestWith(otherSignal) { $0 + [$1] }
+	return signal |> combineLatestWith(otherSignal) |> map { $0.0 + [$0.1] }
 }
 
 /// Delays `Next` and `Completed` events by the given interval, forwarding
@@ -654,82 +650,78 @@ private struct ZipState<T> {
 	}
 }
 
-private func zipWith<T, U, V, E>(otherSignal: Signal<U, E>, transform: ((T, U) -> V))(signal: Signal<T, E>) -> Signal<V, E> {
+/// Zips elements of two signals into pairs. The elements of any Nth pair
+/// are the Nth elements of the two input signals.
+public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) -> Signal<(T, U), E> {
 	return Signal { observer in
 		let initialStates: (ZipState<T>, ZipState<U>) = (ZipState(), ZipState())
 		let states: Atomic<(ZipState<T>, ZipState<U>)> = Atomic(initialStates)
-
+		
 		let flush = { () -> () in
 			var originalStates: (ZipState<T>, ZipState<U>)!
 			states.modify { states in
 				originalStates = states
-
+				
 				var updatedStates = states
 				let extractCount = min(states.0.values.count, states.1.values.count)
-
+				
 				removeRange(&updatedStates.0.values, 0 ..< extractCount)
 				removeRange(&updatedStates.1.values, 0 ..< extractCount)
 				return updatedStates
 			}
-
+			
 			while !originalStates.0.values.isEmpty && !originalStates.1.values.isEmpty {
 				let left = originalStates.0.values.removeAtIndex(0)
 				let right = originalStates.1.values.removeAtIndex(0)
-				sendNext(observer, transform(left, right))
+				sendNext(observer, (left, right))
 			}
-
+			
 			if originalStates.0.isFinished || originalStates.1.isFinished {
 				sendCompleted(observer)
 			}
 		}
-
+		
 		let onError = { sendError(observer, $0) }
 		let onInterrupted = { sendInterrupted(observer) }
-
+		
 		let signalDisposable = signal.observe(next: { value in
 			states.modify { (var states) in
 				states.0.values.append(value)
 				return states
 			}
-
+			
 			flush()
-		}, error: onError, completed: {
-			states.modify { (var states) in
-				states.0.completed = true
-				return states
-			}
-
-			flush()
-		}, interrupted: onInterrupted)
-
+			}, error: onError, completed: {
+				states.modify { (var states) in
+					states.0.completed = true
+					return states
+				}
+				
+				flush()
+			}, interrupted: onInterrupted)
+		
 		let otherDisposable = otherSignal.observe(next: { value in
 			states.modify { (var states) in
 				states.1.values.append(value)
 				return states
 			}
-
+			
 			flush()
-		}, error: onError, completed: {
-			states.modify { (var states) in
-				states.1.completed = true
-				return states
-			}
-
-			flush()
-		}, interrupted: onInterrupted)
-
+			}, error: onError, completed: {
+				states.modify { (var states) in
+					states.1.completed = true
+					return states
+				}
+				
+				flush()
+			}, interrupted: onInterrupted)
+		
 		return CompositeDisposable(ignoreNil([ signalDisposable, otherDisposable ]))
 	}
 }
 
-/// Zips elements of two signals into pairs. The elements of any Nth pair
-/// are the Nth elements of the two input signals.
-public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) -> Signal<(T, U), E> {
-	return signal |> zipWith(otherSignal) { ($0, $1) }
-}
-
 public func zipWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
-	return signal |> zipWith(otherSignal) { $0 + [$1] }
+	return signal |> zipWith(otherSignal) |> map { $0.0 + [$0.1] }
 }
 
 /// Applies `operation` to values from `signal` with `Success`ful results
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -292,7 +292,7 @@ public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal
 	}
 }
 
-public func combineLatestWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
+private func combineLatestWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
 	return signal |> combineLatestWith(otherSignal) |> map { $0.0 + [$0.1] }
 }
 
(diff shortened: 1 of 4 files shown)


jspahrsummers:

I'm kinda skeptical of the need for this, though, and I think matching the number of inputs to the number of outputs (with tuples in the type system) helps catch a good number of errors.

In your particular example, you could just `flatten` or `flatMap` those producers instead, for basically the same effect.

@ReactiveCocoa/reactivecocoa Any feels?

NachoSoto:
Yeah I feel the same way :thumbsup:

larcus94:
@jspahrsummers I was almost expecting an answer like that. I haven't been using RAC in objc a lot but started using it with swift so I missed this.
Would you mind explaining me how I could use `flatMap` then to zip an array of signal producers?

jspahrsummers:
It doesn't really seem like you're using the zipping in your example. If you just want to wait on an array of producers, you can merge them.

larcus94:
@jspahrsummers Why's that? With `zip` I end up with an array of repositories in the next closure. Would you mind explaining me how I could achieve this with `merge` and `flatMap`?

jspahrsummers:
Ah, I see—I didn't realize you were actually using the zipped result.

Okay, we can add this. :+1:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -263,36 +263,43 @@ private func observeWithStates<T, U, E>(signalState: CombineLatestState<T>, othe
...
 }
 
-/// Combines the latest value of the receiver with the latest value from
-/// the given signal.
-///
-/// The returned signal will not send a value until both inputs have sent
-/// at least one value each. If either signal is interrupted, the returned signal
-/// will also be interrupted.
-public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) -> Signal<(T, U), E> {
+private func combineLatestWith<T, U, V, E>(otherSignal: Signal<U, E>, transform: ((T, U) -> V))(signal: Signal<T, E>) -> Signal<V, E> {
```
jspahrsummers: This is no different than (the previous) `combineLatestWith` plus `map`. Can you please revert to that?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -263,36 +263,43 @@ private func observeWithStates<T, U, E>(signalState: CombineLatestState<T>, othe
...
+/// the given signal.
+///
+/// The returned signal will not send a value until both inputs have sent
+/// at least one value each. If either signal is interrupted, the returned signal
+/// will also be interrupted.
+public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) -> Signal<(T, U), E> {
+	return signal |> combineLatestWith(otherSignal) { ($0, $1) }
+}
+
+public func combineLatestWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
```
jspahrsummers: I don't think we'd want this variation to be public. A single `combineLatest` that supports sequences should be good enough.

If users want this specific operator, they can approximate it through `combineLatestWith` and `map`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -883,6 +898,17 @@ public func combineLatest<A, B, C, D, E, F, G, H, I, J, Error>(a: Signal<A, Erro
 		|> map(repack)
 }
 
+/// Combines the values of all the given signals, in the manner described by
+/// `combineLatestWith`.
+public func combineLatest<T, Error>(signals: [Signal<T, Error>]) -> Signal<[T], Error> {
```
jspahrsummers: Can we accept a sequence here, instead of specifically an array? :pray:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -883,6 +898,17 @@ public func combineLatest<A, B, C, D, E, F, G, H, I, J, Error>(a: Signal<A, Erro
...
+	if let first = signals.first {
+		let initial = first |> map { [$0] }
+		return signals[1 ..< signals.count].reduce(initial) { $0 |> combineLatestWith($1) }
+	}
+	
+	return Signal.never
```
jspahrsummers: I'm inclined to say that passing in an empty sequence should be a precondition failure, but I could be convinced otherwise.

@larcus94 Thoughts?

jspahrsummers:
I've added some notes above, most of which apply to `zip` and the producer equivalents too. :zap:

ReactiveCocoa/Swift/Signal.swift
larcus94: I thought about this too and decided to solve it like this dude to the following reasons:
- a precondition would yield the function very inconvenient, since you'd always have to check whether the array is empty
- It makes sense that combining zero signals returns a signal without an event

The thing I'm a bit concerned about is whether it's obvious that when one passes an empty array to this function that the observing closures would never be called. I think I would forget about this quite easily.

larcus94:
@jspahrsummers Cool, I'll fix the code tomorrow :)

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: :ok_hand: Sounds reasonable. As long as we document the behavior on the operator, I think we can leave it be.

larcus94:
@jspahrsummers Thanks for the feedback, I've updated the things you've pointed out. Let me know what you think :)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -273,26 +273,29 @@ public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal
...
+		
 		let signalDisposable = signal |> observeWithStates(signalState, otherState, lock, onBothNext, onError, onBothCompleted, onInterrupted)
 		let otherDisposable = otherSignal |> observeWithStates(otherState, signalState, lock, onBothNext, onError, onBothCompleted, onInterrupted)
-
+		
 		return CompositeDisposable(ignoreNil([ signalDisposable, otherDisposable ]))
 	}
 }
 
+public func combineLatestWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
```
jspahrsummers: Looks like this operator is still public. Can you please make it private?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -653,70 +656,74 @@ public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) ->
...
-			states.modify { (var states) in
-				states.0.completed = true
-				return states
-			}
-
+			
 			flush()
-		}, interrupted: onInterrupted)
-
+			}, error: onError, completed: {
```
jspahrsummers: Looks like the indentation here got messed up.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -653,70 +656,74 @@ public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) ->
...
-		}, error: onError, completed: {
-			states.modify { (var states) in
-				states.1.completed = true
-				return states
-			}
-
-			flush()
-		}, interrupted: onInterrupted)
-
+			}, error: onError, completed: {
```
jspahrsummers: Indentation got messed up here too.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -653,70 +656,74 @@ public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) ->
...
+				}
+				
+				flush()
+			}, interrupted: onInterrupted)
+		
 		return CompositeDisposable(ignoreNil([ signalDisposable, otherDisposable ]))
 	}
 }
 
+public func zipWith<T, E>(otherSignal: Signal<T, E>)(signal: Signal<[T], E>) -> Signal<[T], E> {
```
jspahrsummers: Could you please make this private?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -953,6 +972,18 @@ public func zip<A, B, C, D, E, F, G, H, I, J, Error>(a: Signal<A, Error>, b: Sig
 		|> map(repack)
 }
 
+/// Zips the values of all the given signals, in the manner described by
+/// `zipWith`.
+public func zip<T, Error>(signals: [Signal<T, Error>]) -> Signal<[T], Error> {
```
jspahrsummers: This is still accepting an array (even though the implementation is now perfectly generic). Could you please change it to a sequence?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -883,6 +890,18 @@ public func combineLatest<A, B, C, D, E, F, G, H, I, J, Error>(a: Signal<A, Erro
 		|> map(repack)
 }
 
+/// Combines the values of all the given signals, in the manner described by
+/// `combineLatestWith`.
```
jspahrsummers: Can you please document that this operator will never send events if the sequence is empty?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -953,6 +972,18 @@ public func zip<A, B, C, D, E, F, G, H, I, J, Error>(a: Signal<A, Error>, b: Sig
 		|> map(repack)
 }
 
+/// Zips the values of all the given signals, in the manner described by
+/// `zipWith`.
```
jspahrsummers: Can you please document that this operator will never send events if the sequence is empty?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -474,6 +482,19 @@ public func combineLatest<A, B, C, D, E, F, G, H, I, J, Error>(a: SignalProducer
 		|> map(repack)
 }
 
+/// Combines the values of all the given producers, in the manner described by
+/// `combineLatestWith`.
+public func combineLatest<T, Error>(signalProducers: [SignalProducer<T, Error>]) -> SignalProducer<[T], Error> {
```
jspahrsummers: Can you please use a sequence here too?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -544,6 +565,19 @@ public func zip<A, B, C, D, E, F, G, H, I, J, Error>(a: SignalProducer<A, Error>
 		|> map(repack)
 }
 
+/// Zips the values of all the given producers, in the manner described by
+/// `zipWith`.
+public func zip<T, Error>(signalProducers: [SignalProducer<T, Error>]) -> SignalProducer<[T], Error> {
```
jspahrsummers: Can you please use a sequence here too?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -544,6 +565,19 @@ public func zip<A, B, C, D, E, F, G, H, I, J, Error>(a: SignalProducer<A, Error>
...
+	if let first = signalProducers.first {
+		let initial = first.lift { signal in
```
jspahrsummers: The lift here shouldn't be necessary, since `|>` already does it for you.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -474,6 +482,19 @@ public func combineLatest<A, B, C, D, E, F, G, H, I, J, Error>(a: SignalProducer
 		|> map(repack)
 }
 
+/// Combines the values of all the given producers, in the manner described by
+/// `combineLatestWith`.
+public func combineLatest<T, Error>(signalProducers: [SignalProducer<T, Error>]) -> SignalProducer<[T], Error> {
+	if let first = signalProducers.first {
+		let initial = first.lift { signal in
```
jspahrsummers: The lift here shouldn't be necessary, since `|>` already does it for you.

jspahrsummers:
Thanks for making those changes! Could you please also add a few small unit tests that exercise the new `combineLatest` and `zip` over a sequence of signals, and a sequence of producers? :pray:

ReactiveCocoa/Swift/Signal.swift
larcus94: Oh, I made the equivalent of the signal producer private. Made this private too now :)

robertjpayne:
I'd really appreciate this commit too – I'm doing a very similar thing ( have a set of 10 network signals I need to run in parallel and get the results of all of them ).

larcus94:
@jspahrsummers So I fixed now everything and added some specs too. You might want to check out my comment on an expected value that I don't quite get

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -1566,60 +1569,88 @@ class SignalSpec: QuickSpec {
...
-				|> observe(next: {
-						combinedValues = [$0, $1, $2]
-					}, completed: {
-						completed = true
-					})
 			}
 			
-			it("should forward the latest values from all inputs"){
-				expect(combinedValues).to(beNil())
+			sharedExamples("combineLatest examples") {
```
jspahrsummers: Can you please use a constant string for this, so we can refer to the constant by name instead of duplicating the string contents below?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -1566,60 +1569,88 @@ class SignalSpec: QuickSpec {
...
-				expect(combinedValues).to(beNil())
-
-				sendNext(sinkB, 1)
-				expect(combinedValues).to(beNil())
+			describe("tuple") {
+				beforeEach {
+					combineLatest(signalA, signalB, signalC)
+						|> observe(next: {
+							combinedValues = [$0, $1, $2]
+							}, completed: {
```
jspahrsummers: This can be outdented to match the `|>`.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -1631,58 +1662,83 @@ class SignalSpec: QuickSpec {
...
+					
+					sendNext(sinkA, 0)
+					expect(completed).to(beTruthy())
+				}
+			}
+			
+			describe("tuple") {
+				beforeEach {
+					zip(signalA, signalB, signalC)
+					|> observe(next: {
```
jspahrsummers: Can please indent the `|>` and lines following?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -1631,58 +1662,83 @@ class SignalSpec: QuickSpec {
...
-				sendNext(sinkB, 1)
-				sendNext(sinkC, 2)
-				sendCompleted(sinkB)
-				sendCompleted(sinkC)
-				expect(completed).to(beFalsy())
+			
+			describe("sequence") {
+				beforeEach {
+					zip([signalA, signalB, signalC])
+					|> observe(next: {
```
jspahrsummers: As above.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -1631,58 +1662,83 @@ class SignalSpec: QuickSpec {
...
+				signalA = baseSignalA
+				signalB = baseSignalB
+				signalC = baseSignalC
+				
 				sinkA = baseSinkA
 				sinkB = baseSinkB
 				sinkC = baseSinkC
+			}
+			
+			sharedExamples("zip examples") {
```
jspahrsummers: As above.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -1566,60 +1569,88 @@ class SignalSpec: QuickSpec {
...
-				
-				sendCompleted(sinkA)
-				sendCompleted(sinkB)
-				expect(completed).to(beFalsy())
+			describe("sequence") {
+				beforeEach {
+					combineLatest([signalA, signalB, signalC])
+						|> observe(next: {
+							combinedValues = $0
+							}, completed: {
```
jspahrsummers: As above.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -273,26 +273,29 @@ public func combineLatestWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal
 	return Signal { observer in
 		let lock = NSRecursiveLock()
 		lock.name = "org.reactivecocoa.ReactiveCocoa.combineLatestWith"
-
+
```
jspahrsummers: Can you please fix the whitespace changes in this file, so lines unrelated to the diff don't change?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -653,70 +656,74 @@ public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) ->
...
+			
 			flush()
 		}, error: onError, completed: {
-			states.modify { (var states) in
-				states.0.completed = true
-				return states
-			}
-
-			flush()
+				states.modify { (var states) in
```
jspahrsummers: This new indentation is incorrect.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -653,70 +656,74 @@ public func zipWith<T, U, E>(otherSignal: Signal<U, E>)(signal: Signal<T, E>) ->
...
+			
 			flush()
 		}, error: onError, completed: {
-			states.modify { (var states) in
-				states.1.completed = true
-				return states
-			}
-
-			flush()
+				states.modify { (var states) in
```
jspahrsummers: As above.

jspahrsummers:
Sorry about [the delay](https://twitter.com/jspahrsummers/status/597520138456305665)! I've responded to your question and left a few more minor notes.

larcus94:
No worries :) Fixed the indentation too now :)

PR #1966, opened 2015-05-05
Flatten operations for signal-of-producers

### LINKED ISSUE #1930: Mapping RAC2 hot signals to RAC3 Signal
I'm trying to map between RAC2 and RAC3 concepts, most of which seem to make sense. The one that I am struggling with is the mapping of hot signals.

The ObjC Bridging API allows you to construct a SignalProducer from a RACSignal, however it does not have an API method for constructing a Signal from a RACSignal. 

As a result, any RAC2 signal is bridged to a SignalProducer and as such becomes a cold signal. Personally I consider signals derived from UI control to be hot, however I know that this is debatable.

This doesn't stop me getting the job done, I can certainly construct a SignalProducer in this case:

```
extension UITextField {
  func rac_textSignalProducer() -> SignalProducer<String, NSError> {
    return
      self.rac_textSignal().toSignalProducer()
        |> map { $0 as! String }
  }
}
```

And this works just fine for my purposes, but it does still leave me wondering how to bridge hot signals from RACSignal to Signal?

neilpa:
Will resolve #1930


NachoSoto:
Woot, thanks for tackling this @neilpa!

neilpa:
Ready for review. For now I've left the tests in terms of producers-of-producers but can move/dupe them depending on what people think

NachoSoto:
> For now I've left the tests in terms of producers-of-producers but can move/dupe them depending on what people think

Yeah, ideally we'd reuse the same tests, but probably not possible?

jspahrsummers:
Looks great to me as-is. Thanks @neilpa!

jspahrsummers:
This is implemented correctly, but completely _destroys_ the type inferencer. Not sure what we can do about that. :disappointed:

jspahrsummers:
Splitting `flatMap` into its constituent parts seems to help, as a workaround.

neilpa:
:sob:

One possibility is replacing the lifting overload of `|>` with explicit implementations for the remaining producer operators. Extra boilerplate, but it may also help newcomers when trying to figure out what's available for signals vs producers.

jspahrsummers:
That is certainly one option, but I'd really like to avoid it if possible.

Maybe the bountiful @robrix could take a look? He might know some good workarounds for the type inferencer. :pray:

ikesyo:
It seems that if `flatMap()` of producer is removed, lifting to `flatMap()` of signal works... It's not the case?

robrix:
“destroys” in what sense? Types can’t be inferred at all? Compile times through the roof? Z̳͔͙a͔͕͜l͈͜g͚͉̯̞o͉̪̯̥͈̲͔?̵̩

ikesyo:
@robrix The compiler fails to compile with the message like: `Could not find member 'Merge'`.

```swift
let f = flatMap(.Merge) {
    value in someProducer(value)
} // This is ambiguous, which `flatMap()` function should be used?
let g: SignalProducer<T, E> -> SignalProducer<U, E> = flatMap(.Merge) {
    value in someProducer(value)
} // This is OK.
```

robrix:
Gotcha. Thanks!

PR #1968, opened 2015-05-06
Set Whole Module Optimization on

ikesyo:
Additionaly fix a runtime crash in ActionSpec due to Whole Module Optimization. It is happening at `action.apply(0) |> first`, the affected cases are:

- it("should generate KVO notifications for enabled")
- it("should generate KVO notifications for executing")


ReactiveCocoa.xcodeproj/project.pbxproj
```diff
@@ -1979,6 +1979,7 @@
 				GCC_NO_COMMON_BLOCKS = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				SWIFT_WHOLE_MODULE_OPTIMIZATION = YES;
```
NachoSoto: I imagine this is only for release?

NachoSoto:
So tests are passing on release with the setting on? :)

ReactiveCocoa.xcodeproj/project.pbxproj
ikesyo: This is for Release and Profile, same as Optimization Level.

ikesyo:
@NachoSoto Yes, tests at 82c9e7d are passing, but at 1ee2baa were failing. :grin:

ikesyo:
It seems to me that curried function syntax crashes mainly occurs with protocol type or tuple type. :baby_chick:

The following are revised functions:

```swift
public func observeOn<T, E>(scheduler: SchedulerType) -> (Signal<T, E>) -> Signal<T, E>
public func delay<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType) -> Signal<T, E> -> Signal<T, E>
public func combinePrevious<T, E>(initial: T) -> Signal<T, E> -> Signal<(T, T), E>
public func reduce<T, U, E>(initial: U, combine: (U, T) -> U) -> Signal<T, E> -> Signal<U, E>
public func scan<T, U, E>(initial: U, combine: (U, T) -> U) -> Signal<T, E> -> Signal<U, E>
public func throttle<T, E>(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType) -> Signal<T, E> -> Signal<T, E>
public func timeoutWithError<T, E>(error: E, afterInterval interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType) -> Signal<T, E> -> Signal<T, E>

public func startOn<T, E>(scheduler: SchedulerType) -> SignalProducer<T, E> -> SignalProducer<T, E>
public func start<T, E, S: SinkType where S.Element == Event<T, E>>(sink: S) -> SignalProducer<T, E> -> Disposable
```

jspahrsummers:
Is there a reason to enable this, besides the improved compilation speed (on Xcode 6.3, due to a bug)?

The sketchiness of Swift's optimizations right now makes me really cautious about changes like this.

ikesyo:
@jspahrsummers I didn't think about compilation speed, just for performance optimization. But perhaps I didn't understand the optimization correctly at that time.

It cares reducing dynamic dispatches of non-final `internal class` types. However in the codebase, almost all class types are annotated with `final` other than `CombineLatestState` and `CompositeDisposable.DisposableHandle` I'll tackle the two in another PR, close this out. :sweat_smile:

PR #1970, opened 2015-05-06
[Proposal] Loosen error constraint for Signal/SignalProducer/Action

ikesyo:
It seems to me that `ErrorType` constraint is needed for mainly (or only?) bridging to `RACSignal`, and that could be expressed in bridging APIs.

To remove the constraint from `Signal`/`SignalProducer`/`Action`, I did the following changes:

- `ErrorType.nsError` is now optional type (`-[RACSubscriber sendError:]` can take nil, so I think it's OK).
- `ActionError.nsError` returns `nil` if `E` is not `ErrorType`.
- `NoError` is not `ErrorType` any more.
    - Add `NoError` variants of `toRACSignal()` functions.

How does it look?


jspahrsummers:
> It seems to me that `ErrorType` constraint is needed for mainly (or only?) bridging to `RACSignal`

I can see why it appears that way, but I disagree. `ErrorType` is used for _clarifying intention_ and requiring it to be used a certain way.

It doesn't impose very many restrictions, but I think it helps to clearly indicate error handling code.

ikesyo:
@jspahrsummers Thank you for the explanation! :pray: I understood the direction now, close this. :boom:

kaybutter:
I understand the intention behind having an ErrorType Protocol for all errors. But is it really necessary to be able to transform every error into an NSError instance?

jspahrsummers:
@kaybutter Since Objective-C is still such a big part of Cocoa development, I think it's a reasonable constraint. Your error type may eventually reach a Plain Old Objective-C API, for example, and should have _some_ representation there.

If you don't care, you could just return one `NSError` instance for all of your cases.

PR #1975, opened 2015-05-07
Drop Interrupted events from flattened inner signals

### LINKED ISSUE #1967: Should flatten operations ignore Interrupted events on inner signals
From the discussion started in #1930

>> @neilpa Since Signals send Interrupted if observed after they've already terminated, those operators wouldn't behave as expected.

> I was thinking about pre-Interrupted world. However, this makes me wonder if an Interrupted event on the inner signals should be ignored rather than "poison" the outer signal in the same way as Error.

neilpa:
Resolves #1967


neilpa:
Should actually work now

jspahrsummers:
:gem::sparkles:

PR #1984, opened 2015-05-08
Adds example of using MutableProperty

aventurella:
It was not exactly clear from the CHANGELOG the proper usage of MutableProperty. Added a more specific example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -232,6 +232,54 @@ bound to. If changes to that property should be visible to consumers, it can
 additionally be wrapped in `PropertyOf` (to hide the mutable bits) and exposed
 publicly.
 
+**Using Mutable Properties**
+
+```swift
+
+import ReactiveCocoa
+
+class Foo{
+    var label: MutableProperty<String?>
+
+    func init(){
+        label = MutableProperty(nil)
+    }
+}
+
+
+class SomeViewController{
+    var myFoo: Foo!
+
+    func myInitialization(){
+
+        myFoo = Foo()
+        myFoo.label.value = "ReactiveCocoa"
+
+        // we don't care about the initial value, so we will skip it
+        // we only care about changed values after the 1st value.
+        let mapped = myFoo.label.producer
+            |> skip(1)
+
+        // now lets get notified:
+        mapped.start(next: { value in
+            println("The new value is: \(value)")
+        })
+
+        changeLabelLaterTo("More ReactiveCocoa")
+    }
+
+    func changeLabelLaterTo(value: String){
+        // fake a change:
+        let delay = Int64(1 * Double(NSEC_PER_SEC))
+        let delayTime = dispatch_time(DISPATCH_TIME_NOW, delay)
+        dispatch_after(delayTime, dispatch_get_main_queue()) {
+            self.myFoo.label.value = value
+        }
+    }
+}
+
+```
+
 **If KVC or KVO is required by a specific API**—for example, to observe changes
 to `NSOperation.executing`—RAC 3 offers a `DynamicProperty` type that can wrap
 those key paths. Use this class with caution, though, as it can’t offer any type


jspahrsummers:

I'm not sure that the changelog is the best place for this, though, since the example is not really referring to the _changes_, but just demonstrating how to use the type itself (even if you had no experience with the prior version).

Perhaps a simplified version of this example could be added to the API documentation instead?

iv-mexx:
If only i had seen this PR two days ago when i was trying rac 3.0 with swift the first time... this example would have saved me some time :+1: 

However i agree that this should be somewhere in the docs (thats where i was searching, fully aware that the docs have not yet been update for 3.0 / swift yet)

aventurella:
Ya, it was pretty brutal for me to try to get started with RAC 3. I was more posting that example for any other wayward souls who were also having issues and to verify my assumptions about expected use patterns. I agree that the CHANGELOG is not likely the best place. Is there some kind of central wiki or documentation site where I could submit that to instead? (This would apply for the other PRs I submitted to the CHANGELOG as well).

jspahrsummers:
We have “header” documentation. I think this belongs in a spot that's close to the point-of-use.

iv-mexx:
Usually, the "header" documentation is working well for me, but in this case I didn't know that i needed to use `MutableProperty` in the first place. My problem was more of the "where do i get started" kind.

In my particular case, I had a relatively simple Login Mask implemented in MVVM style with RAC 3.0 in Objective C.
The [GroceryList](https://github.com/jspahrsummers/GroceryList) sample app was very helpful to see how everything works together, which can be found directly from the Readme. 
Then I was trying to do the exact same thing in Swift and I was stuck until I found the [SwiftRACGoogleImages](https://github.com/adly-holler/SwiftRACGoogleImages) project linked in one of the issues here. 

I think, all it needs is to have a sample project written in Swift linked from the Readme, at least to cover the kind of problems that I had.

jspahrsummers:
Header documentation is a perfectly valid place for sample code, IMO, as long as it doesn't expand to the size of a full project. I think `MutableProperty` could be demonstrated in a succinct example.

As for whole projects, we’ve maintained sample projects in the past, and the problem is that they become outdated pretty quickly. We also hesitate to recommend others’ projects, out of concerns that they may not implement the “best” patterns that we want folks to learn from.

Anyways, let’s keep this discussion focused on the `MutableProperty` example specifically.

iv-mexx:
Sorry for becoming off topic but thanks for your answer :hamster: 

So back go `MutableProperty` specifically: I was just re-reading the CHANGELOG again (maybe should have done that before commenting in the first place :blush: ) and noticed that in the next section, [Using Signal.pipe instead of RACSubject](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-development/CHANGELOG.md#using-signalpipe-instead-of-racsubject), there is this short snippet demonstrating the use of `Signal.pipe`:

```Swift
let (signal, sink) = Signal<Int, NoError>.pipe()

signal.observe(next: { value in
    println(value)
})

// Prints each number
sendNext(sink, 0)
sendNext(sink, 1)
sendNext(sink, 2)
```

For me the situation looks very similar to `PropertyType` which could justify the existence of an example for `PropertyType` in the CHANGELOG as well?

Maybe in the form of the _"same"_ example but solved with `MutableProperty` like this:

```Swift
let property = MutableProperty<Int>(0)

property.producer.start( next: { value in
    println(value)
})

// Prints each number starting with `0` as the initial value before changing the value of `property`
property.put(1)
property.put(2)
```

Correct me if I'm wrong, but seeing those two examples in direct comparison might also help to decide which one is the better solution for a given situation.

jspahrsummers:
The primary purpose of the `Signal.pipe()` example in the changelog is to distinguish it from `RACSubject`.

If we were to add a comparable example for properties, it should provide a useful comparison against `RACObserve()` and `RAC()` from the Objective-C API.

While useful, I think that differs from the original example proposed here, which is more of an introduction to the idea of properties in general.

CHANGELOG.md
```diff
@@ -232,6 +232,54 @@ bound to. If changes to that property should be visible to consumers, it can
...
 publicly.
 
+**Using Mutable Properties**
+
+```swift
+
+import ReactiveCocoa
+
+class Foo{
+    var label: MutableProperty<String?>
```
jakecraige: Thoughts on doing something like this instead?

```swit
var label: MutableProperty<String?> = MutableProperty(nil)
```

PR #1985, opened 2015-05-08
Adds Action example to CHANGELOG

aventurella:
The existing example could clarify a bit more how it works. It currently assumes that the reader knows how to build the Action<Input, Output, Error> but does not provide an example.


jspahrsummers:
I have the same thoughts here as in #1984, but thank you for the pull request! :stars:

PR #1988, opened 2015-05-10
Remove flatMap() for producer-of-producers

ikesyo:
Will resolve https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1966#issuecomment-99536117.

The function introduced ambiguity for calling `producer |> flatMap(strategy) { value in ... }`. `flatMap()` for signal-of-producers can be lifted to producer-of-producers.


jspahrsummers:
Ah, makes sense. Thanks! :heart:

PR #1992, opened 2015-05-10
Enable app extension compatibility

### LINKED ISSUE #1990: Make RAC 3 safe for use in app extensions, e.g. today widgets
I'm using RAC 3 in a today widget. The widget is written in Swift and only uses the Swift API. It all seems to work nicely, but the linker always warns me that it is not safe to use ReactiveCocoa in an app extension.

I guess RAC has not "Allow app extension API only" checked in its settings.

I wonder if that could be changed or if - in the case that some part of RAC is really not safe for app extensions - a compatible subset could be extracted.

neilpa:
Resolves #1990. Should have been resolved with e24582e but somewhere along the way the submodule pin for xcconfigs got reverted to an older commit.

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -1989,7 +1989,6 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D047263A19E49FE8006002AA /* Mac-Framework.xcconfig */;
 			buildSettings = {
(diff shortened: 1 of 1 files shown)


NachoSoto:
Thank you! Was did automatically updated with `carthage update` or did you manually update the submodule?

neilpa:
I used `carthage update --no-use-binaries` to workaround Carthage/Carthage#419 so it didn't also delete Box and Result.

NachoSoto:
:tada:

Issue #455, opened 2013-04-30
startWith: current value of signal's observed keypath/value?


zakdances:
I'm setting up a signal with `combineLatest:` that I want to fire immediately. In order for that to happen, each signal passed to `combineLatest:` needs to have fired once already. If I supply each signal with a `startsWith:` value (or if I use `RACAbleWithStart`), it triggers correctly. The problem with that is I need to explicitly feed in the value for that signal's keypath:

```
[[RACSignal combineLatest:@[[
            self.view.rcl_boundsSignal startWith:[NSValue valueWithCGRect:self.view.bounds]] ,
            RACAbleWithStart(btnMode) ,
            [self.orientationSignal startWith:[NSNumber numberWithInteger:self.interfaceOrientation]]]
    subscribeNext:^(id x) {

    NSLog(@"testing 123");

}];
```

Instead of `startsWith:` I want something like `startsWithCurrentKeypathValue` so that I don't need to explicitly fill in the default. This works beautifully with `RACAbleWithStart`....so how can I do this with already-created signals?

`replayLast` is close, but it still only works under certain conditions (also doesn't seem to work with signals created with `distinctUntilChanged`).

Issue #445, opened 2013-04-23
`rac_signalForSelector` never gets fired. Why?


zakdances:
I don't know why my signal isn't getting fired:

```
- (id)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil
{
    self = [super initWithNibName:nibNameOrNil bundle:nibBundleOrNil];
    if (self) {
        // Custom initialization

        [[RACSignal rac_signalForSelector:@selector(myMethod:)]
                            subscribeNext:^(NSString *x) {
            NSLog(@"fired with: %@",x);
        }];
        [self myMethod:@"asd"];
    }
    return self;
}

- (void)myMethod:(id)string
{

}
```

Why does the signal never get fired (even when `myMethod` is called)?

PR #2004, opened 2015-05-14
Replace NSRecursiveLock with Atomic for observers management in Signal

ikesyo:
It seems that the recursive lock causes deadlocks which is happening at Carthage. But I'm not sure this completely fix the issue. Is this truly right direction?

Might fix https://github.com/Carthage/Carthage/issues/445 and https://github.com/Carthage/Carthage/issues/455.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -17,8 +17,7 @@ import Result
 public final class Signal<T, E: ErrorType> {
 	public typealias Observer = SinkOf<Event<T, E>>
 
-	private let lock = NSRecursiveLock()
-	private var observers: Bag<Observer>? = Bag()
+	private var observers: Atomic<Bag<Observer>?> = Atomic(Bag())
 
 	/// Initializes a Signal that will immediately invoke the given generator,
 	/// then forward events sent to the given observer.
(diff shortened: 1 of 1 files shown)


ikesyo:
The old `HotSignal` did use a concurrent queue and `dispatch_barrier_sync/async` for observers modification.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -27,19 +26,17 @@ public final class Signal<T, E: ErrorType> {
...
+			if event.isTerminating {
+				// Disallow any further events (e.g., any triggered recursively).
+				observers = self.observers.swap(nil)
+			} else {
+				observers = self.observers.value
+			}
 
+			if let observers = observers {
 				for sink in observers {
 					sink.put(event)
```
ikesyo: The point was putting event to observers outside the lock...

norio-nomura:
As I tested on #455, this branch introduces another crash.
But applying e806194 to v3.0-beta.4 fixed #455 and did not crash.

With that applying, I tested following Cartfile from #445
```
github "ReactiveCocoa/ReactiveCocoa" "swift-development"
github "thoughtbot/Argo" "master"
github "AliSoftware/OHHTTPStubs"
github "danielgindi/ios-charts" "master"

github "Quick/Quick"
github "Quick/Nimble"
```
I got `carthage update` run with very high cpu usage and building didn't progress. : persevere:
It's needed additional fix.

jspahrsummers:
As much as I would like to do this, it begins to violate the contract of `Signal` (one inherited from `RACSignal`), which says that signal events must be serialized. This change could result in events arriving concurrently, and would violate assumptions made in most of our operators right now.

See [the very long previous discussion](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1120) about recursive locking vs. queue/drain semantics. We ultimately decided that the tradeoffs of recursive locking are preferable.

ikesyo:
Thank you for the reference. :bow: I'll close it out.

norio-nomura:
Oh, I should ask about recursion eight days ago.
I know that original Reactive Extensions using scheduler for preventing recursion, but it did seems RAC did not use scheduler for it. I thought that RAC had different policy…

jspahrsummers:
@norio-nomura Rx does use a queue/drain strategy for schedulers (to avoid a stack overflow from recursive scheduling), but it doesn't extend to Observable events, as far as I know.

norio-nomura:
@jspahrsummers Thank you for answering my stupid question.
Surely Rx uses scheduler on some operators, not on all Observable events.

PR #2005, opened 2015-05-14
Fixed links to RACSignal and RACSequence

mortonfox:


neilpa:
Thanks for the update :beers:

PR #2008, opened 2015-05-14
Allows CocoaAction to be used in release builds.

bromas:
Changing to SignalProducer.start(next, completed.. etc instead
of SignalProducer.start(sink: s) inside of CocoaActions implementation
resolves a runtime crash on device release builds.


jspahrsummers:

If the issue only affects release builds, could we please also disable Swift optimizations for the time being?

siemensikkema:
:+1: this does indeed fix the problem. :tada:

bromas:
:+1: @jspahrsummers That sounds like a good plan, but I'm not sure I understand how to idiomatically make that change. My understanding is that the build configuration settings are being imported with [xcconfigs](https://github.com/jspahrsummers/xcconfigs) and would require creating a branch on that project with the needed changes and then pointing to that?

jspahrsummers:
@bromas Doing it in this project is fine (since it's a temporary workaround for a Swift issue that hopefully will get fixed), but could you please just move the setting to the project level, instead of per-target?

bromas:
@jspahrsummers I believe my update resolves the project settings issue, is there anything else you think I should look over before this is fit to merge?

jspahrsummers:
Sorry, we don't see notifications about new commits—only new comments left on the thread.

This looks good to go. Thanks for making that change! :sparkles:

PR #2010, opened 2015-05-15
Handle Interrupted events in promoteErrors

jspahrsummers:
I couldn't trace any real issues to this in practice, but This Seems Bad™.


ikesyo:
This Seems Good™. :rocket:

PR #2011, opened 2015-05-15
Allow QueueSchedulers to be named for debugging

jspahrsummers:
Also made the initializer arguments explicit, now that we have a couple of them.


NachoSoto:
:ship:

PR #2029, opened 2015-05-22
Fix order of Completed event in .Concat

ikesyo:
Outer signal observer should be notified of Completed event after all inner producers have completed.

It should fix https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2009#discussion_r30481218.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1141,9 +1141,10 @@ private func concat<T, E>(signal: Signal<SignalProducer<T, E>, E>) -> Signal<T, 
 		}, completed: {
 			// Add one last producer to the queue, whose sole job is to
 			// "turn out the lights" by completing `observer`.
-			let completion: SignalProducer<T, E> = .empty |> on(completed: {
+			let completion = SignalProducer<T, E> { innerObserver, _ in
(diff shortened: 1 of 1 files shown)


kastiglione:
:clap: :clap: :clap:

NachoSoto:
Whoops, nice find! 

Do you think you can write a unit test that covers this? :Star:

kastiglione:
This reminds me though of how `on` is actually `will` not `did`, relative to the observer. It's very subtle and easy to miss.

ikesyo:
> Do you think you can write a unit test that covers this?

It seems very difficult or impossible for me to write a test for the behavior at now... (and I'm on a week's leave starting tomorrow) :disappointed_relieved:

jspahrsummers:
Awesome! I'll try to write a test case for this tomorrow, if nobody else gets to it first.

jspahrsummers:
I'm having trouble understanding how this fixes the issue. Is it because completing the outer observer results in `ConcatState` being deallocated, thus breaking any inflight method calls upon it?

jspahrsummers:
Well, regardless, this _does_ fix the crash. I just don't understand how, so I'm unable to write a test.

If anyone else wants to contribute one, please do! :pray::heart:

PR #2016, opened 2015-05-17
Documentation: Framework Overview

iv-mexx:
This is an attempt to update the __Framework Overview__ for RAC 3.0 #1485
Its only a start, but I wanted to get some feedback before continuing.

First off: I'm still new to RAC (i.e. to RAC 3.0). I have used RAC 2.x for a while now, but only for very simple stuff, mostly for KVO / to avoid using plain KVO. Now I'm trying to get more into RAC and thought I could maybe help a little bit while learning. I expect to get criticism on things I might have gotten wrong, but if it is too much work for you to correct my mistakes and you would rather have someone update the documentation who knows more about RAC - no problem, just let me know :stars: 

### Swift / OjbC or _old_ vs _new_

The most important question is how to deal with the Swift/ObjC split (as mentioned in #1483 ).

The [Changelog](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-development/CHANGELOG.md#using-startwithsignal-instead-of-multicasting) states that the ObjC API has almost not been touched, all RAC 3.0 changes are only additiv and 99% of RAC 2.x code will still work with RAC 3.0 without changes. 

However, also from the contents of the Changelog, I was under the impression that the updated documentation should concentrate on the new concepts of RAC 3.0. Anyway, I think the concepts should not be merged together like `Signal Producer (only Swift API): ...` because that would be confusing. Even more, because you could still use the complete _old_ RAC 2.x API in pure Swift without the new concepts (I think). 

Wo what would be the best approach to solve this? Split the Documentation into _old_ and _new_: keep the existing documentation as is since its still valid plus add an updated version of the documentation that focuses only on the new approach. Call it _old_ / _new_, or call it _ObjC_ / _Swift_?

### Work done so far

Anyway, what I've done so far was under the assumption that the new docs should focus on the _new_/_Swift_ concepts of 3.0:

* Remove the section about `Streams` (because its now covered by `Signal` completely) (?)
* Update `Signal` section:
  * briefly mention the concept of streams 
  * explain the concept of __hot__ signals (no side-effects, events sent regardless of subscribers) without mentioning __hot__ (_or should it be called hot here?_)
  * add explanation of `interrupted` event
* Add `SignalProducer` section:
  * mostly taken from the [header documentation](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-development/ReactiveCocoa/Swift/SignalProducer.swift) but less specific (_no mention of types - should that be  mentioned here?_)
[...]

PR #2037, opened 2015-05-22
[WIP] Swift documentation updates

neilpa:
Starts to tackle some of the other documentation updates. Moves all of the obj-c docs to `Documentation/Legacy` (_I think_ I found all the links to update).

Except for the framework overview (#2016), I've only stubbed out the new Swift pages. I'll probably be updating these sporadically so others should feel free to push to this branch too.


jspahrsummers:
Awesome!

If it's okay with you, could we merge this and then continue the work in separate PRs? That way we won't have the rename cluttering up this diff, and it'll be easier to parallelize the other changes.

neilpa:
Works for me, but we may want to wait for #2016 to go in first. Although, I've left a "hole" for it here so I guess this could go in first.

jspahrsummers:
Either way is fine. I'm happy to manually merge #2016 in to resolve the conflict.

neilpa:
Then lets push this and we can start PRs for each individual page.

PR #2038, opened 2015-05-23
Unified events signal on Action

### LINKED ISSUE #2014: [RAC 3.0] Action.completed Signal
`Action` has the signals `values` and `errors`, sending all values/errors generated from applications of the Action.

Why is there no `completed` signal that sends an event every time one of the `SignalProducers` returned from apply() completes?

At first I thought this might be due to the lifetime of of the signals, but then I understood that the situation for the `errors` signal is similar and (i guess) for that reason does not send the errors as `Error` with `sendError` but as values with `sendNext`. 

### My potential application

My potential application for the `completed` signal would be the following simple Login View:
The viewModel has an `Action` that looks like that

```Swift
loginAction = Action { _ in
    return SignalProducer { (sink, _) in
        API.loginWitUsername(self.username.value, andPassword: self.password.value, withSuccess: {
                sink.put(Event.Completed)
            }, andFailure: { error in
                sink.put(Event.Error(Box(error)))
        })
    }
}
```

#### Working manual solution with `apply`

At first, I was starting the `loginAction` manually from the `IBAction` function:

```Swift
@IBAction func login(sender: UIButton) {
    viewModel.loginAction.apply(sender).start( next: { event in
        println("Next: \(event)")
    }, error: {error in
        println("Error: \(error)")
    }, completed: {
        println("Completed")
    })
}
```

Here I had the chance to subscribe to all Events.

#### With `CocoaAction`

Then I wanted to try the same thing without starting the action explicitely by using `CocoaAction`: 

```Swift
loginButtonAction = CocoaAction(viewModel.loginAction)
loginButton.addTarget(loginButtonAction, action: CocoaAction.selector, forControlEvents:.TouchUpInside)
```

With this approach, there is no direct access to the `SignalProducer` for the `loginAction`, but actually I don't care (I think?) because  its still possible to observe for errors via the `errors` signal of the action:

```Swift
viewModel.loginAction.errors.observe { error in
    println("Error: \(error)")
}
```

But then, there's no `completed` signal for the Action.
I would like to observe for the completion of the login (which is equivalent to a successful login):

```Swift
viewModel.loginAction.completed.observe {
    println("Login Successful")
}
```

Maybe I'm approaching the problem from the wrong angle anyway and theres a better way to achieve this?
I guess it would be possible by sending a `next` event at success before sending `completed`, but later there might be additional information about the login progress (the login might consist of several stages like _authenticate_, _load user data_, ...) that would be sent as `next` Events and sending the success of the login the same way feels a little bit off to me.

neilpa:
Resolves #2014


jspahrsummers:
:+1: on the additions here, but I'm inclined to say we should still leave `values` and `errors` as conveniences, since we don't have operators to go from `Event<X, Y>` to `X` or `Y` easily. What do you think?

neilpa:
Sounds like a good plan to me.

neilpa:
So I also added properties for extracting next and error values from Event. There's been a few other places where I've wanted that.

ReactiveCocoa/Swift/Event.swift
```diff
@@ -63,6 +63,43 @@ public enum Event<T, E: ErrorType> {
...
+		case .Completed:
+			return .Completed
+
+		case .Interrupted:
+			return .Interrupted
+		}
+	}
+
+	/// Unwraps the contained `Next` value.
+	public var nextValue: T? {
```
jspahrsummers: Can we just call this `value`?

ReactiveCocoa/Swift/Event.swift
```diff
@@ -63,6 +63,43 @@ public enum Event<T, E: ErrorType> {
...
+		switch self {
+		case let .Next(next):
+			return next.value
+		default:
+			return nil
+		}
+	}
+
+	/// Unwraps the contained `Error` value.
+	public var errorValue: E? {
```
jspahrsummers: Can we just call this `error`, or is that too confusing?

ReactiveCocoa/Swift/Event.swift
neilpa: Sure, I waffled a couple times on `next` vs `value` vs `nextValue`.

jspahrsummers:
:notebook_with_decorative_cover:

ReactiveCocoa/Swift/Event.swift
neilpa: If we're going with `value` for next I'm fine going with `error` here. I also considered calling this `failure`.

jspahrsummers: I'd prefer `error`, since I think most folks would call it an “error object” or something like that.

ReactiveCocoa/Swift/Event.swift
jspahrsummers: It does have a little potential to cause confusion, but in my mind I think of `.Next` as the only event that “has a value,” in the sense of something that I would do processing upon.

Not a very rigorous definition, I admit.

ReactiveCocoa/Swift/Event.swift
neilpa: :+1:

neilpa:
Renamed

jspahrsummers:
:confetti_ball:

PR #2039, opened 2015-05-23
NSNotificationCenter.rac_notifications returns a producer

### LINKED ISSUE #1980: Signal lifetime of NSNotificationCenter.rac_notifications() can't be controlled by consumers side
[The documentation](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/e38812236b37b2d2b8fda98be1e1ed0cdd0e4f0c/ReactiveCocoa/Swift/FoundationExtensions.swift#L14-L15) of `NSNotificationCenter.rac_notifications()` says:

> This signal will not terminate naturally, so it must be explicitly disposed to avoid leaks.

To dispose the signal, consumers should do it by calling `dispose()` to the disposable returned from the generator closure directly, or by sending a termination event to the observer (sink). But because neither the disposable or the observer is not passed to consumers, it seems that there is no way for them to dispose the returned signal.

Do I understand it correctly?

neilpa:
Resolves #1980

### DIFF
--- a/ReactiveCocoa/Swift/FoundationExtensions.swift
+++ b/ReactiveCocoa/Swift/FoundationExtensions.swift
@@ -10,16 +10,16 @@ import Foundation
 import Result
 
 extension NSNotificationCenter {
-	/// Returns a signal of notifications posted that match the given criteria.
+	/// Returns a producer of notifications posted that match the given criteria.
 	/// This signal will not terminate naturally, so it must be explicitly
 	/// disposed to avoid leaks.
-	public func rac_notifications(name: String? = nil, object: AnyObject? = nil) -> Signal<NSNotification, NoError> {
-		return Signal { observer in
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/Swift/FoundationExtensions.swift
```diff
@@ -10,16 +10,16 @@ import Foundation
 import Result
 
 extension NSNotificationCenter {
-	/// Returns a signal of notifications posted that match the given criteria.
+	/// Returns a producer of notifications posted that match the given criteria.
 	/// This signal will not terminate naturally, so it must be explicitly
```
jspahrsummers: *producer

jspahrsummers:
Just the one nitpick. Looks great! :gem:

neilpa:
Done

jspahrsummers:
:shipit:

PR #2040, opened 2015-05-23
Append disposables operator

### LINKED ISSUE #2001: Pipe operator for composite disposables
This may be crazy talk but I've toyed with `|>`ing disposables from `start` or `observe` into a `CompositeDisposable` and have been liking it. Especially for multiline signal chains where wrapping with `addDisposable` or storing the return value feels like it adds "noise".

```swift
// piped
producer
    |> map { ... }
    |> filter { ... }
    |> start { ... }
    |> disposable

// wrapped
disposable.addDisposable(producer
    |> map { ... }
    |> filter { ... }
    |> start { ... })

// stored
let producerDisposable = producer
    |> map { ... }
    |> filter { ... }
    |> start { ... }
disposable.addDisposable(producerDisposable)
```

Thoughts?

neilpa:
Resolves #2001

### DIFF
--- a/ReactiveCocoa/Swift/Disposable.swift
+++ b/ReactiveCocoa/Swift/Disposable.swift
@@ -220,3 +220,15 @@ public final class SerialDisposable: Disposable {
 		orig.innerDisposable?.dispose()
 	}
 }
+
+/// Adds the right-hand-side disposable to the left-hand-side
+/// `CompositeDisposable`.
+///
+///     disposable += producer
+///         |> filter { ... }
+///         |> map    { ... }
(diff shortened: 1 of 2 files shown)


jspahrsummers:
Awesome! :sparkles:

Is this operator allowed to return a value? If so, could it return the `DisposableHandle`?

neilpa:
I'll try, but I'm pretty sure assignment operators are statements and not expressions.

neilpa:
It actually works, I was not expecting that.

jspahrsummers:
:tada:

PR #2041, opened 2015-05-23
Refactor switchToLatest to fix deadlock

### LINKED ISSUE #2035: RAC3: Deadlock in `switchToLatest`
When interrupting a signal where `switchToLatest` is used, the atomic `LatestState` can be modified while the lock is already held. I've narrowed it down to a simple test case:

```swift
SignalProducer<Int, NoError>(value: 1)
	|> flatMap(.Latest) { _ in SignalProducer(value: 10) }
	|> take(1)
	|> wait
```

The `LatestState` lock is held when sending the `Next` event, but that same event causes the signal to be disposed by `take(1)`. The `LatestState` is then modified when processing the `Interrupted` event from the disposal, which causes the deadlock.

I'm not 100% sure, but maybe `sink.put(event)` can be done outside of the lock after validating that `isLatestIncompleteSignal(innerSignal)` succeeds?

jspahrsummers:
Fixes #2035.

I didn’t originally intend to rewrite the operator, but I was worried that the naive fix for the deadlock (13fe954) could introduce a small race condition window where producer `N - 1` could forward an event after producer `N` has started.

/cc @brow @neilpa @stevebrambilla

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -1044,6 +1044,15 @@ class SignalProducerSpec: QuickSpec {
 
 					expect(completed).to(beTruthy())
 				}
+
+				it("should not deadlock") {
+					let result = SignalProducer<Int, NoError>(value: 1)
+						|> flatMap(.Latest) { _ in SignalProducer(value: 10) }
+						|> take(1)
+						|> last
+
+					expect(result?.value).to(equal(10))
+				}
 			}
 
 			describe("interruption") {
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1044,6 +1044,15 @@ class SignalProducerSpec: QuickSpec {
 
 					expect(completed).to(beTruthy())
 				}
+
+				it("should not deadlock") {
```
neilpa: Is this diagnose-able if it fails in the future? E.g. will Xcode timeout and kill the test with some meaningful error?

jspahrsummers: AFAIK it won't time out (though a CI service might), but you'd see in the console log the test that it's waiting indefinitely on. Pausing in the debugger would reflect a deadlock too.

neilpa: That's what I figured, but should be fine.

neilpa:
Nice, I like this approach. Now I think the only remaining operator sending under lock is `combineLatestWith` (but that ones not fixable if we want to guarantee ordering)

PR #2042, opened 2015-05-23
Disallow recursive signals

jspahrsummers:
We’ve [discussed](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1120) [queue](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1308)/[drain](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2004) many times already, and have always concluded that it’s too confusing for too little benefit. I still agree.

However, recursive signals are _also_ very confusing. https://github.com/Carthage/Carthage/pull/474, in particular, is a living testament to this. If a core RAC contributor (:+1: this guy :+1:) can’t disentangle that recursive clusterfuck, how can we expect our users to?

So, why don’t we just explicitly disallow recursive signals? No handling, period. Easy to detect, and easier to understand than before—`NSLock` even logs a helpful message to console when a deadlock occurs!

This change may frustrate at times, but I’d rather be frustrated by something deterministic than a nondeterministic exploding backtrace.


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -190,26 +205,38 @@ public struct SignalProducer<T, E: ErrorType> {
 		// Create a composite disposable that will automatically be torn
 		// down when the signal terminates.
 		let compositeDisposable = CompositeDisposable()
-		compositeDisposable.addDisposable {
-			// Upon early disposable, interrupt the observer and all dependents.
-			sendInterrupted(observer)
```
jspahrsummers: This particular method is an excellent illustration of how confusing recursive sending can be.

Before this PR, disposal could send you an `.Interrupted` event _while you’re processing another_. I seriously doubt any extant RAC code handles this case except by accident.

With the changes here, disposal waits until you’re done with your event processing.

jspahrsummers:
Extracted from https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2042#discussion_r30942872, which got collapsed:

> This particular method implementation is an excellent illustration of how confusing signal recursion can be.
>
> Before this PR, disposal could send you an `Interrupted` event _while you’re processing another_. I seriously doubt any extant RAC code handles this case except by accident.
>
> With the changes here, disposal waits until you’re done with your event processing.

neilpa:
Still need to digest the changes, but I'm :100: for this. Reentrancy sucks and is an utter pain to diagnose and debug.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -27,31 +35,30 @@ public final class Signal<T, E: ErrorType> {
...
-					// recursively).
-					self.observers = nil
-				}
-
-				for sink in observers {
-					sink.put(event)
+			switch event {
+			case .Interrupted:
+				// Normally we disallow recursive events, but
+				// Interrupted is kind of a special snowflake, since it
```
neilpa: :snowflake:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -17,8 +17,16 @@ import Result
...
+	private let atomicObservers: Atomic<Bag<Observer>?> = Atomic(Bag())
+
+	/// Used to ensure that events are serialized during delivery to observers.
+	private let sendLock = NSLock()
+
+	/// Disposes of the generator provided at initialization time.
+	private let generatorDisposable = SerialDisposable()
+
+	/// When disposed of, the Signal should interrupt as soon as possible.
+	private let interruptionDisposable = SimpleDisposable()
```
neilpa: Why a disposable instead of just a boolean for this?

neilpa: Ahh, it's hiding an `Atomic<Bool>`. Might be better to just do that directly since you're not putting this in a `CompositeDisposable`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -86,15 +133,18 @@ public final class Signal<T, E: ErrorType> {
 	public func observe<S: SinkType where S.Element == Event<T, E>>(observer: S) -> Disposable? {
 		let sink = Observer(observer)
 
-		lock.lock()
-		let token = self.observers?.insert(sink)
-		lock.unlock()
+		var token: RemovalToken?
+		atomicObservers.modify { (var observers) in
```
neilpa: Didn't there used to be a version of `modify` that allowed you to pass-thru secondary data. Is it worth bringing that back?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -107,42 +107,57 @@ public struct SignalProducer<T, E: ErrorType> {
...
 		var events: [Event<T, E>] = []
 		var terminationEvent: Event<T,E>?
-		var observers: Bag<Signal<T, E>.Observer>? = Bag()
+		let observers: Atomic<Bag<Signal<T, E>.Observer>?> = Atomic(Bag())
 
 		let producer = self { observer, disposable in
 			lock.lock()
+
+			var token: RemovalToken?
+			observers.modify { (var observers) in
```
neilpa: Doing this under `lock` makes me a bit nervous about the potential for a deadlock. However, I've yet to find a path through the code where that could happen.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -107,42 +107,57 @@ public struct SignalProducer<T, E: ErrorType> {
...
-					lock.unlock()
+					observers.modify { (var observers) in
+						observers?.removeValueForToken(token)
+						return observers
+					}
 				}
 			}
 		}
 
 		let observer = Signal<T, E>.Observer { event in
```
neilpa: Can you rename this sink. It'll help distinguish it even further from the observer in the above closure.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -190,26 +205,28 @@ public struct SignalProducer<T, E: ErrorType> {
 		// Create a composite disposable that will automatically be torn
```
neilpa: This is another case where I think calling the above observer either sink or pipe would improve clarity in this function.

neilpa:
Mostly minor comments but overall I'm happy with this change. It makes control flow easier to trace through in my head too.

robrix:
> This change may frustrate at times, but I’d rather be frustrated by something deterministic than a nondeterministic exploding backtrace.

:+1:

Recursive locks are a code smell, IMO; I’m very much in favour of this.

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: I used a `SimpleDisposable` because we only ever flip this in one direction (never back to `false`), and because I was imagining a world where we might micro-optimize that type.

I agree it's confusing, though, so I can change it.

ReactiveCocoa/Swift/Signal.swift
jspahrsummers: We removed it because of type inferencing errors, and because most readers understand saving a `var` outside of the block better than the complicated tuple usage we had.

ReactiveCocoa/Swift/SignalProducer.swift
jspahrsummers: I'm treating `observers` like an atomic property, where no real work should be done while modifying it, so that should avoid deadlocks.

ReactiveCocoa/Swift/SignalProducer.swift
jspahrsummers: Addressed in 2b15985.

jspahrsummers:
Sorry for introducing so many unrelated changes here. I was working on this in conjunction with https://github.com/Carthage/Carthage/pull/474, which necessitated:

 1. Temporarily reverting https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1966, as discussed with @neilpa in Slack. Unfortunately, the signal-of-producer versions of these operators prevent disposal from propagating correctly, even when lifted and applied directly to producers! We couldn't figure out the issue, so we can return to it in another PR.
 1. No longer materializing `Interrupted` events, since that can prolong a stream that's meant to be cancelled.
 1. Optimizing `Bag` so enumeration of it doesn't dominate the backtrace.
 1. Enabling Swift optimizations that can reduce the incidence of stack overflows from recursion (by inlining and eliminating some calls).

This should be ready for review and merge now.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -367,17 +404,23 @@ public func skip<T, E>(count: Int) -> Signal<T, E> -> Signal<T, E> {
...
-/// manipulated just like any other value.
+/// Treats Events (except for Interrupted) from the input signal as plain
+/// values, allowing them to be manipulated just like any other value.
 ///
-/// In other words, this brings Events “into the monad.”
+/// In other words, this brings the Next, Completed, and Error Events “into the monad.”
 public func materialize<T, E>(signal: Signal<T, E>) -> Signal<Event<T, E>, NoError> {
 	return Signal { observer in
 		return signal.observe(Signal.Observer { event in
-			sendNext(observer, event)
```
neilpa: I think I liked the previous variant where we lifted the `Interrupted` event but still interrupted the resulting signal.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -813,3 +832,335 @@ public func start<T, E>(error: (E -> ())? = nil, completed: (() -> ())? = nil, i
 		return producer.start(next: next, error: error, completed: completed, interrupted: interrupted)
 	}
 }
+
+/// Describes how multiple signals or producers should be joined together.
```
neilpa: Just producers now

neilpa: And more below

neilpa:
:+1: Just a couple comments

jspahrsummers:
:boom:

neilpa:
I'm going to merge this now since it's needed for Carthage/Carthage#474 and there's at least some consensus on the behavior changes.

ReDetection:
@jspahrsummers but what about signal-of-signals? because we don't have the flatten in RAC3, so we don't have any way to merge signals. interesting, that we still have combine and zip.

PR #2046, opened 2015-05-24
Disposal is immediate even when using observeOn

### LINKED ISSUE #2045: Verify that disposal is immediate even when using observeOn
Looking at the code for `observeOn`, I became concerned that disposing of a `SignalProducer` (built using the operator) may not result in the immediate cancellation of upstream work, because of the scheduler hop.

Now I'm less sure that the implementation is wrong, because this concern might get handled properly by `SignalProducer.lift`—but we should write a test case that verifies the behavior regardless.

335g:
Fix #2045 

I have a question.
`ActionDisposable` (returned by `observe`) isn't disposed even upstream signal is disposed.
(= first unit test doesn't pass.)

Is this expected behavior?
I think that `ActionDisposable` should be disposed as soon as observed signal is disposed. But I don't have an idea to solve. :cry:


jspahrsummers:
That is expected behavior. You cannot observe disposables returned _from_ signals or producers to see whether disposal has occurred.

Instead, you'll need to ensure that the `Disposable` returned in `Signal.init`—or the `CompoundDisposable` passed into `SignalProducer.init`—is disposed at the expected time.

jspahrsummers:
Also, can you please test `observeOn` before modifying its implementation? As mentioned in #2045, I believe that it may already have the correct behavior—we just need a test to verify that.

335g:
Thank you for teaching.:bow:

I'm sorry. I had misunderstood.
I'll try that, but there are things that I don't think of immediately (ex. hot to dispose `SignalProducer` (built using the operator))

335g:
This is ready for review.

Does it become the intended test ?
Check please :bow:

jspahrsummers:
Looks good, thanks!

PR #2049, opened 2015-05-26
Remove manual `==` for FlattenStrategy enum

alskipp:
enums without associated values are `Equatable` by default, therefore the `==` func is not required.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -855,16 +855,6 @@ public enum FlattenStrategy: Equatable {
 	case Latest
 }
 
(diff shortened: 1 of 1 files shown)


alskipp:
Further info from Apple can be found [here](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/CollectionTypes.html)

The relevant sentence is: “Enumeration member values without associated values (as described in Enumerations) are also hashable by default.” Hashable implies Equatable by default too.

jspahrsummers:
Yay, unpredictable language rules. :smirk: 

But seriously, thank you for the pull request! :sparkles:

alskipp:
With any luck, a proper ‘deriving’ mechanism for various protocols will be revealed in June 🙏

PR #2051, opened 2015-05-26
Allow recursive gets of `MutableProperty.value` during changes

### LINKED ISSUE #2050: Deadlock using a MutableProperty in a view controller
Due to the change in #2042, I'm hitting a deadlock when using a `MutableProperty` in a view controller.   My setup:

1) In my view controller, I have a `var data = MutableProperty<[MyViewModel]?>(nil)`
2) In `viewDidLoad`, I bind the values from an `Action` to the property with `data <~ viewModel.myAction.values`
3) Also in `viewDidLoad`, right after step 2, I use the mutable property's producer to reload the collection view that is reading my `data` property: `data.producer.start(next: { _ in self.collectionView.reloadData() })`
4) And finally in `viewDidLoad`, right step 3, I invoke the `Action` on my view model: `viewModel.myAction.apply(NSDate())`

For ease of reading, here are steps 2-4:

```
    data <~ viewModel.myAction.values
    data.producer.start(next: { _ in
        self.collectionView.reloadData()
    })
    viewModel.myAction.apply(NSDate())
```

The problem is that in the `UICollectionViewDataSource` methods, I read from the `data` property using the `value` property.  Because I'm inside of the event delivery block higher up the chain, I'm getting the deadlock now that locks are not recursive.  The first workaround I can think of is to not use the `MutableProperty` and instead copy the values as they arrive from the action into variables and use those in the data source methods instead.  I guess that works but really feels stateful.  I'm curious if there any other suggestions for how to structure loading of data from a view model inside of a view controller acting as a data source?

neilpa:
Resolves #2050

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -63,18 +63,29 @@ public final class MutableProperty<T>: MutablePropertyType {
 
 	private let observer: Signal<T, NoError>.Observer
 
+	/// Need a recursive lock around `value` to allow recursive access to
+	/// `value`. Note that recursive sets will still deadlock because the
+	/// underlying producer prevents sending recursive events.
+	private let lock = NSRecursiveLock()
+	private var _value: T
+
 	/// The current value of the property.
 	///
 	/// Setting this to a new value will notify all observers of any Signals
 	/// created from the `values` producer.
 	public var value: T {
 		get {
-			let result = producer |> first
-			return result!.value!
+			lock.lock()
+			let value = _value
+			lock.unlock()
+			return value
 		}
 
-		set(x) {
-			sendNext(observer, x)
+		set {
+			lock.lock()
+			_value = newValue
+			sendNext(observer, newValue)
+			lock.unlock()
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -63,18 +63,29 @@ public final class MutableProperty<T>: MutablePropertyType {
 
 	private let observer: Signal<T, NoError>.Observer
 
+	/// Need a recursive lock around `value` to allow recursive access to
+	/// `value`. Note that recursive sets will still deadlock because the
+	/// underlying producer prevents sending recursive events.
+	private let lock = NSRecursiveLock()
```
jspahrsummers: Can you set `lock.name` in the initializer so deadlocks describe the source more clearly?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -73,6 +73,18 @@ class PropertySpec: QuickSpec {
...
 				expect(signalCompleted).to(beTruthy())
 			}
+
+			it("should not deadlock on recursive value access") {
+				let (producer, sink) = SignalProducer<Int, NoError>.buffer()
+				var property = MutableProperty(0)
+
+				property <~ producer
+				property.producer.start(next: { _ in
+					let test = property.value
```
jspahrsummers: Can you add a flag that ensures this gets called when we expect? (It'd be easy for this test to pass if the block were just never invoked somehow.)

jspahrsummers:
:+1::sparkles:

neilpa:
:bomb:

jspahrsummers:
:shipit:

PR #2055, opened 2015-05-29
Increase <~ operator precedence

neilpa:
Enables assignment of the returned disposable to an optional var without parenthesizing the entire binding expression.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -162,8 +162,8 @@ extension MutableProperty: SinkType {
 infix operator <~ {
 	associativity right
 
-	// Binds tighter than assignment but looser than everything else.
-	precedence 91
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -161,7 +161,9 @@ extension MutableProperty: SinkType {
 
 infix operator <~ {
 	associativity right
-	precedence 90
+
+	// Binds tighter than assignment but looser than everything else.
+	precedence 91
```
neilpa: I could bump this to 95 to leave some space between  but wasn't sure.

jspahrsummers: Maybe 92 or 93? Changing precedence later is technically a breaking change.

Also, can you document that this is (intentionally) less than `|>`?

jspahrsummers:
:+1:

neilpa:
:boom:

PR #2058, opened 2015-06-01
Fix dropping signal bug on multithreading in SignalProducer.buffer()

norio-nomura:
This fix the Carthage hanging I reported on following:
https://github.com/Carthage/Carthage/pull/474#issuecomment-107307773

See also the description of https://github.com/ReactiveCocoa/ReactiveCocoa/commit/5b41bb51320640fc7be95b81ad89283d831e6582

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -180,6 +180,8 @@ public struct SignalProducer<T, E: ErrorType> {
 		}
 
 		let bufferingObserver = Signal<T, E>.Observer { event in
+			lock.lock()
+			
 			let oldObservers = observers.modify { (var observers) in
 				ReactiveCocoa.println("\(ObjectIdentifier(lock).hashValue) bufferingObserver() modify()")
 				if event.isTerminating {
@@ -191,8 +193,6 @@ public struct SignalProducer<T, E: ErrorType> {
 
 			// If not disposed…
 			if let liveObservers = oldObservers {
-				lock.lock()
-
 				ReactiveCocoa.println("\(ObjectIdentifier(lock).hashValue) bufferingObserver() put()")
 				if event.isTerminating {
 					terminationEvent = event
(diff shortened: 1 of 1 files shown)


norio-nomura:
```swift
			if let token = token {
				disposable.addDisposable {
					observers.modify { (var observers) in
						observers?.removeValueForToken(token)
						return observers
					}
				}
			}
```
`addDisposable()`'s action should use `lock` instead of `Atomic<T>.modify()`.
And `observers` can be changed to non `Atomic`.
The codes inside `if let liveObservers = oldObservers {…}` looks strange to me. But I can't determine how it should be…

norio-nomura:
I should write more description of this.

## About dropping signal bug on multithreading in `SignalProducer.buffer()`
`SignalProducer.buffer()` create a pair of `SignalProducer` and `bufferingObserver`.
These share two lock objects
```
		let lock = NSLock()
		let observers: Atomic<Bag<Signal<T, E>.Observer>?> = Atomic(Bag())
```

The issue caused by two threads
- Thread A call `SignalProducer` adding new observer
- Thread B call `bufferingObserver` putting signal to observers

with following scenario:

1. [B] `bufferingObserver` retrieve `observers` from `Atomic<T>`
2. [A] `SignalProducer` acquires `lock`
3. [B] `bufferingObserver` wait `lock`
4. [A] `SignalProducer` add **new observer** to `observers` inside `Atomic<T>`
5. [A] `SignalProducer` relinquishes `lock`
6. [B] `bufferingObserver` acquires lock and put signal to `observers`.

The **new observer** added on 4. will never receive signal when this happen.

For verifying this scenario, I pushed commit as 5b41bb5
For fixing this scenario, I pushed commit as 109da1e
For reverting 5b41bb5, I pushed commit as 3f11cfe

I verified this fix Carthage hanging I reported on following:
https://github.com/Carthage/Carthage/issues/455#issuecomment-106991205
https://github.com/Carthage/Carthage/pull/474#issuecomment-107307773

robrix:
:ok_hand:

norio-nomura:
:bow:

jspahrsummers:
Thanks, @norio-nomura!

PR #2059, opened 2015-06-01
Change void return types to using Void

ikesyo:
See http://ericasadun.com/2015/05/11/swift-vs-void/.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -143,7 +143,7 @@ public final class CocoaAction: NSObject {
 
 	private var _enabled = false
 	private var _executing = false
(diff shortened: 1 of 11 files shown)


neilpa:
Personally I prefer `() -> ()`, but it's probably better to be consistent with apple here.

neilpa:
I want to wait for github/swift-style-guide#39 to shake out before merging.

ikesyo:
Oh, thank you for the opening it! :bow: I'll wait.

jspahrsummers:
Per https://github.com/github/swift-style-guide/issues/39#issuecomment-109443211, I think we'd like to continue using `()`.

Thanks, though!

/cc @robrix

ikesyo:
Yep, thank you all for the discussion! :wink:

PR #2066, opened 2015-06-05
Add overload to |> operator so it kann be used with PropertyType

iv-mexx:
With this change, properties can be manipulated with signal operators without the need to manually retrieve the producer first

```Swift
firstIntProperty <~ secondIntProperty.producer |> map { 2 * $0 }
```

can be simplified to

```Swift
firstIntProperty <~ secondIntProperty |> map { 2 * $0 }
```

I think this makes working with properties a lot more intuitive and removes some noise from the code.

Actually, I would like to be able to do something like

```Swift
enabledProperty <~ combineLatest(condition1Property, condition2Property, condition3Property)
    |> map { $0.0 && $0.1 && $0.2 }
```

But for that to work, suspect all the versions of `combineLatest` would need the corresponding version duplicated, and maybe all other similar functions (`zip`) as well for consistency. 

I wanted to test the waters first with this small PR to get some feedback if this is even wanted and, if so, if theres a better way to do it.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -309,6 +309,16 @@ public func |> <T, E, U, F>(producer: SignalProducer<T, E>, transform: Signal<T,
 	return producer.lift(transform)
 }
 
+/// Applies a Signal operator to a Property.
+///
+/// Example:
+///
+/// 	intProperty
+/// 	|> map { 2 * $0 }
+public func |> <P: PropertyType, X>(property: P, transform: Signal<P.Value, NoError> -> Signal<X, NoError>) -> SignalProducer<X, NoError> {
+	return property.producer.lift(transform)
+}
+
 /// Applies a SignalProducer operator to a SignalProducer.
 ///
 /// Example:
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -322,6 +332,21 @@ public func |> <T, E, X>(producer: SignalProducer<T, E>, @noescape transform: Si
 	return transform(producer)
 }
 
+/// Applies a SignalProducer operator to a Property.
+///
+/// Example:
+///
+/// 	intProperty
+/// 	|> startOn(UIScheduler())
```
iv-mexx: I know this is not really a useful application for an example, but I didn't find an operator that is __only__ available as a Signal Operator AND is more useful than this... :disappointed_relieved:

jspahrsummers:

However, I don't see much advantage in syntactic sugar like this—in general, I'd like for there to be _one_ canonical way to do things, rather than multiple ways that all end up behaving the same.

Since this specific case is just about eliding `.producer`, the potential for additional confusion doesn't seem worth it.

PR #2082, opened 2015-06-08
Add withLatestFrom

endor:
In a ViewModel I have a situation where the value of a `MutableProperty` affects its next value:

```swift
mutablePropertyA.producer
    |> combineLatestWith(mutablePropertyB.producer)
    |> filter({ (a, b) in
        return fulfilsSomeCondition(a, b)
    })
    |> start(next: {result in
        mutablePropertyB.put(result)
    })
```

Other RAC-like frameworks offer a `withLatestFrom` operator for these cases. Could this be added to RAC as well?

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -347,6 +347,38 @@ internal func combineLatestWith<T, E>(otherSignal: Signal<T, E>) -> Signal<[T], 
 	}
 }
 
+/// Combines the latest value of the receiver with the latest value from
+/// the given signal.
+///
+/// The returned signal will not send a value when the otherSignal changes.
+public func withLatestFrom<T, U, E>(otherSignal: Signal<U, E>) -> Signal<T, E> -> Signal<(T, U), E> {
+	return { signal in
+		return Signal { observer in
+			let lock = NSLock()
+			lock.name = "org.reactivecocoa.ReactiveCocoa.withLatestFrom"
+
+			let signalState = CombineLatestState<T>()
+			let otherState = CombineLatestState<U>()
+
+			let onNext = { () -> () in
+				sendNext(observer, (signalState.latestValue!, otherState.latestValue!))
+			}
+
+			let onError = { sendError(observer, $0) }
+			let onCompleted = { sendCompleted(observer) }
+			let onInterrupted = { sendInterrupted(observer) }
+
+			let signalDisposable = observeWithStates(signal, signalState, otherState, lock, onNext, onError, onCompleted, onInterrupted)
+			let signalCompletedDisposable = signal.observe(completed: onCompleted)
+			let otherDisposable = otherSignal.observe(next: { value in
+				otherState.latestValue = value
+			})
+
+			return CompositeDisposable(ignoreNil([ signalDisposable, signalCompletedDisposable, otherDisposable ]))
+		}
+	}
+}
+
 /// Delays `Next` and `Completed` events by the given interval, forwarding
 /// them on the given scheduler.
 ///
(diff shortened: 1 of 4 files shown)


NachoSoto:
Wouldn't that be equivalent to

```swift
SignalProducer(value: [mutablePropertyA.producer, mutablePropertyB.producer])
|> flatMap(.Latest)
```

endor:
@NachoSoto No it wouldn't, because `combineLatest` sends a `next` whenever one of the two changes. `withLatestFrom` only sends a `next` when the first one changes. This has the advantage of not causing a deadlock when changing the second property.

This probably explains it as well: http://rxmarbles.com/#withLatestFrom.

jspahrsummers:
Why don't you just observe `firstProperty.producer`, then read from `secondProperty.value` in a `map`?

I agree that this doesn't seem like something we should add.

endor:
@jspahrsummers the idea was to have all the signals in the chain instead of accessing a signal outside of the chain. Just thought it would make sense to add it as other people seem to want to use it as well.

jspahrsummers:
You can also use `sampleOn` for very similar behavior.

phadej:
@jspahrsummers not really. Please correct me, if I'm wrong.

```swift
// combineLatest
// a: 2---3---1-
// b: --1---6---
// r: --3-4-9-7
signalA.producer
  |> combineLatestWith(signalB.producer)
  |> map ({ (a, b) in
    return a + b;
  })

// withLatestFrom
// a: 2---3---1-
// b: --1---6---
// r: ----4---7
signalA.producer
  |> withLatestFrom(signalB.producer)
  |> map ({ (a, b) in
    return a + b;
  })

// sampleOn, 1. try
signalB.producer
  |> sampleOn(singlaA.producer)
  |> map ({ (b) in
    // only b
  })

// this might do the trick, but...
// a:       2---3---1-
// b:       --1---6---
// sampled  ----1---6
// r:       ----44--77 // will trigger twice!
signalB.producer
  |> sampleOn(signalA.producer)
  |> combineLatestWith(signalA.producer)
  |> map ({ (b, a)
    return a + b
  })
```

jspahrsummers:
You can use `startWithSignal` to avoid starting the producers multiple times.

In general, though, this operator’s behavior can be accomplished in enough other ways that I don’t think we need to add it.

phadej:
Can you give an example how to use `startWithSignal` to achieve
```swift
let resultSignal = signalA.producer
  |> withLatestFrom(signalB.producer)
  |> map ({ (a, b) in
    return a + b;
  })
```
functionality?

jspahrsummers:
Something like this:

```swift
producerB.startWithSignal { signalB, disposableB in
	producerA.startWithSignal { signalA, disposableA in
		signalB
			|> combineLatestWith(signalA)
			|> sampleOn(signalA)
			|> map { $0 + $1 }
			|> observe(next: { sum in
				println(sum)
			})
	}
}
```

staltz:
I helped introduce withLatestFrom into RxJS because this came out of necessity, it was very common to need to combine signals in this manner. It can be understood as `mapTheSignalButPleaseUseTheLatestEventFrom(signalB)`.

And I've seen multiple use cases for it in StackOverflow questions: [[1]](http://stackoverflow.com/questions/31020449/put-in-remove-from-the-basket-using-rx) [[2]](http://stackoverflow.com/questions/27203435/combinelatest-emit-only-when-one-of-the-streams-changes/27207073#27207073) [[3]](http://stackoverflow.com/questions/28153548/rxjs-equivalent-of-bacon-when-with-properties-which-are-sampled-but-not-par/28157642#28157642)

Even though it can be accomplished with some other operators like @jspahrsummers just showed [4], the alias `withLatestFrom` makes code more readable and more productive.

PS [4] In fact in Scala with RxJava it's also easy to accomplish with other operators:
```scala
withLatestFrom[A,B](fast : Observable[A], slow : Observable[B]): Observable[(A,B)] = {
  val hotSlow = slow.publish.refCount
  fast.map({a => hotSlow.map({b => (a,b)})}).switch
}
```

PR #2090, opened 2015-06-09
Missed configuration (crash: dyld: Library not loaded: @rpath/libswiftCore.dylib).

ostapkuzyk:
Missed configuration (crash: dyld: Library not loaded: @rpath/libswiftCore.dylib).


NachoSoto:
Thanks for the pull request @ostapkuzyk!

It might seem that, if a target uses Swift, is should be `true`, right?
However, what that means in practice is that the framework will include the Swift standard library, and application targets that link against this framework will fail when being submitted because it includes an unexpected folder.

I recently filed [a radar](https://openradar.appspot.com/20377499) to complain about this setting.

For this reason, unintuitively, this setting must only be `true` in the outer most target (the application), to make Xcode only include the Swift standard library there.

One day we'll be writing code in an IDE that doesn't expose all these stupid implementation details 😥

PR #2088, opened 2015-06-08
[WIP] Swift 2 updates

neilpa:
This PR currently targets **Xcode 7 Beta 3**

Of significant note is that `|>` is gone. Signal operators are now implemented as protocol extensions so you can use normal dot-method syntax.

PR #2099, opened 2015-06-11
Swift 2 - Use generic extensions for operators

neilpa:
This is about half done but figured people might be curious about this. It converts all the signal operators except `dematerialize` because I couldn't figure out an approach that didn't crash the compiler. I had similar issues with `ignoreNil` but switched that to `filterMap` for now since it can provide the same behavior. The free-functions haven't been removed yet because producers still lift them.

Additionally, I switched observers to straight functions instead of `SinkType` and migrated to `Swift.ErrorType`. The latter was an attempt at mitigating the aforementioned crashes.

Depends on #2088


ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -124,7 +130,7 @@ public func toRACSignal<T: AnyObject, E>(signal: Signal<T?, E>) -> RACSignal {
 		let selfDisposable = signal.observe(next: { value in
 			subscriber.sendNext(value)
 		}, error: { error in
-			subscriber.sendError(error.nsError)
+			subscriber.sendError(error as NSError)
```
NachoSoto: Neat

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -204,25 +219,47 @@ public func filter<T, E>(predicate: T -> Bool) -> Signal<T, E> -> Signal<T, E> {
...
+					sendInterrupted(observer)
+				}
+			}
+		}
+	}
 }
 
 /// Returns a signal that will yield the first `count` values from the
 /// input signal.
 public func take<T, E>(count: Int) -> Signal<T, E> -> Signal<T, E> {
```
NachoSoto: Silly question: how come the free functions are still needed?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -449,7 +513,30 @@ public func dematerialize<T, E>(signal: Signal<Event<T, E>, NoError>) -> Signal<
...
+	public func dematerialize() -> Signal<T.T, T.E> {
+		return Signal { _ in nil }
+		return Signal { (observer: Event<T.T, T.E>.Sink) in
+			return self.signal.observe { event in
+				switch event {
+				case let .Next(innerEvent):
+					observer(innerEvent.event)
+
+				case .Error:
+					fatalError()
```
NachoSoto: I know this is existing code, but maybe do `fatalError("This can't happen")` to make it clearer?

NachoSoto:
Exciting!

MarcoSero:
Looking forward to this! I hope it will ease the learning curve and remove part of the magic behind `|>`.

ReactiveCocoa/Swift/Signal.swift
eliperkins: Looks like Swift 2 removes their free functions for things like `map` from the standard lib as well. Might be worth following suit.

neilpa: Those are needed for `SignalProducer` operator lifting since I haven't updated those yet to use protocol extensions. So far I've only done `Signal`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -211,20 +215,29 @@ public func filter<T, E>(predicate: T -> Bool) -> Signal<T, E> -> Signal<T, E> {
 			}
 		}
 	}
-}
 
-/// Unwraps non-`nil` values from `signal` and forwards them on the returned
-/// signal, `nil` values are dropped.
-public func ignoreNil<T, E>(signal: Signal<T?, E>) -> Signal<T, E> {
```
neilpa: This is one of those instances where attempting to constrain on an `OptionalType` would crash the compiler which is why I switched `filterMap`.

neilpa:
This is ostensibly ready for review modulo some compiler crashes preventing more complex protocol extension constraints. I tried to find workarounds (`ignoreNil` -> `filterMap`) but was mostly unsuccessful in finding alternatives that didn't segfault the compiler (`flatten`, `dematerialize`, better binary `lift`).

cc @robrix in case you have any ideas since I saw you opened radars for similar crashes.

neilpa:
c.f. antitypical/Result#56 and antitypical/Result#60

rpowelll:
Do we put this one on hold awaiting compiler improvements then?

neilpa:
Doesn't matter really since this targets the `swift2` branch which isn't going to be merged until Xcode 7 moves out of beta. I just wanted to separate this from all the other changes given it's scope.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -140,16 +140,25 @@ public final class Signal<T, E: ErrorType> {
...
-	/// Returns a Disposable which can be used to stop the invocation of the
-	/// callbacks. Disposing of the Disposable will have no effect on the Signal
-	/// itself.
-	public func observe(next next: (T -> ())? = nil, error: (E -> ())? = nil, completed: (() -> ())? = nil, interrupted: (() -> ())? = nil) -> Disposable? {
-		return observe(Event.sink(next: next, error: error, completed: completed, interrupted: interrupted))
+public protocol SignalType {
+	/// The type of values being sent on the signal.
+	typealias T
+	/// The type of error that can occur on the signal. If errors aren't possible
+	/// than `NoError` can be used.
```
jspahrsummers: *then

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -140,16 +140,25 @@ public final class Signal<T, E: ErrorType> {
...
+	typealias E: ErrorType
+
+	/// Proxy for the actual signal.
```
jspahrsummers: Can we maybe say something like, “extracts a signal from the receiver?”

I could see this protocol being implemented by stuff like `PropertyType`, for example, in which case “actual signal” might not be quite accurate.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -298,22 +321,217 @@ public struct SignalProducer<T, E: ErrorType> {
 			}
 		}
 	}
+
+	/// Maps each value in the producer to a new value.
+	public func map<U>(transform: T -> U) -> SignalProducer<U, E> {
+		return lift { $0.map(transform) }
```
jspahrsummers: Lack of currying makes me sad. :cry:

jspahrsummers:
I thought I saw a Twitter conversation (don’t remember from whom) implying that protocol extensions can't really use full-fledged constraints on the `typealias`es of the protocol. If true, this would explain issues with `dematerialize` and `flatten`.

If we can't work around it, though, I'm not sure if this is a net improvement. Hopefully the next beta will bring some fixes. :pray:

neilpa:
> If we can't work around it, though, I'm not sure if this is a net improvement. Hopefully the next beta will bring some fixes. :pray:

That's my hope as well.

neilpa:
Sadly, the protocol extensions with more complex type constraints are still crashing the compiler in beta 2.

neilpa:
However, it looks like we can just get rid of `SignalType` and write constrained extensions directly on `Signal`.

neilpa:
This is ready for another review. It uses the new constrained generic extensions in beta 2 which plays nicer with the compiler than some of the protocol extensions I was attempting. In particular, `flatten`, `dematerialize`, and `ignoreNil` are all implemented as extensions now. This does still a number of protocols though for creating the constraints.

neilpa:
Most importantly, I was able to finally delete `|>`

joshaber:
>Most importantly, I was able to finally delete `|>`

JaviSoto:
This is amazing :heart:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -404,52 +404,56 @@ public func skip<T, E>(count: Int) -> Signal<T, E> -> Signal<T, E> {
...
+		return Signal<T.T, T.E> { observer in
+			return self.observe { event in
+				switch event {
+				case let .Next(innerEvent):
+					observer(innerEvent.event)
 
-			case .Error:
-				fatalError()
+				case .Error:
+					fatalError("NoErrorType should be impossible to construct")
```
iv-mexx: "NoErrorType __is__ impossible to construct" ?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -908,218 +908,234 @@ private struct ThrottleState<T> {
...
+	public func promoteErrors<F: ErrorType>(_: F.Type) -> Signal<T, F> {
+		return Signal<T, F> { observer in
+			return self.observe(next: { value in
 				sendNext(observer, value)
 			}, completed: {
 				sendCompleted(observer)
 			}, interrupted: {
 				sendInterrupted(observer)
+			}, error: { _ in
+				fatalError("NoErrorType should be impossible to construct")
```
iv-mexx: "NoErrorType __is__ impossible to construct" ?

ReactiveCocoa/Swift/Signal.swift
neilpa: There's no way to guarantee it via the protocol - 5850095

ReactiveCocoa/Swift/Signal.swift
neilpa: 5850095

schickling:
Great work! The `|>` operator was kinda unintuitive in the beginning.

neilpa:
Given the scope of this change, I'd like a review from @ReactiveCocoa/reactivecocoa before merging.

ReactiveCocoa/Swift/Optional.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  Optional.swift
+//  ReactiveCocoa
+//
+//  Created by Neil Pankey on 6/24/15.
+//  Copyright (c) 2015 GitHub. All rights reserved.
+//
+
+import Foundation
```
ikesyo: This import seems unnecessary.

ReactiveCocoa/Swift/Optional.swift
```diff
@@ -0,0 +1,24 @@
...
+
+/// An optional protocol for use in type constraints.
+public protocol OptionalType {
+	/// The type contained in the otpional.
+	typealias T
+
+	/// Extracts a signal from the receiver.
```
ikesyo: Wrong document.

ReactiveCocoa/Swift/Event.swift
```diff
@@ -159,6 +159,19 @@ extension Event: CustomStringConvertible {
...
 }
 
+/// Event protocol for constraining signal extensions
+public protocol EventType {
+	typealias T
+	typealias E: ErrorType
+	var event: Event<T, E> { get }
+}
+
+extension Event : EventType {
```
ikesyo: Style: There is an extra space before colon.

ikesyo:
We can now remove `ignoreNil()` from `FoundationExtensions.swift`?

ikesyo:
This compiles fine and seems legit! :+1:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -140,7 +140,29 @@ public final class Signal<T, E: ErrorType> {
...
+
+public protocol SignalType {
+	/// The type of values being sent on the signal.
+	typealias T
+	/// The type of error that can occur on the signal. If errors aren't possible
+	/// then `NoError` can be used.
+	typealias E: ErrorType
+
+	/// Extracts a signal from the receiver.
+	var signal: Signal<T, E> { get }
```
jspahrsummers: Why is this property necessary?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -140,7 +140,29 @@ public final class Signal<T, E: ErrorType> {
...
+	func observe(observer: Signal<T, E>.Observer) -> Disposable?
+}
 
+extension Signal: SignalType {
+	public var signal: Signal {
+		return self
+	}
+}
+
+extension Signal {
```
jspahrsummers: Why shouldn't this extend `SignalType`?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -140,7 +140,29 @@ public final class Signal<T, E: ErrorType> {
...
+	typealias T
+	/// The type of error that can occur on the signal. If errors aren't possible
+	/// then `NoError` can be used.
+	typealias E: ErrorType
+
+	/// Extracts a signal from the receiver.
+	var signal: Signal<T, E> { get }
+
+	/// Observes the Signal by sending any future events to the given sink.
+	func observe(observer: Signal<T, E>.Observer) -> Disposable?
```
jspahrsummers: Can `Observer` now be defined in this protocol, instead of `Signal`?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -151,54 +173,32 @@ public final class Signal<T, E: ErrorType> {
...
-}
 
-/// Maps each value in the signal to a new value.
-public func map<T, U, E>(transform: T -> U) -> Signal<T, E> -> Signal<U, E> {
-	return { signal in
-		return Signal { observer in
-			return signal.observe { event in
+	/// Maps each value in the signal to a new value.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func map<U>(transform: T -> U) -> Signal<U, E> {
```
jspahrsummers: Same here and for all these operators—can we extend `SignalType` instead of `Signal` directly?

ReactiveCocoa/Swift/Errors.swift
```diff
@@ -8,27 +8,16 @@
...
-	public var nsError: NSError {
-		return self
-	}
-}
+/// Workaround for constraining generic type extensions with non-protocol types.
+/// Unfortunately there's no way to guarantee that instances of types conforming
+/// to this protocol can't be created.
+///
+/// See http://www.openradar.me/21512469
+public protocol NoErrorType: ErrorType {}
```
jspahrsummers: If we change `NoError` into a `final class` with a private initializer, we can get rid of this protocol:

```swift
public final class NoError: NoErrorType {
	public let _domain: String = ""
	public let _code: Int = 0

	private init() {}
}
```

ReactiveCocoa/Swift/Event.swift
```diff
@@ -159,6 +159,19 @@ extension Event: CustomStringConvertible {
...
+	public var event: Event<T, E> {
+		return self
+	}
+}
+
 /// Puts a `Next` event into the given sink.
 public func sendNext<T, E: ErrorType>(sink: Event<T, E>.Sink, _ value: T) {
```
jspahrsummers: These could be extensions on `SinkType` now.

ReactiveCocoa/Swift/Event.swift
```diff
@@ -159,6 +159,19 @@ extension Event: CustomStringConvertible {
 	}
 }
 
+/// Event protocol for constraining signal extensions
```
jspahrsummers: Can you please fill out the documentation here? :pray:

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -86,7 +88,7 @@ private func optionalize<T, E>(signal: Signal<T, E>) -> Signal<T?, E> {
 ///
 /// Any `Interrupted` events will be silently discarded.
 public func toRACSignal<T: AnyObject, E>(producer: SignalProducer<T, E>) -> RACSignal {
```
jspahrsummers: These can be protocol extensions now.

jspahrsummers:
Looks great! I left some comments, several of which boil down to, “Let’s do these protocols right.”

While I would certainly love for folks to use `Signal` and `SignalProducer` as they are—_if_ we’re going to have protocols that represent that functionality, I’d like them to be as standalone and self-sufficient as possible (even if they ultimately reduce back down into the built-in types).

ReactiveCocoa/Swift/Errors.swift
kballard: If you really want to have a type that can't be created, just say

```swift
public enum NoError: ErrorType {}
```

With no variants, it has literally no inhabiting values.

neilpa:
Most of these should be easy to switch back to protocol extensions. I only switched them for the sake of consistency. However, I suspect some will have to remain as signal and/or producer extensions directly due to compiler crashes. I'll try now that we have beta 3 but have little hope things have been fixed since beta 2.

ReactiveCocoa/Swift/Signal.swift
neilpa: This is probably a leftover from before I added `observe` to the protocol. I'll see if this can be removed.

ReactiveCocoa/Swift/Event.swift
neilpa: We're not using `SinkType` anymore, sinks are just closures now.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
neilpa: The optional overloads are making this tricky to work as extensions.

ReactiveCocoa/Swift/Signal.swift
neilpa: Probably. However, other implementors could theoretically change it since we can only give it a default `typealias` without enforcing it.

jspahrsummers: We can't do `typealias Observer = ` in a protocol? (Just asking, because I don't actually know.)

ReactiveCocoa/Swift/Event.swift
jspahrsummers: Oh yeah. That's kind of unfortunate for this case. :confused:

ReactiveCocoa/Swift/ObjectiveCBridging.swift
jspahrsummers: Ah, no worries, then.

ReactiveCocoa/Swift/Signal.swift
neilpa: Turns out there are a few places relying on this that would otherwise require creating a new signal simply to `observe` the original.

neilpa:
This addresses some of the feedback. I've been unable to get the compiler to bend to my will for other pieces.

Issue #2053, opened 2015-05-27
What is the point of SinkOf


CosynPa:
Since `SinkOf<T>` is just a function `T -> Void`, why don't just use the plain function?

kastiglione:
`SinkOf` is a standard Swift type. It conforms to the protocol `SinkType`, which says:

> Useful mainly when the optimizer's ability to specialize generics outstrips its ability to
specialize ordinary closures. For example, you may find that instead of:
>
> `func f(g: (X)->Void) { ... g(a) ...}`
>
> the following generates better code:
>
> `func f<T: Sink where T.Element == X>(g: T) { ... g.put(a) ...}`

See http://swiftdoc.org/protocol/SinkType/

CosynPa:
@kastiglione Why is the later code better? Because it use generic type?

kastiglione:
I don't have an answer other than the compiler produces more optimal code, as the docs say. If anyone has more info here I'd love to learn more too.

jspahrsummers:
At one point, our observers were `SinkType`s _plus_ some special sauce. That's since changed, so I guess we could conceivably use plain functions now.

I'd be interested in a pull request trying it out, if anyone wants to give that a shot.

brow:
Per @neilpa's [comment](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2101#issuecomment-111848421) on #2101, this should probably be moved out of the 3.0 milestone.

Is there anything keeping us from creating a milestone for the API-breaking release that will support Swift 2? Would it be designated 4.0?

jspahrsummers:
@brow A separate milestone sounds like a good idea. Maybe just title it “Swift 2,” since it may be incorporated into RAC 3.0 if Swift 2 hits production before we've shipped officially.

neilpa:
I've added a milestone and added current issues and PRs to it.

brow:
Nice. Shall we close this now that #2102 is merged to `swift2`, or do we prefer to close only once the change is in a release?

jspahrsummers:
We can close this out, so the milestone progress is accurate.

PR #2101, opened 2015-06-14
Functions as observers

brow:
This is an experiment with representing observers with functions `Event<T, E> -> ()` rather than with `SinkType`s. The experiment was suggested in #2053.


ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -44,17 +44,17 @@ class SignalProducerSpec: QuickSpec {
...
 
-				weak var testSink: TestSink?
+				weak var objectRetainedByObserver: NSObject?
 				producer.startWithSignal { signal, _ in
-					var sink = TestSink()
-					testSink = sink
-					signal.observe(sink)
+					let object = NSObject()
+					objectRetainedByObserver = object
+					signal.observe { _ in object }
```
brow: I wonder if `object` in this closure is at risk of being compiled out.

neilpa:
I made the same change in 4a30f08 as part of #2099 (although I didn't fix the lifetime tests yet). I'd want to wait on this change until we have a Swift 2 release rather than introduce another breaking change before 3.0.

brow:
Sounds good to me. I'll close this PR and reopen it against `swift2`.

Issue #1999, opened 2015-05-12
How to connect an Action to UIButton


gabro:
Hi,
I'm trying to connect an `Action` to a `UIButton` using `CocoaAction` but I can't seem to make this compile. Here's my attempt:

## ViewModel
```
let email = MutableProperty<String>("")
let password = MutableProperty<String>("")
var loginAction: Action<Void, (String, String, String), NoError>!

init() {
    loginAction = Action(enabledIf: validLoginData) { (_: Void) in
      return combineLatest(self.email.producer, self.password.producer)
    }
    
    loginAction.values.observe { (email, password) in
      println("\(email) \(password)")
    }
}
```

## ViewController
```
    override func viewDidLoad() {
      viewModel.email           <~ loginView.emailTextField.rac_textSignalProducer()
      viewModel.password        <~ loginView.passwordTextField.rac_textSignalProducer()
      viewModel.activationCode  <~ loginView.activationCodeTextField.rac_textSignalProducer()
      
      let loginCocoaAction = CocoaAction(viewModel.loginAction)  
      loginView.loginButton.addTarget(loginCocoaAction, action: CocoaAction.selector, forControlEvents: UIControlEvents.TouchUpInside)      
    }
```

I get a compiler error on the line

```
let loginCocoaAction = CocoaAction(viewModel.loginAction)
```

Here's the message:
> /Users/gabro/Work/buildo/hailadoc/ios/hailadoc-ios/LoginViewController.swift:42:25: Cannot find an initializer for type 'CocoaAction' that accepts an argument list of type '(Action<Void, (String, String, String), NoError>!)'

According to the [changelog for RAC 3.0](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/08a73c888bf02b372e02ed338ecf3352afad3b77/CHANGELOG.md) I should be able to wrap my action with `CocoaAction` in order to attach it to a `UIControl` such as `UIButton`, but I seem to be unable to make this compile.

What am I doing wrong here?

ikesyo:
The [`CocoaAction` initializer you are using](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/9ffe766aa5ed2a50a2a5ac4abcdec4a31ff65f0a/ReactiveCocoa/Swift/Action.swift#L193) can accepts `Input` which is `AnyObject`, but your action `Input` type is `Void` which is not `AnyObject`. You might be able to use the following initializer instead:

```swift
[...]

PR #2104, opened 2015-06-15
Add cocoaAction property to Action

neilpa:
As discussed in #1999

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -44,9 +44,12 @@ public final class Action<Input, Output, Error: ErrorType> {
 	/// Whether the instantiator of this action wants it to be enabled.
 	private let userEnabled: PropertyOf<Bool>
 
-	/// Lazy creation and storage of a UI bindable CocoaAction. The default
-	/// version force casts the AnyObject? input from CocoaAction to match
-	/// `Input` but this property can be overriden for alternative behavior.
+	/// Lazy creation and storage of a UI bindable CocoaAction.
+	///
+	/// The default version force casts the AnyObject? input from CocoaAction to
+	/// match `Input` but this property can be overriden for alternative behavior.
+	/// In cases where the input type is ignored use `Any` instead of `Void` or
+	/// `()` because the bound control sends self as input.
 	public lazy var cocoaAction: CocoaAction = { _ in
 		CocoaAction(self) { $0 as! Input }
 	}()
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -44,6 +44,13 @@ public final class Action<Input, Output, Error: ErrorType> {
 	/// Whether the instantiator of this action wants it to be enabled.
 	private let userEnabled: PropertyOf<Bool>
 
+	/// Lazy creation and storage of a UI bindable CocoaAction. The default
+	/// version force casts the AnyObject? input from CocoaAction to match
+	/// `Input` but this property can be overriden for alternative behavior.
+	public lazy var cocoaAction: CocoaAction = { _ in
+		CocoaAction(self) { $0 as! Input }
+	}()
+
 	/// This queue is used for read-modify-write operations on the `_executing`
 	/// property.
 	private let executingQueue = dispatch_queue_create("org.reactivecocoa.ReactiveCocoa.Action.executingQueue", DISPATCH_QUEUE_SERIAL)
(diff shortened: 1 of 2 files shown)


rpowelll:
:+1: This seems like a better approach than making users manage `CocoaAction` instances themselves

ReactiveCocoa/Swift/Action.swift
```diff
@@ -44,6 +44,16 @@ public final class Action<Input, Output, Error: ErrorType> {
 	/// Whether the instantiator of this action wants it to be enabled.
 	private let userEnabled: PropertyOf<Bool>
 
+	/// Lazy creation and storage of a UI bindable CocoaAction.
+	///
+	/// The default version force casts the AnyObject? input from CocoaAction to
+	/// match `Input` but this property can be overriden for alternative behavior.
```
jspahrsummers: This property can't be overridden, since the class is `final`.

jspahrsummers:
:+1: in general, but I'm nervous about the implicit (hidden) use of `as!`.

Can we create the `CocoaAction` with an overloaded `private` function that automatically handles at least the specific case of `Input == ()`? That would avoid the most common potential crash.

ReactiveCocoa/Swift/Action.swift
neilpa: Poor wording, I meant re-assigned. However, this warning is mostly moot if the `private` function works.

neilpa:
I'll give that a try

neilpa:
I'm not sure the overloaded function approach is going to work. I tried multiple variations of the following but the wrong wrapper is always chose for `()` input. Is there some way to better discriminate these cases to effectively get `Input = ()` and `Input != ()`?

```swift
lazy var cocoaAction: CocoaAction = { _ in Action.wrapAction(self) }()

static func wrapAction(action: Action<(), Output, Error>) -> CocoaAction {
	return CocoaAction(action, input: ())
}

// Also tried `Action<Any, Output, Error>` here since that should cover all cases with above
// but then the compiler fails to infer types.
static func wrapAction(action: Action) -> CocoaAction {
	return CocoaAction(action) { $0 as! Input }
}
```

jspahrsummers:
Maybe it'd be easier to have an overloaded function for transforming the input, `-> ()` or `-> T`.

neilpa:
No go :sob:. I even tried this and found a compiler error I hadn't seen before.

screen shot 2015-06-16 at 12 02 40 am

ikesyo:
Yep, `Void` is a empty (0 element) tuple, and tuples are not extendable...

jspahrsummers:
Sorry, I meant we could have some functions like so:

```swift
static func transformCocoaActionInput(input: AnyObject?) -> () {
    return ()
}

static func transformCocoaActionInput(input: AnyObject?) -> T? {
    return input as? T
}
```

etc., as needed.

neilpa:
Yea, I tried that first before trying the protocol (that's what my "No go" comment was referring to). The problem is it still requires a catch-all `AnyObject? -> Input` with force casting and the compiler always picks that over the specialized overloads.

jspahrsummers:
In lieu of a perfect solution, I guess, maybe we can call this `unsafeCocoaAction` or something like that?

neilpa:
I'd be fine with that.

On a related note, I was experimenting last week with some [UIKit bindings](https://github.com/neilpa/Rex/pull/15) that would obviate the need for this property.

```swift
extension UIButton {
    public var rac_pressed: MutableProperty<CocoaAction> { ... }
}

class MyView: UIView {
    let model: MutableProperty<MyModel>
    let button: UIButton!

    public func viewDidLoad() {
        button.rac_pressed <~ model.producer
            |> map { CocoaAction($0.someAction, input: () }
    }
}

neilpa:
Updated to `unsafeCocoaAction`

jspahrsummers:
:shipit:

almassapargali:
I'm rewriting some parts of my app because I found memory leaks lately. And now I'm thinking, won't current implementation of `unsafeCocoaAction` lead to memory leak, since `Action` stores `CocoaAction`, who retains `Action` in it's `_execute` closure?

guidomb:
What is the "best" way to define an action in a view model that does not care about the input? Using `AnyObject?` or `Void` generates a runtime error when using the `unsafeCocoaAction` property

almassapargali:
Yes. Because unsafeCocoaAction force casts to Input type (that's why it's unsafe). If your action takes Void as Input you may assign unsafeCocoaAction property with your own CocoaAction. Like:

myAction.unsafeCocoaAction = { [unowned myAction] in
    return CocoaAction(myAction, input: ())
}()

guidomb:
@almassapargali Thanks! Looks like is more typing than just using `AnyObject!` as the Action's input type.

PR #2111, opened 2015-06-18
Add @warn_unused_result annotations

neilpa:
Note that this PR targets #2099

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -86,6 +86,7 @@ public final class Action<Input, Output, Error: ErrorType> {
 	/// If the action is disabled when the returned SignalProducer is started,
 	/// the produced signal will send `ActionError.NotEnabled`, and nothing will
 	/// be sent upon `values` or `errors` for that particular signal.
+	@warn_unused_result
 	public func apply(input: Input) -> SignalProducer<Output, ActionError<Error>> {
 		return SignalProducer { observer, disposable in
 			var startedExecuting = false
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/Swift/Action.swift
```diff
@@ -86,6 +86,7 @@ public final class Action<Input, Output, Error: ErrorType> {
 	/// If the action is disabled when the returned SignalProducer is started,
 	/// the produced signal will send `ActionError.NotEnabled`, and nothing will
 	/// be sent upon `values` or `errors` for that particular signal.
+	@warn_unused_result(message="Did you forget to `start` the producer?")
```
JaviSoto: Woot this is really cool!!!!

JaviSoto:
This is awesome. I wish there was a way to factor this out somehow, there's a lot of "duplication", but it's not really code per-se so it's not a huge deal.

:shipit: 

(I'll merge this already because it targets a different PR. If anybody else disagrees with this we can always revert it)

neilpa:
:sparkles:

neilpa:
> (I'll merge this already because it targets a different PR. If anybody else disagrees with this we can always revert it)

Figured it would be nice to have a separate history for this which is I why I didn't include it directly in the protocol extension branch.

neilpa:
Oh, and here's what it actually looks like

screen shot 2015-06-18 at 2 24 40 pm

nomothetis:
If I may, the "`" looks a bit like a quote, which reads wrong. I'd say change it either "Did you forget to call `start` on the producer?" or simply "Did you forget to start the producer?"

neilpa:
Done 30ba8ad

nomothetis:
:bow:

NachoSoto:
🎊

ReactiveCocoa/Swift/Action.swift
MarcoSero: :100:

jspahrsummers:
:tada:

PR #2120, opened 2015-06-23
Add two macros named RACOnMainThread and RACCell.

WilliamZang:
Lots of people use RAC macro in UI. But it's not only deliver on main thread. Sometime it occur some bugs. 

So I added two macros for these question. Is it useful?

### DIFF
--- a/ReactiveCocoa/Objective-C/RACSubscriptingAssignmentTrampoline.h
+++ b/ReactiveCocoa/Objective-C/RACSubscriptingAssignmentTrampoline.h
@@ -42,13 +42,28 @@
         (RAC_(TARGET, __VA_ARGS__, nil)) \
         (RAC_(TARGET, __VA_ARGS__))
 
-/// Do not use this directly. Use the RAC macro above.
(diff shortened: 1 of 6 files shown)


jspahrsummers:
Thanks for the pull request!

In general, however, we try to minimize the number of macros that the framework exposes, keeping them only when they _significantly_ improve readability versus the alternative. Compare `RAC() = ` with `-[RACSignal setKeyPath:onObject:]`, for example.

In this case, it's really easy to use the `-deliverOnMainThread` operator before binding. If you forget to use that operator, why would a second macro help you remember?

We also don't want to add macro variants for every possible combination. For example, it's important to catch errors when binding to a property, but a special macro that did that for you would bloat the API even further.

So although I appreciate the intention behind this PR, I don't think we can justify it in the framework. Thank you anyways! :sparkles:

WilliamZang:
You're welcome. I know what you mean.
I'll try to contribute as another way in the future.

WilliamZang:
You're welcome. I know what you mean.
I'll try to contribute as another way in the future.

PR #2133, opened 2015-06-29
Refactor NSError creation to use message parameter of Result.static.error()

ikesyo:
That can be used from [0.4.4](https://github.com/antitypical/Result/blob/0.4.4/Result/Result.swift#L102-L115).


jspahrsummers:
:sparkles:

PR #2134, opened 2015-06-29
Replace shouldBeEnabled() with _enabled.value in Action.apply()

### LINKED ISSUE #2078: RAC3: Using shouldBeEnabled instead of enabled in Action's apply method
Hi all,

I've been reading through RAC3's (very elegant) Swift code, partly to understand how to use the framework, and partly just to learn by immersion how excellent Swift code is written. I've been learning a lot. 

I had a pedantic question about `Action`'s apply method, where you use this closure:

<pre>
if Action.shouldBeEnabled(userEnabled: self.userEnabled.value, executing: self._executing.value) {
    self._executing.value = true
    startedExecuting = true
}
</pre>

Is there any benefit to calling `shouldBeEnabled`, instead of using the existing `enabled` property, which is bound to the return value of `shouldBeEnabled` in `init`? Does it have to do with the possibility of reading the `enabled` property before it's updated by the binding?

Thanks!

ikesyo:
Resolves #2078.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -91,7 +91,7 @@ public final class Action<Input, Output, Error: ErrorType> {
 			var startedExecuting = false
 
 			dispatch_sync(self.executingQueue) {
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Thanks!

PR #2148, opened 2015-07-06
Add ProducedSignal typealias in SignalProducer

### LINKED ISSUE #2043: Add more typealiases
Given a `SignalProducer<T, E>`, there's no easy way to get its associated `Signal` type, or even sink type IIRC. We should add some public typealiases to make stuff like this easier.

Same thing probably applies to `Action`, `PropertyType`, and others.

ikesyo:
Resolves #2043.

Should we also add `public typealias Observer = ProducedSignal.Observer`?


jspahrsummers:
This is fine as-is. Thanks! :sparkles:

PR #2151, opened 2015-07-06
Fix documentation for scan operator

MichaelMcGuire:
Change an incorrect reference from `reduce` to `scan` in the documentation


jspahrsummers:
Thanks!

PR #2158, opened 2015-07-09
Updates for Xcode 7 beta 3

ikesyo:
Type inference about single-expression closures with non-Void return types seems to be improved.


neilpa:
Thanks :tada:

PR #2186, opened 2015-07-19
Remove CocoaAction initializer that doesn't work

### LINKED ISSUE #2176: CocoaAction input is always nil
let action : Action<AnyObject?,AnyObject?,NoError> = Action { input in
            return SignalProducer(value: input)
        }
   
        cocoaAction = CocoaAction(action)
        button.addTarget(cocoaAction, action: CocoaAction.selector, forControlEvents: .TouchUpInside)
        
        action.values |> observe(next: println) // prints 'nil'

i just conver the input to output, I was expecting the button itself as input on the action, but i always get returned nil.

jspahrsummers:
Resolves #2176.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -188,13 +188,6 @@ public final class CocoaAction: NSObject {
 		self.init(action, { _ in input })
 	}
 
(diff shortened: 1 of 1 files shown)


jspahrsummers:
This builds just fine. Travis is complaining about a race-y Objective-C test. :unamused:

PR #2191, opened 2015-07-19
Change scheduler on README

almassapargali:
In [this](https://github.com/ReactiveCocoa/ReactiveCocoa#throttling-requests) section of README it makes use of `throttle` like this:
```
    |> throttle(0.5, onScheduler: UIScheduler())
```
However, `throttle` takes `DateSchedulerType` as second parameter, which `UIScheduler` doesn't confirms to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ search strings:
 let searchStrings = textField.rac_textSignal()
     |> toSignalProducer()
     |> map { text in text as! String }
-    |> throttle(0.5, onScheduler: UIScheduler())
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Oops, thanks! Can you fix the reference in the passage directly below too? :pray:

almassapargali:
I'm sorry, not exactly sure what to fix else?

jspahrsummers:
Ah, I was thinking of a different section. Nevermind.

Thanks for the PR!

PR #2185, opened 2015-07-18
[WIP] propertyOf operator

brow:
I [observed](https://reactivex.slack.com/archives/reactivecocoa/p1437252176000002) today that @neilpa and I have both written almost the same function `propertyOf<T>(initialValue: T)(signal: Signal<T, NoError>) -> PropertyOf<T>` in RAC3 projects. I've actually pasted it into two of my projects.

This adds `propertyOf` to RAC, using @neilpa's documentation and function signature from [Rex](https://github.com/neilpa/Rex) (which I prefer because it's curried for use with `|>`) and my implementation from [Sukhasana](https://github.com/brow/Sukhasana) (which I don't strongly prefer over @neilpa's).

@jspahrsummers did once give some reasons for not including a function like this: https://reactivex.slack.com/archives/reactivecocoa/p1424118385000915

This is currently crashing tests, possibly because of memory corruption introduced by @noescape.

Issue #2006, opened 2015-05-14
Property from a 1-buffer?


andymatuschak:
What do you folks think about this API proposal? I often find myself wanting to make signal-based APIs, but to allow for easy synchronous "sampling" in cases where that makes sense.

```swift
/// A property which makes its signal's most recently emitted value always
/// available for instantaneous access
public struct SignalProperty<T, E>: PropertyType {
  /// Initializes the property with `initialValue`, immediately starts
  /// `signalProducer` with an observing buffer of size 1. Values emitted
  /// on `signalProducer` will replace `initialValue` as the property's value.
  public init(initialValue: T, signalProducer: SignalProducer<T, E>)
}
```

neilpa:
I've written something very similar to this but in terms of reduce which offers more flexibility. I hadn't considered wrapping it in a `PropertyType` though (I was reducing collection deltas to a collection view data source)

```swift
public struct ReducibleProperty<T, U, E>: PropertyType {
    public init(initial: U, combine: (U,T) -> U, producer: SignalProducer<T, E>)
}
```

Regardless, I'm :+1: for something like this.

jspahrsummers:
Is this just a wrapper around `MutableProperty` plus `<~`? Not that there would necessarily be anything wrong with that, I just want to understand where it would fit in.

andymatuschak:
Ah, yep! I hadn't noticed `<~`!

OK. Then probably the common pattern would look like this:

```
public class SomeController {
  public let someProperty: PropertyOf<Bacon> { return PropertyOf(_someProperty) }
  private let _someProperty: MutableProperty<Bacon>
  
  init(dependency: Dependency) {
    _someProperty = MutableProperty(initialValue: .Uncooked)
    _someProperty <~ dependency.baconSignalProducer
  }
}
```

My suggested API would remove some of the "ceremony" but isn't a *huge* difference:

```
public class SomeController {
  public let someProperty: PropertyOf<Bacon>

  init(dependency: Dependency) {
    someProperty = PropertyOf(SignalProperty(initialValue: .Uncooked, signalProducer: dependency.baconSignalProducer))
  }
}
```

jspahrsummers:
I like that it's declarative, though, and expresses intention better. I'd be :+1: on adding something like that.

andymatuschak:
Cool. I'll send a pull request after I return from honeymoon (late June).

MichaelMcGuire:
[...]

PR #2192, opened 2015-07-19
propertyOf operator

brow:
Reopening #2185 against `master`:

> I observed today that @neilpa and I have both written almost the same function propertyOf<T>(initialValue: T)(signal: Signal<T, NoError>) -> PropertyOf<T> in RAC3 projects. I've actually pasted it into two of my projects.

> This adds propertyOf to RAC, using @neilpa's documentation and function signature from Rex (which I prefer because it's curried for use with |>) and my implementation from Sukhasana (which I don't strongly prefer over @neilpa's).

> @jspahrsummers did once give some reasons for not including a function like this: https://reactivex.slack.com/archives/reactivecocoa/p1424118385000915

> This is currently crashing tests, possibly because of memory corruption introduced by @noescape.

@jspahrsummers said:
> Could this instead be a PropertyOf initializer?

To do:
- [x] Documentation for new functions/methods
- [x] More tests 
- [x] Separate issue and repro for the memory corruption issue

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -228,3 +228,19 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
 public func <~ <Destination: MutablePropertyType, Source: PropertyType where Source.Value == Destination.Value>(destinationProperty: Destination, sourceProperty: Source) -> Disposable {
 	return destinationProperty <~ sourceProperty.producer
 }
+
+
+/// Creates a new property bound to `signal` starting with `initialValue`.
+public func propertyOf<T>(initialValue: T)(signal: Signal<T, NoError>) -> PropertyOf<T> {
+	let mutableProperty = MutableProperty(initialValue)
+	mutableProperty <~ signal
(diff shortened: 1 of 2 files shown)


brow:
I like the initializers better, and not just because they sidestep the memory corruption issue I was triggering with `|>`. (I'll open a separate issue for that.)

@jspahrsummers This does make it possible to avoid using a `MutableProperty` in the implementation, though I haven't tried that yet. Is that what you had in mind?

jspahrsummers:
Looks good! Using `MutableProperty` is fine.

I'll let this sit for a short while, in case other collaborators have opinions (before we add this to the API that can't break), but I'm :+1: as it stands.

335g:
Looks good :sparkles:

neilpa:
I'm obviously in favor of letting this sneak in pre-3.0 :wink:

neilpa:
This has sat for long enough, I'm gonna go ahead and merge this.

almassapargali:
I think #2006 can be closed too. Great addition, by the way 👍

neilpa:
Good call

PR #2032, opened 2015-05-22
Replace explicit currying syntax with functions returning functions

jspahrsummers:
I still occasionally see crashes that appear to be caused by currying. @robrix has mentioned in the past that refactoring curried functions into this form typically fixes it.

Unfortunately, the crashes are fairly nondeterministic, so I can't truly verify this fix, but I don't see how it could get any worse.

PR #1824, opened 2015-03-13
Enable test fixed by Swift 1.2b3

neilpa:
Also fixes `timeoutWithError` since that was still crashing in the tests like a few other curried functions.

/cc @jspahrsummers

PR #2194, opened 2015-07-19
[WIP] Memory corruption

brow:
While working on #2192, I accidentally wrote a test that crashes with memory corruption. In this branch, I'm trying to reduce the crashing test to understand what's going on.

So far, it seems like the prerequisites for the crash include defining a polymorphic, curried function and using `|>` to apply it to a `Signal` or `SignalProducer`. The fact that `a |> f(b)` triggers the crash while `f(b)(a)` does not led me to discover #1953, but I'm not sure yet that this is a `@noescape` issue.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -228,3 +228,19 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
 public func <~ <Destination: MutablePropertyType, Source: PropertyType where Source.Value == Destination.Value>(destinationProperty: Destination, sourceProperty: Source) -> Disposable {
 	return destinationProperty <~ sourceProperty.producer
 }
+
+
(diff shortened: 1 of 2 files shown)


brow:
The crash has been difficult to isolate because it's not deterministic, but each of the commits on this branch produced a crash in at least one test run.

brow:
The crash still occurs if I revert 99b999509ad1b8647d86e66aa005eb9def0a76c4, so this seems to be a separate issue from #1953.

ikesyo:
We had many crashes related to curried functions in operator definitions. Because of that, we now don't use curried functions anymore to avoid that undeterministic crashes. I belive this is a compiler bug and fixed in Xcode 7/Swift 2 beta.

brow:
@ikesyo That resolves the crash. Thanks!

ikesyo:
@brow No problem!

Ref: #1775(eed1485, 04af6ca, 43808aa), #1824, #1902 and #2032.

PR #2121, opened 2015-06-23
watchOS support

MarcoSero:
This PR is against the [`protocol extensions`](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2099) branch.

- [X] watchOS target successfully building
- [ ] ~~watchOSTests passing all tests~~

Quick does not seem to support watchOS at the moment.

UPDATE: unfortunately the watchOS test target can't be added until Apple fixes [rdar://21760513](https://openradar.appspot.com/21760513)

PR #2206, opened 2015-07-22
Added framework for WatchOS

JARinteractive:
Updated Result to version supporting WatchKit.
Added New WatchKit target (no test target, as tests are not supported against watch targets at the moment)
Disable DTrace on WatchKit due to compilation errors relating to embedded assembly code

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "antitypical/Result" ~> 0.5
+github "antitypical/Result" "0.6-beta.1"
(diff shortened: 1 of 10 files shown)


MarcoSero:
I had already opened https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2121 but then I got stuck trying to add the test target for the watch.
Did Apple mention somewhere that tests on the watch are not supported yet?

JARinteractive:
I've encountered the same problem reported in this radar: https://openradar.appspot.com/21760513

PR #2207, opened 2015-07-23
Update README

zhigang1992:
`toSignalProducer` is only available on Swift Signal


ikesyo:
Nice catch! :+1: Thank you for the fix.

ikesyo:
I added one more commit on master: e3c3101

Issue #2075, opened 2015-06-07
Proposal: revive debounce (the throttle of old)


heiberg:
The new `throttle` in RAC 3 has a behavior which matches its name better. But sometimes the old RAC 2 `throttle` behavior is just what you need. 

[This issue](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1348) raised the question a while back, and it was suggested that a new issue be opened if anyone wanted a `debounce`-like operation. 

I'd like to upvote the idea.

Example real-world use cases:
- Hide the page control for a scroll view when the user hasn't scrolled for a while.
- Postpone kicking off a slow network search operation until the user has stopped typing in the search field.

neilpa:
I'm in favor of this operator, too many real world scenarios where it's useful. However, I'd want to wait until after a proper 3.0 release since this is additive and we're basically locked on the initial API.

jspahrsummers:
Agreed with @neilpa. :zap:

carlosefonseca:
That would be great! Also updating the docs would be sweet too: https://github.com/ReactiveCocoa/ReactiveCocoa/tree/swift2#throttling-requests :)

natestedman:
Is this a correct implementation of `debounce`?

    public extension SignalProducerType
    {
        @warn_unused_result
        public func debounce(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType) -> SignalProducer<Value, Error>
        {
            return flatMap(.Latest, transform: { next in
                SignalProducer(value: next).delay(interval, onScheduler: scheduler)
            })
        }
    }

The [Objective-C implementation](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/Objective-C/RACSignal+Operations.m#L139) is much more complex, despite `map`, `switchToLatest`, and `delay` all being available, so I feel like I might be missing something here. However, it appears to work, although I've only used it on `mainQueueScheduler`.

NachoSoto:
That looks correct!

mdiep:
That looks like an implementation of `throttle` to me, not `debounce`.

IIUC, If you have a signal that sends 10 values, one every `0.5` seconds, then:

* `throttle(1)` would send 1 value—1 second after the last value was sent
* `debounce(1) would send something like 5 values—1 every second

natestedman:
I think in that case, Swift `throttle` would send two values - let's say the signal's sending ascending integers:

[...]

Issue #1348, opened 2014-05-22
Throttling signals.


zdavison:
A common problem I have is that I have very high throughput signals that are sending values at a very fast rate. I really don't need these to send so frequently, so I'd like to say 'only send me a new value is 2 seconds or more have elapsed'.

My understanding of `throttle:` is that it actually does the reverse, only sending me values when the signal hasn't sent anything in X seconds. 

How do I get the behaviour I'm looking for?

PR #2208, opened 2015-07-23
Throttling is sending next within interval

zhigang1992:
Just added tests here to see if I understand `throttling` correctly

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -1072,6 +1072,9 @@ class SignalSpec: QuickSpec {
 				expect(values).to(equal([0]))
 
 				sendNext(observer, 1)
+				scheduler.advance()
+				expect(values).to(equal([0]))
+
 				scheduler.advanceByInterval(0.5)
 				expect(values).to(equal([0]))
 			}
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -1058,6 +1058,27 @@ class SignalSpec: QuickSpec {
 				expect(signal).notTo(beNil())
 			}
 
+			it("should not send values when it's less than interval") {
```
brow: "Should not send values when it's less than interval" seems synonymous with "should send values on the given scheduler at no less than the interval", the name of the next example. It's not clear to me if they are testing distinct properties of `throttle`. Would it make sense to combine the test code into a single example?

jspahrsummers:
It seems like you are expecting `throttle` to delay _all_ values by at least the given interval (like RAC 2.x)—is that correct?

`throttle` no longer behaves that way. Instead, its only role is to ensure that _at least_ the given interval elapses between successive values. If the interval has already elapsed when a value is received, it can be forwarded immediately.

zhigang1992:
@jspahrsummers @brow hm, I see... Guess I should read the doc a bit more careful. :)

lingoer:
:)

brow:
> It seems like you are expecting throttle to delay all values by at least the given interval (like RAC 2.x)—is that correct? throttle no longer behaves that way.

It sounds like it would be appropriate to mention this in the [CHANGELOG](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/CHANGELOG.md).

brow:
Also, `enum ThrottleStrategy`, anyone? :trollface:

zhigang1992:
@brow I also found those two
https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1348
https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2075

Maybe debounce would be a good idea

PR #2240, opened 2015-08-01
Add bindings for properties with an optional type

JARinteractive:
Allow use of <~ operator to bind to MutableProperty<T?> to Signal<T>, SignalProducer<T>, or Property<T>


neilpa:
Thanks for the PR, but I'm not sure adding additional overloads to the `<~` operator is a good idea. Does anyone else on @ReactiveCocoa/reactivecocoa have an opinion here?

jspahrsummers:
I agree—seems easy enough to apply `map { $0 }` or something similar. I kinda dislike how “magic” optionals are in general. :confused:

PR #2283, opened 2015-08-24
Fix backticks in unsafeCocoaAction documentation

juliangrosshauser:
Xcode's Quick Help (for example) doesn't like the second backtick:

screenshot 2015-08-24 23 18 05


neilpa:
Thanks for catching and fixing that :tada:

PR #2289, opened 2015-08-26
Added Signal & SignalProducer silent feature

pepibumur:
### Context?
ReactiveCocoa supports a custom ErrorType defined as `NoError`. Some features make use of that kind of signals like for example the property binding between Signal and SignalProducers and properties. 

If you work with Signal/Producers that report errors they cannot be directly binded to the properties. In that case you have to convert it into a new Signal/Producer<T, NoError> in order to being able to bind it. 

### Proposal
I was thinking about the best way to integrate it with ReactiveCocoa and I thought about a method called *silent* that propagates all the events except the error event. That way the developer can do something like:

```swift
property <~ my_signal().silent()
```

### Doubts
- Don't know if it's the best approach, and if it's aligned with the API at all but I've faced that problem and I guess more developers are doing.
- I thought about passing an error block that way we force the developer to decide what to do with errors before silencing the Signal/Producer

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1068,6 +1068,28 @@ public func zip<S: SequenceType, T, Error where S.Generator.Element == Signal<T,
 	return Signal.never
 }
 
+/// Silents the Signal errors.
+/// It's useful to bind a Signal into a property (binding operator supports only Signals of ErrorType == NoError)
+///
+/// Returns a Signal that propagates all the source Signal events but the error event
+public func silent<T, E>() -> Signal<T, E> -> Signal<T, NoError> {
+    return { signal in
+        return Signal { observer in
+            signal.observe { (event) -> Void in
+                switch event {
+                    case .Error(_): break
+                case .Completed:
+                    observer(.Completed)
+                case .Interrupted:
+                    observer(.Interrupted)
+                case .Next(let next):
+                    observer(.Next(next))
+                }
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


NachoSoto:
I'd be worried about it encouraging a bad practice: ignoring errors. Even `RxJava` doesn't have an operator like this.

neilpa:
Yea, this discussion has come up before and this operator makes it _too_ easy to ignore errors without thinking about them.

Also, the implementation you have creates an infinite signal if an error is ever sent which probably isn't what you want.

pepibumur:
Ohps you're right with the @neilpa .
I didn't know about these discussions before. I totally agree with your suggestions @NachoSoto  and @neilpa . I also propose forcing the developer passing an error closure to create that signal.

```swift
property <~ my_signal().silent { (error) -> Void in 

}
```
It's just a proposal and don't know if it follows at all the principles of the ReactiveCocoa's API design. If you consider not, feel free to close it.

Thanks :)

NachoSoto:
We're trying to keep the 3.0 API surface as slim as possible, but thanks for the PR @pepibumur!

For that reason @neilpa started https://github.com/neilpa/Rex, maybe this can be useful there?

neilpa:
Yup, and it already even has [`ignoreError`](https://github.com/neilpa/Rex/blob/6d2d95e55daede4c9f26c5e7c3d7169ec8ddba07/Source/Signal.swift#L29-L35)

pepibumur:
Wow! thanks @NachoSoto and @neilpa , I'm closing this branch.
I'll use Rex, great contribution

PR #2189, opened 2015-07-19
Swift 2

neilpa:
This PR currently targets **Xcode 7**. It should also work on Xcode 7.1 beta.

Of significant note is that |> is gone. Signal operators are now implemented as protocol extensions so you can use normal dot-method syntax.

----

Continuation of #2088 since the `swift-development` branch has been merged to master

PR #2292, opened 2015-08-27
Always use code signing for iOS target

rpowelll:
This fixes an issue that would arise when trying to use the framework with Carthage, as brought up in #2189

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -2461,7 +2461,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D047263419E49FE8006002AA /* iOS-Framework.xcconfig */;
 			buildSettings = {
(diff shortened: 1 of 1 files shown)


JaviSoto:
Yeah, I had this issue as well. I was patching it locally while I figured out if there was a reason for it to be the way it is, but I don't know :D 

I'll wait for another contributor to give the thumbs up, but this looks good to me :+1:

neilpa:
I was also hoping for someone more versed in the nuances of code signing would chime in which is why I hadn't merged it yet.

rpowelll:
Yeah this solution doesn't seem ideal. I'd love some input from someone more knowledgeable on the topic, but I thought it made sense to open this PR to give the issue some more visibility.

NachoSoto:
I'm afraid to call myself "versed in the nuances of code signing", but I'm willing to say this diff looks correct.

neilpa:
I'll go ahead and merge this then, especially since this is still a beta branch (even though I think a fair number of people are using it).

rpowelll:
:rocket:

PR #2295, opened 2015-08-27
[WIP/Feedback] Simplify/modernize QueueScheduler API

liscio:
To take advantage of the improvements in libdispatch that appeared in iOS 8 and OS X 10.10, and to heed the suggestion in the headers for the same, I am proposing a change to the QueueScheduler class. Here's the relevant quote from the docs:

> The `dispatch_get_global_queue()` function may be used to obtain a target queue of a specific quality of service class, however the use of `dispatch_queue_attr_make_with_qos_class()` is recommended instead.

(I wish they'd specify why it's recommended. Was the other approach problematic? Is it just a general cleanup? So many questions!)

Anyway, the API would be simplified such that you can either:

1. Ask for the mainQueueScheduler, or
2. Specify your requested QOS class on initialization

I keep trying to think through good reasons why we might still want to offer the ability to supply another queue on initialization, and the one that remains is that the caller might want to supply their own queue for protecting access to a shared resource, and RAC's Signal{,Producer}s could better co-exist with "legacy code" that accesses the same resource.

In _that_ case, perhaps a different kind of scheduler is warranted for wrapping existing queues, and we can advise against using it in the common documentation/examples/patterns?

An alternate approach (my preferred one) would be to allow access to the underlying queue instance variable, so legacy code can simply replace their own queue creation with a QueueScheduler() to suit their fancy, and then access its queue property.

Thoughts?

(This PR points at the `swift2` branch because that's where I've been working from lately.)


jspahrsummers:
There are still some consumers of RAC that support OS X 10.9 (GitHub Desktop, most notably).

cc @joshaber @mdiep @robrix

liscio:
@jspahrsummers Fair enough, but using swift2 will require folks to adopt the 10.11 SDK (as the current Xcode 7 doesn't seem to allow swift2 compilations against the 10.10 SDK the last I checked), so perhaps by the time the `swift2` becomes current, they might drop that requirement?

That said, I can also use swift2's availability feature to offer the existing behavior on API consumers requiring that behavior. I'll try that out next time I get a chance to bang on this.

neilpa:
> swift2 will require folks to adopt the 10.11 SDK

This is independent of the minimum deployment target. You can build against a newer SDK but still target older OS's. I even thought I read somewhere that Xcode 7 is finally adding warnings if you use new APIs that aren't available in your minimum deployment target.

liscio:
@neilpa This is true, however I have found in practice that it can be more painful to be on the new SDKs attempting to target too-old OSes. That said, swift2 is supposed to alleviate this pain—we have to wait and see, I guess.

Anyway, I'll take that as a 2nd "vote" (including my own) to explore the reintroduction of the status quo for 10.9 consumers using the new availability macros.

(So far no violent disagreement with the meat & potatoes of the change is a good thing, right? :P)

neilpa:
> I have found in practice that it can be more painful to be on the new SDKs attempting to target too-old OSes

Yup, been bitten by that one several times.

> Anyway, I'll take that as a 2nd "vote" (including my own) to explore the reintroduction of the status quo for 10.9 consumers using the new availability macros.

Sounds good.

liscio:
I appear to have successfully "set things right" for the OS X 10.9 consumers with this latest changeset, however there is one point of awkwardness that is introduced if someone wants to construct a `QueueScheduler` with no parameters. Note how in the `QueueSchedulerSpec` I had to specify a `priority` or `qos` in order to "break the tie" between the two constructor definitions. Not sure whether that's a big deal.

Similarly I renamed the internal `init()` method to take an `internalQueue` parameter to avoid confusion with the deprecated `init(queue:, name:)` definition.

neilpa:
> I had to specify a priority or qos in order to "break the tie" between the two constructor definitions. Not sure whether that's a big deal.

Lets remove the overload taking `priority` now that there's the `qos` variant. The former behavior can still be achieved directly with the deprecated `queue` function. You may even be able to use `obsoleted` to point people in that direction.

liscio:
I'm OK with this, however I feel it's a bit heavy-handed to outright remove/obsolete it right away, and could result in unnecessary work for the consumer of the API. 

Consider this: A user is building existing code that still targets OSX 10.9, and uses the `priority` based API. Removing the function means that they would have to modify their code to use the dispatch_queue_t equivalent just to _compile_ their app right away.

I'd offer that a more graceful transition would be to obsolete the function only in the next RAC release after giving them a warning in RAC 4, and outright remove it all in RAC 5.

Granted, if you folks feel that RAC's approach _should_ be more heavy-handed, I'm OK with that as well. I just feel that consumers using RAC 3 right now will already have a lot of work ahead of them when porting to RAC 4 as it is…

neilpa:
This is a very low impact breaking change compared to the rest of the breaking changes already happening between 3 and 4.

neilpa:
Also, RAC has historically supported _current OS - 1_ and 10.9 isn't going to be that much longer.

liscio:
> This is a very low impact breaking change compared to the rest of the breaking changes already happening between 3 and 4. I don't want to carry forward deprecated APIs if it can be avoided.

Cool with me. 

> Also, RAC has historically supported current OS - 1 and 10.9 isn't going to be that much longer.

Would RAC 4 be the time to axe 10.9 support, given that 10.11 will be considered current in a few weeks' time? If that's the case, we can go back to before I even added any availability macros/etc, and simplify the API/code even more. :P

neilpa:
> Would RAC 4 be the time to axe 10.9 support

Since I do almost all iOS dev I'll let someone else with more context on the Mac side make that call.

liscio:
I think this combination of deprecation and obsolescence seems to work well. Folks that require targeting back to OS X 10.9 will get a gentle warning about the init(queue:) API, and once OS X 10.10 is targeted that warning turns into an error.

The side-effect of this change is that the Mac test target will fire errors for the deprecated API use. This happens because the test target has a minimum deployment target of 10.9, and I use availability macros to ensure that the available API is used instead.

neilpa:
Awesome. Thanks for putting the extra thought into this one.

robrix:
@liscio: This is an excellent contribution. Thank you!

mhuusko5:
I'm obviously very late to this convo, alas, the warning->error just kicked in/I noticed...

What should users using dispatch queues for synchronisation do now?

andersio:
`QueueScheduler.queue` exposes the underlying queue (Swift 3.0+?). As for constructing a scheduler that targets an arbitrary queue, perhaps @liscio would know better than me.

liscio:
> What should users using dispatch queues for synchronisation do now?

@mhuusko5 I'm not sure what the question is. What is the pattern that you're no longer able to implement?

mhuusko5:
@liscio creating a scheduler that targets an existing queue. My framework two dispatch queues, one for synchronisation, and one for response to the client. I am no longer to represent dispatching to these queues with QueueScheduler.

liscio:
@mhuusko5 Yeah, this makes sense. See the referenced pull request, and my related fork if you'd like to try this new API before it's merged.

mhuusko5:
@liscio perfect! Thanks

PR #2299, opened 2015-08-29
Use range based looping

Nirma:
Thought this way of writing the loops was a bit more modern. 

Hope it helps!

### DIFF
--- a/ReactiveCocoa/Swift/Bag.swift
+++ b/ReactiveCocoa/Swift/Bag.swift
@@ -46,7 +46,7 @@ internal struct Bag<T> {
 	mutating func removeValueForToken(token: RemovalToken) {
 		if let identifier = token.identifier {
 			// Removal is more likely for recent objects than old ones.
-			for i in reverse(0..<elements.count) {
+			for i in reverse(0..<elements.endIndex) {
 				if elements[i].identifier == identifier {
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/Swift/Bag.swift
```diff
@@ -60,7 +60,7 @@ internal struct Bag<T> {
 	/// will reset all current identifiers to reclaim a contiguous set of
 	/// available identifiers for the future.
 	private mutating func reindex() {
-		for var i = 0; i < elements.count; i++ {
+		for i in 0..<elements.count {
```
NachoSoto: This is already better, but there's an even safer way of creating this `Range`: using [`indices`](http://swiftdoc.org/func/indices/).

Nirma: @NachoSoto Thanks for the tip! I swapped out count with endIndex. 
For the startIndex I think leaving that as 0 is fine but endIndex is much better than what I had before with using count.

NachoSoto: It's only fine because it's an array, though `indices` would be more correct.
But sure, this is okay.

NachoSoto:
Failing test is unrelated

PR #2303, opened 2015-08-30
Use optional chaining in place of nil check

Nirma:
Thought this may be a more modern way of handling this logic piece. 

Hope it helps!


NachoSoto:
Thanks! This was probably like this due to an old compiler crash.

NachoSoto:
(Failing test is unrelated)

PR #2304, opened 2015-08-31
modern swift fix

Nirma:
I believe this is a cleaner way to write this.

Hope it helps!

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -21,7 +21,7 @@ internal final class Atomic<T> {
 			return v
 		}
 	
-		set(newValue) {
+		set {
 			lock()
 			_value = newValue
 			unlock()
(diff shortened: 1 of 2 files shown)


NachoSoto:
Yup, `newValue` is unnecessary, but I wonder if this was meant for clarity?

Nirma:
@NachoSoto Yeah it is a bit of a puzzle as to why it was written since it isnt needed.
I think deleting it will make the codebase more uniform but thats just my two cents.

What do you think?

neilpa:
I prefer to still specify `newValue` for setters because it's a "magic" name but doesn't look like it (e.g I wish the language had used `$0` for this like with closures).

NachoSoto:
Now that you say it @neilpa, I've been bitten by using the propety name instead of `newValue` so many times, that I think being explicit here is better :+1: 

Thanks for the PR though @Nirma!

Issue #2311, opened 2015-09-03
Swift 2 sink type changes behavior of calling start with trailing closure


sebreh:
In Swift 2 (and the *swift2* branch of this repo), the `SinkOf<T>` type is gone and instead a normal closure `T -> ()` is used. This is great but it introduces a potential risk due to the behaviour change when calling SignalProducer<T,E>.start() with a trailing closure. In Swift 1.x, the trailing closure would default to the `next` argument, and in order to provide a sink you would need to pass in a `SinkOf<Event<T>>`. In Swift 2 however, the trailing closure will match the new sink type `T -> ()`. From the call site, the call looks identical after migrating to the swift2 branch but the behaviour is different. This could potentially introduce issues in cases where the argument to start is ignored (`_`), as the type checker will not catch it and the closure will be called on *next/interrupted/error/completed* and not just on *next*.

Example:

Swift 1.x:

```swift
producer |> start { _ in
	// Block only called for next events, `_` is of type Bool
}
```

Swift 2:

```swift
producer.start { _ in
	// Block is called for next/completed/interrupted/error events, `_` is of type (Event<Bool, NoError>)
}
```

You could just argue to "deal with it", but I was considering to just re-add a `SinkOf<T>` type to wrap the sink closure and call `start(SinkOf { x in ... })`, like this:

```swift
public struct SinkOf<T> {
	let sink: T -> ()

	init(_ sink: T -> ()) {
		self.sink = sink
	}
}

SignalProducer<Bool, NoError>(value: true).start(SinkOf { e in
    // e is an Event<Bool, NoError>
})
```

It would also allow you to continue using `producer.start { next in ... }` without having to specify explicitly type of the `next` parameter to help the type system find the right `start` function.

What do you think? I'd be happy to submit a PR if you agree this needs fixing.

neilpa:
Yes, this needs to be fixed. However, I'd like to find a way that doesn't require bringing back `SinkType`.

sebreh:
I can see that. You could have a different function name, e.g. `startWithSink(sink: T-())`. Although that would make the API biased towards using the shorter and cleaner `start` with closures. At the same time, `startWithSink(sink: T-())` is not that much heavier than the `start(sink: SinkOf<T>)` we had before.

neilpa:
[...]

PR #2318, opened 2015-09-06
Replace start/observe closure syntax with distinct functions

sebreh:
Issues #2311 contains the background discussion for this change.

In Swift 2 (and the swift2 branch of this repo), the `SinkOf<T>` type is gone and instead a normal closure `T -> ()` is used. This is great but it introduces a potential risk due to the behaviour change when calling `SignalProducer.start()` with a trailing closure. In Swift 1.x, the trailing closure would default to the next argument, and in order to provide a sink you would need to pass in a `SinkOf<Event<T>>`. In Swift 2 however, the trailing closure will match the new sink type `T -> ()`. From the call site, the call looks identical after migrating to the swift2 branch but the behaviour is different. This could potentially introduce issues in cases where the argument to start is ignored `(_)`, as the type checker will not catch it and the closure will be called on next/interrupted/error/completed and not just on next.

Since Box is no longer needed with Swift 2, this PR replaces the `start(next:_:completed:_:error:_:interrupted)` syntax with only one `start` and `observe` that takes a sink when handling multiple cases, as well as separate functions `startNext()`, `startCompleted()`, `startError()`, `startInterrupted()` for handling a single case.

### DIFF
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -94,9 +94,9 @@ public func toRACSignal<T: AnyObject, E>(producer: SignalProducer<T?, E>) -> RAC
 	return RACSignal.createSignal { subscriber in
 		let selfDisposable = producer.start { event in
 			switch event {
-			case .Next(let value):
+			case let .Next(value):
 				subscriber.sendNext(value)
-			case .Error(let error):
+			case let .Error(error):
 				subscriber.sendError(error as NSError)
 			case .Completed:
 				subscriber.sendCompleted()
@@ -125,9 +125,9 @@ public func toRACSignal<T: AnyObject, E>(signal: Signal<T?, E>) -> RACSignal {
 	return RACSignal.createSignal { subscriber in
 		let selfDisposable = signal.observe { event in
 			switch event {
-			case .Next(let value):
+			case let .Next(value):
 				subscriber.sendNext(value)
-			case .Error(let error):
+			case let .Error(error):
 				subscriber.sendError(error as NSError)
 			case .Completed:
 				subscriber.sendCompleted()
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -92,13 +92,18 @@ public func toRACSignal<T: AnyObject, E>(producer: SignalProducer<T, E>) -> RACS
 /// Any `Interrupted` events will be silently discarded.
 public func toRACSignal<T: AnyObject, E>(producer: SignalProducer<T?, E>) -> RACSignal {
 	return RACSignal.createSignal { subscriber in
-		let selfDisposable = producer.start(next: { value in
-			subscriber.sendNext(value)
-		}, error: { error in
-			subscriber.sendError(error as NSError)
-		}, completed: {
-			subscriber.sendCompleted()
-		})
+		let selfDisposable = producer.start { event in
+			switch event {
+			case .Next(let value):
+				subscriber.sendNext(value)
+			case .Error(let error):
+				subscriber.sendError(error as NSError)
+			case .Completed:
+				subscriber.sendCompleted()
+			default:
+				break
+			}
+		}
 
 		return RACDisposable {
 			selfDisposable.dispose()
@@ -118,14 +123,19 @@ public func toRACSignal<T: AnyObject, E>(signal: Signal<T, E>) -> RACSignal {
 /// Any `Interrupted` event will be silently discarded.
 public func toRACSignal<T: AnyObject, E>(signal: Signal<T?, E>) -> RACSignal {
 	return RACSignal.createSignal { subscriber in
-		let selfDisposable = signal.observe(next: { value in
-			subscriber.sendNext(value)
-		}, error: { error in
-			subscriber.sendError(error as NSError)
-		}, completed: {
-			subscriber.sendCompleted()
-		})
-
+		let selfDisposable = signal.observe { event in
+			switch event {
+			case .Next(let value):
+				subscriber.sendNext(value)
+			case .Error(let error):
+				subscriber.sendError(error as NSError)
+			case .Completed:
+				subscriber.sendCompleted()
+			default:
+				break
+			}
+		}
+		
 		return RACDisposable {
 			selfDisposable?.dispose()
 		}
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -328,7 +328,7 @@ private final class CombineLatestState<T> {
 private func observeWithStates<T, U, E>(signal: Signal<T, E>, _ signalState: CombineLatestState<T>, _ otherState: CombineLatestState<U>, _ lock: NSLock, _ onBothNext: () -> (), _ onError: E -> (), _ onBothCompleted: () -> (), _ onInterrupted: () -> ()) -> Disposable? {
 	return signal.observe { event in
 		switch event {
-		case .Next(let value):
+		case let .Next(value):
 			lock.lock()
 			
 			signalState.latestValue = value
@@ -337,7 +337,7 @@ private func observeWithStates<T, U, E>(signal: Signal<T, E>, _ signalState: Com
 			}
 			
 			lock.unlock()
-		case .Error(let error):
+		case let .Error(error):
 			onError(error)
 		case .Completed:
 			lock.lock()
@@ -520,12 +520,12 @@ extension SignalType {
 
 			disposable += self.observe { event in
 				switch event {
-				case .Next(let value):
+				case let .Next(value):
 					state.modify { (var st) in
 						st.latestValue = value
 						return st
 					}
-				case .Error(let error):
+				case let .Error(error):
 					sendError(observer, error)
 				case .Completed:
 					let oldState = state.modify { (var st) in
@@ -729,7 +729,7 @@ extension SignalType {
 
 			return self.observe { event in
 				switch event {
-				case .Next(let value):
+				case let .Next(value):
 					// To avoid exceeding the reserved capacity of the buffer, we remove then add.
 					// Remove elements until we have room to add one more.
 					while (buffer.count + 1) > count {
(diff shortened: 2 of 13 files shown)


sebreh:
You could also make the case that `startNext(T -> ())` should also allow optional completion and error closures, i.e. make it `startNext(next: T -> (), completion: () -> (), error: E -> (), interrupted: () -> ())`, but I am not sure how heavily the RAC team wants to encourage using the `Event<T, E>` type in handlers.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -92,13 +92,18 @@ public func toRACSignal<T: AnyObject, E>(producer: SignalProducer<T, E>) -> RACS
...
-		})
+		let selfDisposable = producer.start { event in
+			switch event {
+			case .Next(let value):
+				subscriber.sendNext(value)
+			case .Error(let error):
+				subscriber.sendError(error as NSError)
+			case .Completed:
+				subscriber.sendCompleted()
+			default:
```
NachoSoto: I like this PR for this line right here: users will be aware of their need to explicitly ignore events, which wasn't the case before due to the default arguments!

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -163,15 +163,45 @@ extension Signal: SignalType {
...
+	}
+	
+	/// Observes the Signal by invoking the given callback when an `interrupted` event is
+	/// received. If the Signal has already terminated, the callback will be invoked 
+	/// immediately.
+	///
+	/// Returns a Disposable which can be used to stop the invocation of the
+	/// callback. Disposing of the Disposable will have no effect on the Signal
+	/// itself.
+	public func observeInterrupted(interrupted: () -> ()) -> Disposable? {
```
NachoSoto: I like these names

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -311,13 +311,43 @@ extension SignalProducerType {
...
+		return start(Event.sink(next: next))
+	}
+
+	/// Creates a Signal from the producer, then adds exactly one observer to
+	/// the Signal, which will invoke the given callback when an `error` event is
+	/// received.
+	///
+	/// Returns a Disposable which can be used to interrupt the work associated
+	/// with the Signal.
+	public func startError(error: E -> ()) -> Disposable {
```
NachoSoto: These names however are a little bit weird: `startError`, `startCompleted`... it sounds like you're "starting an error signal" or "starting a completed signal".

sebreh: I agree the naming is weird, and I think your suggestion is better. I think it is still a bit awkward to distinguish them by forcing you to name the first parameter, but I don't have a better suggestion really :)

ReactiveCocoa/Swift/ObjectiveCBridging.swift
sebreh: It is definitely not as compact on the caller side, but its probably worth it to be more explicit.

ReactiveCocoa/Swift/SignalProducer.swift
neilpa: The whole point of calling them `startNext`, etc. was to avoid the overload ambiguity while still allowing trailing closure syntax. Forcing you to name the params is going to break that and force an extra set of parens. I'm open to better names, but I think the `observe*` ones are good and am ok with the `start*` variants even if there not perfect it if means trailing closures and consistency.

sebreh: What about `startWithNext`, `startWithCompleted` etc?

neilpa:
> You could also make the case that startNext should also allow optional completion and error closures

Interesting idea. If you only provide a `next` block do trailing closures still work? If so, I'm not even sure we'd want the other `startError`, etc. methods.

And for the sake of comparison, the different ways for dealing with 2 event cases:

```swift
// startNext:completed:
producer.startNext(next: { value in
    print("value: \(value)")
}, completed: { _ in
    print("completed")
})

// start switch
producer.start {
    switch $0 {
    case let .Next(value):
        print("value: \(value)")
    case let .Completed:
        print("completed")
    default:
        break
    }
}

// start if-let
producer.start { event in
    if case let .Next(value) = event {
        print("value: \(value)")
    } else if case let .Completed = event {
        print("completed")
    }
}
```

ReactiveCocoa/Swift/SignalProducer.swift
neilpa: That's probably better since these are essentially wrappers for `startWithSignal` attaching the provided sink as the observer.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -92,13 +92,18 @@ public func toRACSignal<T: AnyObject, E>(producer: SignalProducer<T, E>) -> RACS
...
-		let selfDisposable = producer.start(next: { value in
-			subscriber.sendNext(value)
-		}, error: { error in
-			subscriber.sendError(error as NSError)
-		}, completed: {
-			subscriber.sendCompleted()
-		})
+		let selfDisposable = producer.start { event in
+			switch event {
+			case .Next(let value):
```
neilpa: I _think_ most of the RAC code does `case let .Next(value):` instead of moving the `let` inside the case name.

sebreh:
Yes, the trailing closure works. It would make it basically identical to the previous closure based `start` behavior, but without the naming ambiguity. I personally like having the option of only providing closures as I think it makes for cleaner code at the call site, but I was unsure if you guys wanted to force people to use the event based `start()` to make sure all cases are explicitly handled.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
sebreh: Good catch, I'll change it.

neilpa:
I tend to agree after looking at the different approaches side by side. My proposal would be to have 3 convenience methods (don't think interrupted is needed).

```swift
startWithNext(next: T -> (), error: (E -> ())? = nil, completed: (() -> ())? = nil, interrupted: (() -> ())? = nil)
startWithError(error: E -> (), completed: (() -> ())? = nil, interrupted: (() -> ())? = nil)
startWithCompleted(completed: () -> (), error: (E -> ())? = nil, interrupted: (() -> ())? = nil)
// observeNext, etc. too
```
The `error` param could be removed from `startWithCompleted` (or vice-versa) but having them both enables either order at the call-site.

>  was unsure if you guys wanted to force people to use the event based start()

Nope, since you can still use `switch` with `default` or `if case`.

sebreh:
I agree that API would make the most sense :+1:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -316,8 +316,28 @@ extension SignalProducerType {
 	///
 	/// Returns a Disposable which can be used to interrupt the work associated
 	/// with the Signal, and prevent any future callbacks from being invoked.
-	public func start(error error: (E -> ())? = nil, completed: (() -> ())? = nil, interrupted: (() -> ())? = nil, next: (T -> ())? = nil) -> Disposable {
-		return start(Event.sink(next: next, error: error, completed: completed, interrupted: interrupted))
+	public func startWithNext(error error: (E -> ())? = nil, completed: (() -> ())? = nil, interrupted: (() -> ())? = nil, next: (T -> ())? = nil) -> Disposable {
```
neilpa: I didn't realize that the `next` param was going to be optional and have to be last. This is what I meant if the trailing closure syntax would still work in the case where it was first but the only non-optional param. If that's not possible, I don't think this is the right change and is potentially more confusing. At this point it may be back to the individual `startWithNext`, `startWithError`, etc. approach and no optional params.

neilpa: Although, I would sacrifice the trailing closure syntax to require the associated closure be the first param and required.

```swift
// this still seems reasonable to me ...
producer.startWithNext({
    println("Next \($0)")
})
```

sebreh: Hmm. The next-closure doesn't have to be optional but then the ordering will matter (e.g. `completed` must precede `next`), so that is not an option IMO. It does have to be the last argument to support the trailing closure.

I'm not sure. I think keeping it clean at the call site could be worth an somewhat un-intuitive argument ordering. This was the same style used with the previous overloading of `start`, but I agree it gets more confusing with the new `startWithNext` naming.

Looks like we have the following options:

1.
```swift 
producer.startWithNext({
    println("Next \($0)")
})

producer.startWithNext({
    println("Next \($0)")
  }, 
  error: {
    println("Error \($0)")
})
```

2.
```swift
producer.startWithNext {
    println("Next \($0)")
}

producer.startWithNext(
  next: {
    println("Next \($0)")
  }, 
  error: {
    println("Error \($0)")
})
```

3.
```swift 
producer.startWithNext {
    println("Next \($0)")
}

producer.startWithError {
    println("Error \($0)")
}
```

neilpa: > I'm not sure. I think keeping it clean at the call site could be worth an somewhat un-intuitive argument ordering.

This is really why I don't like (2) because it doesn't work when you have more than one closure argument. I think I'm back to the (3) camp with single closure argument and use `start` if you want to handle multiple.

sebreh: After thinking about it a bit more, I tend to agree it might be the least confusing option. Time will tell if it would be too cumbersome to use.

So if we agree on this, I'll go ahead and revert my last commit and go back to the original suggestion.

neilpa: :+1:

sebreh:
@neilpa that should be it. Also change the "case let" style to be consistent across the code base.

neilpa:
Awesome! Thanks for putting all the time/discussion into this one.

robertjpayne:
I didn't reallse see this convo but I'd like to say the new API in my opinion is inferior to the old API because there is no way to chain the subscriptions and now if we *ever* have to listen in on more than one event type we have to run a ugly as switch statement.

I'd much much much prefer to have the optional closures back.

robertjpayne:
Actually maybe it's not too bad having the switch statement… guess it depends on how much logic has to go into these blocks and suppose it should probably be minimal.

PR #2320, opened 2015-09-08
SignalProducer.merge: checking if event isTerminating outside of switch

NachoSoto:
I noticed that we were checking `event.isTerminating` inside of the `switch`, but only for `.Completed` and `.Interrupted`, missing `.Error`!

I also simplified the definition of this function.

*NOTE: I haven't been able to run tests because `carthage update` is failing at the moment: something about one of the `Cartfile.private` frameworks being built for `watchOS`. I'll take a look at that tomorrow.*

### DIFF
--- a/ReactiveCocoa/Swift/Event.swift
+++ b/ReactiveCocoa/Swift/Event.swift
@@ -36,13 +36,7 @@ public enum Event<T, E: ErrorType> {
 		case .Next:
 			return false
 
-		case .Error:
-			return true
-
-		case .Completed:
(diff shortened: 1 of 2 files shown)


neilpa:
Is there a case where this actually matters? Errors on the inner signal should propagate out and eventually tear down the entire `merge`d signal.

neilpa:
If anything, the `isTerminating` check is unnecessary and could be removed since `Completed` and `Interrupted` both satisfy that by default.

PR #2328, opened 2015-09-09
Don't lose the concatted disposable.

joshaber:
If the source signal completes synchronously (e.g., `-startWith:`), then we overwrite the disposable for the new signal (`concattedDisposable`) with the disposable for the already completed source signal.

This means we end up with disposables that are never disposed of, even if the subscription is disposed of.

I tried to write a test for this, but it’s Really Tricky® as the disposables we leak are down in RAC’s innards.

### DIFF
--- a/ReactiveCocoa/Objective-C/RACSignal.m
+++ b/ReactiveCocoa/Objective-C/RACSignal.m
@@ -186,7 +186,7 @@
 
 - (RACSignal *)concat:(RACSignal *)signal {
 	return [[RACSignal createSignal:^(id<RACSubscriber> subscriber) {
-		RACSerialDisposable *serialDisposable = [[RACSerialDisposable alloc] init];
+		RACCompoundDisposable *compoundDisposable = [[RACCompoundDisposable alloc] init];
 
 		RACDisposable *sourceDisposable = [self subscribeNext:^(id x) {
 			[subscriber sendNext:x];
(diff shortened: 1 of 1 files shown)


joshaber:
>I tried to write a test for this, but it’s Really Tricky® as the disposables we leak are down in RAC’s innards.

(That said, I did verify this with a small sample app using Instruments.)

neilpa:
> (That said, I did verify this with a small sample app using Instruments.)

Might be worthwhile to include this app in the project for cases where unit testing is prohibitively hard.

joshaber:
I'm happy to include it, but frankly I doubt it'd see any use beyond this PR.

For posterity:

```objc
@interface AppDelegate ()

@property (weak) IBOutlet NSWindow *window;
@property (nonatomic, strong) RACSubject *subject1;
@property (nonatomic, strong) RACDisposable *disposable;

@end

@implementation AppDelegate

- (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
	_subject1 = [RACSubject new];
}

- (IBAction)clicky:(id)sender {
	self.disposable = [[self.subject1
		startWith:nil]
		subscribeNext:^(id x) {

		}];
}

- (IBAction)stop:(id)sender {
	[self.disposable dispose];
	self.disposable = nil;
}

@end
```

And then use the Allocation's instrument's `Mark Generation` to measure what sticks around after clicking start and stop.

neilpa:
Works for me now that we have a record of it somewhere.

PR #2333, opened 2015-09-10
Added support for Apple TV

NachoSoto:
### Changes:

- Changed `xcconfig` reference to include https://github.com/jspahrsummers/xcconfigs/pull/42.
- Changed `Result` reference to include https://github.com/antitypical/Result/pull/80.
- Created framework target (made sure it uses the same `plist` and `product name`).
- Created scheme for new target.

### TODO:

- [x] Update `antitypical/Result` to include `tvOS` target. *[Done](https://github.com/antitypical/Result/pull/80)*.
- [x] Update `xcconfig` reference to `master` once https://github.com/jspahrsummers/xcconfigs/pull/42 is merged.
- [x] Update `Result` reference to `master` (or a release) once https://github.com/antitypical/Result/pull/80 is merged.
- [x] Work around http://www.openradar.me/22639221? The framework fails to compile right now because `TARGET_OS_IPHONE` is defined when compiling the `AppleTV` target, so it tries to include an [iOS header](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift2/ReactiveCocoa/ReactiveCocoa.h#L66).

Related: https://github.com/Carthage/Carthage/pull/742

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,4 +9,4 @@
 	url = https://github.com/nachosoto/xcconfigs.git
 [submodule "Carthage/Checkouts/Result"]
 	path = Carthage/Checkouts/Result
-	url = https://github.com/antitypical/Result.git
+	url = https://github.com/nachosoto/Result.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/Quick/Quick.git
 [submodule "Carthage/Checkouts/xcconfigs"]
 	path = Carthage/Checkouts/xcconfigs
-	url = https://github.com/jspahrsummers/xcconfigs.git
+	url = https://github.com/nachosoto/xcconfigs.git
 [submodule "Carthage/Checkouts/Result"]
 	path = Carthage/Checkouts/Result
 	url = https://github.com/antitypical/Result.git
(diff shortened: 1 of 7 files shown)


NachoSoto:
Thanks!

ashfurrow:
Pardon my ignorance, but I've got a question about the TV support. As far as I can tell, building ReactiveCocoa using Carthage on the current `swift2` branch requires the use of Xcode 7.1 beta, right? Since Xcode 7 GM doesn't recognize the tvOS target, it fails to build under Carthage. So that would mean no one can currently submit apps targeting iOS 9, since that requires to GM. Is there something I'm missing?

NachoSoto:
> Pardon my ignorance, but I've got a question about the TV support. As far as I can tell, building ReactiveCocoa using Carthage on the current swift2 branch requires the use of Xcode 7.1 beta, right? 
> Since Xcode 7 GM doesn't recognize the tvOS target, it fails to build under Carthage

Only to build the `tvos` framework. Otherwise, you can build the other 3 with 7.0 GM.

> So that would mean no one can currently submit apps targeting iOS 9, since that requires to GM. Is there something I'm missing?

You can simply filter the architectures you build using `--platform`. You almost never want to build everything **four** times anyway.

ashfurrow:
Ok, that makes sense. I didn't see the --platform option anywhere in the Carthage docs.

Issue #2327, opened 2015-09-09
RAC3 = Swift 1.2 and RAC4 = Swift 2


neilpa:
~~Having not shipped v3.0 final and assuming Xcode 7 GM's today, I propose changing our target to Swift 2. We can keep around the existing `RC.1` release and move that code to a `swift-1.2` branch. When we're ready to cut an `RC.2` we can merge `swift2` into `master` (_should_ be soon™). If we need to backport fixes to 1.2 we can always release an `RC.1.1` w/out breaking semver.~~

It's going to be cleaner if we just split major versions of RAC along the Swift version boundary. So RAC3 will target Swift 1.2 and RAC4 will target Swift 2. Going forward we'll only be backporting fixes to RAC3 on demand and future development will be on RAC4.

@ReactiveCocoa/reactivecocoa Thoughts?

NachoSoto:
I'm in. I don't think any of us will have much time to maintain the Swift 1.2 version anyway.

neilpa:
Also, the docs are mostly in sync with the changes in that branch. The only thing I know that's missing is updating the changelog for 1.2 -> 2.0.

ikesyo:
:+1: for the direction, and also we should use lowercase `rc` for tagging in favor of https://github.com/CocoaPods/cocoapods.org/issues/185#issuecomment-123419140.

kastiglione:
Why not? :+1:

brow:
We did [tell users](https://github.com/ReactiveCocoa/ReactiveCocoa/releases/tag/v3.0-RC.1):

> With this release candidate, RAC 3.0 is API stable, and all of our desired documentation is in place. Please report any bugs you may find!

> Breaking changes may still be made if absolutely necessary, but—for all intents and purposes—this is the API we intend to ship.

If I may be obtuse, what's the downside of incrementing the major version to 4? I agree that people probably wouldn't find time to maintain RAC 3.

neilpa:
> If I may be obtuse, what's the downside of incrementing the major version to 4?

I'm not against this. Although it may be confusing to release v3 and turnaround a week later with a v4-alpha release. In practice, we're effectively doing the same thing. It's just a matter of labeling the releases and communicating that appropriately.

brow:
I suppose I don't lean either way; both approaches seem slightly confusing to humans, and neither approach is confusing to dependency managers. So, :+1: for incrementing the RC number.

neilpa:
[...]

PR #2334, opened 2015-09-10
Don't use KVO in RACCommand

mdiep:
`RACCommand` has always been slow/expensive to initialize. Due to some absurd slowness in our app's unit tests, I started profiling and found that in our tests we were spending a lot of time rehashing inside KVO. 

This PR changes `RACCommand` so that it no longer uses KVO, saving on initialization and disposal. The changes here make our app's unit tests run ~15% faster. (We create a lot of `RACCommand`s.) I'm sure the savings won't be as noticeable in the real world, but I think these changes should still improve things.

### DIFF
--- a/ReactiveCocoa/Objective-C/RACCommand.m
+++ b/ReactiveCocoa/Objective-C/RACCommand.m
@@ -54,6 +54,8 @@ const NSInteger RACCommandErrorNotEnabled = 1;
 }
 
 - (void)setAllowsConcurrentExecution:(BOOL)allowed {
+	[self willChangeValueForKey:@keypath(self.allowsConcurrentExecution)];
+	
 	if (allowed) {
 		OSAtomicOr32Barrier(1, &_allowsConcurrentExecution);
 	} else {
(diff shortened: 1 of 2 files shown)


joshaber:
This _seems_ legit to me, but I'd love to have a second set of :eyes:.

neilpa:
Haven't looked at the code closely yet, but this would technically be a breaking change for RAC3. This should probably be targeted at the `swift2` branch for now (which I may rename `RAC4` depending on the outcome of #2327).

ashfurrow:
Personally, it feels strange to remove a feature – possibly a breaking change – solely for the sake of unit test run duration.

mdiep:
How is this a breaking change? It changes the implementation, but maintains the current behavior and interface.

ashfurrow:
`allowsConcurrentExecution` is public property that someone may be observing. It's unlikely, given that there's a subject to use as well, but it seems like the kind of subtle change that could drive someone nuts trying to figure out why their code doesn't work anymore.

mdiep:
> `allowsConcurrentExecution` is public property that someone may be observing. It's unlikely, given that there's a subject to use as well, but it seems like the kind of subtle change that could drive someone nuts trying to figure out why their code doesn't work anymore.

Ah, okay. I just added that back (with a test). That's not an important part of the PR—the important part is not using `RACObserve` internally.

ashfurrow:
Coooool – looks good :100: 

I'm not the most judicious code-reviewer, though. Someone else should double-check juuuust to be sure.

neilpa:
This _looks_ right to me to so here goes nothing

mdiep:
:sparkling_heart:

PR #2337, opened 2015-09-11
Fix watchOS compilation due to NSURLConnection

alessandroorru:
`+[NSURLConnection sendAsynchronousRequest:queue:completionHandler:]` is deprecated on iOS9 and unavailable on watchOS, so watchOS framework fails to build. I think that NSURLConnection support should be directly removed from ReactiveCocoa, since the deprecation.

In case you prefer to keep it for retrocompatibility, I fixed it by replacing it with `-[NSURLSession dataTaskWithRequest:completionHandler:]` method. It isn't a clean solution since this is a category on NSURLConnection, anyway.

By the way, since there are differences between iOS SDK and watchOS SDK, I think that it should be better to have a dedicated unit test target: this problem for example would have been discovered instantly.

### DIFF
--- a/ReactiveCocoa/Objective-C/NSURLConnection+RACSupport.m
+++ b/ReactiveCocoa/Objective-C/NSURLConnection+RACSupport.m
@@ -16,39 +16,37 @@
 
 + (RACSignal *)rac_sendAsynchronousRequest:(NSURLRequest *)request {
 	NSCParameterAssert(request != nil);
-
+	
 	return [[RACSignal
-		createSignal:^(id<RACSubscriber> subscriber) {
-			NSOperationQueue *queue = [[NSOperationQueue alloc] init];
-			queue.name = @"org.reactivecocoa.ReactiveCocoa.NSURLConnectionRACSupport";
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-			[NSURLConnection sendAsynchronousRequest:request queue:queue completionHandler:^(NSURLResponse *response, NSData *data, NSError *error) {
-				// The docs say that `nil` data means an error occurred, but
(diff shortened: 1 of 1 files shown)


alessandroorru:
Anyway, I think that it should be even better to totally remove NSURLConnection support from ReactiveCocoa, since the deprecation.

alessandroorru:
This is another possibility #2339

ReactiveCocoa/Objective-C/NSURLConnection+RACSupport.m
```diff
@@ -16,39 +16,37 @@ @implementation NSURLConnection (RACSupport)
...
-				// but we can at least prevent _some_ unnecessary work --
-				// without writing all the code for a proper delegate, which
-				// doesn't really belong in RAC.
-				queue.suspended = YES;
-				[queue cancelAllOperations];
-			}];
-		}]
-		setNameWithFormat:@"+rac_sendAsynchronousRequest: %@", request];
+			 createSignal:^(id<RACSubscriber> subscriber) {
+				 NSURLSessionDataTask * task = [[NSURLSession sharedSession] dataTaskWithRequest:request completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
```
NachoSoto: This is still in `NSURLConnection+RACSupport.m`, so we'd have to move.

However, I'd say this is legacy code. I would stick to #2339 (minus changing `EMBEDDED_CONTENT_CONTAINS_SWIFT`), and if needed we can always add another extension for `NSURLSession` in Objective-C. We [already have one for Swift](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/Swift/FoundationExtensions.swift#L29-L50).

NachoSoto:
Thanks for the fix though!

PR #2339, opened 2015-09-11
Removed including of NSURLConnection+RACSupport on watchOS target

marchinram:
Removed from Compile Build Phase and from ReactiveCocoa.h umbrella header when compiling for watchOS

Also needed to set EMBEDDED_CONTENT_CONTAINS_SWIFT = YES or would receive a crash on launch saying that libSwiftCore could not be loaded


NachoSoto:
> Removed from Compile Build Phase and from ReactiveCocoa.h umbrella header when compiling for watchOS

Thanks!

> Also needed to set EMBEDDED_CONTENT_CONTAINS_SWIFT = YES or would receive a crash on launch saying that libSwiftCore could not be loaded

Sorry that's actually not correct. That setting should only be enabled at the application level, not on each framework. But I don't blame you, that's an implementation detail in Xcode that we shouldn't have to deal with.

alessandroorru:
I solved it in another way, take a look and choose the one you prefer #2337

marchinram:
@NachoSoto Thanks for the tip, I removed the build setting

marchinram:
@alessandroorru I would prefer my way because the only code change is in the umbrella header

NachoSoto:
Awesome this looks good :thumbsup:

PR #2345, opened 2015-09-12
Change @warn_unused_result message for blocking operators

stevebrambilla:
This PR removes the "Did you forget to call `start` on the producer?" warnings from the blocking operators that return Results instead of SignalProducers.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1038,7 +1038,6 @@ extension SignalProducerType {
 	}
 
 	/// Starts the producer, then blocks, waiting for the first value.
-	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func first() -> Result<T, E>? {
 		return take(1).single()
 	}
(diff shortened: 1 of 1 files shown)


NachoSoto:
Oh right, the message is wrong in these, but they're still useful, aren't they?

Maybe we can just change the message?

stevebrambilla:
These we're popping up for me since in my tests I'd often have something like this:

```swift
device.open().wait() // Setup code that runs before each test
...
device.someOtherRequest().start() // The actual producer being tested, which depends on the device being open
```

Using `wait()` would warn about the unused result, which was unused because it's not needed here.

But we could keep a warning with a different message for `first` / `single` / `last` since those imply to me that the consumer intends to use the result, but I don't think that `wait` does. What do you think?

NachoSoto:
I'd rather keep the warnings, that way users need to explicitly ignore the result, and be 100% aware that they're doing so:
```swift
_ = device.open.wait()
```

neilpa:
:+1: for keeping the warning on `wait` too. The producer can still error even if you don't care about the value of the result.

neilpa:
As an aside, it may be beneficial to split your `device` into multiple objects where `open` returns a producer of a new type that offers the `someOtherRequest` producer.

stevebrambilla:
Alright sure, let's keep them. How about I change these messages to "Did you forget to check the result?" then?

The `device` / `open` code was _a bit_ of a contrived example, but thanks!

NachoSoto:
> How about I change these messages to "Did you forget to check the result?" then?

👍🏻

neilpa:
Awesome! Thanks for PR.

PR #2347, opened 2015-09-13
Remove redundant import statements

dapenggao:
These frameworks are not used by any code in the files.


neilpa:
Thanks for the PR!

Issue #2212, opened 2015-07-23
Rename `T` and `E` to `Value` and `Error`


neilpa:
Should we follow the lead that the stdlib made with collection changes in beta 3?

NachoSoto:
I wouldn't be against this.

kastiglione:
What about additional value types, `U`, `V`?

neilpa:
I was initially thinking about the event/signal generics. However, we could probably come up with better context dependent names for those as well. The one exception being the `combineLatest`/`zip` overloads.

JaviSoto:
I think this is the type of thing that aids the newcomers. I'm really glad that Apple is doing this for the Swift stdlib as well.
I'm not as worried about the generic types for the rest of the methods, sometimes maybe `V` will be good enough™, but we can always considered names such as `TransformedValue`

PR #2349, opened 2015-09-13
[4.0] Rename `T` to `Value` and `E` to `Error`

bencochran:
This mostly completes #2212. I left the `V` and `F` parameters alone… it’s what the standard library mostly does, and things like `TransformedValue` make for some long function signatures. But I could be convinced otherwise.

`Event` presented a challenge: it already has a static member named `Error` (namely, the `case Error(E)`). So it cannot be generic with `Error` as the parameter name. I left it as `E` for now. Anyone have ideas for a name?


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -722,9 +722,9 @@ extension SignalType {
 	/// Waits until `self` completes and then forwards the final `count` values
 	/// on the returned signal.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func takeLast(count: Int) -> Signal<T, E> {
+	public func takeLast(count: Int) -> Signal<Value, Error> {
 		return Signal { observer in
-			var buffer = [T]()
+			var buffer: [Value] = []
```
bencochran: I snuck in this change to be more stylistically self-consistent.

neilpa:
The only reasonable name I've got is `Err`, but abbreviating like that isn't idiomatic.

Alternatively, the `Error` case could be renamed `Failure`. It would mirror the naming in `Result`, but I would need more convincing about making this change.

bencochran:
Hmm, interesting thought about changing the case name. I could also see `Failed` which mirrors `Completed`/`Interrupted` in its tense (and since those are all terminating events, having the same sense is nice).

neilpa:
:+1: on `Failed` over `Failure` if we go down that route.

bencochran:
Ok, I rebased this to pull in @NachoSoto’s stuff. And I used `Err` in `Event` for now.

I’m going to do a branch with `case Failed` and see how that feels, but I’ll let that be a separate pull request and discussion.

neilpa:
> but I’ll let that be a separate pull request and discussion

:+1:

bencochran:
Opened that thing as a separate pull request targeting the swift2 branch. So this should be good to go on its own.

PR #2354, opened 2015-09-14
Use `if case let` instead of single-case `switch`es

dapenggao:
- It is more idiomatic to use `if case let` rather than single-case `switch`es
- `Optional<T>` is now `Optional<Wrapped>` in the Swift standard library.


ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -20,27 +20,27 @@ internal final class Atomic<T> {
 
 			return v
 		}
-	
-		set(newValue) {
```
NachoSoto: This is on purpose (see #2304)

ReactiveCocoa/Swift/Bag.swift
```diff
@@ -18,7 +18,7 @@ internal final class RemovalToken {
 
 /// An unordered, non-unique collection of values of type T.
 internal struct Bag<T> {
-	private var elements: [BagElement<T>] = []
```
NachoSoto: I think the alternative is clearer: it separates the type definition from the value.

ReactiveCocoa/Swift/Event.swift
```diff
@@ -81,24 +81,22 @@ public enum Event<T, E: ErrorType> {
 
 	/// Unwraps the contained `Next` value.
 	public var value: T? {
-		switch self {
-		case let .Next(value):
+		if case let .Next(value) = self {
```
NachoSoto: Cool

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -15,15 +15,15 @@ extension RACScheduler: DateSchedulerType {
 	}
 
 	public func schedule(action: () -> ()) -> Disposable? {
-		return self.schedule(action)
+		return schedule(action)
```
NachoSoto: I'd rather keep the `self`: it makes it clear that this is an instance property instead of a global, local variable or parameter.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -54,11 +54,11 @@ extension RACSignal {
 	/// each invocation of start().
 	public func toSignalProducer(file: String = __FILE__, line: Int = __LINE__) -> SignalProducer<AnyObject?, NSError> {
 		return SignalProducer { observer, disposable in
-			let next = { (obj: AnyObject?) -> () in
+			let next = { obj in
```
NachoSoto: This were here to workaround compiler crashes, most likely.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -17,7 +17,7 @@ import Result
 public final class Signal<T, E: ErrorType> {
 	public typealias Observer = Event<T, E>.Sink
 
-	private let atomicObservers: Atomic<Bag<Observer>?> = Atomic(Bag())
+	private let atomicObservers = Atomic(Bag<Observer>?())
```
NachoSoto: Hmm `Observer?` isn't here anymore?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -104,7 +103,7 @@ public final class Signal<T, E: ErrorType> {
 
 	/// Interrupts all observers and terminates the stream.
 	private func interrupt() {
-		if let observers = self.atomicObservers.swap(nil) {
+		if let observers = atomicObservers.swap(nil) {
```
NachoSoto: Ditto

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -229,13 +228,9 @@ extension SignalType {
...
 		return Signal { observer in
 			return self.observe { event in
-				switch event {
-				case let .Next(value):
-					if predicate(value) {
-						sendNext(observer, value)
-					}
-
-				default:
+				if case let .Next(value) = event where predicate(value) {
```
NachoSoto: Nice

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -351,7 +346,7 @@ private final class ConcatState<T, E: ErrorType> {
 	let disposable: CompositeDisposable
 
 	/// The active producer, if any, and the producers waiting to be started.
-	let queuedSignalProducers: Atomic<[SignalProducer<T, E>]> = Atomic([])
+	let queuedSignalProducers = Atomic([SignalProducer<T, E>]())
```
NachoSoto: Ditto: this was clearer before

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -568,12 +563,11 @@ private struct LatestState<T, E: ErrorType> {
 	var replacingInnerSignal: Bool = false
 }
 
-// Have to extend `Signal` directly to avoid a compiler crash.
```
NachoSoto: Awesome

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -615,7 +608,7 @@ extension SignalType {
 /// A reference type which wraps an array to avoid copying it for performance and
 /// memory usage optimization.
 private final class CollectState<T> {
-	var values: [T] = []
+	var values = [T]()
```
NachoSoto: Ditto

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -822,9 +811,9 @@ extension Signal where T: EventType, E: NoError {
 }
 
 private struct SampleState<T> {
-	var latestValue: T? = nil
-	var signalCompleted: Bool = false
-	var samplerCompleted: Bool = false
+	var latestValue: T?
+	var signalCompleted = false
+	var samplerCompleted = false
```
NachoSoto: Same in these 3: explicit is good.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -865,10 +854,10 @@ extension SignalType {
 					sendInterrupted(observer)
 				}
 			}
-
```
NachoSoto: Can you try to get all these space changes out of the diff? Looks like you might have a different setting.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1025,11 +1014,9 @@ extension SignalType {
 			let disposable = CompositeDisposable()
 
 			let signalDisposable = self.observe { event in
-				switch event {
-				case .Completed:
-					break
-
-				case .Next, .Error, .Interrupted:
```
NachoSoto: Explicit here is good: if we ever add a `case` to `Event` this will fail to compile, and we'll be forced to make a decision for the new `case`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1100,7 +1081,7 @@ extension SignalType {
 }
 
 private struct ZipState<T> {
-	var values: [T] = []
+	var values = [T]()
```
NachoSoto: Ditto

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1216,9 +1196,9 @@ extension SignalType {
 				case let .Next(value):
 					operation(value).analysis(ifSuccess: { value in
 						sendNext(observer, value)
-						}, ifFailure: { error in
+						}) { error in
```
NachoSoto: This should be explicit, trailing closure syntax doesn't make much sense here

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1290,8 +1269,8 @@ extension SignalType {
 }
 
 private struct ThrottleState<T> {
-	var previousDate: NSDate? = nil
-	var pendingValue: T? = nil
+	var previousDate: NSDate?
+	var pendingValue: T?
```
NachoSoto: Ditto

NachoSoto:
Thanks for the PR! :star:

dapenggao:
@NachoSoto Wow! A lot of helpful feedback! Thanks a lot! Perhaps I'm too obsessed with type inference and other kinds of fancy features. I will fix the mentioned issues as soon as possible. : )

dapenggao:
@NachoSoto And for the spacing changes I'm actually enforcing this repo's coding style guide, which states that empty lines should not be indented.

neilpa:
> I'm actually enforcing this repo's coding style guide, which states that empty lines should not be indented.

:+1: , but it might be better suited as a separate PR since there's already numerous changes here.

dapenggao:
@neilpa I see. This can probably be done tomorrow cuz there are quite a few lines of code to change.

dapenggao:
And there is also this method with a single-case switch and a `fallthrough` that is quite hard to refactor into another simpler form. I wonder if anyone else can come up with an elegant solution. 
```swift
public func skipWhile(predicate: T -> Bool) -> Signal<T, E> {
	return Signal { observer in
		var shouldSkip = true

			switch event {
			case let .Next(value):
				shouldSkip = shouldSkip && predicate(value)
					if !shouldSkip {
						fallthrough
					}

			default:
				observer(event)
			}
		}
	}
}
```

ReactiveCocoa/Swift/Signal.swift
NachoSoto: Bump :)

ReactiveCocoa/Swift/ObjectiveCBridging.swift
dapenggao: @NachoSoto I tested on Xcode 7 GM and the compiler doesn't crash on this now.

dapenggao:
@NachoSoto @neilpa The indentation changes are removed now. I will create another pull request.

ReactiveCocoa/Swift/Disposable.swift
```diff
@@ -164,13 +164,13 @@ public final class ScopedDisposable: Disposable {
...
 	}
 
-	deinit {
-		dispose()
-	}
-
 	public func dispose() {
 		innerDisposable.dispose()
 	}
+
```
NachoSoto: I think this was better before: keeping `init` and `deinit` together.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
NachoSoto: I believe you :D

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -88,7 +88,7 @@ public struct SignalProducer<T, E: ErrorType> {
 
 	/// A producer for a Signal that never sends any events to its observers.
 	public static var never: SignalProducer {
-		return self.init { _ in () }
+		return self.init { _ in  }
```
NachoSoto: Explicit was good here

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -203,8 +203,8 @@ public struct SignalProducer<T, E: ErrorType> {
 			operation().analysis(ifSuccess: { value in
 				sendNext(observer, value)
 				sendCompleted(observer)
-			}, ifFailure: { error in
-				sendError(observer, error)
+				}, ifFailure: { error in
+					sendError(observer, error)
```
NachoSoto: Can you fix this indentation?

NachoSoto:
Cool just a few things left :+1: 

> I will create another pull request.

ReactiveCocoa/Swift/SignalProducer.swift
dapenggao: In this case, I suggest that we simply use `return` rather than an empty tuple

dapenggao:
@NachoSoto I use `Ctrl+I`. Xcode indents empty lines by default, so I have to change a setting to make it  work.

NachoSoto:
What setting?
I'm worried that we're not using default indentation settings, so those lines will get changed again. I think we should favor automatic indentation over anything.

dapenggao:
I tick the `Including whitespace-only lines` option and Xcode will remove indentations on empty lines.
screen shot 2015-09-16 at 10 51 10 am

NachoSoto:
Yeah that's just going to cause problems because nobody has that setting enabled.

NachoSoto:
This looks good now though, thanks! :star:

0xpablo:
Isn't  https://github.com/dapenggao/ReactiveCocoa/commit/07c164f3f9f511f035a7206040c39138785ecf9a against Swift's coding conventions?
Thanks

NachoSoto:
What part @0xpablo ?

PR #2357, opened 2015-09-15
Update changelog for v4 alpha

neilpa:
Almost ready

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,11 @@ error and completed. See #2311 and #2318 for more details.
 
 #### Renamed try and catch operators
 
+The `try` and `catch` operators were renamed because of the addition of the
+error handling keywords with the same name. They are now `attempt` and
+`flatMapError` respectively. Also, `tryMap` was renamed to `attemptMap` for
+consistency.
+
 #### Renaming T and E generic parameters
 
 Coming Soon to a later alpha near you.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ changes](#30). This section only covers the differences when between 3.0 and
 ReactiveCocoa 4.0 targets Swift 2 and the current focus is on leveraging the
 improvements from Swift 1.2 to provide a simpler API.
 
-### Signal operators are protocol extensions
+#### Signal operators are protocol extensions
 
 The biggest change from RAC3 to RAC4 is that signal and producer operators are
 implemented as protocol extensions instead of global functions. This is similar
@@ -32,19 +32,24 @@ source looking for these operators. Now as protocol extensions, the
 `SignalProducer` operators are explicitly implementated in terms of their
 `Signal` counterpart when available.
 
-### Removal of `|>` custom operator
+#### Removal of `|>` custom operator
 
 As already alluded to above, the custom `|>` operator for chaining signals has
 been removed. Instead standard method calling syntax is used for chaining
 operators.
 
-### Replacement of the `start` and `observer` overloads
+#### `Event.Sink` is now a function
 
-### Replacements for `catch` and `try`
+With the removal of `SinkType` in Swift 2, the `Event.Sink` type is now just a
+function `Event -> ()`.
 
-### Renaming `T` and `E` generic parameters (WIP)
+#### Replacements for the `start` and `observer` overloads
 
-Coming Soon but not in the current alpha
+#### Renamed `try` and `catch`
+
+#### Renaming `T` and `E` generic parameters
+
+Coming Soon to a later alpha near you.
 
 
 # 3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,7 +73,7 @@ error handling keywords with the same name. They are now `attempt` and
 `flatMapError` respectively. Also, `tryMap` was renamed to `attemptMap` for
 consistency.
 
-#### Added flatten and flatMap for signals-of-prodcuers
+#### Added flatten and flatMap for signal-of-producers
 
 This fills a gap that was missing in RAC 3. It's a common pattern to have a
 hot `Signal` of values that need to be mapped to "work" -- `SignalProducer`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
-# 4.0 (alpha-1)
+# 4.0 
+
+**The RAC 4 APIs are work in progress**. There may be significant breaking
+changes in later alphas so be prepared for that before taking a dependency.
 
 If you're new to the Swift API and migrating from RAC 2, start with the
 [3.0 changes](#30). This section only covers the differences between 3.0 and
@@ -6,6 +9,8 @@ If you're new to the Swift API and migrating from RAC 2, start with the
 
 ReactiveCocoa 4.0 targets Swift 2 and the current focus is on leveraging the
 improvements from Swift 1.2 to provide a simpler API.
+
+## Alpha 1
 
 #### Signal operators are protocol extensions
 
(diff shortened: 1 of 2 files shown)


ikesyo:
Why Quick and Nimble submodules are removed?

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
+# 4.0 (alpha-1)
+
+If you're new to the Swift API and migrating from RAC2, starting with the [3.0
```
ikesyo: `RAC2` should be  `RAC 2` for consistency in the document.

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
...
+changes](#30). This section only covers the differences when between 3.0 and
+4.0.
+
+ReactiveCocoa 4.0 targets Swift 2 and the current focus is on leveraging the
+improvements from Swift 1.2 to provide a simpler API.
+
+#### Signal operators are protocol extensions
+
+The biggest change from RAC3 to RAC4 is that signal and producer operators are
```
ikesyo: As above.

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
...
+implemented as protocol extensions instead of global functions. This is similar
+to many of the collection protocol changes in the Swift 2 standard library.
+
+This enables chaining signal operators with normal dot-method calling syntax.
+Previously the custom `|>` was required to enable chaining global functions
+without a mess of nested calls and parenthesis.
+
+```swift
+/// RAC3
```
ikesyo: As above.

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
...
+signal |> filter { $0 % 2 == 0 } |> map { $0 * $0 } |> observe { print($0) }
+
+/// RAC4
```
ikesyo: As above.

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
...
+signal.filter { $0 % 2 == 0 } .map { $0 * $0 } .observe { print($0) }
+```
+
+Additionally, this means that `SignalProducer` operators are less "magic". In
+RAC3 the `Signal` operators were implicitly lifted to work on `SignalProducer`
```
ikesyo: As above.

CHANGELOG.md
```diff
@@ -1,3 +1,63 @@
...
+#### Event.Sink is now a function
+
+With the removal of `SinkType` in Swift 2, the `Event.Sink` type is now just a
+function `Event -> ()`.
+
+#### Event cases are no longer boxed
+
+The improvements to associated enum values in Swift 2 mean that `Event` cases
+no longer need to be `Box`ed. In fact, the `Box` dependency has been removed
+completed from RAC4.
```
ikesyo: As above.

neilpa:
> Why Quick and Nimble submodules are removed?

Unintentional.

CHANGELOG.md
ikesyo: `completely`?

neilpa:
Should be good to go

CHANGELOG.md
```diff
@@ -1,3 +1,76 @@
+# 4.0 (alpha-1)
+
+If you're new to the Swift API and migrating from RAC 2, starting with the
```
raylillywhite: `starting` -> `start`

CHANGELOG.md
```diff
@@ -1,3 +1,76 @@
+# 4.0 (alpha-1)
+
+If you're new to the Swift API and migrating from RAC 2, start with the
+[3.0 changes](#30). This section only covers the differences when between 3.0
```
bencochran: Remove “when”

CHANGELOG.md
```diff
@@ -1,3 +1,76 @@
...
+#### Event.Sink is now a function
+
+With the removal of `SinkType` in Swift 2, the `Event.Sink` type is now just a
+function `Event -> ()`.
+
+#### Event cases are no longer boxed
+
+The improvements to associated enum values in Swift 2 mean that `Event` cases
+no longer need to be `Box`ed. In fact, the `Box` dependency has been removed
+completed from RAC 4.
```
bencochran: “…has been removed *completely*…”

CHANGELOG.md
```diff
@@ -1,3 +1,76 @@
...
+
+#### Replacements for the start and observer overloads
+
+The `observe` and `start` overloads taking `next`, `error`, etc. optional
+function parameters have been removed. This was necessitated by the change to
+`Event.Sink` becoming a function type which introduced an unresolvable
+ambiguity. They've been replaced with methods taking a single function with
+the target `Event` case -- `observeNext`, `startWithNext`, and the same for
+error and completed. See #2311 and #2318 for more details.
```
bencochran: Would it be worth putting a before/after example here? Like:
```swift
/// RAC 3
producer
    |> start(next: { println($0) },
        error: { println("Error: \($0)") })

/// RAC 4
producer.start { event in
    if case let .Next(value) = event {
        print(value)
    } else case let .Error(error) = event {
        print("Error:", error)
    }
}
// or
producer
    .onError { print("Error:", $0) }
    .startWithNext { print($0) }
```

neilpa: After attempting to use this new style in practice (upgrading [neilpa/Rex](https://github.com/neilpa/rex)), I'm not a fan. I think we need another pass at this change, potentially bringing back the `start` overload but with `interrupted` as the last param instead of `next`. This is one place where trailing closure syntax is getting in the way.

NachoSoto: I actually think this is cleaner:
```
producer.start { event in 
  switch event {
    case let .Next(value):
      print(value)

    case let .Error(error):
      print("Error: \(error)")

    default:
      break
  }
}
```

CHANGELOG.md
```diff
@@ -1,3 +1,76 @@
+# 4.0 (alpha-1)
+
+If you're new to the Swift API and migrating from RAC 2, start with the
+[3.0 changes](#30). This section only covers the differences between 3.0 and
+4.0.
+
+ReactiveCocoa 4.0 targets Swift 2 and the current focus is on leveraging the
+improvements from Swift 1.2 to provide a simpler API.
+
+#### Signal operators are protocol extensions
```
NachoSoto: I think it would be worth mentioning that we brought back `flatten`ing `Signal`s of `SignalProducer`s?

CHANGELOG.md
neilpa: I think the answer is probably `startWith(next next: T -> () = { _ in }, error: E -> () = { _ in }, completed: () -> () = { _ in }, interrupted: () -> () = { _ in })` to avoid the overload ambiguity. This also limits the possibility of shooting yourself in the foot with trailing closure syntax since you're forced to specify `next` and `interrupted` is the least common. Just not sure what to call the `observe` equivalent since `observeWith` doesn't sound right.

CHANGELOG.md
neilpa: These never actually shipped, but mentioning the addition is good.

NachoSoto: s/brought back/added

CHANGELOG.md
neilpa: I added an explicit disclaimer to the top of this section for now and will open another issue to discuss this further. I don't want to block an initial alpha on this though.

neilpa:
@NachoSoto Assuming this looks good to you go ahead and merge this then push the button on the release.

CHANGELOG.md
```diff
@@ -1,3 +1,94 @@
...
+error and completed. See #2311 and #2318 for more details.
+
+#### Renamed try and catch operators
+
+The `try` and `catch` operators were renamed because of the addition of the
+error handling keywords with the same name. They are now `attempt` and
+`flatMapError` respectively. Also, `tryMap` was renamed to `attemptMap` for
+consistency.
+
+#### Added flatten and flatMap for signals-of-prodcuers
```
NachoSoto: Perfect!

Just a small typo: `producers`.

neilpa: Fixed

NachoSoto:
:tada:

PR #2360, opened 2015-09-15
Event.Error → Event.Failed

bencochran:
As part of #2349 there was a discussion of the name of the `Error` case of `Event`s. That led me down the path of thinking how when explaining RAC to people there’s always a fun dance of “values across a signal, or errors, but only one error”.

What if instead of an “error” event it was a “failed” event, mirroring the other terminating events, “completed” and “interrupted”, in tense?

This would be a decently large API change to happen mid-release, but I wanted to put the idea out there for discussion.


neilpa:
Personally I really like this change. However, I'm not convinced it's worth the churn for consumers. Unlike the removal of `|>`, I don't see the same level of benefit even though it would increase clarity.

@ReactiveCocoa/reactivecocoa other opinions?

bencochran:
(I’m going wait to update this until #2388 is merged)

kastiglione:
I like the idea of being more precise. Perhaps `Failed`, perhaps something else. If we were ever going to do changes like this, it's arguable that it's better to do it while Swift is also churning.

nomothetis:
:+1: It was very confusing for me in the early days that I couldn't send multiple errors along a signal. `.Failed` is a better name for it, I think.

mdiep:
I really like the consistency of `Failed` with `Completed` and `Interrupted`. And not all failures are errors.

:+1:

robb:
I like how it fails _with_ an error, making signals of errors a little less confusing imo. :+1:

robrix:
Very much in favour of more accurately capturing the intent of the feature in the API. Nice one.

rpowelll:
Would we also want to rename `flatMapError` to `flatMapFailure`?

bencochran:
@rpowelll I struggled with that… if it’s `flatMapFailure` should it also be `mapFailure`? I guess another way of phrasing it is: does the operation conceptually operate on the value (lowercase v) or the event?

With that in mind, to me `mapFailure` seems like it would return another type of “event” entirely (making it essentially the same as `flatMapFailure` in my mind). So I favored keeping the two using the same wording.

bencochran:
This can now merge cleanly. I also updated documentation, readme, and changelog.

There are a few bits I still think could use some clarity / outside opinions:

1. `flatMapError` vs `flatMapFailure`, and relatedly, `mapError` vs  `mapFailure`
1. The name `promoteErrors` throws me off a bit. `promoteFailures` doesn’t seem right either. ¯\\\_(ツ)_/¯

neilpa:
Sorry for dropping the ball on this one. At this point, I'd like to delay it until an alpha 3 build given the other interim changes and the unknowns you call out.

mdiep:
Should this be in the 4.0 milestone?

neilpa:
> Should this be in the 4.0 milestone?

Done

bencochran:
No worries. I agree bumping this makes sense.

bencochran:
👉 #2505

PR #2363, opened 2015-09-15
Workaround to prevent NSError from losing its UserInfo in toRACSignal

raylillywhite:
This is a workaround for [rdar://22708537](http://www.openradar.me/radar?id=6068957668704256) Casting an NSError that's in a generic function with type parameter `<E: ErrorType>` to an NSError strips the error's UserInfo dictionary.

One downside to this workaround is that subclasses of NSError will still be broken. The only workaround that I could think of that would handle that is:

```
    var nsError = error as NSError
    var mirror: Mirror? = Mirror(reflecting: error)
    while mirror != nil {
        if mirror!.subjectType == NSError.self {
            nsError = unsafeBitCast(error, NSError.self)
            break
        }
        mirror = mirror!.superclassMirror()
    }
    subscriber.sendError(nsError)
```

I wasn't sure if there was a significant runtime cost to creating the Mirror struct, or if there was a better way to handle that, so I just left this handling NSError and not subclasses. Let me know if you'd like me to implement the code above instead though.

### DIFF
--- a/ReactiveCocoaTests/Swift/ObjectiveCBridgingSpec.swift
+++ b/ReactiveCocoaTests/Swift/ObjectiveCBridgingSpec.swift
@@ -13,6 +13,7 @@ import ReactiveCocoa
 import XCTest
 
 class ObjectiveCBridgingSpec: QuickSpec {
+	
 	override func spec() {
 		describe("RACSignal.toSignalProducer") {
 			it("should subscribe once per start()") {
@@ -44,6 +45,9 @@ class ObjectiveCBridgingSpec: QuickSpec {
 		}
 
 		describe("toRACSignal") {
+			let key = "TestKey"
+			let userInfo: [String: String] = [key: "TestValue"]
+			let testNSError = TestNSError(domain: "TestDomain", code: 1, userInfo: userInfo)
 			describe("on a Signal") {
 				it("should forward events") {
 					let (signal, sink) = Signal<NSNumber, NoError>.pipe()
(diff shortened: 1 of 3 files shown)


NachoSoto:
That bug is unfortunate, I just duped your radar!

I'm not too convinced that we should include this kind of workaround in RAC though, to be honest.
At least not without proper test coverage.

What does everyone else think?

raylillywhite:
@NachoSoto Thanks for duping the radar :)

I've updated the workaround to not involve any runtime checks and properly handle NSError subclasses. I also added some related tests. Mind taking another look and seeing if this new workaround looks like it should be in RAC?

ReactiveCocoaTests/Swift/ObjectiveCBridgingSpec.swift
```diff
@@ -85,6 +89,23 @@ class ObjectiveCBridgingSpec: QuickSpec {
...
+					
+					racSignal.subscribeError {
+						error = $0
+						return
+					}
+					
+					sendError(sink, testNSError)
+					
+					let userInfoValue = error?.userInfo[key] as? String
+					expect(userInfoValue).to(equal(userInfo[key]))
```
NachoSoto: Awesome

ReactiveCocoaTests/Swift/TestError.swift
```diff
@@ -14,3 +16,6 @@ enum TestError: Int {
 
 extension TestError: ErrorType {
 }
+
+class TestNSError: NSError {
```
NachoSoto: It's discouraged to subclass `NSError`. Instead, you should differentiate errors using the `domain`, like you're doing in the test.

Can you change the test to just use `NSError`?

NachoSoto:
This looks great! Just one final comment :star:

ReactiveCocoaTests/Swift/TestError.swift
raylillywhite: I thought the same thing, but when I checked the docs to verify that, they seem to actually _encourage_ it.

> NSError is not an abstract class, and can be used directly. Applications may choose to create subclasses of NSError to provide better localized error strings by overriding localizedDescription.

Regardless, it sounds like it's at least a bad practice, so it's probably better to test the common case here, like you suggested. So I just removed `TestNSError` and updated the tests.

NachoSoto: Oh wow, I'm surprised the docs say that. You can provide a localized description by setting the `NSLocalizedDescriptionKey` key in the user info. No need to subclass! :)

NachoSoto:
Perfect :star:

PR #2366, opened 2015-09-16
Remove indentations to match coding guidelines

dapenggao:


NachoSoto:
AFAIK the current indentation is based on Xcode default settings. We shouldn't fight auto indentation...

dapenggao:
Actually the current indentation is quite mixed, i.e. some lines are indented and some are not. Perhaps we should fix these things just before release. So never mind...

PR #2351, opened 2015-09-13
[WIP] Readded travis.yml

NachoSoto:
Based on https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/.travis.yml

### TODO:
- [x] Add `watchOS` test target and set up scheme.
- [ ] Either wait for `xctool` or cleanup `cibuild` script (https://github.com/NachoSoto/ReactiveCocoa/commit/e8e03cae87e4fbd32b4521dc165fc979425abdd8).
- [ ] Updated `Nimble` and `Quick` to include `watchOS` built frameworks.
- [ ] Make `watchOS` test target use the right `Nimble` and `Quick`.
- [ ] Add `tvOS` test target.

PR #2367, opened 2015-09-16
Fix `filter` operation

bencochran:
A bug snuck in with #2354 causing `filter` to forward events when the predicate was false. This brings back some nesting to fix that.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -228,8 +228,10 @@ extension SignalType {
 	public func filter(predicate: T -> Bool) -> Signal<T, E> {
 		return Signal { observer in
 			return self.observe { event in
-				if case let .Next(value) = event where predicate(value) {
-					sendNext(observer, value)
+				if case let .Next(value) = event {
+					if predicate(value) {
+						sendNext(observer, value)
+					}
 				} else {
 					observer(event)
 				}


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -228,8 +228,10 @@ extension SignalType {
 	public func filter(predicate: T -> Bool) -> Signal<T, E> {
 		return Signal { observer in
 			return self.observe { event in
-				if case let .Next(value) = event where predicate(value) {
-					sendNext(observer, value)
+				if case let .Next(value) = event {
```
NachoSoto: Huh, wasn't the previous implementation equivalent? 😰

NachoSoto: Oh, duh, that was going to the `else` branch!

NachoSoto:
Thanks a lot! I imagine the tests caught this though, right? (Even though we're not running them yet, see #2351).

ReactiveCocoa/Swift/Signal.swift
dapenggao: Oops! Sorry for the bug...

bencochran:
Tests first crash in `ignoreNil`, but they also fail on `filter` if I leave those tests out.

NachoSoto:
Makes sense!

PR #2370, opened 2015-09-16
Change `NoError` to an `enum`

dapenggao:
- The `NoError` type is currently a `final class`. But it is more natural for an `ErrorType` to be an `enum`.
- This PR also removes the free functions used as workaround for compiler crashes. (The project builds successfully on Xcode 7 GM.)
- This PR is unit tested with three failed cases (which is the same as before).


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -435,7 +435,7 @@ extension SignalProducerType {
 	/// will also be interrupted.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func combineLatestWith<U>(otherProducer: SignalProducer<U, E>) -> SignalProducer<(T, U), E> {
-		return lift(ReactiveCocoa.combineLatestWith)(otherProducer)
+		return lift { signal -> Signal<T, E> -> Signal<(T, U), E> in { $0.combineLatestWith(signal) } }(otherProducer)
```
NachoSoto: Can you explain this?

NachoSoto: Oh never mind

ReactiveCocoa/Swift/Property.swift
```diff
@@ -181,10 +181,10 @@ public final class MutableProperty<T>: MutablePropertyType {
 }
 
 infix operator <~ {
-	associativity right
+associativity right
 
-	// Binds tighter than assignment but looser than everything else
-	precedence 93
+// Binds tighter than assignment but looser than everything else
+precedence 93
```
NachoSoto: Can you fix this indentation?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -948,7 +948,7 @@ extension SignalType {
 				observer(event.map { value in
 					accumulator = combine(accumulator, value)
 					return accumulator
-				})
+					})
```
NachoSoto: Indentation

NachoSoto:
Looks great! Can you just restore the indentation changes? :)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -161,7 +161,7 @@ public final class MutableProperty<T>: MutablePropertyType {
 				.flatMapError { error in
 					assert(false, "Received unexpected error from KVO signal: \(error)")
 					return .empty
-				}
+			}
```
jlawton: I would think that the indentation should match the `.flatMapError`

dapenggao:
Done!

ReactiveCocoa/Swift/Action.swift
```diff
@@ -170,7 +170,7 @@ public final class CocoaAction: NSObject {
 				self?.willChangeValueForKey("enabled")
 				self?._enabled = value
 				self?.didChangeValueForKey("enabled")
-			}
+		}
```
ikesyo: This should be reverted too.

ReactiveCocoa/Swift/Action.swift
```diff
@@ -178,7 +178,7 @@ public final class CocoaAction: NSObject {
 				self?.willChangeValueForKey("executing")
 				self?._executing = value
 				self?.didChangeValueForKey("executing")
-			}
+		}
```
ikesyo: This should be reverted too.

dapenggao:
Now done!

I propose that we change the method signature of `lift` from
```swift
func lift<U, F, V, G>(transform: Signal<U, F> -> Signal<T, E> -> Signal<V, G>) -> SignalProducer<U, F> -> SignalProducer<V, G>
```
to
```swift
func lift<U, F, V, G>(transform: Signal<T, E> -> Signal<U, F> -> Signal<V, G>) -> SignalProducer<U, F> -> SignalProducer<V, G>
```
Swapping the positions of `Signal<U, F>` and `Signal<T, E>` not only makes the code more intuitive (because `Signal<T, E>` is the type of `self`, it is more reasonable to make it the first argument), but also allows code simplifications such as the following:

From: 
```swift
func combineLatestWith<U>(otherProducer: SignalProducer<U, E>) -> SignalProducer<(T, U), E> {
	return lift { signal -> Signal<T, E> -> Signal<(T, U), E> in { $0.combineLatestWith(signal) } }(otherProducer)
}
```
To:
```swift
func combineLatestWith<U>(otherProducer: SignalProducer<U, E>) -> SignalProducer<(T, U), E> {
	return lift(Signal.combineLatestWith)(otherProducer)
}
```
The simplified code is possible because [Swift instance methods are curried functions](http://oleb.net/blog/2014/07/swift-instance-methods-curried-functions/).

NachoSoto:
Totally! That's welcome in another PR :)

NachoSoto:
Thanks for doing this!

PR #2361, opened 2015-09-15
Re-added tvOS scheme

NachoSoto:
Reverts ReactiveCocoa/ReactiveCocoa#2359

This is incompatible with Xcode 7, so we need to wait until 7.1 is GM.

PR #2372, opened 2015-09-16
Update Result to new beta version

gfontenot:
I tagged 0.6 beta 2 of Result last night. I don't think RAC is
particularly interested in the `materialize` function, but it includes
the tvOS target and it'll help other libs with Result as a common
dependency to point at a tag instead of a sha.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "antitypical/Result" "76f9a97"
+github "antitypical/Result" "0.6.0-beta.2"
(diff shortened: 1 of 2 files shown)


NachoSoto:
Yup, perfect!
Thanks :heart:

neilpa:
Isn't this going to break Xcode 7 + carthage because of tvOS

NachoSoto:
Oh shit, you're right. Let me revert this and just include @gfontenot in #2361

neilpa:
Seems like the `tvOS` bits will need to go in a `0.7` for Result so that it can be split from `0.6` and still target Xcode 7

gfontenot:
Oops, sorry. Missed that convo about the `tvOS` release.

I'll take a look at fixing up Result to have a 0.7 series of releases that include `tvOS`.

Glad Apple has made this so simple to reason about!

neilpa:
Yea, I just opened antitypical/Result#82 to track

neilpa:
> Glad Apple has made this so simple to reason about!

Thankfully we can mostly workaround their failures with careful versioning on our side.

PR #2375, opened 2015-09-16
README.md update. Changed old style of handling next even to a new one with startWithNext.

nikita-leonov:
Updated documentation according to the recent change in starting SignalProducers


NachoSoto:
Thanks!

PR #2378, opened 2015-09-17
Change `lift` signature for better code

dapenggao:
As mentioned in #2370, I propose that we change the method signature of `lift` from
```swift
func lift<U, F, V, G>(transform: Signal<U, F> -> Signal<T, E> -> Signal<V, G>) -> SignalProducer<U, F> -> SignalProducer<V, G>
```
to
```swift
func lift<U, F, V, G>(transform: Signal<T, E> -> Signal<U, F> -> Signal<V, G>) -> SignalProducer<U, F> -> SignalProducer<V, G>
```
Swapping the positions of `Signal<U, F>` and `Signal<T, E>` not only makes the code more intuitive (because `Signal<T, E>` is the type of `self`, it is more reasonable to make it the first argument), but also allows code simplifications such as the following:

From: 
```swift
func combineLatestWith<U>(otherProducer: SignalProducer<U, E>) -> SignalProducer<(T, U), E> {
	return lift { signal -> Signal<T, E> -> Signal<(T, U), E> in { $0.combineLatestWith(signal) } }(otherProducer)
}
```
To:
```swift
func combineLatestWith<U>(otherProducer: SignalProducer<U, E>) -> SignalProducer<(T, U), E> {
	return lift(Signal.combineLatestWith)(otherProducer)
}
```
The simplified code is possible because [Swift instance methods are curried functions](http://oleb.net/blog/2014/07/swift-instance-methods-curried-functions/).

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -435,7 +435,7 @@ extension SignalProducerType {
 	/// will also be interrupted.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func combineLatestWith<U>(otherProducer: SignalProducer<U, E>) -> SignalProducer<(T, U), E> {
-		return lift { sg -> Signal<T, E> -> Signal<(T, U), E> in { $0.combineLatestWith(sg) } }(otherProducer)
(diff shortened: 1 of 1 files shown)


neilpa:
Excellent! I had tried this on an early beta but it used to crash the compiler.

Issue #2377, opened 2015-09-17
RAC4 question. Recommended approach to migrate free functions and replace |>


nikita-leonov:
What would be suggested approach to move code like following to a RAC4?
```
    public func connectDevice(deviceType: HuamiType) -> SignalProducer<Device, PeripheralError> {
        return prepareForScanning()
            |> findNearbyDevice(deviceType, RSSI: -60)
            |> connectDevice
            |> discoverServicesForDevice
            |> discoverCharacteristicsForDevice
    }
```

Before I used a |> to apply locally defined functions, it was a pleasant experience. Now I am not sure how to chain it. It definitely would not work with extension for SignalProducerType, as there is a strict type of  SignalProducer expected, as well as chined functions have side effects (opsie, I know this is bad) as they use some entity variables like centralManager etc.

NachoSoto:
Can you show how those functions were defined in `RAC 3`?

> Now I am not sure how to chain it. It definitely would not work with extension for SignalProducerType, as there is a strict type of SignalProducer expected

What do you mean by "as there is a strict type of SignalProducer expected"? If you mean that the `T` has to match a certain type, you can still do that! :) For example:

```swift
extension SignalProducerType where T: YourProtocol {
	public func findNearbyDevice(deviceType: DeviceType, RSSI: Int) -> SignalProducer<T, E> {
		return self.map { value in 
                       // here `value` conforms to `YourProtocol`
                       return ...
                }
	}
}
```

calebd:
I have a similar problem. I have view model properties (`SignalProducer<ViewModel?, NoError>`) that have image producers (`SignalProducer<UIImage, NSError>`). If I `map({ $0?.imageProducer })` I end up with a `SignalProducer<SignalProducer<UIImage, NSError>?, NoError>`. I use the following function to flatten that result:

```swift
func promoteNil<T, E>(producer: SignalProducer<SignalProducer<T, E>?, E>) -> SignalProducer<T?, E> {
    return producer |> flatMap(.Latest, { thing in
        if let producer = thing {
            return producer |> map({ .Some($0) })
        }
        else {
            return SignalProducer<T?, E>(value: nil)
        }
    })
}
```

So now I can do `viewModel.producer |> promoteErrors(NSError) |> map({ $0?.imageProducer }) |> promoteNil` which results in a `SignalProducer<UIImage?, NSError>`.

I have yet to find a way to make that chainable.

[...]

PR #2380, opened 2015-09-17
Moved flatMap intro SignalProducerType extension.

nikita-leonov:
According to @neilpa message in #2377 there was language and compiler limitations that didn't allowed this before. On a Xcode 7.0 (7A220) it complies well and all unit-tests are passing. Does it mean it can me merged in, or there is a special case that still may crash such implementation?

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -598,7 +598,7 @@ extension SignalProducerType {
 	}
 }
 
-extension SignalProducer where T: OptionalType {
+extension SignalProducerType where T: OptionalType {
 	/// Unwraps non-`nil` values and forwards them on the returned signal, `nil`
 	/// values are dropped.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
@@ -973,13 +973,11 @@ extension SignalProducerType {
 			}
 		}
 	}
-}
 
-extension SignalProducer {
 	/// `concat`s `next` onto `self`.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
-	public func concat(next: SignalProducer) -> SignalProducer {
-		return SignalProducer<SignalProducer, E>(values: [self, next]).flatten(.Concat)
+	public func concat(next: SignalProducer<T, E>) -> SignalProducer<T, E> {
+		return SignalProducer<SignalProducer<T, E>, E>(values: [self.producer, next]).flatten(.Concat)
 	}
 }
 
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -913,7 +913,35 @@ public func zip<S: SequenceType, T, Error where S.Generator.Element == SignalPro
 	return .empty
 }
 
+extension SignalProducer where T: SignalProducerType, E == T.E {
+	/// Flattens the inner producers sent upon `producer` (into a single producer of
+	/// values), according to the semantics of the given strategy.
+	///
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -913,7 +913,35 @@ public func zip<S: SequenceType, T, Error where S.Generator.Element == SignalPro
...
+	/// Flattens the inner producers sent upon `producer` (into a single producer of
+	/// values), according to the semantics of the given strategy.
+	///
+	/// If `producer` or an active inner producer emits an error, the returned
+	/// producer will forward that error immediately.
+	///
+	/// `Interrupted` events on inner producers will be treated like `Completed`
+	/// events on inner producers.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func flatten(strategy: FlattenStrategy) -> SignalProducer<T.T, E> {
```
NachoSoto: Can this be moved to `SignalProducerType` too?

nikita-leonov: @NachoSoto good question, let me check, as I was trying to solve only my case and didn't thought that someone may require flatten as well.

neilpa: I'll be pleasantly surprised if it works w/out making the compiler die. IIRC I tried to convert this in multiple betas up to 5 and was disappointed each time.

neilpa: If it does work, you should also be able to change `filterNil` because that was the other problem operator.

nikita-leonov: @neilpa what sort of issue did you have with compiler? In my case I just have an issues with types in ```lift``` call which looks like a legit call:
```
Cannot invoke 'lift' with an argument list of type '((Signal<SignalProducer<Self.T, Self.E>, Self.E>) -> Signal<Self.T, Self.E>)
```

neilpa: Previously it would seg-fault the compiler. That looks like a normal compiler error to me.

neilpa: I recommend storing intermediate steps in locals with explicit type annotations. That's usually how I track down the actual source of errors when there's otherwise a lot of type-inference involved.

nikita-leonov: Hm, yes I was over-engineering. Actual change was as simple as replace ```SignalProducer``` to ```SignalProducerType``` and retaining the rest for flatten as is. it compiles, it works as it pass the tests.

There are many other things that could be moved to ```SignalProducerType``` and ```SignalType``` in the similar manner, but I would prefer to keep it in a separate PR in other case it will grow to big.

nikita-leonov:
Actually wait a little bit and I move reminders in ```SignalProducerType``` and ```SignalType```. It seems it is not as big as I expected.

ReactiveCocoa/Swift/SignalProducer.swift
nikita-leonov: ps practice of breaking down into smaller steps with an expected type clearly defined is a good one, I use it often as well. The problem in my case that there was no further way to breakdown... I actually found the place of an error but because of a very complex signature of signals etc was too blind to notice it.

nikita-leonov:
I think it is done. By done I mean done-done :)

neilpa:
Awesome! Everything's implemented in terms of protocol extensions now :tada:

NachoSoto:
Awesome! :grinning:

PR #2386, opened 2015-09-19
Improvements on `Bag` and others

dapenggao:
#### Bag
- Manual overflow check is replaced by automatic check using `UInt.addWithOverflow()`
- Remove the implementation of `generate()` because this method is default-implemented

#### Others
- `modf()` has an overload that returns a 2-tuple
- `CLong` is replaced by `Int` because the compiler infers `Int` for `DISPATCH_QUEUE_PRIORITY_DEFAULT` and other values

### DIFF
--- a/ReactiveCocoa/Swift/Bag.swift
+++ b/ReactiveCocoa/Swift/Bag.swift
@@ -24,8 +24,8 @@ internal struct Bag<T> {
 	/// Inserts the given value in the collection, and returns a token that can
 	/// later be passed to removeValueForToken().
 	mutating func insert(value: T) -> RemovalToken {
-		let nextIdentifier = currentIdentifier &+ 1
-		if nextIdentifier == 0 {
+		let (nextIdentifier, overflow) = UInt.addWithOverflow(currentIdentifier, 1)
+		if overflow {
 			reindex()
 		}
 
@@ -33,7 +33,7 @@ internal struct Bag<T> {
 		let element = BagElement(value: value, identifier: currentIdentifier, token: token)
 
 		elements.append(element)
-		currentIdentifier++
+		currentIdentifier = nextIdentifier
 
 		return token
 	}
@@ -42,14 +42,14 @@ internal struct Bag<T> {
 	///
 	/// If the value has already been removed, nothing happens.
 	mutating func removeValueForToken(token: RemovalToken) {
-		if let identifier = token.identifier {
-			// Removal is more likely for recent objects than old ones.
-			for i in (0..<elements.endIndex).reverse() {
-				if elements[i].identifier == identifier {
-					elements.removeAtIndex(i)
-					token.identifier = nil
-					break
-				}
+		guard let identifier = token.identifier else { return }
+		// Removal is more likely for recent objects than old ones.
+		for i in (elements.startIndex..<elements.endIndex).reverse() {
+			if elements[i].identifier == identifier {
+				swap(&elements[i], &elements[elements.endIndex - 1])
+				elements.removeLast()
+				token.identifier = nil
+				break
 			}
 		}
 	}
@@ -58,7 +58,7 @@ internal struct Bag<T> {
 	/// will reset all current identifiers to reclaim a contiguous set of
 	/// available identifiers for the future.
 	private mutating func reindex() {
-		for i in 0..<elements.endIndex {
+		for i in elements.startIndex..<elements.endIndex {
 			currentIdentifier = UInt(i)
 
 			elements[i].identifier = currentIdentifier
@@ -67,30 +67,15 @@ internal struct Bag<T> {
 	}
 }
 
-extension Bag: SequenceType {
-	func generate() -> AnyGenerator<T> {
-		var index = 0
-		let count = elements.count
-
-		return anyGenerator {
-			if index < count {
-				return self.elements[index++].value
-			} else {
-				return nil
-			}
-		}
-	}
-}
-
 extension Bag: CollectionType {
 	typealias Index = Array<T>.Index
 
 	var startIndex: Index {
-		return 0
+		return elements.startIndex
 	}
-	
+
 	var endIndex: Index {
-		return elements.count
+		return elements.endIndex
 	}
 
 	subscript(index: Index) -> T {
--- a/ReactiveCocoa/Swift/Bag.swift
+++ b/ReactiveCocoa/Swift/Bag.swift
@@ -42,14 +42,14 @@ internal struct Bag<T> {
 	///
 	/// If the value has already been removed, nothing happens.
 	mutating func removeValueForToken(token: RemovalToken) {
-		guard let identifier = token.identifier else { return }
-		// Removal is more likely for recent objects than old ones.
-		for i in (elements.startIndex..<elements.endIndex).reverse() {
-			if elements[i].identifier == identifier {
-				swap(&elements[i], &elements[elements.endIndex - 1])
-				elements.removeLast()
-				token.identifier = nil
-				break
+		if let identifier = token.identifier {
+			// Removal is more likely for recent objects than old ones.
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/Swift/Bag.swift
```diff
@@ -24,8 +24,8 @@ internal struct Bag<T> {
 	/// Inserts the given value in the collection, and returns a token that can
 	/// later be passed to removeValueForToken().
 	mutating func insert(value: T) -> RemovalToken {
-		let nextIdentifier = currentIdentifier &+ 1
-		if nextIdentifier == 0 {
+		let (nextIdentifier, overflow) = UInt.addWithOverflow(currentIdentifier, 1)
```
NachoSoto: Oooh yes so much better

ReactiveCocoa/Swift/Bag.swift
```diff
@@ -58,7 +58,7 @@ internal struct Bag<T> {
 	/// will reset all current identifiers to reclaim a contiguous set of
 	/// available identifiers for the future.
 	private mutating func reindex() {
-		for i in 0..<elements.endIndex {
+		for i in elements.startIndex..<elements.endIndex {
```
NachoSoto: This would be equivalent to `for i in elements.indices`

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -119,7 +119,7 @@ public func toRACSignal<T: AnyObject, E: NSError>(producer: SignalProducer<T?, E
 				subscriber.sendError(error)
 			case .Completed:
 				subscriber.sendCompleted()
-			default:
+			case .Interrupted:
```
NachoSoto: +1

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -167,7 +167,7 @@ public func toRACSignal<T: AnyObject, E: NSError>(signal: Signal<T?, E>) -> RACS
                 subscriber.sendError(error)
             case .Completed:
                 subscriber.sendCompleted()
-            default:
+            case .Interrupted:
```
NachoSoto: +1

NachoSoto:
This is great!

dapenggao:
Also, for efficiency we can remove an element from `Bag` by exchanging the element with the last element and then `removeLast()`, which is an amortized O(1) operation.

```swift
guard let identifier = token.identifier else { return }
// Removal is more likely for recent objects than old ones.
for i in elements.indices.reverse() {
	if elements[i].identifier == identifier {
		swap(&elements[i], &elements[elements.endIndex - 1])
		elements.removeLast()
		token.identifier = nil
		break
	}
}
```

But I am not sure whether this is premature optimization.

dapenggao:
To push even further we can use a contiguous array, which is supposed to be faster for `class` types.
```swift
private var elements: ContiguousArray<BagElement<T>> = []
```

dapenggao:
Never mind. I forgot that `BagElement` is a `struct` so `ContiguousArray` is equivalent to `Array`.

Here is an alternate implementation I came up with that does not even require an `identifier`. This implementation simply uses pointer identity to identify elements.

```swift
internal final class RemovalToken {
	private var valid = true
}

/// An unordered, non-unique collection of values of type T.
internal struct Bag<T> {
	private var elements: [BagElement<T>] = []

	/// Inserts the given value in the collection, and returns a token that can
	/// later be passed to removeValueForToken().
	mutating func insert(value: T) -> RemovalToken {
		let token = RemovalToken()
		let element = BagElement(value: value, token: token)

		elements.append(element)

		return token
	}

	/// Removes a value, given the token returned from insert().
	///
	/// If the value has already been removed, nothing happens.
	mutating func removeValueForToken(token: RemovalToken) {
		guard token.valid else { return }
		for i in elements.indices.reverse() {
			if elements[i].token === token {
				elements.removeAtIndex(i)
				token.valid = false
				break
			}
		}
	}
}

private struct BagElement<T> {
	let value: T
	let token: RemovalToken
}
```

neilpa:
I'm be in favor of the new implementation, but I'd like to see measurements first because `Bag` provides the backing store for all signal subscriptions.

dapenggao:
Yup. I'm building some performance tests right now. Will show the results when done.

neilpa:
:100:

dapenggao:
I'm back! So I have written some performance tests on mass insertion and deletion (1000 random integers). As expected, removal is much slower than insertion. Surprisingly though, the greatest performance bottleneck for removal is not the search & remove operations, but the `reverse()` call in `elements.indices.reverse()`. If I remove of 1000 items at random order, the implementation with `reverse()` typically uses ~230ms, whereas the one without it uses only ~65ms.

All the implementations I tested perform well on insertion: they all use ~1ms for inserting 1000 elements. 

For deletion, the current version of `Bag` (with the `reverse()` call removed) actually performs better by a marginal ~10ms than pointer identity based implementations. Swapping the element to the back of the array and then `removeLast()` does not improve performance and using `ContiguousArray` can only improve performance by only ~10ms. 

However, I have not tested inserting objects or protocol types.

[Here](https://drive.google.com/file/d/0B75ok0XdsJQdYVcwTEtDalowdG8/view?usp=sharing) is my test project.

neilpa:
Wow! Thanks for really digging into this, this is excellent work. I'm also surprised by the results so I'm glad you measured it. I'll take a closer look when I have a bit more time.

neilpa:
> However, I have not tested inserting objects or protocol types.

RAC's usage is with function options (`Event<T, E> -> ()`) since that's what all the observers are.

dapenggao:
Hello! I am back! I made some further tests on `Bag` for storing `class` types as well as closures. While the current version performs better at storing `Int`s, it is slower at storing objects and closures than identity-based implementations. For removing 1000 items at random, the current implementation generally uses 20 more milliseconds than do the others, which is again very surprising to me.
The test project is [here](https://drive.google.com/open?id=0B75ok0XdsJQdWUVRblBsTUNKR28).

dapenggao:
For #2417: I changed `Bag<T>` to a class and re-ran the benchmark again. There is no observable performance impact in regard to insertion/deletion. However, a `class Bag` seems to crash the test. I guess that has to do with shared mutable state.
screen shot 2015-09-27 at 8 12 37 pm
screen shot 2015-09-27 at 8 12 57 pm

NachoSoto:
> For #2417: I changed Bag<T> to a class and re-ran the benchmark again. There is no observable performance impact in regard to insertion/deletion

Well in that case you'd need to put it inside of an `Atomic<Bag>`. That'd be a good idea in any case to fully replicate how it's used in `RAC` anyway.

NachoSoto:
@dapenggao I just ran your tests in Xcode 7.1 Beta 2 (it's still unclear whether #2417 only reproduces with Swift 2.1) and everything passes :/

dapenggao:
> Well in that case you'd need to put it inside of an Atomic<Bag>. That'd be a good idea in any case to fully replicate how it's used in RAC anyway.
`struct`-based `Bag`s indeed get significantly slower when they are inside `Atomic`. The `class` based `Bag` is only slightly slower, but I wonder if this is still an option for us (due to #2421).
Test project is [here](https://drive.google.com/file/d/0B75ok0XdsJQdTG9QQlU0OVhrNXM/view?usp=sharing).

NachoSoto:
Reopened the changes in #2577. Sorry for duplicating the commits but your remote seemed to be gone :/

@dapenggao let us know if you're still interested in those performance improvements! :)

RuiAAPeres:
@dapenggao is this something you want to merge?

dapenggao:
@RuiAAPeres I think we can close this now. The important bits have already been merged in #2577.

PR #2387, opened 2015-09-20
Publish optionalize() for signal/producer

335g:
I PR, because I don't know the reason that `optionalize()` must be private. And I think that it is better to provide type conversion. Well, but I solve it if I use `map()` :sweat_smile:


neilpa:
Lets not add this to the public API surface. It's trivial to implement and unlike other simple operators it's not generally needed.

335g:
:ok_hand:

PR #2388, opened 2015-09-20
Rename generic T and E params

neilpa:
Merged and resolved conflicts building from #2349 since it was outdated

/cc @bencochran


ReactiveCocoa/Swift/Action.swift
```diff
@@ -202,18 +202,18 @@ public final class CocoaAction: NSObject {
...
 	/// disabled.
 	case NotEnabled
 
 	/// The producer returned from apply() sent the given error.
-	case ProducerError(E)
+	case ProducerError(Error)
 }
 
-public func == <E: Equatable>(lhs: ActionError<E>, rhs: ActionError<E>) -> Bool {
+public func == <Error: Equatable>(lhs: ActionError<Error>, rhs: ActionError<Error>) -> Bool {
```
NachoSoto: Nit: should there be a space between `==` and `<`?

ReactiveCocoa/Swift/Event.swift
```diff
@@ -10,13 +10,13 @@
 ///
 /// Signals must conform to the grammar:
 /// `Next* (Error | Completed | Interrupted)?`
-public enum Event<T, E: ErrorType> {
+public enum Event<Value, Err: ErrorType> {
```
NachoSoto: I think I missed this discussion: why `Err` instead of `Error`?

neilpa: Because it conflicts w/ the `Error` case. See #2360

ReactiveCocoa/Swift/Action.swift
neilpa: Huh? There is already

ReactiveCocoa/Swift/Event.swift
NachoSoto: Ooooh right

ReactiveCocoa/Swift/Action.swift
NachoSoto: It was a rhetorical question :P

neilpa: Not for normal functions, but for operators I find it really helps with readability.

bencochran:
Thanks for giving this an update

neilpa:
Updated the changelog too, should be ready to go.

PR #2397, opened 2015-09-22
Ensure that PRODUCT_NAME is set when used as a subproject

### LINKED ISSUE #2395: ReactiveCocoa.framework showing as ".framework" in my projects
On a brand new iOS SDK Xcode 7 project, ReactiveCocoa.framework is being shown as `.framework`. Seems like build settings might be a bit skewiff. Installed via Carthage with the `--no-build` option and then manually adding RAC as a subproject of my app's project.

tonyarnold:
When used as a subproject, the `$PRODUCT_NAME` variable was not being inferred from the xcconfigs. Setting the project level product name works around the problem (the tests use `$TARGET_NAME` and the frameworks all use `$PROJECT_NAME`). 

I'm sure there's a problem here deeper than what I've investigated, but this PR fixes the symptoms reasonably cleanly.

Fixes #2395.


tonyarnold:
Argh, my mistake — ignore this!

PR #2398, opened 2015-09-22
Explicitly set the PRODUCT_NAME to TARGET_NAME

### LINKED ISSUE #2395: ReactiveCocoa.framework showing as ".framework" in my projects

tonyarnold:
When used as a subproject, the `$PRODUCT_NAME` variable was not being inferred from the xcconfigs. Setting the project level product name works around the problem (the tests use `$TARGET_NAME` and the frameworks all use `$PROJECT_NAME`). 

I'm sure there's a problem here deeper than what I've investigated, but this PR fixes the symptoms reasonably cleanly.

Fixes #2395.

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -2767,6 +2767,7 @@
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
 				ONLY_ACTIVE_ARCH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = "org.reactivecocoa.$(PRODUCT_NAME:rfc1034identifier)";
+				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_WHOLE_MODULE_OPTIMIZATION = NO;
 				TARGETED_DEVICE_FAMILY = "1,2";
 				VERSIONING_SYSTEM = "apple-generic";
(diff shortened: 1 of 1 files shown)


NachoSoto:
Awesome!
Just making sure: this is changed in the **project**, not the targets, right? That way we avoid duplication :+1:

tonyarnold:
Absolutely — I've even removed the customised entries at the target levels on the tests.

NachoSoto:
Awesome!

NachoSoto:
Actually, one second. Using `$TARGET_NAME` means that other platforms, like `watchOS` would have a different product name, which is wrong!

Am I misisng something?

tonyarnold:
It looks like it's an Xcode bug? Check the derived names of the framework targets — they're correctly getting their values from the xcconfigs, but only if the project has a PRODUCT_NAME set. 

This bears some deeper investigation.

neilpa:
You should be able to use `$(PROJECT_NAME)` instead of target name

neilpa:
And then `$(PROJECT_NAME)Tests` for all the tests.

tonyarnold:
@neilpa this is more work/maintenance (modifying the setting in each test target, versus just the project), and I don't think it really changes the outcome. Happy to go with whatever, but I'd favour less maintenance. 

Maybe we need our own local xcconfig on top of those included in the submodule?

neilpa:
I don't really care about the tests, can do whatever works and is easy there. However, I don't see how `$TARGET_NAME` works since that should resolve to ReactiveCocoa-Mac, ReactiveCocoa-iOS, etc.

tonyarnold:
`$TARGET_NAME` _is_ being ignored on the non-tests. Like I said, it seems like a bug in Xcode and the fix seems ridiculous, but unless `$PRODUCT_NAME` is set a the project level, the value in the xcconfigs is not flowing through. Completely nuts :nut_and_bolt:

neilpa:
Ahh. I understand now. Lulz Xcode.

tonyarnold:
I've made a terrible mistake! This "feature" is fixed in Xcode 7.1, so now we end up with ReactiveCocoa-Mac.framework and a whole swag of problems. I'll submit a new PR to hardcore the name when I'm home later :frowning:

NachoSoto:
Oh yeah that explains what I was seeing today!

tonyarnold:
:metal: *hardcore* the name

PR #2405, opened 2015-09-23
Change `catch` to `flatMapError` in README

rpowelll:
Just a quick fix for the README. We were still referring to the old `catch` method in the search example, which has since been renamed to `flatMapError`.


neilpa:
Thanks!

Issue #2412, opened 2015-09-24
Question: README.md example code not working? (swift2 branch)


RefuX:
Following along the README.md and this basic example fails to compile....

```
class ViewController: UIViewController {

    @IBOutlet weak var textField: UITextField!
    
    override func viewDidLoad() {
        super.viewDidLoad()
        // Do any additional setup after loading the view, typically from a nib.
        
        //RACSignal
        let searchSignal : RACSignal = textField.rac_textSignal()
        
        let searchStrings : SignalProducer<String, NSError> = textField.rac_textSignal()
            .toSignalProducer()
            .map { text in text as! String }
        
        let searchResults = searchStrings
            .flatMap(.Latest) { query in
                let URLRequest = self.searchRequestWithEscapedQuery(query)
                return NSURLSession.sharedSession().rac_dataWithRequest(URLRequest)
            }
            .map { data, URLResponse in
                let string = String(data: data, encoding: NSUTF8StringEncoding)!
                return parseJSONResultsFromString(string)
            }
            .observeOn(UIScheduler())
        
    }

    func searchRequestWithEscapedQuery(query: String) -> NSURLRequest {
        return NSURLRequest()
    }
    
    func parseJSONResultsFromString(json: String) -> String {
        return ""
    }
}
```

With the error
   Cannot invoke 'map' with argument list of type ((_, _) -> _)

iv-mexx:
Seems like that is too much for type inference to work, when adding type annotations to the `map` and `flatMap` calls it works for me in Xcode 7.0:

```Swift
class ViewController: UIViewController {
    
    @IBOutlet weak var textField: UITextField!
    
    override func viewDidLoad() {
        super.viewDidLoad()
        // Do any additional setup after loading the view, typically from a nib.
        
        //RACSignal
        let searchSignal : RACSignal = textField.rac_textSignal()
        
        let searchStrings : SignalProducer<String, NSError> = textField.rac_textSignal()
            .toSignalProducer()
            .map { text in text as! String }
        
        let searchResults = searchStrings
            .flatMap(.Latest) { (query: String) -> SignalProducer<(NSData, NSURLResponse), NSError> in
                let URLRequest = self.searchRequestWithEscapedQuery(query)
                return NSURLSession.sharedSession().rac_dataWithRequest(URLRequest)
            }
            .map { (data, URLResponse) -> String in
                let string = String(data: data, encoding: NSUTF8StringEncoding)!
[...]

PR #2413, opened 2015-09-24
Add explicit type annotations to examples in Readme

iv-mexx:
This is in response to #2412 

In order to make the examples compile in the current Xcode Version (7.0), explicit type annotations are needed at some places where Xcode seems to not be able to infer the correct types automatically.

This clutters up the code, but I guess the examples should work when pasted into Xcode. Hopefully these changes can be reverted anytime soon...

Furthermore, `println` is now just `print` and I've added an explicit `self.` to the `self.parseJSONResultsFromString` call under the assumption that `parseJSONResultsFromString` is a method rather than a free function as was assumed in #2412

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ That’s easy enough:
 
 ```swift
 searchResults.startWithNext { results in
-    println("Search results: \(results)")
+    print("Search results: \(results)")
 }
 ```
 
@@ -129,7 +129,7 @@ quickest solution would be to log them, then ignore them:
         return NSURLSession.sharedSession()
             .rac_dataWithRequest(URLRequest)
             .flatMapError { error in
-                println("Network error occurred: \(error)")
+                print("Network error occurred: \(error)")
                 return SignalProducer.empty
             }
     }
@@ -152,7 +152,7 @@ let searchResults = searchStrings
             .rac_dataWithRequest(URLRequest)
             .retry(2)
             .flatMapError { error in
-                println("Network error occurred: \(error)")
+                print("Network error occurred: \(error)")
                 return SignalProducer.empty
             }
     }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,13 +77,13 @@ With each string, we want to execute a network request. Luckily, RAC offers an
 
 ```swift
 let searchResults = searchStrings
-    .flatMap(.Latest) { query in
+    .flatMap(.Latest) { (query: String) -> SignalProducer<(NSData, NSURLResponse), NSError> in
```
NachoSoto: I think this is useful in documentation in any case: type inference is good, but better to be explicit to help people understand sample code!

README.md
```diff
@@ -123,15 +123,15 @@ To remedy this, we need to decide what to do with errors that occur. The
...
-
+        
         return NSURLSession.sharedSession()
             .rac_dataWithRequest(URLRequest)
             .flatMapError { error in
-                println("Network error occurred: \(error)")
+                print("Network error occurred: \(error)")
                 return SignalProducer.empty
-            }
+        }
```
NachoSoto: This indentation was correct before (I know, Xcode still doesn't understand how to indent this...)

README.md
```diff
@@ -145,20 +145,20 @@ Our improved `searchResults` producer might look like this:
...
+        
         return NSURLSession.sharedSession()
             .rac_dataWithRequest(URLRequest)
             .retry(2)
             .flatMapError { error in
-                println("Network error occurred: \(error)")
+                print("Network error occurred: \(error)")
                 return SignalProducer.empty
-            }
+        }
```
NachoSoto: This indentation was correct

NachoSoto:
Looks good, thank you so much!
Can you just fix the indentation in those 2 places? :)

iv-mexx:
> Can you just fix the indentation in those 2 places? :)

Sure, should be correct no I think.

iv-mexx:
Ups, NOW it should be correct :smile:

NachoSoto:
Perfect! :star:

PR #2428, opened 2015-09-29
Move `sendLock` into the initializer

dapenggao:
`sendLock` is only used by `sink`, so it doesn't have to be made an instance variable and the class does not have to retain this lock either--the `sink` will retain it properly.


NachoSoto:
Good idea!

Just to make sure, since we don't have them running on Github yet: do tests still pass?

dapenggao:
Yes, all tests are passed.

Issue #2431, opened 2015-09-29
get, set, willSet, didSet for properties


paulyoung:
Given some old code that I'm trying to migrate, I'm at a loss on how to achieve the same thing without drastically changing the interface.

```swift
// Old
let _foos = [Foo]()
let foos =  [Foo]() {
    get {
        return _foos
    }
    set {
        _foos.sort { ... }
    }
}
```

```swift
// New
let foos = MutableProperty<[Foo]>([])
```

Could anyone offer any advice on the best way to do this?

NachoSoto:
I'm :+1: for adding `willSet` as a `T -> T` pure function that can perform transformations before setting a new value in `MutableProperty`.

However, I don't think I understand your example. `get` allows you to declaratively implement a getter, but there is no way to know when the value would implicitly change. If you want to map a property to an existing one you can always use `<~` operator.

paulyoung:
Yeah, I mostly just mentioned `get` since it appears in the original code.

PR #2432, opened 2015-09-29
[WIP] willSet

paulyoung:
This pull request is an attempt to find a solution to the problem described in #2431.

I consider this to be super rough, so welcome any and all feedback.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -116,7 +116,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 
 		self.willSet = willSet
 		_value = self.willSet(initialValue)
-		sendNext(observer, initialValue)
+		sendNext(observer, _value)
 	}
 
 	deinit {
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -109,7 +109,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public let producer: SignalProducer<Value, NoError>
 
 	/// Initializes the property with the given value to start.
-	public init(_ initialValue: Value, willSet: ((value: Value) -> Value)? = { $0 }) {
+	public init(_ initialValue: Value, willSet: ((Value -> Value)? = { $0 }) {
 		lock.name = "org.reactivecocoa.ReactiveCocoa.MutableProperty"
 
 		(producer, observer) = SignalProducer<Value, NoError>.buffer(1)
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -80,6 +80,8 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// underlying producer prevents sending recursive events.
 	private let lock = NSRecursiveLock()
 	private var _value: Value
+	
+	private let willSet: (Value) -> Value
 
 	/// The current value of the property.
 	///
@@ -95,7 +97,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 
 		set {
 			lock.lock()
-			_value = newValue
+			_value = willSet(newValue)
 			sendNext(observer, newValue)
 			lock.unlock()
 		}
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -80,6 +80,8 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// underlying producer prevents sending recursive events.
 	private let lock = NSRecursiveLock()
 	private var _value: Value
+	
+	private let willSet: (Value) -> Value
```
NachoSoto: I'd recommend using a different name since `willSet` is a reserved keyword.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -107,12 +109,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public let producer: SignalProducer<Value, NoError>
 
 	/// Initializes the property with the given value to start.
-	public init(_ initialValue: Value) {
+	public init(_ initialValue: Value, willSet: ((value: Value) -> Value)? = { $0 }) {
```
NachoSoto: The `value` parameter name shouldn't be necessary.

ReactiveCocoa/Swift/Property.swift
paulyoung: How about `beforeSet`?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -107,12 +109,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
 		lock.name = "org.reactivecocoa.ReactiveCocoa.MutableProperty"
 
 		(producer, observer) = SignalProducer<Value, NoError>.buffer(1)
 
-		_value = initialValue
+		self.willSet = willSet
+		_value = self.willSet(initialValue)
 		sendNext(observer, initialValue)
```
NachoSoto: Oops, this should send the transformed `intialValue`, right?
Which is good that we found because that means that's not covered by a test :D

ReactiveCocoa/Swift/Property.swift
NachoSoto: How about something like `valueTransformer`?

ReactiveCocoa/Swift/Property.swift
paulyoung: :see_no_evil:

paulyoung:
Pushed some updates based on the feedback so far.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -106,6 +106,36 @@ class PropertySpec: QuickSpec {
...
+				context("with the default value transformer") {
+					it("should not transform the value set") {
+						let property = MutableProperty(0)
+						var sentValue: Int?
+						
+						property.producer.startWithNext { value in
+							sentValue = value
+						}
+						
+						expect(property.value).to(equal(0))
```
NachoSoto: This is only testing the initial value. Maybe change the property a few times to verify that this remains true?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -106,6 +106,36 @@ class PropertySpec: QuickSpec {
...
+				context("with a custom value transformer") {
+					it("should transform the value set") {
+						let property = MutableProperty(0) { $0 + 1 }
+						var sentValue: Int?
+						
+						property.producer.startWithNext { value in
+							sentValue = value
+						}
+						
+						expect(property.value).to(equal(1))
```
NachoSoto: Ditto here, this is only verifying the initial value.

NachoSoto:
Looks great! Just a few details left to test.

paulyoung:
There's some interesting discussion on Slack about this approach that makes me think this isn't a great idea after all.

neilpa:
Agreed, this isn't something we should be adding to the API surface. There are plenty of existing signal operators that may be able to help but I don't think we should be trying to mimic KVO because multiple sources of truth suck.

ReactiveCocoaTests/Swift/PropertySpec.swift
paulyoung: Yep. Not even sending the transformed value inside of `set`.

raylillywhite:
I'd be concerned that most people would think that when they see a MutableProperty that what they put in is what they'd get out. I'd prefer that if we added something like this, that it at least be it's own class — If you see a `MappedMutableProperty` (which could conform to MutablePropertyType), then it's a lot more obvious that the value will/may be transformed.

neilpa:
> I'd prefer that if we added something like this, that it at least be it's own class

Still against this as part of RAC core. The prescribed way to handle value transformations is signal operators while properties are about caching the most recent value, generally for UI binding.

liscio:
I'm also on board with the "don't modify `MutableProperty` to do this" gang. I feel that trying to accomplish what KVO does is better handled by the kind of approach proposed in https://github.com/ReactiveCocoa/EditableProperty (where values may be modified/merged by an `Editor`).

paulyoung:
Thanks for participating everyone. I think this pull request has served it purpose which was to generate more discussion around this topic.

That said, I'm going to close since I agree with all of the points being made _against_ this approach.

NachoSoto:
:+1: thanks for the idea @paulyoung!

PR #2433, opened 2015-09-30
Fix -flatten: circular reference.

joshaber:
Previously we only nil’d our reference out on completion. So if we were disposed of instead of terminating naturally, we’d keep this block alive and everything it referenced.

Much like https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2328, I’m not sure how to test this as the objects we leak are deep within `-flatten:`.

I was (finally!) able to reproduce this and measure it in Instruments using these methods:

```objc
- (RACSignal *)doAThing {
	return [[[RACSignal
		return:@1]
		map:^id(id value) {
			return [[RACSignal return:@1] subscribeOn:[RACScheduler mainThreadScheduler]];
		}]
		flatten:1];
}

- (IBAction)start:(id)sender {
	[[[self doAThing]
		take:1]
		subscribeNext:^(id x) {

		} completed:^{

		}];
}
```

The key being that the signal has to be disposed of synchronously with its send (done by the `-take:` in this case), but subscribed to asynchronously (by the `-subscribeOn:`).

### DIFF
--- a/ReactiveCocoa/Objective-C/RACSignal+Operations.m
+++ b/ReactiveCocoa/Objective-C/RACSignal+Operations.m
@@ -520,10 +520,6 @@ static RACDisposable *subscribeForever (RACSignal *signal, void (^next)(id), voi
 		void (^completeIfAllowed)(void) = ^{
 			if (selfCompleted && activeDisposables.count == 0) {
 				[subscriber sendCompleted];
-
-				// A strong reference is held to `subscribeToSignal` until completion,
-				// preventing it from deallocating early.
-				subscribeToSignal = nil;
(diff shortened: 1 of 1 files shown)


mdiep:
Makes sense to me. :+1:

mdiep:
I talked through this a bit more with @joshaber to make sure I understood it.

In his example:

1. `take:` completes before `flatten:` because it completes when the value is sent, while `flatten:` has to wait for the signal to complete
2. In that case, the `flatten:` is disposed—it doesn't `sendCompleted`
3. Because it wasn't completing, `flatten:` was leaking

PR #2435, opened 2015-09-30
Reduce dependency on the `signal` property of `SignalType`

dapenggao:
In many places `signal` can actually be replaced by `self`, which can give rise to looser coupling. 
Other changes are taken from #2386.
All tests are passed.
PS: The indentation changes are actually replacing spaces with tabs.

### DIFF
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -99,7 +99,7 @@ public func toRACSignal<Value: AnyObject, Error: NSError>(producer: SignalProduc
 ///
 /// Any `Interrupted` events will be silently discarded.
 public func toRACSignal<Value: AnyObject, Error>(producer: SignalProducer<Value?, Error>) -> RACSignal {
-	return toRACSignal(producer.mapError { $0 as NSError })
+    return toRACSignal(producer.mapError { $0 as NSError })
 }
 
 /// Creates a RACSignal that will start() the producer once for each
@@ -107,9 +107,9 @@ public func toRACSignal<Value: AnyObject, Error>(producer: SignalProducer<Value?
 ///
 /// Any `Interrupted` events will be silently discarded.
 public func toRACSignal<Value: AnyObject, Error: NSError>(producer: SignalProducer<Value?, Error>) -> RACSignal {
-	// This special casing of `Error: NSError` is a workaround for rdar://22708537
-	// which causes an NSError's UserInfo dictionary to get discarded
-	// during a cast from ErrorType to NSError in a generic function
+    // This special casing of `Error: NSError` is a workaround for rdar://22708537
+    // which causes an NSError's UserInfo dictionary to get discarded
+    // during a cast from ErrorType to NSError in a generic function
 	return RACSignal.createSignal { subscriber in
 		let selfDisposable = producer.start { event in
 			switch event {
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -119,7 +119,7 @@ public func toRACSignal<T: AnyObject, E: NSError>(producer: SignalProducer<T?, E
 				subscriber.sendError(error)
 			case .Completed:
 				subscriber.sendCompleted()
-			default:
+			case .Interrupted:
```
NachoSoto: :thumbsup:

NachoSoto:
Looks good!

Could you separate the indentation change though?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -311,11 +311,11 @@ extension SignalType where Value: SignalProducerType, Error == Value.Error {
 	/// emitted from `signal` complete.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	private func concat() -> Signal<Value.Value, Error> {
-		return Signal<Value.Value, Error> { [signal = self.signal] observer in
+		return Signal<Value.Value, Error> { observer in
```
iv-mexx: I'm not sure about this - doesn't this now capture `self` with a strong reference (thus leading to a potential strong reference cycle), whereas it previously had captured only `signal` and not `self`?

andersio: Both captures the same reference. `Signal`'s implementation of `SignalType.signal` just returns a reference of the instance itself.

dapenggao: @iv-mexx `self` is the same object as `self.signal` at runtime. So capturing either `self.signal` or `self` will have the same effect.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -167,7 +167,7 @@ public func toRACSignal<Value: AnyObject, Error: NSError>(signal: Signal<Value?,
                 subscriber.sendError(error)
             case .Completed:
                 subscriber.sendCompleted()
-            default:
+			case .Interrupted:
```
dapenggao: This line displays perfectly well on Xcode, I don't know why it got shifted on GitHub...

ReactiveCocoa/Swift/Signal.swift
iv-mexx: Ok, thanks for the clarification :+1:

ReactiveCocoa/Swift/ObjectiveCBridging.swift
iv-mexx: Tabs vs. Spaces.
The old line was indented wit a total of 12 spaces, the new line is indented with 3 tabs.

dapenggao:
`observeWithStates` is changed to a private method for better consistency with `observe`.

ReactiveCocoa/Swift/ObjectiveCBridging.swift
NachoSoto: In any case, it's always a good idea to not mix indentation changes with other functional changes to make reviewing code easier. I know about the `w=0` flag, but still, it's a good idea to keep the git history clean :)

NachoSoto:
In any case, it's always a good idea to not mix indentation changes with other functional changes to make reviewing code easier. I know about the w=0 flag, but still, it's a good idea to keep the git history clean :)

In particular, it's going to be really hard to look back at the history and figure out what changed in here:
screen shot 2015-10-01 at 12 09 11

Could you amend the commits to undo the indentation changes?

dapenggao:
@NachoSoto So in that particular case, the entire function is shifted because it became nested in an extension.

NachoSoto:
> But in the case above, the entire function is shifted because it became nested in an extension.

Oh my bad, makes sense :+1:

PR #2439, opened 2015-10-02
SignalProducer: added overloads to all operators for Signal

NachoSoto:
This also introduces a `lift` overload that allows us to implement similar overloads to other operators if we need to.
Note that I haven't added any tests since these operator are written in terms of `lift` which **is** tested.

## Overloaded operators:
- `takeUntil`
- `combineLatestWith`
- `sampleOn`
- `takeUntilReplacement`
- `zipWith`

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -397,7 +397,15 @@ extension SignalProducerType {
 	/// yielded from start().
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func lift<U, F, V, G>(transform: Signal<Value, Error> -> Signal<U, F> -> Signal<V, G>) -> Signal<U, F> -> SignalProducer<V, G> {
-		return self.lift(transform)
+		return { otherSignal in
+			return SignalProducer { observer, outerDisposable in
+				self.startWithSignal { signal, disposable in
+					outerDisposable.addDisposable(disposable)
+
+					outerDisposable += transform(signal)(otherSignal).observe(observer)
+				}
+			}
+		}
 	}
 	
 	/// Maps each value in the producer to a new value.
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -389,6 +389,17 @@ extension SignalProducerType {
 		}
 	}
 
+	/// Lifts a binary Signal operator to operate upon a Signal and a SignalProducer instead.
+	///
+	/// In other words, this will create a new SignalProducer which will apply
+	/// the given Signal operator to _every_ Signal created from the two
+	/// producers, just as if the operator had been applied to each Signal
+	/// yielded from start().
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func lift<U, F, V, G>(transform: Signal<Value, Error> -> Signal<U, F> -> Signal<V, G>) -> Signal<U, F> -> SignalProducer<V, G> {
+		return self.lift(transform)
+	}
+	
 	/// Maps each value in the producer to a new value.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func map<U>(transform: Value -> U) -> SignalProducer<U, Error> {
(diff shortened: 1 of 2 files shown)


andymatuschak:
We should probably get an opinion on this from @ReactiveCocoa/reactivecocoa.

The implications here are broad: this pull request suggests that for each binary signal operator we add to `Signal`, we should add a `SignalProducer`-vs-`Signal` convenience overload which lifts the operator into `SignalProducer`.

The motivation is strong: as we started to adopt RAC 4 in our project, I noticed that we were seriously overusing `SignalProducer` in places where `Signal` would do. We should be strongly encouraging people to use `Signal` when possible, but mixing and matching requires developers to understand `lift`. I'd argue that developers should only need to understand `lift` when they get to the savviness point of creating their own signal operators.

More general comment: if we go ahead with this, I think we should land this for all signal operators simultaneously, rather than adding them piecemeal.

NachoSoto:
> More general comment: if we go ahead with this, I think we should land this for all signal operators simultaneously, rather than adding them piecemeal.

Totally agreed!

NachoSoto:
- takeUntilReplacement

NachoSoto:
This is ready for review! :star2:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -481,6 +503,20 @@ extension SignalProducerType {
...
 
+	/// Forwards the latest value from `self` whenever `sampler` sends a Next
+	/// event.
+	///
+	/// If `sampler` fires before a value has been observed on `self`, nothing
+	/// happens.
+	///
+	/// Returns a producer that will send values from `self`, sampled (possibly
+	/// multiple times) by `sampler`, then complete once both input producers have
+	/// completed, or interrupt if either input producer is interrupted.
```
kastiglione: Technically "both input producers" and "either input producer" isn't precise for this `Signal` overload.

kastiglione:
I'm :+1:. We should totally add a cpp pass to the build and make a macro to define these operators :trollface:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -389,6 +389,17 @@ extension SignalProducerType {
...
 
+	/// Lifts a binary Signal operator to operate upon a Signal and a SignalProducer instead.
+	///
+	/// In other words, this will create a new SignalProducer which will apply
+	/// the given Signal operator to _every_ Signal created from the two
+	/// producers, just as if the operator had been applied to each Signal
+	/// yielded from start().
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func lift<U, F, V, G>(transform: Signal<Value, Error> -> Signal<U, F> -> Signal<V, G>) -> Signal<U, F> -> SignalProducer<V, G> {
+		return self.lift(transform)
```
andersio: I'm not sure if I've missed something, but this seems an infinite recursion.

NachoSoto: No, this is using the other `lift` overload (and thankfully `clang` would warn if that weren't true! :))

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Good catch!

ReactiveCocoa/Swift/SignalProducer.swift
andersio: May I ask if there is one in `SignalProducerType`? I can see only one `lift` implementation which takes a `Signal -> Signal -> Signal` transform, but it returns `SignalProducer -> SignalProducer` and thus the return type is not matched. It doesn't seem there is an overload, and IMO the compiler doesn't complain because the argument type and return type both matches. I tried to reproduce it (http://pastebin.com/gLy1dwHt) in playground and it crashes.

NachoSoto: Ha you're absolutely right, @andersio.
I'm surprised this isn't a warning :(

NachoSoto: Fixed! And I wrote tests for this new `lift` operator.

NachoSoto:
Bump. Who wants to be the hero who presses the green button? :D

mdiep:
> Who wants to be the hero who presses the green button?

:hand: 



This all checks out! :+1:

NachoSoto:
Thank you sir! 
lol-christian-bale

PR #2449, opened 2015-10-06
flatten and flatMap are now possible for all 4 combinations of Signal+SignalProducer

### LINKED ISSUE #2422: Flatten for Signal<Signal<T, E>, E>
Is there a reason why a function like
```swift
extension SignalType where Value: SignalType, Error == Value.Error {
    public func flatten(strategy: FlattenStrategy) -> Signal<Self.Value.Value, Self.Error>
}
```

would not be possible?

NachoSoto:
Fixes #2422. This is a follow up to #2439.

## Description:

Before this change, `flatten` and `flatMap` were only defined for `Signal<SignalProducer<T, E>, E>` and `SignalProducer<SignalProducer<T, E>, E>`. However, after a long conversation with @andymatuschak (who helped a lot and with whom I pair programmed to make this change) we realize that there's no reason why these operators should not exist for `Signal<Signal<T, E>, E>` and `Signal<Signal<T, E>, E>`.

## Why:
We came up with some examples for when it would be useful:
- `Signal<SignalProducer<T, E>, E>.flatten`:
  - `.Merge`: You want to flatten the results of some network request that gets created upon tapping on a button.
  - `.Concat`: same as previous example, except you want results to come in order. 
  - `.Latest`: same as previous example, except you want to cancel any request that's already running

- `Signal<Signal<T, E>, E>.flatten`:
  - `.Merge`: You want to flatten the values created by some clock created upon tapping on a button. Both of these things, button taps, and a "clock" are *hot* streams.
  - `.Concat`: Same as the previous example, but you want to wait for previous "clocks" to finish ticking.
  - `.Latest`: Same as previous example, but you only want to observe the _ticks_ from the last "clock".

Due to the absence of these operators (as well as those introduced in #2439), we believe that `SignalProducer` was being favored, which is the opposite of what we wanted to accomplish with `RAC 3/4`: limit side effects by encouraging the usage of *hot* `Signal`s.

## Notes:
- I haven't added any tests for these new operators since we managed to keep them all defined in terms of the canonical implementation: `Signal<SignalProducer<T, E>, E>.flatten`.
- I added an initializer to `SignalProducer` that constructs a producer given a `Signal`, with no side-effects.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -262,6 +262,30 @@ public enum FlattenStrategy: Equatable {
 	case Latest
 }
 
+extension SignalType where Value: SignalType, Error == Value.Error {
+	/// Flattens the inner signals sent upon `signal` (into a single signal of
+	/// values), according to the semantics of the given strategy.
+	///
+	/// If `signal` or an active inner signal emits an error, the returned
+	/// signal will forward that error immediately.
+	///
+	/// `Interrupted` events on inner signals will be treated like `Completed`
+	/// events on inner signals.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Error> {
+		switch strategy {
+		case .Merge:
+			return self.map(SignalProducer.init).merge()
+
+		case .Concat:
+			return self.map(SignalProducer.init).concat()
+
+		case .Latest:
+			return self.map(SignalProducer.init).switchToLatest()
+		}
+	}
+}
+
 extension SignalType where Value: SignalProducerType, Error == Value.Error {
 	/// Flattens the inner producers sent upon `signal` (into a single signal of
 	/// values), according to the semantics of the given strategy.
@@ -295,6 +319,17 @@ extension SignalType {
 	/// signal will forward that error immediately.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func flatMap<U>(strategy: FlattenStrategy, transform: Value -> SignalProducer<U, Error>) -> Signal<U, Error> {
+		return map(transform).flatten(strategy)
+	}
+
+	/// Maps each event from `signal` to a new signal, then flattens the
+	/// resulting signals (into a signal of values), according to the
+	/// semantics of the given strategy.
+	///
+	/// If `signal` or any of the created signals emit an error, the returned
+	/// signal will forward that error immediately.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func flatMap<U>(strategy: FlattenStrategy, transform: Value -> Signal<U, Error>) -> Signal<U, Error> {
 		return map(transform).flatten(strategy)
 	}
 }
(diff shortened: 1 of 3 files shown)


paulyoung:
:+1: I found this trying to figure out how to use `flatten` on `Signal<Signal<T, E>, E>`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -262,6 +262,30 @@ public enum FlattenStrategy: Equatable {
...
+		switch strategy {
+		case .Merge:
+			return self.map(SignalProducer.init).merge()
+
+		case .Concat:
+			return self.map(SignalProducer.init).concat()
+
+		case .Latest:
+			return self.map(SignalProducer.init).switchToLatest()
+		}
```
mdiep: Since `flatten` is already defined for `Signal<SignalProducer<T, E>, E>`, you can just call `flatten` directly here.

```swift
return self.map(SignalProducer.init).flatten(strategy)
```

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -19,6 +19,17 @@ public struct SignalProducer<Value, Error: ErrorType> {
...
+	/// Initializes a SignalProducer that will emit the same events as the given signal.
+	///
+	/// If the Disposable returned from start() is disposed or a terminating
+	/// event is sent to the observer, the given signal will be
+	/// disposed.
+	public init<S: SignalType where S.Value == Value, S.Error == Error>(signal: S) {
+		self.init { observer, disposable in
+			disposable += signal.observe(observer)
+		}
+	}
```
mdiep: It seems a little weird to create a `SignalProducer` from a `Signal`. Not necessarily wrong, but a little weird. :confused:

I think you've done this because `merge()`, `switchToLatest()`, and `contact()` are defined on `Signal<SignalProducer<T, E>, E>`. Would it be better to define those on `Signal<Signal<T, E>, E>` instead and implement the `SignalProducer` version with the `Signal` version instead?

mdiep:
This looks correct, but I had a suggestion and a question. :smile:

ReactiveCocoa/Swift/Signal.swift
NachoSoto: Ooooh good idea :D

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: > It seems a little weird to create a SignalProducer from a Signal. Not necessarily wrong, but a little weird. :confused:

I don't think it's that weird. You can think about this constructor as the `pure` or `unit` functions: it creates a `SignalProducer` with only the `Signal` part of it, but without side effects.

> I think you've done this because merge(), switchToLatest(), and contact() are defined on Signal<SignalProducer<T, E>, E>. Would it be better to define those on Signal<Signal<T, E>, E> instead and implement the SignalProducer version with the Signal version instead?

That would mean that we need a constructor of `Signal` that takes a `SignalProducer` (the opposite version of this), right?
Unfortunately that wouldn't be "pure", because in order to construct that we'd have side effects, unlike in this version.

I hope that answers your question? It was a great question though!

ReactiveCocoa/Swift/Signal.swift
NachoSoto: Done!

NachoSoto:
Rebased!

ReactiveCocoa/Swift/SignalProducer.swift
mdiep: :+1:

NachoSoto:
Thank you!

PR #2450, opened 2015-10-07
Fixed infinite recursion in RACScheduler.schedule

NachoSoto:
We ran into this infinite recursion in the Khan Academy app, and @andymatuschak traced down the issue to this.
I added (previously lacking!) test coverage for this and fixed the implementation.

### DIFF
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -15,7 +15,8 @@ extension RACScheduler: DateSchedulerType {
 	}
 
 	public func schedule(action: () -> ()) -> Disposable? {
-		return (self.schedule(action) as RACDisposable) as Disposable?
+		let disposable: RACDisposable = self.schedule(action) // Call the Objective-C implementation
+		return disposable as Disposable?
 	}
 
 	public func scheduleAfter(date: NSDate, action: () -> ()) -> Disposable? {
--- a/ReactiveCocoa/Swift/ObjectiveCBridging.swift
+++ b/ReactiveCocoa/Swift/ObjectiveCBridging.swift
@@ -15,7 +15,7 @@ extension RACScheduler: DateSchedulerType {
 	}
 
 	public func schedule(action: () -> ()) -> Disposable? {
-		return self.schedule(action)
+		return (self.schedule(action) as RACDisposable) as Disposable?
 	}
 
 	public func scheduleAfter(date: NSDate, action: () -> ()) -> Disposable? {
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/ObjectiveCBridging.swift
```diff
@@ -15,7 +15,7 @@ extension RACScheduler: DateSchedulerType {
 	}
 
 	public func schedule(action: () -> ()) -> Disposable? {
-		return self.schedule(action)
+		return (self.schedule(action) as RACDisposable) as Disposable?
```
andymatuschak: I think it might be slightly clearer what you're doing if you were to separate this into two lines and add a comment like:

```swift
let disposable: RACDisposable = self.schedule(action) // Call the Objective-C implementation
return disposable as Disposable?
```

Now that I look at it… why is this method necessary? If someone calls the original Obj-C implementation and gets a `RACDisposable` back, can't that automatically be assigned to a `Disposable?` l-val as is happening here?

NachoSoto: Totally, I'll separate each part.

I think, even if the method IS redundant, it serves as documentation. Otherwise users might wonder why this implements to `DateSchedulerType`.

Though now that I think about it, it probably is necessary since the return type is technically different.

andymatuschak: This does work in a playground:

```
protocol P {}
class TestP: P {}

class A {
	static func test() -> TestP! {
		return TestP()
	}
}

let p: P? = A.test()
```

NachoSoto: I'm confused, I think that example is slightly different?

I updated the PR with your suggested implementation though :)

andymatuschak: Hm, why's it different? Isn't it identical? Name transformations:

```
protocol Disposable {}
class RACDisposable: Disposable {}

class RACScheduler {
  func schedule() -> RACDisposable! {
    return RACDisposable()
  }
}

let disposable: Disposable? = RACScheduler().schedule()
```

I guess what I'm trying to show is: this kind of covariant coercion to l-val types is the norm in Swift. e.g. we do things like this all the time:

```
zip(someFunctionReturningAnArray(), aFunctionReturningASequence()
```

The return values from those functions are being transparently covariantly cast to `SequenceType`.

NachoSoto: Ooooh yes, we were talking about different things.
I meant that in this example:

```swift
protocol Disposable {}
class RACDisposable: Disposable {}

class RACScheduler: DateSchedulerType {
  func schedule() -> Disposable! {
     // This coercion works.
     // But I think it's good to implement the method
     // explicitly for documentation purposes.
     // And I'm not 100% sure Swift would understand that
     // that this DOES implement DateSchedulerType without it.
     return RACDisposable()
  }
}
```

andymatuschak: Indeed, you're right, just tried in a playground, and swiftc's not smart enough to see that.

PR #2452, opened 2015-10-08
Don't sign framework for iOS simulator

gfontenot:
There's no real need to sign the iOS framework for simulator targets.

This will also let users build RAC on CI systems using Carthage that don't have a developer cert installed.

This is probably worth applying to `master` as well, wasn't sure which direction I should make this change, or when this branch was going to end up getting merged.

I'm also not sure what the best way to handle watchOS and tvOS targets are. I'm assuming we need to make this same change there. Maybe someone with more insight into those platforms knows the answer?

related: antitypical/Result#97

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -2644,6 +2644,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 57A4D2461BA13F9700F7D4B1 /* tvOS-Framework.xcconfig */;
 			buildSettings = {
+				"CODE_SIGN_IDENTITY[sdk=appletvsimulator*]" = "";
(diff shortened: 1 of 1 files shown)


gfontenot:
Updated this to also not sign tvOS and watchOS simulators.

paulyoung:
Could/should this be added to https://github.com/jspahrsummers/xcconfigs?

neilpa:
> This is probably worth applying to master as well, wasn't sure which direction I should make this change, or when this branch was going to end up getting merged.

This branch should be fine. There's already talk of merging to `master` soon. I may even have a chance this weekend.

gfontenot:
This is still an issue for me. Is there anything I can do to move it forward?

neilpa:
Sorry about that. Can you re-target against `master` and I'll merge it.

gfontenot:
Yep, incoming.

PR #2462, opened 2015-10-13
added support for more Type flexible takeUntil and sampleOn

mishagray:
The Event type can be anything now.  So if you are using a Signal that isn' t (), than it still works fine.
ErrorType must still be NoError.

Added an assertionFailure of you you to call SendError() on a Signal.Sink of NoError

### DIFF
--- a/ReactiveCocoa/Swift/Event.swift
+++ b/ReactiveCocoa/Swift/Event.swift
@@ -175,6 +175,12 @@ public func sendError<Value, Err: ErrorType>(sink: Event<Value, Err>.Sink, _ err
 	sink(.Error(error))
 }
 
+/// Puts an `Error` event into the given sink.
+public func sendError<Value>(sink: Event<Value, NoError>.Sink, _ error: NoError) {
+	assertionFailure("Don't send an Error to a Signal of type NoError NoError!")
+	sink(.Error(error))
+}
+
 /// Puts a `Completed` event into the given sink.
 public func sendCompleted<Value, Err: ErrorType>(sink: Event<Value, Err>.Sink) {
 	sink(.Completed)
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/Swift/Event.swift
```diff
@@ -175,6 +175,12 @@ public func sendError<Value, Err: ErrorType>(sink: Event<Value, Err>.Sink, _ err
 	sink(.Error(error))
 }
 
+/// Puts an `Error` event into the given sink.
+public func sendError<Value>(sink: Event<Value, NoError>.Sink, _ error: NoError) {
+	assertionFailure("Don't send an Error to a Signal of type NoError NoError!")
```
NachoSoto: This is less strict than before: it's now a *runtime* assertion instead of something we can *know at compile time*. That's why this method does not make much sense.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -829,7 +829,7 @@ extension SignalType {
 	/// multiple times) by `sampler`, then complete once both input signals have
 	/// completed, or interrupt if either input signal is interrupted.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func sampleOn(sampler: Signal<(), NoError>) -> Signal<Value, Error> {
+	public func sampleOn<S: SignalType where S.Error == NoError>(sampler: S) -> Signal<Value, Error> {
```
NachoSoto: This is on purpose: the current signature forces users to explicitly ignore values rather than that happening transparently.

NachoSoto:
Thanks for the PR! Hopefully my comments help clarify why those method signatures are the way they are :+1:

PR #2463, opened 2015-10-13
Added support for using faster OSAtomicCompareAndSwap32

mishagray:
Added support for using faster OSAtomicCompareAndSwap32 to Atomic.swift

For sharing Ints, Bools, it's even faster to use OSAtomicCompareAndSwap32 etc.  Which should result in faster composition of Signals.

Added AtomicInt32, AtomicInt64 and AtomicBool.
Changed the Signal disposable Atomic<Bool> to use AtomicBool()

Added the @testable import and added AtomicSpec.swift BACK into the test targets.

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0995ECFE1BCC7C4B000AE3F1 /* AtomicSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = D0C312EE19EF2A7700984962 /* AtomicSpec.swift */; settings = {ASSET_TAGS = (); }; };
+		0995ECFF1BCC7C4C000AE3F1 /* AtomicSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = D0C312EE19EF2A7700984962 /* AtomicSpec.swift */; settings = {ASSET_TAGS = (); }; };
 		314304171ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.h in Headers */ = {isa = PBXBuildFile; fileRef = 314304151ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		314304181ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.m in Sources */ = {isa = PBXBuildFile; fileRef = 314304161ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.m */; };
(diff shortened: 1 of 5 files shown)


dapenggao:
How much performance gain can this get?

NachoSoto:
This seems like a ton of extra complexity for a non-measured performance boost. Have you ran any test to get an idea of how much faster this is?

mishagray:
You can read some specifics.  In general OSSpinLocks themselves are actually implemented using OSAtomic Operators with  a loop on around a an atomic compare.   This makes sense when you need to share a class or object because you need to searlize access to a complex structure. 

But it's kind of triple overkill to use lock/unlock operation to lock access for a Bool or Int.  You are actually allocating the SPINLock which itself is using OSAtomic increment/decrement/swap so you can coordinate another Int operation.   So it's probably 2-3x the memory, and 2-3x the number operations.  (Lock, modify. Unlock) vs (modify).     In practice we are talking 5ms vs 10-12ms per "lock".   

https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Multithreading/ThreadSafety/ThreadSafety.html#//apple_ref/doc/uid/10000057i-CH8-124887

NachoSoto:
If it's really THAT much performance gain then it's certainly  something to be considered. However, I'd recommend that you write a few performance tests with `XCTest` and see how much this added complexity gets us.

liscio:
At a high level, this change by @mishagray is something I can get behind, because those `OSSpinLock*` calls have always bugged me for similar reasons (to guard an int or bool flag?! really?!).

To address @NachoSoto's concern about added complexity, I think that @mishagray can accomplish the same performance win by factoring out an `AtomicType` protocol that works with the existing `Atomic` class, but adds specializations for when the storage type is an Int or Bool, and only in the cases that we _need_ at the moment (e.g. Nobody is calling compare&swap yet, etc.)

And yes, :+1: to the suggestion that a benchmark would be great.

kastiglione:
> In practice we are talking 5ms vs 10-12ms per "lock".

I think these numbers should be removed, or updated with backing info. Which device(s)? Under which conditions? Benchmarks? Those numbers struck me as high, and without anything else to go on, this seems like premature optimization.

JaviSoto:
To get behind @NachoSoto's concern, I think it's especially scary to have all this complexity in code that's really hard to test for correctness. We can check that the code works in all of RAC's unit tests, but not if we have race conditions. However, it's easier to reason about correctness and prevent introducing bugs with just *one* Atomic type.
However, `Atomic` is used a lot, so if there's a performance gain, I'm definitely in favor of this.

>Those numbers struck me as high, and without anything else to go on, this seems like premature optimization.

Yeah, especially what's going to affect whether this has a positive effect in performance in whether Atomic is a hot-spot or not. If this code is rarely called, a micro-optimization like this, even if it saves 20ms won't make a difference.

neilpa:
> factoring out an AtomicType protocol that works with the existing Atomic class, but adds specializations for when the storage type is an Int or Bool

This sounds like a good approach to explore (assuming benchmarks show we need a change).

Issue #2384, opened 2015-09-18
CombineLastest a lot of signals on a background QueueScheduler causes a crash


jeremiegirault:
You can reproduct with the following project:

https://github.com/jeremiegirault/rac-bug
- Without the scheduler, combineLatest works
- Without a lot of signals combineLatest works
  But
- With a lot of signals (100 in my case) + QueueScheduler, combineLatest crashes

jdauphant:
:+1:

NachoSoto:
Can you attach a log of the crash?

jeremiegirault:
You can find the two crashes here:

https://gist.github.com/jeremiegirault/2ebae389b7964fff1c89

neilpa:
Looks like stack overflow, there's > 500 frames on the crashing thread. We need to re-implement the batched versions of `combineLatest` to not build signal chains iteratively which is what caused this problem.

jeremiegirault:
what is strange, is that it occurs only when using startOn.
When just creating the signal, it does not stack overflow. Any idea why this would occur ?

neilpa:
Yea, it's the actual sending of events that's problematic which is synchronously calling through the entire chain. You end up with ~5 frames on the stack per signal in the chain when an event is sent.

As a workaround for now you could try smaller batches of ~10 signals per `combineLatest` and then wrap all those in an outer `combineLatest`

joshaber:
The main thread also has a bigger stack by default, so it'd take longer to overflow.

neilpa:
> The main thread also has a bigger stack by default, so it'd take longer to overflow.

Interesting, learned something new today. Also, those stacks were bigger than 500 frames, but it looks like the crash report truncates them.

jdauphant:
We can see around 3000 stack frames in the debugger

neilpa:
I've got a partial fix locally that I hope to wrap up later tonight.

jdauphant:
Have you news about the issue ?

neilpa:
I've been AFK for the last few weeks so haven't been able to handle this

mishagray:
Ug.  I may have a fork I played with that re-implemented combineLatest and coverted the tuple into an array[Any] and used a single semaphore (casting things to Any first, and then Back to T).   
It's how I did the combine() implementation in FutureKit.   

Let me see if I can post it somewhere.

mishagray:
[...]

PR #2466, opened 2015-10-13
Update Result to 0.4.5

ikesyo:
See https://github.com/antitypical/Result/issues/90 for the details.


NachoSoto:
I don't know if we'll be making more releases of RAC 3. I'm not against this change, but I think we should merge #2189 into master already.

NachoSoto:
This is not relevant anymore, right?

neilpa:
I was planning on back-porting the fix for #2384 to RAC 3 since it's pretty bad.

mdiep:
This doesn't seem relevant anymore. At the very least, this would need to target a maintenance branch.

PR #2474, opened 2015-10-17
Update `SignalProducerSpec` to the new `Observer` API

dapenggao:

PR #2475, opened 2015-10-17
Change `sink` to `observer`

dapenggao:
Continuation of #2474.


NachoSoto:
Good catch @paulYoung!

PR #2477, opened 2015-10-17
Update BasicOperators.md

chenjsa:
Replaced 'println' with 'print'.


neilpa:
Awesome! Thanks for the PR.

PR #2485, opened 2015-10-19
MutableCollectionProperty

pepibumur:
### What?
I thought it would be useful having a custom property in ReactiveCocoa for collections that notified about changes in the collection with more details *(elements updated, removed, inserted, ...)*. That way if you use the property to represent for example a TableViewCollection you can update on the view only the elements that have actually changed.

I originally implemented the feature on this repository https://github.com/gitdoapp/RAC-MutableCollectionProperty

The component  is *fully tested*.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -130,6 +130,155 @@ extension MutableProperty: SinkType {
 	}
 }
 
+public enum CollectionChange<T> {
+    case Deletion(Int, T)
+    case Addition(Int, T)
+    case Insertion(Int, T)
+    case Replaced(Int, T)
+    case Replacement([T])
+    case StartChange
+    case EndChange
+}
+
+public final class MutableCollectionProperty<T>: PropertyType {
+
+    public typealias Value = [T]
+
+    
+    // MARK: - Private attributes
+
+    private let _valueObserver: Signal<Value, NoError>.Observer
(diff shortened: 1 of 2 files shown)


NachoSoto:
Thanks for opening a PR!

I don't think this belongs as part of `RAC` though, I think it makes much more sense as a separate library (which I just starred, it looks awesome!). What do you think?

neilpa:
Agreed, this should remain separate from RAC.

Also, this is more complicated than necessary. I've written some collection change notification systems that looked similar. It took me forever to realize that insert and remove is all you need with element and index. Everything else (move, replace, etc.) is a composite and should be defined in terms of them.

Often the reason for extra events is batching and animating. I'd recommend keeping these concerns separate and limiting yourself to this.

```swift
enum Event<T> {
    case Insert(T, Int)
    case Remove(T, Int)
    case Composite([Event])
}
```

The `Composite` case could even be removed if you buffer before mapping to animations.

pepibumur:
Hey @NachoSoto and @neilpa , I can keep it in that library. I wasn't sure if it could fit with the core library components. Also thanks for the tips @neilpa , I'll apply your suggestions. I was actually thinking about the same but ended up adding events which aren't needed at all.

Thanks you both, closing this PR then.

neilpa:
@pepibumur Awesome! I also opened an issue in that project in case you have other questions for me about this stuff. I've given it quite a bit of thought in past projects.

PR #2489, opened 2015-10-20
Make Bag and Atomic public

neilpa:
I've needed access to these a few times and was about to resort to copying again


NachoSoto:
This is good for now.
If we determine that #2463 is actually a huge performance boost I would strongly suggest taking `Atomic` out as a dependency. Then I wouldn't be too worried about `Atomic` being too complex if it's tested in isolation.

Issue #2453, opened 2015-10-09
How do I cancel the work of inner producers?


bdolman:
Let's say I have parallel work I want to do, such as downloading multiple images. Each of these downloads is represented by a SignalProducer. I create them and flatten them using the Merge strategy. When I start the outer producer and get a disposable I expected that disposing of that would also dispose of the inner producers. What actually happens is that after disposal, my outer producer stops receiving events, but the actual work being done by the inner producers continues, because they have not been disposed.

How do I get the actual work being done by the inner producers to stop? Here's some test code that illustrates the scenario:

```swift
func producerOfValues(values: [String]) -> SignalProducer<String, NoError> {
    return SignalProducer<String, NoError> {observer, disposable in
        for value in values {
            sleep(1)
            print("doing work: \(value)")
            sendNext(observer, value)
            if disposable.disposed { // 'disposed' is never true
                print("producer disposed")
                break
            }
        }
        sendCompleted(observer)
    }
}

let (signal, sink) = SignalProducer<SignalProducer<String, NoError>, NoError>.buffer(2)
let flattened = signal.flatten(.Merge)

let letterProducer = producerOfValues(["a", "b", "c"])
let numberProducer = producerOfValues(["1", "2", "3"])

sendNext(sink, letterProducer.startOn(QueueScheduler()))
sendNext(sink, numberProducer.startOn(QueueScheduler()))
sendCompleted(sink)

let disposable = flattened.startWithNext {
    print($0)
}
disposable.dispose()
```

Output:

```text
doing work: a
doing work: 1
doing work: b
doing work: 2
doing work: c
doing work: 3
```

NachoSoto:
Hmm this seems like a bug. I would expect the disposal to cancel the inner producers as well, but I'm not 100% sure. @andy_matuschak do you know if this is the expected behavior? It feels wrong.

andymatuschak:
Agreed: that's how `-[RACSignal flatten:]` worked.

stevebrambilla:
I was about to create a new issue before seeing this one, but I think what I've been digging into this afternoon is the same underlying issue.

After updating my RAC dependency I noticed that some of my tests were failing due to inner producers continuing to be started after their outer producers have been disposed of (due to a `take(1)` or similar).

I've been using this test case to debug the issue:

```swift
it("should not start further inner producers after outer producer is disposed") {
[...]

PR #2492, opened 2015-10-21
Added operations not, and, or to SignalProducerType

mpurland:


neilpa:
Thanks for the PR. However, as standalone signal operators these aren't that useful. Generally you'll already have `map` block you can incorporate this logic into. 

I'm experimenting with similar ideas for boolean properties which could get rolled into RAC if they prove useful. c.f. neilpa/Rex#48.

mpurland:
@neilpa I like the idea. It would be nice if those predicate properties init allowed inclusion of their SignalProducer counterparts. 

Is the idea going forward to only improve existing `core` functionality in ReactiveCocoa but leave other operations to extension frameworks?

neilpa:
> Is the idea going forward to only improve existing core functionality in ReactiveCocoa but leave other operations to extension frameworks?

Not strictly, but we don't want to add new operators that increase the API surface when they can easily be achieved with existing operators.

neilpa:
> @neilpa I like the idea. It would be nice if those predicate properties init allowed inclusion of their SignalProducer counterparts.

I've got a few ideas to improve that PR, just haven't gotten around to it yet.

NachoSoto:
> Not strictly, but we don't want to add new operators that increase the API surface when they can easily be achieved with existing operators.

Especially when the focus is on getting RAC ~~3~~4 stable :)

mpurland:
> Especially when the focus is on getting RAC 34 stable :)

I would love to help with that. What needs attention?

NachoSoto:
We're tracking that [here](https://github.com/ReactiveCocoa/ReactiveCocoa/milestones/4.0). There's really only one major thing blocking: #2453 (and maybe #2360)

neilpa:
I'd also add waiting for Xcode 7.1 to GM in case it requires more breaking changes. I don't want to be forced to release a 5.0 anytime soon.

NachoSoto:
That's out already :)

neilpa:
Oh, it was still 7.0 in the appstore a few days ago. I haven't kept on top of these things as much lately.

PR #2501, opened 2015-10-23
Rename schemes to match with watchOS one

ikesyo:
Removing a space would be good.


NachoSoto:
Yup! Thanks :)

ikesyo:
:sparkling_heart:

PR #2503, opened 2015-10-23
Remove SWIFT_WHOLE_MODULE_OPTIMIZATION build setting

ikesyo:
That was for Swift 1.2.


NachoSoto:
Can you verify that this is still set as part of `-O` from the `xcconfig`?

ikesyo:
Do you mean that we should still disable WMO (set `-O`, not `-Owholemodule` to `SWIFT_OPTIMIZATION_LEVEL `)?

NachoSoto:
No no, just making sure that `-O` still contains `-Owholemodule` :)

ikesyo:
Oh, thank you. `SWIFT_OPTIMIZATION_LEVEL` is `-Owholemodule`, which is from `*.xcconfig`s:

NachoSoto:
Yup, perfect!

PR #2505, opened 2015-10-23
Event.Error → Event.Failed

bencochran:
#2360 again, but against master

My last status was:
> There are a few bits I still think could use some clarity / outside opinions:
> 
> 1. `flatMapError` vs `flatMapFailure`, and relatedly, `mapError` vs `mapFailure`
> 1. The name `promoteErrors` throws me off a bit. `promoteFailures` doesn’t seem right either. ¯\\\_(ツ)_/¯

As it stands, I’m personally feeling no huge need to change `flatMapError` or `promoteErrors` to be `failure`-based. So unless others feel strongly, this should be safe to review.


NachoSoto:
:shipit: this looks great

bencochran:
Thank you!

PR #2509, opened 2015-10-24
Hardcode the framework name to fix issues arising from #2398

tonyarnold:
So the "feature" in Xcode that made my "fix" in #2398 was fixed in Xcode 7.1. 

Come at me bro

Yes, I should have hardcoded this from the start. Come at me.

PR #2510, opened 2015-10-24
Revert #2509

neilpa:
This is to further discuss #2509. I don't understand what it's supposed to fix. I can't build without reverting it since Xcode automatically edits and breaks the schemes each time I open the project.


tonyarnold:
The tagged 4.0a2 produces framework bundles named "ReactiveCocoa-Mac.framework", "ReactiveCocoa-iOS.framework", and so on. The framework's module names reflect the bundle naming — it's wrong. The PR we're having trouble with was reverting a mistake I made with my earlier fix (to rely on TARGET_NAME). 

I've pushed further fixes to #2509, but I'm not sure if we can just reopen and merge those guys in. I'll open another PR — reverting these commits via this PR won't fix the underlying issue, sadly.

neilpa:
I see

PR #2511, opened 2015-10-24
Fix the test target bundle identifiers and names

tonyarnold:
This is a follow on to #2509 — the test targets ended up broken because they have no PRODUCT_NAME and no PRODUCT_BUNDLE_IDENTIFIER. 

This PR supersedes #2510.


ReactiveCocoa.xcodeproj/project.pbxproj
```diff
@@ -2992,6 +2996,8 @@
 				GCC_NO_COMMON_BLOCKS = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
+				PRODUCT_BUNDLE_IDENTIFIER = "org.reactivecocoa.$(PRODUCT_NAME:rfc1034identifier)-Tests";
+				PRODUCT_NAME = "$(PROJECT_NAME)";
```
neilpa: You need to prefix `Tests` here as well. I checked out your branch and the tests don't build without it.

neilpa: Or you need to set `TARGET_NAME` differently for the tests because they're getting the same name as the framework

neilpa:
Hmm, my above comments still don't fix it. Maybe we just need to ditch some of the xcconfig stuff because it seems utterly broken in 7.1

ReactiveCocoa.xcodeproj/project.pbxproj
tonyarnold: Which should be OK? `ReactiveCocoa.framework` and `ReactiveCocoa.xctest`, right? I can't get the dependencies to build with Xcode 7.1 :(

tonyarnold: _Backs away slowly, alcohol in hand_

neilpa: > The thing that is broken is that ENABLE_BITCODE is enabled for the OS X targets

This needs to be explicitly disabled for the mac tests. It already is for the framework

Issue #2515, opened 2015-10-26
Deadlock on `buffer(...)`


AlexeyDemedetskiy:
Hi. Sorry for giant stack trace. I've caught deadlock when I am try to bind property to signal in signal fire.

At short: 
1) My timer is fired. (#68)
2) I assign new value to mutable property (#62)
3) Controller 1 observe change and decide to push Controller 2. (#19)
4) Controller 2 try to bind to the same signal as Controller 1 minded. (#12)
5) ???.
6) Deadlock.

RAC version 4.0 alpha 1
```
#0	0x000000019ab34c6c in semaphore_wait_trap ()
#1	0x000000019ac0e97c in _os_semaphore_wait ()
#2	0x0000000101d4d428 in _dispatch_barrier_sync_f_slow ()
#3	0x0000000100abe5d0 in static SignalProducer.(buffer<A, B where ...> (SignalProducer<A, B>.Type) -> (Int) -> (SignalProducer<A, B>, (Event<A, B>) -> ())).(closure #1) at /Users/project/Carthage/Carthage/Checkouts/ReactiveCocoa/ReactiveCocoa/Swift/SignalProducer.swift:154
#4	0x0000000100aa9050 in partial apply for static SignalProducer.(buffer<A, B where ...> (SignalProducer<A, B>.Type) -> (Int) -> (SignalProducer<A, B>, (Event<A, B>) -> ())).(closure #1) ()
#5	0x0000000100a9f424 in SignalProducer.startWithSignal<A, B where ...>((Signal<A, B>, Disposable) -> ()) -> () at /Users/project/Carthage/Carthage/Checkouts/ReactiveCocoa/ReactiveCocoa/Swift/SignalProducer.swift:247
#6	0x0000000100a63308 in protocol witness for SignalProducerType.startWithSignal<A where ...>((Signal<A.T, A.E>, Disposable) -> ()) -> () in conformance <A, B where ...> SignalProducer<A, B> at /Users/project/Carthage/Carthage/Checkouts/ReactiveCocoa/ReactiveCocoa/Swift/SignalProducer.swift:218
[...]

PR #2519, opened 2015-10-28
Fix 2515 (deadlock in SignalProducer.buffer)

Vadim-Yelagin:
https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2515

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -122,47 +122,24 @@ public struct SignalProducer<Value, Error: ErrorType> {
 	public static func buffer(capacity: Int = Int.max) -> (SignalProducer, Signal<Value, Error>.Observer) {
 		precondition(capacity >= 0)
 
-		// This is effectively used as a synchronous mutex, but permitting
-		// limited recursive locking (see below).
-		//
-		// The queue is a "variable" just so we can use its address as the key
-		// and the value for dispatch_queue_set_specific().
-		var queue = dispatch_queue_create("org.reactivecocoa.ReactiveCocoa.SignalProducer.buffer", DISPATCH_QUEUE_SERIAL)
-		dispatch_queue_set_specific(queue, &queue, &queue, nil)
-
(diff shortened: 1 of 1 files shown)


NachoSoto:
It's hard to tell what this is trying to fix without a new failing unit test. Do you think you could come up with one so that:
- We can verify that there is in fact a dead-locking scenario.
- We can verify that this change fixes it, and it doesn't break any existing test.
- We can be sure that the problem is never introduced again in the future.

neilpa:
I haven't looked at the code yet but I'm assuming this uses queue-drain to resolve. It's something which has been discussed in the past but has been decided against because it has it's own pitfalls.

neilpa:
cc @jspahrsummers since I wasn't really involved back then.

vadim-fueled:
I haven't tried adding any new tests, but all tests pass even if I simply remove the lock in question. If there's been a lot of consideration of different pitfalls of different approaches, that would be really helpful to have tests that illustrate those pitfalls. I've posted a snippet deadlocking on nested observing on a single queue in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2515
I think I can see how the lock can help when adding an observer on one thread and inserting an event on another thread. But either was this half-hearted implementation of recursive locking (which by itself is an anti-pattern and can't comprehensibly protect from deadlocks when calling arbitrary callback code while locked) doesn't seem like the way to go.
I'm sorry I don't have more time to write tests or otherwise work on this matter at this moment.

vadim-fueled:
What if instead of having a single lock per buffer we make a lock for every observer and store it in the state in pair with the observer? This way when a new observer is added, only its newly created lock gets locked, which will deadlock if the observer tries to insert a new event, but not if it adds a new observer. When inserting a new event, the locks get locked one-by-one when iterating over observers. This will too deadlock when trying to insert an event, but not when adding a new observer. It's the same desired behavior but without the need for recursive locks.

Vadim-Yelagin:
Oops, that was me.
I'm going to close this PR and move the conversation back to https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2515

PR #2520, opened 2015-10-28
Don't sign framework for iOS simulator

gfontenot:
There's no real need to sign the iOS framework for simulator targets.

This will also let users build RAC on CI systems using Carthage that don't
have a developer cert installed.

Replaces #2452

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -2654,6 +2654,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 57A4D2461BA13F9700F7D4B1 /* tvOS-Framework.xcconfig */;
 			buildSettings = {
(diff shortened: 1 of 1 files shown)


neilpa:
@gfontenot Did this actually work for you? I'm trying to setup Travis for Rex and it's still [trying to sign iOS simulator builds](https://travis-ci.org/neilpa/Rex/builds/92610050#L139). I'm wondering if it has to do with RAC's xcconfig usage since Result only skipped the device build.

gfontenot:
It's working for me as of eedcf94a401bcccd0686584a1f919e6213a76390

neilpa:
Do you happen to have an OSS project using Travis w/ RAC as a dependency? I'd like to take a look at how it's setup to see if I can spot any differences.

gfontenot:
I don't have anything open source, unfortunately. If there's something specific you'd like me to post, I can do that.

PR #2521, opened 2015-10-29
refactored and re-implemented combineLatest() to use a single OSSpinLock regardless of tuple size

mishagray:
Added simple Tuple extensions for packing and unpacking Tuples into Arrays.
    
completely reimplemented  the combineLatest(signals : [Signal<Value,Error]) -> Signal<[Value],Error> to use a single array of values, and and a single spin lock, regardless of the number of elements in the array.   
    
No more 'climbing' chain of   .combineLatest().combineLatestWith().map(repack) for each additional tuple size.

Old Solution had  O(N) NSLocks, and copies of the 'latestValue' (where N = tuple or array size).

New solution uses a single array of latest Values, a single spin lock, a won't create a set of blocks that could overflow the stack.  O(1) stack depth!

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -680,7 +680,7 @@ extension SignalType {
 	/// at least one value each. If either signal is interrupted, the returned signal
 	/// will also be interrupted.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func combineLatestWith<O : SignalType where Error == O.Error>(otherSignal: O) -> Signal<(Value, O.Value), Error> {
+	public func combineLatestWith<U>(otherSignal: Signal<U, Error>) -> Signal<(Value, U), Error> {
 		return combineLatest(self, otherSignal)
 	}
 
@@ -1252,16 +1252,103 @@ private struct ThrottleState<Value> {
 	var pendingValue: Value? = nil
 }
 
-
 extension SignalType {
-	private final func mapToAny() -> Signal<Any,Error> {
-		return self.map { $0 }
+	
+	/// observer this signal and forward any change to an Atomically wrapped CombineLatestStates object.
+	/// CombineLatestStates.observerForIndex will create a type safe version dependeing on the Value type of this SignalType
+	/// this signal will update a specific index Collector using a Type specific set operation
+	private func observeForCombineLatest<Collector:IndexedCollectorType>(values:Atomic<CombineLatestStates<Collector>>, index:Int,observer:Signal<Collector.Value, Error>.Observer) -> Disposable? {
+		
+		// let's make a SignalType specific observer that will forward to a combineLatest observer when needed
+		// this oberver will be based on Value, and the Collector's setValueAtIndex will be generated in a type safe manner
+		let innerObserver: Observer<Value,Error> = values.value.observerForIndex(index, observer: observer)
+		
+		return self.observe { e in
+			values.withValue { _ in  // let's observer the event with the type safe lock on.
+				innerObserver.action(e)
+			}
+		}
 	}
+	
 }
+
+// manage a CombineLatestStatesProtocol for Tuples, using the TupleValueType protocol
+private final class CombineLatestStates<Collector:IndexedCollectorType>  {
+	
+	typealias Value = Collector.Value
+	
+	var collector: Collector
+	
+	var completions: [Bool]
+	var signalsStarted = 0
+	
+	init(_ collector: Collector) {
+		self.collector = collector
+		self.completions = [Bool](count: self.collector.count,repeatedValue: false)
+	}
+	
+	// this is generic so it will become type-specific any given value T.
+	// it will call the type specific-generic function setValueAtIndex
+	final func onNext<T,Error>(index:Int,value:T,observer:Signal<Value, Error>.Observer) {
+		
+		let isAStart = collector.setValueAtIndex(index,value: value)
+		if (isAStart) {
+			signalsStarted++
+		}
+		if (signalsStarted == collector.count) {
+			observer.sendNext(collector.toValue())
+		}
+	}
+	
+	final func onComplete<Error>(index:Int,observer:Signal<Value,Error>.Observer) {
+		completions[index] = true
+		
+		// we could count completions like we do starts, but it's a little safer to verify the whole list before we complete.
+		let sendComplete = !completions.contains { !$0 }
+		
+		if (sendComplete) {
+			observer.sendCompleted()
+		}
+	}
+	
+	// this is generic so it will become type-specific any given value T.
+	// it will call the type specific-generic function setValueAtIndex on the supplied collector
+	final func observerForIndex<T,Error>(index:Int,observer:Signal<Value, Error>.Observer) -> Observer<T,Error> {
+		
+		return Observer<T,Error> { event in
+			switch event {
+			case let .Next(value):
+				self.onNext(index,value: value,observer: observer)
+				
+			case .Completed:
+				self.onComplete(index, observer: observer)
+				
+			case let .Error(err):
+				observer.sendError(err)
+				
+			case .Interrupted:
+				observer.sendInterrupted()
+			}
+		}
+	}
+	
+	
+}
+
 @warn_unused_result(message="Did you forget to call `observe` on the signal?")
 public func combineLatest<A : SignalType, B : SignalType where A.Error == B.Error>(a: A, _ b: B) -> Signal<(A.Value, B.Value), A.Error> {
-	let anySignals: [Signal<Any,A.Error>] = [a.mapToAny(),b.mapToAny()]
-	return combineLatest(anySignals).map { $0.toTuple() }
+	
+	let collector = TupleCollector2<A.Value,B.Value>()
+	let states = Atomic(CombineLatestStates(collector))
+	
+	return Signal{ observer in
+		let disposable = CompositeDisposable()
+		
+		disposable += a.observeForCombineLatest(states, index: 0, observer: observer)
+		disposable += b.observeForCombineLatest(states, index: 1, observer: observer)
+		
+		return disposable
+	}
 }
 
 /// Combines the values of all the given signals, in the manner described by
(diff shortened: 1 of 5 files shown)


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1309,98 +1252,236 @@ private struct ThrottleState<Value> {
...
 
 /// Combines the values of all the given signals, in the manner described by
 /// `combineLatestWith`.
 @warn_unused_result(message="Did you forget to call `observe` on the signal?")
 public func combineLatest<A, B, C, D, E, F, Error>(a: Signal<A, Error>, _ b: Signal<B, Error>, _ c: Signal<C, Error>, _ d: Signal<D, Error>, _ e: Signal<E, Error>, _ f: Signal<F, Error>) -> Signal<(A, B, C, D, E, F), Error> {
-	return combineLatest(a, b, c, d, e)
-		.combineLatestWith(f)
-		.map(repack)
+	
+	let anySignals: [Signal<Any,Error>] =
```
NachoSoto: Can you explain this? We're losing type information here

ReactiveCocoa/Swift/TupleExtensions.swift
```diff
@@ -40,3 +40,125 @@ internal func repack<A, B, C, D, E, F, G, H, I>(t: (A, B, C, D, E, F, G, H), val
 internal func repack<A, B, C, D, E, F, G, H, I, J>(t: (A, B, C, D, E, F, G, H, I), value: J) -> (A, B, C, D, E, F, G, H, I, J) {
 	return (t.0, t.1, t.2, t.3, t.4, t.5, t.6, t.7, t.8, value)
 }
+
+
+// This will extend 'indexable' collections that can use an integer Index, 
+/// like Arrays to allow some conviences tuple implementations
+public extension SequenceType where Self:Indexable, Self.Index : IntegerLiteralConvertible {
+	
+	private func _get_element<T>(index : Index) -> T {
```
NachoSoto: This would have to use `Self.Generator.Element` instead of redefining a new type.

ReactiveCocoa/Swift/TupleExtensions.swift
```diff
@@ -40,3 +40,125 @@ internal func repack<A, B, C, D, E, F, G, H, I>(t: (A, B, C, D, E, F, G, H), val
...
+		let x = self[index]
+		let t = x as? T
+		precondition(t != nil, "\(x) is not of type \(T.self) at index \(index) of \(Self.self)")
+		return t!
+	}
+	
+	public func toTuple<A,B>() -> (A,B) {
+		return
+			(self._get_element(0),
```
NachoSoto: `SequenceType` is homogenous, so all elements inside would have the same type.

NachoSoto:
Thanks for the PR! There are a few very important issues with the new implementation though that I'm afraid we would not be able to accept.
The idea of reducing the number of copies and locks is important, however. Did you happen to measure the impact of this change?

ReactiveCocoa/Swift/Signal.swift
mishagray: We 'lose' the type information momentarily during processing, but than we get it back via .map { $0.toTuple() }

We aren't losing any type information at all.  We are just converting things to Any so we can process all the signals into a single Value array, and then we convert the array BACK to the Type specific tuple.

So if people use the tuple form, they will still get Type specific return values.

ReactiveCocoa/Swift/TupleExtensions.swift
mishagray: That's fine.  The typical use is that SequenceType is of type Any, and we want to covert the Array<Any> back into a type specific tuple.

ReactiveCocoa/Swift/TupleExtensions.swift
mishagray: I don't know what you mean here...    This is an Extension to SequenceTypes to let it convert itself BACK into a type specific Tuple, with a specific tuple signature.

ReactiveCocoa/Swift/Signal.swift
NachoSoto: I don't doubt this doesn’t behave the same way, but the compiler is not making the same guarantees anymore due to the type erasure going on here.

mishagray:
So the biggest bug that I seem to have fixed is in here: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2384

What I can do is add a test case to the Test Suite that will actually test Signal and SignalProducer arrays of the size 100+  (which will fail under the current implementation).

ReactiveCocoa/Swift/TupleExtensions.swift
NachoSoto: I see: you're storing a heterogeneous set of data in a homogeneous type. I very strongly believe that no optimization is worth the risks that this introduces.

mishagray:
Some review of the current implementation...

It uses 'pairs' of CombineLatestState<Value> structs..
So if you have a 'tuple' or even an array of N items, you end up with 2*(N-1) CombineLatestState<Value>
structs, each managing two halfs of a partial tuple.

You end up composing N-1 'half-tuples'.  But the first value in your tuple is actually stored N-1 times, the second N-2, etc.   This is O(N^2) storage.    As people switch from using reference types to Value types this can get pretty bad, since it will copy values each time it 'adds' another value to 'tuple'

For each Next event, there can be worst case N-1 'observer' calls to compose the answer.  Even on average that's N/2 observer calls for each .Next() that is sent. 

The 'stack' depth of StartProducers 'lifting' combineLatest everytime 'start' is called can produce 4-5 calls PER tuple, but definitely causes a O(N) based stack depth.    
It uses NSLock, which does slower thread locking, and allocates N-1 NSLocks for a tuple/array of size N.

If you look at this code:

```
var generator = producers.generate()
if let first = generator.next() {
	let initial = first.map { [$0] }
		return GeneratorSequence(generator).reduce(initial) { producer, next in
		producer.combineLatestWith(next).map { $0.0 + [$0.1] }
	}
}
```
you are basically composing a 'chain' of blocks that will always be O(N) deep long, and will require a deep stack just to execute start() on the composed signal.  

It's 'bad' for the Signal.
But for SignalProducer it's even worse, because most calls are 'lift' from Signal, and that adds 2-4X as many stack depth calls.    It may be that the compiler optimizations can flatten the stack loop for the 'Tuple' implementations, but but for the Sequence_Type, that can't be done, and will try to create a closure that requires O(N) stack size to handle each start() and each .Next,

Whereas this branch gives you O(N) storage vs O(N)^2, using a single shared class to manage the latestValues.
Only one observer execution per .Next event sent from a sub-signal, not an average of N/2 observer calls.
O(1) stack depth not O(N).  feel free to compose hundreds of Signal Producers.  The current code can crash somewhere between 25-50 Signal Producers with a stack overflow (especially on background thread).
Uses superfast OSSpinLock, since we are just doing fast array updates.

As for the type erasure step on tuples:

- If you are using a combineLatest(SequenceType) version, than no Any conversion is done at all.  It's perfectly type safe.  It's also where you are most likely to see the performance issues.

- As for the 'type' safety - type conversion in Swift is only done 'safely' in Debug builds.  Once code is compiled in Release, any runtime safety measures are removed anyways, and there is no disadvantage to treating the Values as an Any array.
 
- The compiler will still perform all the same type-safe tuple checking you already have in the current version. No compile-time type safe checking has been removed at all.   combineLatest will still perform all the same compile-time type checking you need.  

I've ported this app to the latest Rac 4 and reproduced the crashes (and then fixed them).
https://github.com/mishagray/rac-bug/tree/swift2

With the current branch, it crashes (especially if using Debug builds).
Currently I'm seeing stack overflow crashes on background queues with total Heap usage around 24MB of Heap before it dies.  Not sure if the Heap would have gotten bigger before it died.

With my branch, no crashes.  17MB max heap.

neilpa:
I agree w/ @NachoSoto about the type erasure here. Instead we can define `combineLatestWith` as the 10-arity version with params 3-10 being optional. Then invert the lower arity overloads to be defined in terms of the higher ones.

I started down this path at one point but there's a few difficulties to making this work. I never had a chance to finish it but I'll dig it up.

mishagray:
Hmm.  Actually using a fixed size tuple has given me an idea here. 

I think I can use a protocol to switch between using an array or a type safe tuple using a simple protocol on my existing shared struct.   And use an array for when you have a type safe array of matching signals producers, or a single tuple when using tuples...   

So we can have the exact same performance but no type erasure ...

Okay ... Give me a few hours....

neilpa:
> Hmm. Actually using a fixed size tuple has given me an idea here.

Awesome! I'll let you run with it then.

mishagray:
Okay!  All typesafe and efficient.  Created a IndexCollectorType protocol that abstracts away from the logic whether it's an array or a tuple.  So the 'combine' logic is identical, and uses a protocol to abstract away the storage method.

A type safe 'observe' is generated for each sub-signal to match it's specific Value type. Which uses a type-safe 'set' and an index value, abstracting away the 'tuple' or 'array' implementation.

combineLatest implementations now create a Collector for their tuple type, create a CombineLatestStates<Collector>() class and then creates the correct type-safe observer for each sub-signal.

NachoSoto:
Do you think you can write a quick benchmark to get an idea of how much more efficient and less memory consuming is?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1309,98 +1252,325 @@ private struct ThrottleState<Value> {
...
+
+		return disposable
+	}
 }
 
 /// Combines the values of all the given signals, in the manner described by
 /// `combineLatestWith`.
 @warn_unused_result(message="Did you forget to call `observe` on the signal?")
 public func combineLatest<A, B, C, D, E, F, G, H, Error>(a: Signal<A, Error>, _ b: Signal<B, Error>, _ c: Signal<C, Error>, _ d: Signal<D, Error>, _ e: Signal<E, Error>, _ f: Signal<F, Error>, _ g: Signal<G, Error>, _ h: Signal<H, Error>) -> Signal<(A, B, C, D, E, F, G, H), Error> {
-	return combineLatest(a, b, c, d, e, f, g)
```
NachoSoto: This is A LOT of repeated code. We wouldn't want to maintain so much duplication.

neilpa: This is why I suggested `combineLatestWith` be implemented once at the max-arity. Then define the lower-arity overloads in terms of it. I've got it worked out in my head, but I haven't spent time actually trying to implement it. If I do this afternoon I'll open an alternative PR.

mishagray: I started down that thinking.  

But this code is really the optimum sized memory and usage for the every situation.   

You could do everything with "max arrity" but then you use more memory than you need (for the typical 2,3,4 case). 

The reality is that while this may look like more "lines of code" it's the bare minimum number of actual operators needed to do any sized tuple up to max arrity.   

The current code  "looked smaller" but since each  version of the operation uses the version before, you are just satisfying some human need to see fewer lines of code.  Not thinking about the number of compiled instructions. 

I'm pretty sure this is the smallest amount of compiled code needed.   And will always use the smallest amount of memory.  

While maintaining support for both reference and value types, and maintaining type-correctness.

neilpa:
> But this code is really the optimum sized memory and usage for the every situation.

At the expense of human maintenance though. The max-arity approach should have similar performance characteristics with a slight overhead beyond this approach. I don't want to micro-optimize further if it produces this much duplication. At least not without benchmarks saying we need it.

NachoSoto:
> At least not without benchmarks saying we need it.

:100:

PR #2524, opened 2015-10-30
use #import, not @import

ianthehenry:
As far as I can tell, there is no way to enable module support in a project that uses *any* Objective-C++. This means that importing the ReactiveCocoa headers -- even from a pure Objective-C file -- will result in an error on this line.

Since it's the only use of `@import` in the codebase, it seems nice from a consistency standpoint as well.


NachoSoto:
This is 100% equivalent as it will get automatically converted to a module `@import` if they're enabled, so okay!

PR #2526, opened 2015-11-01
Implemented Signal.mergeSignals. Fixes #2487

### LINKED ISSUE #2487: Signal<Signal<T, E>, E>.merge()
This is a continuation to #2314. [We now have a way to flatten signals of signals](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2449). However, it's still not trivial to do the equivalent of `SignalProducer(values: [producer1, producer2]).flatten(.Merge)` for a `Signal` of `Signal`s, because creating the outer signal is not trivial.

There was a discussion recently about this, and this is my proposed implementation:
```swift
extension SignalType {
	/// Merges the given signals into a single `Signal` that will emit all values
	/// from each of them, and complete when all of them have completed.
	public static func mergeSignals(signals: [Signal<Self.Value, Self.Error>]) -> Signal<Self.Value, Self.Error> {
		let producer: SignalProducer<Signal<Value, Error>, Self.Error> = SignalProducer(values: signals)
		var result: Signal<Self.Value, Self.Error>!

		producer.startWithSignal { (signal, _) in
			result = signal.flatten(.Merge)
		}

		return result
	}
}
```

What does everything think of adding this to the API?

NachoSoto:
See discussion in #2487 for more details.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -361,7 +361,7 @@ extension SignalType {
 extension SignalType {
 	/// Merges the given signals into a single `Signal` that will emit all values
 	/// from each of them, and complete when all of them have completed.
-	public static func mergeSignals(signals: [Signal<Value, Error>]) -> Signal<Value, Error> {
+	public static func merge(signals: [Signal<Value, Error>]) -> Signal<Value, Error> {
 		let producer = SignalProducer<Signal<Value, Error>, Error>(values: signals)
 		var result: Signal<Value, Error>!
 
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -358,6 +358,21 @@ extension SignalType {
 	}
 }
 
+extension SignalType {
+	/// Merges the given signals into a single `Signal` that will emit all values
+	/// from each of them, and complete when all of them have completed.
+	public static func mergeSignals(signals: [Signal<Value, Error>]) -> Signal<Value, Error> {
+		let producer = SignalProducer<Signal<Value, Error>, Error>(values: signals)
+		var result: Signal<Value, Error>!
+
+		producer.startWithSignal { (signal, _) in
+			result = signal.flatten(.Merge)
+		}
+
+		return result
+	}
+}
+
 extension SignalType where Value: SignalProducerType, Error == Value.Error {
 	/// Returns a signal which sends all the values from producer signal emitted from
 	/// `signal`, waiting until each inner producer completes before beginning to
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -361,7 +361,7 @@ extension SignalType {
 extension SignalType {
 	/// Merges the given signals into a single `Signal` that will emit all values
 	/// from each of them, and complete when all of them have completed.
-	public static func merge(signals: [Signal<Value, Error>]) -> Signal<Value, Error> {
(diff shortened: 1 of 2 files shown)


NachoSoto:
Bump! :)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -331,6 +331,21 @@ extension SignalType {
 	}
 }
 
+extension SignalType {
+	/// Merges the given signals into a single `Signal` that will emit all values
+	/// from each of them, and complete when all of them have completed.
+	public static func mergeSignals(signals: [Signal<Value, Error>]) -> Signal<Value, Error> {
```
JaviSoto: The word "signals" seems redundant. Could this be called simply `merge`?

JaviSoto:
:+1: Just a comment on the name, but I think this operator is basic enough that we should provide it 1st party IMO.

ReactiveCocoa/Swift/Signal.swift
bencochran: I agree about `merge`. Also, could this take a `SequenceType` containing `Signal`s instead of an array?

NachoSoto: Both points are really good :+1: changing!

NachoSoto: I changed the name.
Unfortunately I couldn't get the `SequenceType` change to compile:

```swift
public static func merge<T: SignalType, S: SequenceType where S.Generator.Element == T, T.Value == Self.Value, T.Error == Self.Error>
```

That's fine, but then it can't find a definition of `flatten`... and I'm having a hard time figuring out why because `SourceKit` keeps crashing :fire:

bencochran: I had no idea this was possible, but I just tried

```swift
func merge<S : SequenceType where S.Generator.Element == Signal<Value, Error>>(signals: S) -> Signal<Value, Error>
```

… and that worked? Swift. 🙃

NachoSoto: Oh, derp. So much simpler than everything I tried :sweat_smile: Thanks!

NachoSoto:
Addressed comments!

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -496,6 +495,71 @@ class SignalSpec: QuickSpec {
...
+				expect(lastValue) == 1
+
+				observer2.sendNext(2)
+				expect(lastValue) == 2
+
+				observer1.sendNext(3)
+				expect(lastValue) == 3
+			}
+
+			it("should not stop when one signal completes") {
```
JaviSoto: Would it make sense to write a similar tests for when one of the signals sends an error?

NachoSoto: Sure, but I just wanted high level tests. That should be covered by `flatten(.Merge)`, so I  didn't want to duplicate too much of that coverage here.

JaviSoto: Yeah, although that's an implementation detail, it's not really encoded in the API.

NachoSoto: Actually, 
> Would it make sense to write a similar tests for when one of the signals sends an error?

That test would verify that no values are emitted after an error? That's basically testing the `Signal` contract, which should be tested elsewhere :)

NachoSoto:
Any further thoughts? :D

NachoSoto:
Bump :P

JaviSoto:
:shipit:

NachoSoto:
Thanks!

PR #2531, opened 2015-11-03
Moved flatMapError to Signal

### LINKED ISSUE #2414: Signal<AnyObject, NSError> to Signal<AnyObject, NoError>?
We have `catch` (or `flatMapError` now in RAC4) for converting a `SignalProducer` that sends `NSError` to `NoError`. How would you do the same with a `Signal`? I see `mapError` but as far as I can tell, it's not possible to use that operator to achieve this? Or am I wrong? 😳

NachoSoto:
Fixes #2414.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -229,6 +229,33 @@ extension SignalType {
 		}
 	}
 
+	/// Catches any failure that may occur on the input signal, mapping to a new producer
+	/// that starts in its place.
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -229,6 +229,33 @@ extension SignalType {
 		}
 	}
 
+	/// Catches any failure that may occur on the input signal, mapping to a new producer
```
neilpa: mapping to a new **signal**

neilpa: nevermind

PR #2536, opened 2015-11-04
Enable "Require Only App-Extension-Safe API" for RAC 2

brow:
This ports #1891 back to RAC 2. I couldn't find a living branch for RAC 2, so I created `2.x-maintenance`.

I've tested this so far by pointing to this branch from the Cartfile of the project where I originally experienced the "linking to dylib not safe" warning.

I opened an equivalent PR against Mantle 1: https://github.com/Mantle/Mantle/pull/639

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,3 +1,3 @@
-github "jspahrsummers/xcconfigs" >= 0.7.1
+github "jspahrsummers/xcconfigs" >= 0.8
 github "Quick/Quick"
 github "Quick/Nimble"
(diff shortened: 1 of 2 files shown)


brow:
@ReactiveCocoa/reactivecocoa Should I be trying to get Travis to build this branch? Travis seems unable to build the version of Nimble that was already in `Cartfile.resolved` prior to this change, seemingly because Travis is building with Xcode 6.1 and Nimble 0.4.2 uses Swift 1.2 features.

If not, then this is ready for review. This is a build config change not covered by tests, anyway.

PR #2542, opened 2015-11-05
Added overloads for flatMap and flatten for Signals/Producers which can't emit errors

NachoSoto:
Added overloads for `flatMap` and `flatten` for `Signal`s or `SignalProducer`s which can't emit errors.
This makes the methods friendlier to newcomers who might struggle with type errors because the outer type is `NoError`.

It is unfortunate that this requires so much duplication, but it's probably for the best to put the burden on the framework, rather than users struggling with this.

*Thanks to @andymatuschak for the original idea!*

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -388,6 +388,30 @@ extension SignalType {
 	}
 }
 
+extension SignalType where Error == NoError {
+	/// Maps each event from `signal` to a new producer, then flattens the
+	/// resulting producers (into a signal of values), according to the
+	/// semantics of the given strategy.
+	///
+	/// If `signal` or any of the created producers fail, the returned signal
+	/// will forward that failure immediately.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func flatMap<NewValue, NewError>(strategy: FlattenStrategy, transform: Value -> SignalProducer<NewValue, NewError>) -> Signal<NewValue, NewError> {
+		return map(transform).flatten(strategy)
+	}
+
+	/// Maps each event from `signal` to a new signal, then flattens the
+	/// resulting signals (into a signal of values), according to the
+	/// semantics of the given strategy.
+	///
+	/// If `signal` or any of the created signals emit an error, the returned
+	/// signal will forward that error immediately.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func flatMap<NewValue, NewError>(strategy: FlattenStrategy, transform: Value -> Signal<NewValue, NewError>) -> Signal<NewValue, NewError> {
+		return map(transform).flatten(strategy)
+	}
+}
+
 extension SignalType where Value: SignalProducerType, Error == Value.Error {
 	/// Returns a signal which sends all the values from producer signal emitted from
 	/// `signal`, waiting until each inner producer completes before beginning to
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -310,6 +310,21 @@ extension SignalType where Value: SignalType, Error == Value.Error {
 	}
 }
 
+extension SignalType where Value: SignalType, Error == NoError {
+	/// Flattens the inner signals sent upon `signal` (into a single signal of
+	/// values), according to the semantics of the given strategy.
+	///
+	/// If `signal` or an active inner signal emits an error, the returned
+	/// signal will forward that error immediately.
+	///
+	/// `Interrupted` events on inner signals will be treated like `Completed`
+	/// events on inner signals.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Value.Error> {
+		return self.promoteErrors(Value.Error.self).flatten(strategy)
+	}
+}
+
 extension SignalType where Value: SignalProducerType, Error == Value.Error {
 	/// Flattens the inner producers sent upon `signal` (into a single signal of
 	/// values), according to the semantics of the given strategy.
@@ -331,6 +346,21 @@ extension SignalType where Value: SignalProducerType, Error == Value.Error {
 		case .Latest:
 			return self.switchToLatest()
 		}
+	}
+}
+
+extension SignalType where Value: SignalProducerType, Error == NoError {
+	/// Flattens the inner producers sent upon `signal` (into a single signal of
+	/// values), according to the semantics of the given strategy.
+	///
+	/// If `signal` or an active inner producer fails, the returned signal will
+	/// forward that failure immediately.
+	///
+	/// `Interrupted` events on inner producers will be treated like `Completed`
+	/// events on inner producers.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Value.Error> {
+		return self.promoteErrors(Value.Error.self).flatten(strategy)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


andymatuschak:
Yay! To maybe provide some color: we've noticed that a pretty common pattern is to shovel signals imperatively into some signal of signals (e.g. for network requests). Then you want to `flatten(.Latest)` or whatever.

The outer signal can't have errors: it's a glorified queue; the inner signals can error.

neilpa:
I worry this will exacerbate the ambiguity problem with `flatten` and `flatMap`. Also, I don't think it's obvious that this is just a glorified queue. Maybe there's a better way to expose this pattern than further overloading `flatten` and `flatMap` with subtly different semantics.

NachoSoto:
> I worry this will exacerbate the ambiguity problem with flatten and flatMap.

It won't because the two operators are defined on different types. The original implementations are only defined for when `Error == Value.Error`.

> Maybe there's a better way to expose this pattern than further overloading flatten and flatMap with subtly different semantics.

This doesn't really have anything to do with the pattern though, it simply makes using these operators more straightforward. As they are right now, if you try to do `SignalProducer<SignalProducer, SomeError>, NoError>.flatten(...)` it won't compile, and the kinds of errors you'll see will be very confusing to people.

neilpa:
> It won't because the two operators are defined on different types.

You're still introducing another case where the inner types are varying and the compiler may not be able to resolve the ambiguity.

>  This doesn't really have anything to do with the pattern though

I was speaking to @andymatuschak's comment about it effectively being a queue. Should we call these operators something different because you're flattening over both the error and the value. I'm probably just being overly cautious though.

NachoSoto:
> You're still introducing another case where the inner types are varying and the compiler may not be able to resolve the ambiguity.

In that case there's nothing we can do... But I think that potential error case is simpler to solve for most people than figuring out that they need `promoteErrors`.

> Should we call these operators something different because you're flattening over both the error and the value. I'm probably just being overly cautious though.

Oh you mean the `flatMap` version. Yeah, I wasn't completely sure of that overload myself for that very reason, but I don't think we have to change the name. In fact that would be even more confusing. Consider this example:

```swift
let signal: Signal<String, NoError>

signal
    .promoteErrors(SomeError.self)
    .flatMap(.Merge) { string -> SignalProducer<String, SomeError> in
        return f(string)
    }
```

Most people won't understand why they need `promoteErrors` there. And the thing is, they shouldn't have to. I think this limitation is accidental complexity that we can solve:

```swift
let signal: Signal<String, NoError>

// No need to `promoteErrors`.
// Yes, technically you're flattening over both error and value, but only because the outer signal could not fail.
signal.flatMap(.Merge) { string -> SignalProducer<String, SomeError> in
  return f(string)
}
```

neilpa:
Fair enough, I'm convinced.

neilpa:
We should probably add a quick test for the `NoError` case though.

NachoSoto:
Sounds good :)
You mean for `SignalProducer<T, NoError>.flatMap`?

neilpa:
Yea

NachoSoto:
Actually, @neilpa you were right with your first point, this is legitimately ambiguous now:
```swift
let producer: SignalProducer<SignalProducer<Int, NoError>, NoError>
let flattened = producer.flatten(.Concat) // There are 2 definitions for this now! :(
```

And because there's no way to do:
```swift
extension SignalProducerType where Error == NoError, Value.Error != NoError
```

We can disambiguate this :( So unfortunately this is a blocker for this feature. We could provide a different name for these overloaded `flatten` and `flatMap`, but I think that would be even more confusing. What do you think?

I filed a radar for this: http://openradar.me/23432210.

andymatuschak:
Wait, why is this ambiguous? The NoError-specifc implementation should be strictly more specific, so it should win… am I misunderstanding the overloading precedence rules? :(

NachoSoto:
So now `flatten` (same for `flatMap`) has these 2 definitions:
```swift
extension SignalType where Value: SignalType, Error == Value.Error {
	public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Error>
}

extension SignalType where Value: SignalType, Error == NoError {
	public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Value.Error>
}
```

Unfortunately, the first one includes the second one if `Error == Value.Error == NoError`, but there's no way to do
```swift
extension SignalType where Value: SignalType, Error == Value.Error, Error != NoError {
	public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Error>
}
```
or
```swift
extension SignalType where Value: SignalType, Value.Error != Error {
	public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Value.Error>
}
```.

> Wait, why is this ambiguous? The NoError-specifc implementation should be strictly more specific, so it should win

I don't think method resolution takes specificity into account to resolve the method (which I actually like... the rules in C++, for example, are so complex that nobody can actually reason about them).

**Comment by andymatuschak** [comment]
> On Nov 6, 2015, at 10:52 AM, NachoSoto <notifications@github.com> wrote:
> 
> So now flatten (same for flatMap) has these 2 definitions:
> 
> extension SignalType where Value: SignalType, Error == Value.Error {
>     public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Error>
> }
> 
> extension SignalType where Value: SignalType, Error == NoError {
>     public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Value.Error>
> }
> Unfortunately, the first one includes the second one if Error == Value.Error == NoError, but there's no way to do
> 
> extension SignalType where Value: SignalType, Error == Value.Error, Error != NoError {
>     public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Error>
> }
> or
> 
> extension SignalType where Value: SignalType, Value.Error != Error {
>     public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Value.Error>
> }
> ```.
I just tested, and this’ll work fine if you define another variant like:

extension SignalType where Value: SignalType, Error == NoError, Value.Error == NoError {}
> 
> > Wait, why is this ambiguous? The NoError-specifc implementation should be strictly more specific, so it should win
> 
> I don't think method resolution takes specificity into account to resolve the method (which I actually like... the rules in C++, for example, are so complex that nobody can actually reason about them).
It does! I figured out the issue. The problem is that the constraints in the two definitions are *equally specific* (`Value: SignalType, Error == Value.Error`, and `Value:SignalType, Error == NoError`).

But this clause *is* more specific, so the compiler does the right thing: `Value: SignalType, Error == NoError, Value.Error == NoError`

**Comment by NachoSoto** [comment]
> But this clause *is* more specific, so the compiler does the right thing: `Value: SignalType, Error == NoError, Value.Error == NoError`

That's different definition though :( we want to be able to flatten `Signal<Signal<T, SomeError>, NoError>`.

**Comment by NachoSoto** [comment]
I'm going to take this out of 4.0 since it's likely we won't be able to solve it with `Swift 2.1`.

**Comment by Vadim-Yelagin** [comment]
This use case with a "glorified queue", how does it handle the situation when an inner signal fails? Won't it fail the entire flattened signal? Don't the users expect that this flattened signal will carry on?
If it's a signal of signals, not signal producers, won't they miss the events that happen in the inner signal before they are shoved to the queue?
Probably much better alternative is to use an `Action` that takes signal producers as input for its application closure (one being essentially an identity function).

**Comment by mdiep** [comment]
> > But this clause is more specific, so the compiler does the right thing: Value: SignalType, Error == NoError, Value.Error == NoError
>
> That's different definition though :( we want to be able to flatten Signal<Signal<T, SomeError>, NoError>.

I think @andymatuschak was saying that your code works for that if you add a 3rd variant that _is_ more specific for that particular case.

```swift
extension SignalType where Value: SignalType, Error == Value.Error {
    public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Error>
}

extension SignalType where Value: SignalType, Error == NoError {
    public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, Value.Error>
}

extension SignalType where Value: SignalType, Error == NoError, Value.Error == NoError {
    public func flatten(strategy: FlattenStrategy) -> Signal<Value.Value, NoError>
}
```

@NachoSoto Do you have any interest in pursuing this?

mdiep:
Superseded by #2885.

PR #2544, opened 2015-11-09
Changed toRACSignal from free function to protocol extension

NachoSoto:
This is a leftover from the `RAC 3 -> RAC 4` (`Swift 1.2 -> Swift 2.0`) migration.
Note that this is a breaking change, but luckily we're in time before the `4.0` release.

Given that a lot of it is indentation changes, I recommend reviewing it [ignoring white spaces](https://github.com/NachoSoto/ReactiveCocoa/commit/465d2cc39fe9b83ef22f0477a7c9973057f4e386?w=0).


neilpa:
:+1:

PR #2545, opened 2015-11-10
Re-enable continuous integration

felixjendrusch:
This re-enables continuous integration builds on Travis CI by [explicitly setting build dependencies](e396263) as [described by `xctool`](https://github.com/facebook/xctool#continuous-integration) and [restricting tests to available targets](2bc64e0) (Mac and iOS).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: objective-c
+osx_image: xcode7
+before_install: true
+install: true
+git:
+  submodules: false
+script: script/cibuild
+notifications:
+  email: false
+  slack:
+    secure: C9QTry5wUG9CfeH3rm3Z19R5rDWqDO7EhHAqHDXBxT6CpGRkTPFliJexpjBYB4sroJ8CiY5ZgTI2sjRBiAdGoE5ZQkfnwSoKQhWXkwo19TnbSnufr3cKO2SZkUhBqOlZcA+mgfjZ7rm2wm7RhpCR/4z8oBXDN4/xv0U5R2fLCLE=
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode7
+osx_image: xcode7.1
 before_install: true
 install: true
 git:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install: true
 install: true
 git:
   submodules: false
-script: script/cibuild
+script: script/cibuild ReactiveCocoa-Mac ReactiveCocoa-iOS
 notifications:
   email: false
   slack:
--- a/ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa-Mac.xcscheme
+++ b/ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa-Mac.xcscheme
@@ -3,9 +3,23 @@
    LastUpgradeVersion = "0700"
    version = "1.3">
    <BuildAction
-      parallelizeBuildables = "YES"
-      buildImplicitDependencies = "YES">
+      parallelizeBuildables = "NO"
+      buildImplicitDependencies = "NO">
       <BuildActionEntries>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "D45480561A9572F5009D7229"
+               BuildableName = "Result.framework"
+               BlueprintName = "Result-Mac"
+               ReferencedContainer = "container:Carthage/Checkouts/Result/Result.xcodeproj">
+            </BuildableReference>
+         </BuildActionEntry>
          <BuildActionEntry
             buildForTesting = "YES"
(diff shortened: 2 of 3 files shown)


ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa-Mac.xcscheme
```diff
@@ -3,8 +3,8 @@
    LastUpgradeVersion = "0700"
    version = "1.3">
    <BuildAction
-      parallelizeBuildables = "YES"
-      buildImplicitDependencies = "YES">
+      parallelizeBuildables = "NO"
+      buildImplicitDependencies = "NO">
```
NachoSoto: Can you explain why this change?

ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa-Mac.xcscheme
```diff
@@ -3,8 +3,8 @@
    LastUpgradeVersion = "0700"
    version = "1.3">
    <BuildAction
-      parallelizeBuildables = "YES"
-      buildImplicitDependencies = "YES">
+      parallelizeBuildables = "NO"
```
NachoSoto: Can you explain the reason for this change?

.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: objective-c
+osx_image: xcode7.1
+before_install: true
+install: true
+git:
+  submodules: false
+script: script/cibuild ReactiveCocoa-Mac ReactiveCocoa-iOS
```
NachoSoto: Ooooh nice, I didn't realize we could specify these :+1:

NachoSoto:
Thanks for the PR! I'm excited to have this back again :D

ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa-Mac.xcscheme
felixjendrusch: As mentioned, `xctool` requires dependencies to be specified explicitly, meaning there should be no implicit dependencies. Disabling the finding of implicit dependencies should also help avoiding any confusion.

More related (possibly outdated) background:
- http://www.openradar.me/20490378#ag9zfm9wZW5yYWRhci1ocmRyFAsSB0NvbW1lbnQYgICA4LKcjwsM

  > I found that Implicit Dependencies would only mess up if I didn't have all dependencies explicitly configured. I I made sure that all dependencies where explicit, the problem would go away.

- http://pewpewthespells.com/blog/managing_xcode.html#scheme-action

  > However there are some edge-cases to this behavior. If you find yourself in a situation where you cannot rely on Xcode's ability to resolve the necessary target dependencies correctly, then both "Parallelize Build" and "Find Implicit Dependencies" should be disabled on the scheme.

- http://andymolloy.net/workspace-interproject-dependencies-and-xcodebuild/

  > If you have an Xcode workspace containing multiple projects, and a scheme in that workspace depends on targets in more than one of those projects, Xcode’s own automatic dependency detection may well work, but xcodebuild will fail. [...] I also had to uncheck the Parallelize Targets box.

ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa-Mac.xcscheme
felixjendrusch: See https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2545#discussion_r44449283.

NachoSoto: That explains `buildImplicitDependencies` (which I agree with, I'm :+1: for explicitness), but why `parallelizeBuildables`?

NachoSoto:
Part of me wonders why we're still using `xctool`. It does much more than we need, ideally we just use a tool that works **on top of xcodebuild** (same philosophy as `Carthage`), like [`scan`](https://github.com/fastlane/scan).

ReactiveCocoa.xcodeproj/xcshareddata/xcschemes/ReactiveCocoa-Mac.xcscheme
felixjendrusch: Xcode might decide to build `Result` and `ReactiveCocoa` in parallel because there is no target dependency for the former on the latter. This should never happen because `ReactiveCocoa` depends on `Result`. A proper target dependency would require a project reference of the `Result` project on the `ReactiveCocoa` project, which would also add the schemes of the former to the latter.

felixjendrusch:
I agree on `xctool` vs. something that works on top of `xcodebuild` but would prefer to not depend on Ruby, Gems, Fastlane, ... so yeah, something "like scan".

NachoSoto:
We can just invoke `xcodebuild` directly then, it's not that much work.

felixjendrusch:
I like the idea of having reusable build scripts though, like [`jspahrsummers/objc-build-scripts`](https://github.com/jspahrsummers/objc-build-scripts), so maybe the switch from `xctool` to `xcodebuild` should be proposed there?

NachoSoto:
That's a fair point.

I'm going to merge this since it's useful nonetheless! We can consider updating the build script at some other point.

Thanks a lot :star: :green_heart:

PR #2547, opened 2015-11-10
Signal: changed class methods/properties to static

NachoSoto:
The `Signal` class is `final` so I don't expect this to actually make a difference, but at least semantically, it doesn't make sense to want *dynamic dispatching* on these.


ikesyo:
Makes sense. :+1:

PR #2549, opened 2015-11-11
Implement `skipRepeats` for SignalType with optional value

iv-mexx:
Before, it was not possible to just use `skipRepeats()` in such a scenario:

```
let signal = Signal<Int?, NoError>
let signalSkipRepeates = signal.skipRepeats()
```

:exclamation: `Cannot invoke 'skipRepeats' with no arguments`
:exclamation: `Expected an argument list of type '((Self.Value, Self.Value) -> Bool)'`

Thats because the version of `skipRepeats` that needs no argument is only implemented for Signals where the Value conforms to `Equatable`, but Optionals do not conform to `Equatable` (even though comparison is actually implemented and "just works")

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -658,6 +658,48 @@ class SignalSpec: QuickSpec {
 				expect(values).to(equal([ true, false, true ]))
 			}
 
+			it("should skip duplicate optional values") {
+				let (baseSignal, observer) = Signal<Bool?, NoError>.pipe()
+				let signal = baseSignal.skipRepeats()
+
+				var values: [Bool?] = []
+				signal.observeNext { values.append($0) }
+
+				expect(values.count).to(equal(0))
+
+				observer.sendNext(true)
+				expect(values.count).to(equal(1))
+				expect(values[0]).to(equal(true))
+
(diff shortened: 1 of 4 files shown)


ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -658,6 +658,48 @@ class SignalSpec: QuickSpec {
...
+				let signal = baseSignal.skipRepeats()
+
+				var values: [Bool?] = []
+				signal.observeNext { values.append($0) }
+
+				expect(values.count).to(equal(0))
+
+				observer.sendNext(true)
+				expect(values.count).to(equal(1))
+				expect(values[0]).to(equal(true))
```
iv-mexx: Unfortunately its not possible to do
```
expect(values).to(equal([true]))
```

because arrays with optional values are not equatable

:exclamation: `error: binary operator '==' cannot be applied to two '[Bool?]' operands`

NachoSoto:
Thanks for the PR!
This is a limitation of the language, as there's currently no way to express that `Optional<T: Equatable>` is itself `Equatable`.

For that reason I don't think we should add an overload) for every type that suffers from this problem (like `Array`).

Instead, I recommend simply doing this:

```Swift
signal 
   .skipRepeats(==)
```

Issue #2543, opened 2015-11-09
[Feature request]: Add a signal to PropertyType


NachoSoto:
I think this was debated at some point but I can't find any other issue.

It would be nice if `PropertyType` also exposed a `Signal<Value, NoError>`. Of course we can do `property.producer.skip(1)`, but it would be nice to have its own method to expose this with the right *no-side-effects* semantics.

neilpa:
I would take this a step further to say that `PropertyType` should conform to both `SignalType` and `SignalProducerType`.

NachoSoto:
Yes, absolutely! But that's going to require higher-kinded types so that things like `SignalType.map` can return `Self<Value, Error>`.

neilpa:
Nope, [`SignalType` exposes `signal: Signal<Value, Error>`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/49d75631c0daf4c470cb5d9c5efb8b1d19aab05c/ReactiveCocoa/Swift/Signal.swift#L144-L156) and doesn't use `Self` because of that.

NachoSoto:
> I would take this a step further to say that `PropertyType` should conform to both `SignalType` and `SignalProducerType`.

What I mean is that operators, like `map`, [are implemented on `SignalType`, but they can't return `Self<Value, Error>`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/Swift/Signal.swift#L214).

If `PropertyType` were to conform to `SignalType`, for example, doing `property.map` would return a `Signal`, which is wrong. It should return another property of the same type.

That's why I'm saying that unfortunately what you suggest is not possible right now.

neilpa:
> they incorrectly return Signal.

I wouldn't call this incorrect given that this is a language limitation and there's nothing we can do to workaround it.

NachoSoto:
For sure, hence the italic :)

paulyoung:
It was originally brought up in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2146. I wish I had this all the time but never got around to it.

NachoSoto:
@neilpa do you have any further thoughts on this?

Given that, as I explained, we can't make it conform to `SignalType` yet, seems like an appropriate temporary solution to simply add a `Signal`.

This isn't blocking `4.0` though, of course.

neilpa:
Nope, we should just go ahead and add it.

PR #2550, opened 2015-11-11
MutablePropertyType: added "immutableView" property

NachoSoto:
This simplifies a very common use case where types want to have a `MutablePropertyType`, but they only want to expose it as immutable.

```swift
class Example {
    public var currentValue: AnyProperty<Int> {
        return self.currentValueImmutableProperty.immutableView
    }

    private let currentValueImmutableProperty = MutableProperty<Int>(0)
}
```

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -121,6 +121,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	}
 }
 
+extension MutablePropertyType {
+	/// Returns a "view" of this property that can only be observed, but not modified.
+	public var immutableView: AnyProperty<Value> {
+		return AnyProperty(self)
+	}
+}
+
 /// Wraps a `dynamic` property, or one defined in Objective-C, using Key-Value
 /// Coding and Key-Value Observing.
 ///


NachoSoto:
There's a case to be made for this not being crucial for `RAC`'s public API, so I'd like to hear everyone's thoughts!

liscio:
What about sticking with the reversed naming such that a property is immutable, and only a mutableProperty is mutable? E.g.:

```
class Example {
    public var currentValueProperty: AnyProperty<Int> {
        return self.currentValueMutableProperty.property
    }

    private let currentValueMutableProperty = MutableProperty<Int>(0)
}
```

That is, you change the name of 'immutableProperty` to simply `property`. Thoughts?

But I also wonder: What is the benefit of doing this, versus what you propose in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2543? I mean, you could use your mechanism there to do:

```
class Example {
    public var currentValueProperty: AnyProperty<Int> {
        return AnyProperty(initialValue: currentValueMutableProperty.value, producer: currentValueMutableProperty.signal)
    }

    private let currentValueMutableProperty = MutableProperty<Int>(0)
}
```

Or just replace `currentValueMutableProperty.signal` with `currentValueMutableProperty.producer` and you can get this functionality today (albeit with side-effects.)

Just thinking aloud alongside you… :)

NachoSoto:
I mean, the `AnyProperty(property: PropertyType)` constructor already exists for just this purpose. The only new thing that I'm proposing here is to expose it from `MutableProperty`, which I think is a more approachable API.

neilpa:
While it's a more approachable API, I'm not a fan. It encourages tracking additional state outside of the stream model and injecting it imperatively at arbitrary times. Ideally, `MutableProperty` should live at the edges of your program acting as the bridge between the stream and imperative world. Not as an intermediary that gets sprinkled across a code base.

Basically, I want to be more prescriptive about the use-cases for properties but others may disagree.

NachoSoto:
> It encourages tracking additional state outside of the stream model and injecting it imperatively at arbitrary times.

Can you elaborate on what you mean by this?

> Ideally, MutableProperty should live at the edges of your program acting as the bridge between the stream and imperative world. Not as an intermediary that gets sprinkled across a code base.

What makes you say that? `PropertyType` was added **to solve a very concrete use case** (the lack of which was one of my confusions in #1650): **representing data as both its current value and any subsequent changes**. This was solved in `RAC 2` with things like `replayLast`, and same in `RX`: with *multicasting*. However, that has huge consequences in complexity because this behavior was never encoded in the type system.

Thus, I'm not sure why you think that `PropertyType` should live "at the edges of your program". It's a very useful construct. One that I'm trying to make more approachable with this small change.

brow:
I use the pattern of `Example` often. Most of my view models primarily expose `AnyProperty`s, and some view models are simple enough that I prefer to implement them using `MutableProperty`s.

However, I don't see `MutablePropertyType.immutableView` as any simpler than `AnyProperty(property: PropertyType)`, and it introduces two new articles of vocabulary into the code -- "immutable" and "view".

andersio:
It doesn't seem bringing any extra goods that explicitly constructing an `AnyProperty` would fail at. If the intention is just to find a shorter way, `lazy` would probably be the simplest solution while maintaining the access control. More importantly, it allows lazy initialisation, and also a point of side effect injection if you use a closure.

```
class Example {
    public lazy var currentValue: AnyProperty<Int> = AnyProperty(property: self._currentValue)
    // Or `.init(property:)` if you want the line to be shorter.

    private let _currentValue = MutableProperty<Int>(0)
}
```

NachoSoto:
> However, I don't see MutablePropertyType.immutableView as any simpler than AnyProperty(property: PropertyType)

Not at all, that's why I wanted to hear everyone's opinion. I agree it doesn't add anything that couldn't be done before.

@andersio the problem with that example is that the property becomes mutable though!

andersio:
@NachoSoto You can use `public private(set)`. ;)

NachoSoto:
It's still mutable :D

andersio:
@NachoSoto Keeping it immutable to the outside world would be fairly enough, wouldn't it? `public private(set)` already prevents the property from being replaced at compile time.

NachoSoto:
Closing this since it doesn't seem to be of interest.

Issue #2425, opened 2015-09-28
Binding to a dynamic property doesn't compile


czechboy0:
Hi, I'm trying to bind a signal producer sending over `String`s to a `textField`'s `stringValue` property in `NSTableCellView`. However, the compiler refuses to compile this
```
DynamicProperty(object: view.textField!, keyPath: "stringValue") <~ syncerViewModel.status
```
where `syncerViewModel.status` is of type `SignalProducer<String, NoError>`. However, what **does** compile is when I add an extra pointless map at the end, like this
```
DynamicProperty(object: view.textField!, keyPath: "stringValue") <~ syncerViewModel.status.map { $0 }
```

Can someone please explain to me what I'm doing wrong here? I assumed that just mapping to the same type should not change the underlying producer type.

Anyway, what is also frustrating that this doesn't compile:
```
let source: SignalProducer<String, NoError> = syncerViewModel.status
let destination = DynamicProperty(object: view.textField!, keyPath: "stringValue")
destination <~ source
```
error: `Binary operator '<~' cannot be applied to operands of type 'DynamicProperty' and 'SignalProducer<String, NoError>'`

I'm on OS X 10.10.5, Xcode 7 (release version). What am I doing wrong?

ikesyo:
`DynamicProperty` needs `SignalProducer<AnyObject?, NoError>` and `String` itself is not `AnyObject` but can be bridged to that through `NSString`. `map { $0 }` is really bridging from `String` to `NSString?` by the type inferencer. That is the trick.

czechboy0:
Ah, I see, that makes sense, thanks so much! I'll just bridge it explicitly I guess to keep it readable. Thanks again!

Issue #2198, opened 2015-07-20
DynamicProperty <~ SignalProducer requires trailing map


calebd:
I have a `SignalProducer<UIImage?, NoError>` that I want to bind to a `DynamicProperty`. For some reason the binding fails saying that the `<~` operator cannot be applied to those objects. The only way to make it work is to add a trailing `|> map { $0 }` which doesn't make any sense to me. The type of the returned `SignalProducer` is identical with or without the map. Has anyone else seen this behavior?

jspahrsummers:
`DynamicProperty` must be bound to `AnyObject?` _exactly_. A type matching it isn't good enough.

neilpa:
> The type of the returned SignalProducer is identical with or without the map.

Not exactly. The compiler's type-inference is kicking and implicitly upcasting to `AnyObject` and wrapping in an optional.

Issue #1884, opened 2015-04-10
Why is mapping required when using <~ to bind Signal to DynamicProperties


ohwutup:
I am attempting to bind a `Signal<String,NoError>` to a `DynamicProperty` of a UILabel's text.

It works great, but _only_ when a call to `map` is inline.

Given the following `Signal<String,NoError>`,
```
lazy var proximityLabelText: Signal<String,NoError> = {
        return self.proximityManager.beacons |> map { beacons in
            switch beacons[0].proximity {
            case .Immediate:
                return "immediate"
            case .Near:
                return "near"
            case .Far:
                return "far"
            case .Unknown:
                return "unknown"
            }
        }
    }()
```

This works:
```
DynamicProperty(object: self.locationLabel, keyPath: "text") <~ self.viewModel.proximityLabelText |> map { $0 }
```
But this throws a compiler error that `Bindary operator '<~' cannot be applied to operands of type 'DynamicProperty' and 'Signal<String,NoError>`:
```
let producer = self.viewModel.proximityLabelText |> map { $0 }
DynamicProperty(object: self.locationLabel, keyPath: "text") <~ producer
```

Is something wrong here or am I missing something?

jspahrsummers:
`DynamicProperty` only allows binding values that conform to `AnyObject`. You'll have to use `NSString` or otherwise map to an `AnyObject` value.

ohwutup:
Thank makes sense, but when I do this:
```
let producer = self.viewModel.proximityLabelText |> map { $0 as NSString }
DynamicProperty(object: self.locationLabel, keyPath: "text") <~ producer
```
it just gives me the same error as before with `NSString` in place of `String`.

jspahrsummers:
Sorry, looks like the type specifically needs to be `AnyObject`. Can you try typing `producer` as `SignalProducer<AnyObject, NoError>`?

ohwutup:
Same error, just now with `AnyObject`

jspahrsummers:
Sorry again, it should be `AnyObject?` (i.e., `Optional<AnyObject>`).

ohwutup:
that works! thanks @jspahrsummers!

ohwutup:
and just for clarity here, the reason why it is needing a `Signal<AnyObject?, NoError>` is because of this? https://github.com/ReactiveCocoa/ReactiveCocoa/blob/swift-development/ReactiveCocoa/Swift/Property.swift#L120-L128

[...]

Issue #1858, opened 2015-03-26
Binding to dynamic signal without casting to `AnyObject?`


Ben-G:
It seems that because the type of `Value` for `DynamicProperty` is `AnyObject?` every signal that wants to bind to a dynamic property needs to be casted to `AnyObject?`.

Looking at the method signature of the bind operator this makes sense:
`public func <~ <T: AnyObject, P: MutablePropertyType, DynamicProperty where P.Value == T>(property: P, signal: Signal<T, NoError>) -> Disposable`

The type of the signal and the value of the property need to match exactly. **Is there no way in Swift to express that any subtype of `AnyObject?` should be acceptable as well?**

Here's a snippet I've been playing around with:

      let signal = post!.likes.producer.lift(map { $0.reduce("") { $1.username } as AnyObject? })
      textProperty = DynamicProperty(object: likesLabel, keyPath: "text")
      textProperty! <~ signal

When casting to `NSString?` or `NSString` instead of `AnyObject?` the compiler complains that the `<~` operator is not applicable for the given types.

Could someone with a better understanding of Swift's type system explain wether the generic constraint I'm looking for is available or not?

Ben-G:
Interesting. I just found out that the issue does not occur when binding the signal directly, without creating an intermediate variable:

    textProperty! <~ post!.likes.producer.lift(map { return self.stringFromUserlist($0) })

jspahrsummers:
On the latest version of `swift-development`, `<~` is defined generically over any `MutablePropertyType`. I'm afraid there's no way to loosen its type constraints specifically for `DynamicProperty`.

PR #2551, opened 2015-11-12
Overload `<~` to automatically bridge values to Objective-C

sharplet:
There seems to a [good](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2425) [amount](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2198) [of](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1884) [evidence](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1858) that doing this kind of thing is no fun:

```swift
DynamicProperty(object: myLabel, keyPath: "text") <~ gimmeStrings.map { $0 } // 😞
```

I worked out a way to overload `<~` for types that are `_ObjectiveCBridgeable` and automatically bridge them.

```swift
DynamicProperty(object: myLabel, keyPath: "text") <~ gimmeStrings // \o/
```

I'm not sure how I feel about the use of `_ObjectiveCBridgeable` and `_bridgeToObjectiveC()`. But I think this is pretty nice to have, and it should be equivalent to the implicitly bridged version.

-----

P.S.

I didn't include it in the PR, but I also add this function to my project which brings it back to RAC 2 levels of awesomeness:

```swift
func RAC(object: NSObject?, _ keyPath: String) -> DynamicProperty {
  return DynamicProperty(object: object, keyPath: keyPath)
}

RAC(myLabel, "text") <~ gimmeStrings
```

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -254,6 +254,11 @@ public func <~ <Value: _ObjectiveCBridgeable>(property: DynamicProperty, produce
 	return property <~ producer.map { $0._bridgeToObjectiveC() }
 }
 
+/// Binds `destinationProperty` to the latest values of `sourceProperty`, automatically bridging values to Objective-C.
+public func <~ <Value: _ObjectiveCBridgeable, Source: PropertyType where Source.Value == Value>(destinationProperty: DynamicProperty, sourceProperty: Source) -> Disposable {
+	return destinationProperty <~ sourceProperty.producer
+}
+
 
 /// Binds `destinationProperty` to the latest values of `sourceProperty`.
 ///
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -244,6 +244,17 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
 }
 
 
+/// Binds a signal to a `DynamicProperty`, automatically bridging values to Objective-C.
+public func <~ <Value: _ObjectiveCBridgeable>(property: DynamicProperty, signal: Signal<Value, NoError>) -> Disposable {
+	return property <~ signal.map { $0._bridgeToObjectiveC() }
+}
+
+/// Binds a signal producer to a `DynamicProperty`, automatically bridging values to Objective-C.
+public func <~ <Value: _ObjectiveCBridgeable>(property: DynamicProperty, producer: SignalProducer<Value, NoError>) -> Disposable {
+	return property <~ producer.map { $0._bridgeToObjectiveC() }
+}
+
+
 /// Binds `destinationProperty` to the latest values of `sourceProperty`.
 ///
 /// The binding will automatically terminate when either property is
--- a/ReactiveCocoaTests/Swift/PropertySpec.swift
+++ b/ReactiveCocoaTests/Swift/PropertySpec.swift
@@ -672,6 +672,35 @@ class PropertySpec: QuickSpec {
 					expect(bindingDisposable.disposed) == true
 				}
 			}
+
+			describe("to a dynamic property") {
+				var object: ObservableObject!
+				var property: DynamicProperty!
+
+				beforeEach {
+					object = ObservableObject()
+					expect(object.rac_value) == 0
+
+					property = DynamicProperty(object: object, keyPath: "rac_value")
+				}
+
+				afterEach {
+					object = nil
+				}
+
+				it("should bridge values sent on a signal to Objective-C") {
+					let (signal, observer) = Signal<Int, NoError>.pipe()
+					property <~ signal
+					observer.sendNext(1)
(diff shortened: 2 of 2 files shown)


neilpa:
While this is cool, I'm :-1: on adding it. Each additional operator overload almost always creates a significant negative impact to compile times. I've seen exponential or worse degradations on other projects.

I also don't want to encourage further use of `DynamicProperty` because it effectively circumvents Swift's type safety. To that end, I've been maintaining [Rex](https://github.com/neilpa/Rex) which at this point is mostly about providing [type-safe cocoa bindings](https://github.com/neilpa/Rex/tree/a64d1967bfb44ff1fb0b8f52862a388d0ed3718a/Source/UIKit). At this point these should probably be moved into RAC but I haven't found the time to port them, flush out the missing properties, add documentation, etc.

sharplet:
> Each additional operator overload almost always creates a significant negative impact to compile times. I've seen exponential or worse degradations per-overload on other projects.

Are you referring to ReactiveCocoa compile times, or compile times for callers of the operator?

neilpa:
Existing usage of the operator

neilpa:
Although I'll admit that most of my data on compile-time increases when calling overloaded operators is anecdotal from twitter and other github issues.

NachoSoto:
Thanks for the PR!

> I also don't want to encourage further use of DynamicProperty because it effectively circumvents Swift's type safety.

For sure, but we need to improve the current situation because right now it's really hard to use with `UIKit`/`AppKit`.

I think this is interesting too, but it only solves a tiny portion of the problems. What I brought up in #2535 was a more general perspective of the issue, which how hard to compose `PropertyType` is. If we can solve that problem in a general way then we'll have a framework to improve things like this :)

sharplet:
So, for posterity at least, I've just pushed a commit that adds one more overload for binding a source `PropertyType` with bridgeable values to `DynamicProperty`.

Compile-time performance is really important, so I think it would be really interesting to measure the difference these overloads would make to a project that uses a lot of bindings. Validating our assumptions there could at least be data to back a radar about compile-time performance for overloaded operators.

I don't feel like my current project is of a sufficient size to really validate the performance question (at least at its current early stage). Is there perhaps anyone we could invite to try out this branch and report back?

sharplet:
Also, regarding the future design questions about composition of properties, is the hope to solve that for RAC 4 (in case breaking changes are required)? In my use, I'm finding that it adds a lot of value to my codebase in terms of readability, so I think it would be great to have *some* kind of enhancement available in 4.0.

RuiAAPeres:
@neilpa @NachoSoto should this be merged or closed?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -232,6 +232,22 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
...
 
 
+/// Binds a signal to a `DynamicProperty`, automatically bridging values to Objective-C.
+public func <~ <Value: _ObjectiveCBridgeable>(property: DynamicProperty, signal: Signal<Value, NoError>) -> Disposable {
+  return property <~ signal.map { $0._bridgeToObjectiveC() }
+}
+
+/// Binds a signal producer to a `DynamicProperty`, automatically bridging values to Objective-C.
+public func <~ <Value: _ObjectiveCBridgeable>(property: DynamicProperty, producer: SignalProducer<Value, NoError>) -> Disposable {
+  return property <~ producer.map { $0._bridgeToObjectiveC() }
```
mdiep: The indentation here and above is off. RAC uses tabs. ☺️

mdiep:
I agree both that (a) the current situation is not great and (b) the type-unsafely of `DynamicProperty` is not great. Furthermore, I think @NachoSoto's idea to make `DynamicProperty` generic over `Value` (as proposed in #2535) is a great one. AIUI, that would eliminate the need for the operators proposed here.

In light of that, I'd suggest:

1. That we merge these operators to help with the situation in RAC 4.

2. That we make `DynamicProperty` generic over `Value` for RAC 5 and remove these operators.

What does everyone think of that? @sharplet, would you be interested in opening a PR for (2)?

sharplet:
@mdiep I've fixed up the indentation and rebased on master. I took a stab at (2), and I'll be pushing a PR for it shortly!

sharplet:
@mdiep I've opened https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2838, with the breaking changes to `DynamicProperty`.

ikesyo:
#2852 is merged for RAC 5 with breaking changes, but should this still be merged for current RAC 4?

mdiep:
👍 to merging this. Let's get the CI build to pass though.

sharplet:
Looks like the build passed on the last commit?

mdiep:
If failed the first time, but I reran it. 😄

sharplet:
Excellent, thanks everyone!

PR #2571, opened 2015-11-25
Associative lift operators to fix synchronous sampleOn

neilpa:
I decided to keep `liftLeft` and `liftRight` private for now. I don't think this is a complexity we should expose in the API given that these only matter for some operators and only when both producers are synchronous.


NachoSoto:
> I decided to keep liftLeft and liftRight private for now. I don't think this is a complexity we should expose in the API

Completely :+1:

NachoSoto:
Awesome :shipit: 

Great job!

PR #2573, opened 2015-11-26
Update Result to 1.0

gfontenot:
This makes the constraint on Result much looser (since we're no longer
targeting a specific tag), so this should make it nicer for users as
well.


NachoSoto:
Good times!

NachoSoto:
(Merging as soon as tests finish)

PR #2576, opened 2015-11-29
Add skipUntil operator

### LINKED ISSUE #2575: skipUntil and time-framed Producers
I'm struggling to get this right. I have a SignalProducer that will send Quality of Service prediction values every 20 milliseconds. My requirements are:

* Discard all the values sent during the first `filteringTime` seconds of every experiment.
* Throttle the output to `throttlingTime` seconds.
* Restart the signal producer every `experimentDuration` seconds.
* Perform `maxExperiments` restarts, then complete.

This is my best approach for the moment:

```swift
func qualityOfServiceSignalProducer(filteringTime filteringTime: NSTimeInterval, throttlingTime: NSTimeInterval, maxExperiments: Int, experimentDuration: NSTimeInterval) -> SignalProducer<QualityLevel, NoError>
{
    let filterTimerProducer = SignalProducer<(), NoError> { sink, disposable in
        self.scheduler.scheduleAfter(self.scheduler.currentDate.dateByAddingTimeInterval(filteringTime)) {
            sink.sendCompleted()
        }
    }
    
    let endExperimentTimerProducer = SignalProducer<(), NoError> { sink, disposable in
        self.scheduler.scheduleAfter(self.scheduler.currentDate.dateByAddingTimeInterval(experimentDuration)) {
            sink.sendCompleted()
        }
    }

    return self.tesseractReactiveMonitor.forecastSignalProducer() //the signal producer that sends QoS predictions on start
        .skipUntil(filterTimerProducer)
        .throttle(throttlingTime, onScheduler: scheduler)
        .takeUntil(endExperimentTimerProducer)
        .times(maxExperiments)
}

```

What bothers me the most is to have to use these two timers, `filterTimerProducer` and `endExperimentTimerProducer`, in combination with `takeUntil`-like operators, which feels a little bit artificial. I wonder if there is a better approach using existing RAC operators.

My other concern is that RAC doesn't provide a `skipUntil` operator. Given the symmetry between:

* `take` ~> `skip`
* `takeWhile` ~> `skipWhile`

I was expecting to also have `takeUntil` ~> `skipUntil`. Perhaps I'm missing something and there is a better way to implement this `skipUntil` behavior with existing operators?

Anyway, I've created a `skipUntil` implementation by pretty much mimicking `takeUntil` in this commit: https://github.com/cesteban/ReactiveCocoa/commit/fbf467cbecec8b50122b1e385d3051ed29c47321

I will send a pull request if you think this could be useful for anybody else. It is definitely useful for me unless anybody can suggest a better approach.

Thanks a lot. Any advice will be greatly appreciated.

cesteban:
Add `skipUntil` operator, which will skip all values from the `signal` until a `trigger` sends `.Next` or `.Completed`. From that point all values from the `signal` will be emitted normally.

Any feedback is highly appreciated.

Fixes #2575.


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1016,6 +1016,27 @@ extension SignalType {
 			return disposable
 		}
 	}
+	
+	/// Does not forward any values from `self` until `trigger` sends a Next or
```
NachoSoto: sends a Next or Completed **event**.

ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
```diff
@@ -298,6 +298,48 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+					switch event {
+					case let .Next(value):
+						lastValue = value
+					default:
+						break
+					}
+				}
+			}
+			
+			it("should skip values until the trigger fires") {
```
NachoSoto: Can you write another test for when the `trigger` completes without sending any events?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -784,6 +784,48 @@ class SignalSpec: QuickSpec {
...
+					switch event {
+					case let .Next(value):
+						lastValue = value
+					default:
+						break
+					}
+				}
+			}
+			
+			it("should skip values until the trigger fires") {
```
NachoSoto: Can you write another test for when the trigger completes without sending any events?

NachoSoto:
This looks good! The failing test is unrelated. I opened [an issue](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2578) to fix that flaky test.

ReactiveCocoa/Swift/Signal.swift
cesteban: Yup, sorry.

ReactiveCocoaTests/Swift/SignalSpec.swift
cesteban: Of course! I also added the same tests to `takeUntil` spec, to keep the correspondence. Have a look and let me know your thoughts.

NachoSoto:
:shipit:

Issue #2044, opened 2015-05-23
Removing Disposable


jspahrsummers:
`Disposable`s in general, but especially the management of them in signal/producer lifetimes, feel like kind of a hack now that we have a proper `.Interrupted` event.

It’s probably too late to do anything about this for RAC 3, but maybe we should rethink what disposal means for future releases.

Thoughts?

/cc @nomothetis @andymatuschak

nomothetis:
Sorry, I was on vacation.

I think this is conceptually a great idea. I'd have to think carefully about the problems they solve and how `.Interrupted` responds to them to see what the real cost would be, though.

neilpa:
We should experiment with this for RAC 4 and see how well it works.

NachoSoto:
I'd still love to consider this, but how does everyone feel about bumping it out of the initial `4.0` release?

andymatuschak:
Agreed.

neilpa:
> I'd still love to consider this, but how does everyone feel about bumping it out of the initial 4.0 release?

My plan was to take a first crack at this either this weekend or next week so I'd like to keep it for now.

neilpa:
I'd also like Xcode 7.1 to GM and stabilize before we attempt to finalize 4.0 given all the other shit it broke.

NachoSoto:
@neilpa how do you feel about this? what do you think about leaving it for `4.1`?

neilpa:
I explored this some about a month ago. I didn't get very far and it's going to take more serious thought and work to pull off. I'm fine delaying it assuming we can figure out other workarounds for the `flatten` disposal bugs.

NachoSoto:
:+

andersio:
I have tried to put together a prototype. But my assumption is that the only replacement to `Disposable` _once and for all_ would be circular signals in `Signal.observe`, which would be extended to accept a `trigger` that can dispose of the signal.

The most important tidbit is that `Signal.observe` would no longer return a "detacher", which is crucial to the entire lifetime management model. The top-to-bottom `interrupted` wouldn't help here, because detaching an observer is an action in a reverse direction.

mdiep:
Closing this since it's now a ReactiveSwift issue, not a ReactiveCocoa one.

PR #2582, opened 2015-12-01
Flatten fixes

neilpa:
Fixes the disposal problems from `flatten`ing producers-of-producers brought up in #2453. I've split out the operators affected by this because they can't be implemented in terms of `lift`. Instead I've had to do some ad-hoc sharing of implementations across the different operators. Maybe there's a more general way to abstract them that I missed though.

Also fixes #2548.

~~As a bonus I've started to address #2044 by removing the optional disposal when creating signals. This simplifies some of the disposal management in signal operators which had to aggregate these. In practice, I've never seen a signal not return nil. Unlike producers, signal cleanup is directly tied to termination which they directly control since they can't be prematurely interrupted.~~


neilpa:
Flaky test again, but should be good otherwise

NachoSoto:
> In practice, I've never seen a signal not return nil.

I checked our codebase and the only place where I found this was this use case:
```swift
let signal = Signal { observer in
	return producer.start(observer)
}
```

How would disposal be forwarded to the `SignalProducer` without being able to explicitly return a `Disposable`?

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
+//
+//  Flatten.swift
```
NachoSoto: :+1:

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
...
+	/// The returned producer completes only when `producer` and all producers
+	/// emitted from `producer` complete.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	private func concat() -> SignalProducer<Value.Value, Error> {
+		return SignalProducer<Value.Value, Error> { observer, disposable in
+			let state = ConcatState(observer: observer, disposable: disposable)
+
+			self.startWithSignal { signal, signalDisposable in
+				disposable += signalDisposable
+				signal.observeConcat(state)
```
NachoSoto: Nice

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
...
+	/// The returned signal completes when `signal` and the latest inner
+	/// signal have both completed.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	private func switchToLatest() -> Signal<Value.Value, Error> {
+		return Signal<Value.Value, Error> { observer in
+			self.observeSwitchToLatest(observer, Atomic(LatestState<Value, Error>()), SerialDisposable())
+		}
+	}
+
+	private func observeSwitchToLatest(observer: Observer<Value.Value, Error>, _ state: Atomic<LatestState<Value, Error>>, _ latestInnerDisposable: SerialDisposable) {
```
NachoSoto: I think this makes a lot of sense!!

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -889,13 +478,9 @@ extension SignalType where Value: EventType, Error == NoError {
...
 	/// Injects side effects to be performed upon the specified signal events.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func on(event event: (Event<Value, Error> -> ())? = nil, failed: (Error -> ())? = nil, completed: (() -> ())? = nil, interrupted: (() -> ())? = nil, terminated: (() -> ())? = nil, disposed: (() -> ())? = nil, next: (Value -> ())? = nil) -> Signal<Value, Error> {
+	public func on(event event: (Event<Value, Error> -> ())? = nil, failed: (Error -> ())? = nil, completed: (() -> ())? = nil, interrupted: (() -> ())? = nil, terminated: (() -> ())? = nil, next: (Value -> ())? = nil) -> Signal<Value, Error> {
 		return Signal { observer in
-			let disposable = CompositeDisposable()
-
-			_ = disposed.map(disposable.addDisposable)
-
-			disposable += signal.observe { receivedEvent in
```
NachoSoto: Can you explain why this is okay? Won't this subscription leak now?

NachoSoto:
This makes a lot of sense, great job @neilpa.
I have some concerns about removing the returned `Disposable` but it might be because I'm not seeing how this is equivalent.

neilpa:
> How would disposal be forwarded to the SignalProducer without being able to explicitly return a Disposable?

It doesn't need to because you can't early terminate a signal via disposal. Your example effectively makes the started producer un-cancelable (even before this change). So we only care about ensuring it gets invoked on termination which [producers already handle](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/4faa7bc75cd30682e9e5615bb6359988f0339800/ReactiveCocoa/Swift/SignalProducer.swift#L251-L255).

ReactiveCocoa/Swift/Signal.swift
neilpa: In this case and a lot of other's it's safe since the two signal lifetimes are identical. Although things like `take` still technically work they'll end up subscribe longer than needed.I thought I could get away with partially doing #2044 here but I guess not.

neilpa: Well technically this could be fixed for the operators where it matters, but not worth it without a holistic solution.

neilpa:
I've reverted the disposable bits so now it's just the flatten changes.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1004,6 +1004,27 @@ class SignalProducerSpec: QuickSpec {
...
+				let (baseProducer, baseObserver) = SignalProducer<Int, TestError>.buffer()
+
+				var (disposed, interrupted) = (false, false)
+				let disposable = baseProducer
+					.flatMapError { (error: TestError) -> SignalProducer<Int, TestError> in
+						return SignalProducer<Int, TestError> { _, disposable in
+							disposable += ActionDisposable { disposed = true }
+						}
+					}
+					.start(Observer(interrupted: { _ in
```
NachoSoto: Nit: you don't need `_ in`

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1390,6 +1411,97 @@ class SignalProducerSpec: QuickSpec {
...
+			describe("disposal") {
+				var completeOuter: (Void -> Void)!
+				var disposeOuter: (Void -> Void)!
+				var execute: (FlattenStrategy -> Void)!
+
+				var innerDisposable = SimpleDisposable()
+				var interrupted = false
+
+				beforeEach {
+					execute = { strategy in
```
NachoSoto: Cool

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1390,6 +1411,97 @@ class SignalProducerSpec: QuickSpec {
...
+					}
+
+					it("should cancel inner work when disposed after the outer producer completes") {
+						execute(.Concat)
+
+						completeOuter()
+						disposeOuter()
+
+						expect(innerDisposable.disposed).to(beTruthy())
+						expect(interrupted).to(beTruthy())
```
NachoSoto: Maybe simplify all these with the new syntax?
```
expect(innerDisposable.disposed) == true
expect(interrupted) == true
```

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1390,6 +1411,97 @@ class SignalProducerSpec: QuickSpec {
...
+					}
+				}
+
+				describe("Merge") {
+					it("should cancel inner work when disposed before the outer producer completes") {
+						execute(.Merge)
+
+						disposeOuter()
```
NachoSoto: Maybe verify that
```swift
expect(innerDisposable.disposed) == false
expect(interrupted) == false
```
before disposing?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1390,6 +1411,97 @@ class SignalProducerSpec: QuickSpec {
...
+
+						expect(innerDisposable.disposed).to(beTruthy())
+						expect(interrupted).to(beTruthy())
+					}
+
+					it("should cancel inner work when disposed after the outer producer completes") {
+						execute(.Latest)
+
+						completeOuter()
+						disposeOuter()
```
NachoSoto: Maybe verify the flags before disposing here too (and maybe after completing), for completeness?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1390,6 +1411,97 @@ class SignalProducerSpec: QuickSpec {
...
+						expect(innerDisposable.disposed).to(beTruthy())
+						expect(interrupted).to(beTruthy())
+					}
+				}
+
+				describe("Latest") {
+					it("should cancel inner work when disposed before the outer producer completes") {
+						execute(.Latest)
+
+						disposeOuter()
```
NachoSoto: Ditto here

NachoSoto:
Awesome!! :+1: I think this is good to go, just a few suggestions for the tests.
It would be awesome to get another pair of eyes on this though since it's a pretty important change @ReactiveCocoa/reactivecocoa.

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
...
+			self.observeMerge(MergeState(observer: relayObserver, disposable: nil))
+		}
+	}
+
+	private func observeMerge(state: MergeState<Value.Value, Error>) -> Disposable? {
+		return self.observe { event in
+			switch event {
+			case let .Next(producer):
+				producer.startWithSignal { innerSignal, innerDisposable in
+					state.inFlight.modify { $0 + 1 }
```
liscio: I feel bad for not having an alternate suggestion here, but the exposure of `inFlight` and requiring external modifications (not to mention the fact that modifications must happen within a `modify {}` block) seems "smelly" to me. Especially when paired with `decrementInFlight()` which adds a side-effect (completing the signal) as well as decrementing.

At the very least, maybe clean up `MergeState`'s public API? Ideally with new naming to hide the fact that there is an increment/decrement at all: `inFlightDidBegin()`, `inFlightDidEnd()`? Or `addInFlight()`, `removeInFlight()`?

NachoSoto: While I appreciate your suggestion to improve this, I think it would be great to leave that for a separate change so that this is purely rearranging the existing implementation. Otherwise we risk potentially not knowing what broke this operator.

mdiep:
Does this also fix #2548?

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
...
+	/// The returned signal completes only when `signal` and all producers
+	/// emitted from `signal` complete.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	private func concat() -> Signal<Value.Value, Error> {
+		return Signal<Value.Value, Error> { observer in
+			self.observeConcat(ConcatState(observer: observer, disposable: nil))
+		}
+	}
+
+	private func observeConcat(state: ConcatState<Value.Value, Error>) -> Disposable? {
```
mdiep: If I were to write this, I would construct the `ConcatState` inside `observeConcat`—let the caller pass in the observer and an optional disposable. This more closely matches `Signal.observe()` and narrows the usage of `ConcatState` to inside `observeConcat`. (AFAICT it's an implementation detail—you don't need to extract those bits to a separate type.)

NachoSoto:
Maybe! @neilpa do you want to cherry-pick that test? :) 7575aae286d69a8fec07ebdaf9b85fa6b0540180

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
...
+		self.disposable = disposable
+	}
+
+	func decrementInFlight() {
+		let orig = inFlight.modify { $0 - 1 }
+		if orig == 1 {
+			observer.sendCompleted()
+		}
+	}
+}
```
mdiep: I don't think `MergeState` buys us very much in terms of abstraction. IMO it makes things a little harder to read since the `sendCompleted` in `decrementInFlight` is the only real logic. I would create and use the `Atomic` directly from `observeMerge` and get rid of this class.

ReactiveCocoa/Swift/Flatten.swift
mdiep: I left [a similar comment](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2582/files#r46477295) below before reading this one. `MergeState` is new in this PR. I think the existing implementation would actually be preferred.

ReactiveCocoa/Swift/Flatten.swift
mdiep: I just noticed this discussion is already happening [above](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2582/files#r46462825).

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
...
+	/// The returned signal completes when `signal` and the latest inner
+	/// signal have both completed.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	private func switchToLatest() -> Signal<Value.Value, Error> {
+		return Signal<Value.Value, Error> { observer in
+			self.observeSwitchToLatest(observer, Atomic(LatestState<Value, Error>()), SerialDisposable())
+		}
+	}
+
+	private func observeSwitchToLatest(observer: Observer<Value.Value, Error>, _ state: Atomic<LatestState<Value, Error>>, _ latestInnerDisposable: SerialDisposable) -> Disposable? {
```
mdiep: AFAICT there's not a good reason to pass in `state`. It looks like it's always initialized with a `Atomic(LatestState<Value, Error>())`.

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,571 @@
...
+	/// Catches any failure that may occur on the input signal, mapping to a new producer
+	/// that starts in its place.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func flatMapError<F>(handler: Error -> SignalProducer<Value, F>) -> Signal<Value, F> {
+		return Signal { observer in
+			self.observeFlatMapError(handler, observer) { _ in }
+		}
+	}
+
+	private func observeFlatMapError<F>(handler: Error -> SignalProducer<Value, F>, _ observer: Observer<Value, F>, onErrorStart: Disposable -> ()) -> Disposable? {
```
mdiep: Can we have callers pass in a disposable instead of a block that takes a disposable?

mdiep:
Looks functionally good to me. :+1: Just a few suggestions about about some of the private interfaces.

ReactiveCocoa/Swift/Flatten.swift
liscio: @NachoSoto "broke" is a relative term at this point, isn't it? `MergeState` is brand-new and not in any production code, so I imagine that any adjustments in the future will come with even higher scrutiny. Just trying to flag it as early as possible, since it stood out to me as confusing/strange on the initial read.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
neilpa: I'd rather wait and do this for all instances in one big change.

ReactiveCocoa/Swift/Flatten.swift
neilpa: I was trying to minimize churn by keeping these as close to the original as possible. That said, I'm happy to clean this and the other private interfaces up.

neilpa:
> Does this also fix #2548?

Yup (and I grabbed the previously failing case)

neilpa:
:boom:

jdauphant:
What is the state of this MR ?

Apparently an Objective-C test failed "/Users/travis/build/ReactiveCocoa/ReactiveCocoa/ReactiveCocoaTests/Objective-C/RACTargetQueueSchedulerSpec.m:42: failed - expected to eventually match (Stall on main thread)., got <1.0000>"

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,552 @@
...
+extension SignalProducerType where Value: SignalProducerType, Error == Value.Error {
+	/// Flattens the inner producers sent upon `signal` (into a single signal of
+	/// values), according to the semantics of the given strategy.
+	///
+	/// If `signal` or an active inner producer fails, the returned signal will
+	/// forward that failure immediately.
+	///
+	/// `Interrupted` events on inner producers will be treated like `Completed`
+	/// events on inner producers.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
```
jdauphant: The comment should be about producer.
warn_unused_result should be "Did you forget to call `start` on the producer?"

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,552 @@
...
+extension SignalProducerType where Value: SignalType, Error == Value.Error {
+	/// Flattens the inner signals sent upon `signal` (into a single signal of
+	/// values), according to the semantics of the given strategy.
+	///
+	/// If `signal` or an active inner signal emits an error, the returned
+	/// signal will forward that error immediately.
+	///
+	/// `Interrupted` events on inner signals will be treated like `Completed`
+	/// events on inner signals.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
```
jdauphant: The comment should be about producer.
warn_unused_result should be "Did you forget to call `start` on the producer?"

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,552 @@
...
+	/// Returns a producer which sends all the values from each producer emitted from
+	/// `producer`, waiting until each inner producer completes before beginning to
+	/// send the values from the next inner producer.
+	///
+	/// If any of the inner producers emit an error, the returned producer will emit
+	/// that error.
+	///
+	/// The returned producer completes only when `producer` and all producers
+	/// emitted from `producer` complete.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
```
jdauphant: warn_unused_result should be "Did you forget to call `start` on the producer?"

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,552 @@
...
+}
+
+extension SignalProducerType {
+	/// Maps each event from `self` to a new producer, then flattens the
+	/// resulting producers (into a producer of values), according to the
+	/// semantics of the given strategy.
+	///
+	/// If `self` or any of the created producers fail, the returned producer
+	/// will forward that failure immediately.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
```
jdauphant: warn_unused_result should be "Did you forget to call `start` on the producer?"

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -0,0 +1,552 @@
...
+extension SignalProducerType where Value: SignalProducerType, Error == Value.Error {
+	/// Returns a signal that forwards values from the latest signal sent on
+	/// `signal`, ignoring values sent on previous inner signal.
+	///
+	/// An error sent on `signal` or the latest inner signal will be sent on the
+	/// returned signal.
+	///
+	/// The returned signal completes when `signal` and the latest inner
+	/// signal have both completed.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
```
jdauphant: warn_unused_result should be "Did you forget to call `start` on the producer?"

ReactiveCocoa/Swift/Flatten.swift
neilpa: This one is correct, it's a signal extension.

ReactiveCocoa/Swift/Flatten.swift
neilpa: This one is correct, it's a signal extension.

neilpa:
Should be good for another review

ReactiveCocoa/Swift/Flatten.swift
jdauphant: Yep you have correct it. It's just a github bug, my comment have been attached to a different line in the "Files changed" tab.
screenshot from 2015-12-13 04-47-33

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1004,6 +1004,27 @@ class SignalProducerSpec: QuickSpec {
...
+				let (baseProducer, baseObserver) = SignalProducer<Int, TestError>.buffer()
+
+				var (disposed, interrupted) = (false, false)
+				let disposable = baseProducer
+					.flatMapError { (error: TestError) -> SignalProducer<Int, TestError> in
+						return SignalProducer<Int, TestError> { _, disposable in
+							disposable += ActionDisposable { disposed = true }
+						}
+					}
+					.start(Observer(interrupted: {
```
NachoSoto: Nit: maybe make this just one line?
```
.start(Observer(interrupted: { interrupted = true }))
```

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1004,6 +1004,27 @@ class SignalProducerSpec: QuickSpec {
...
+						interrupted = true
+					}))
+
+				baseObserver.sendFailed(.Default)
+				disposable.dispose()
+
+				expect(interrupted).to(beTruthy())
```
NachoSoto: Maybe change all these to the new syntax?
```
expect(interrupted) == true
expect(disposed) == true
```

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1390,6 +1411,112 @@ class SignalProducerSpec: QuickSpec {
...
+						expect(innerDisposable.disposed).to(beTruthy())
+						expect(interrupted).to(beTruthy())
+					}
+
+					it("should cancel inner work when disposed after the outer producer completes") {
+						execute(.Merge)
+
+						completeOuter()
+
+						expect(innerDisposable.disposed).to(beFalsy())
```
NachoSoto: Ditto

NachoSoto:
This looks good to me! :shirt:

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
neilpa: I'd rather not mix and match and just do a big find/replace fix once

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
neilpa: I'd rather not mix and match and just do a big find/replace fix once

neilpa:
One last tweak

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: Sounds good

NachoSoto:
(I edited the description to reference #2582 as well.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1707,6 +1832,35 @@ class SignalProducerSpec: QuickSpec {
...
+				}.startOn(scheduler)
+
+				var started = false
+				let producer2 = SignalProducer<Int, NoError>() { handler, _ in
+					started = true
+					handler.sendNext(2)
+					handler.sendCompleted()
+				}
+
+				let result = producer1.concat(producer2).take(1).collect().first()
```
NachoSoto: I mentioned this on #2548: maybe put each operator in a separate line?
```swift
let result = producer1
  .concat(producer2)
  .take(1)
  .collect()
  .first()
```

NachoSoto:
Anyone else want to take a final look? Otherwise I'll merge this tomorrow :)

With this I think we're ready for a `RAC 4` RC 1 :D

mdiep:
Looks good to me! :+1:

NachoSoto:
:tada:

mdiep:
:fireworks:

Issue #2586, opened 2015-12-02
Mapped signal leaks if it doesn't terminate


Vadim-Yelagin:
This code causes a genuine memory leak

	let (signal, _) = Signal<Int, NoError>.pipe()
	let _ = signal.map { $0 }

<img width="985" alt="screenshot 2015-12-02 16 27 30" src="https://cloud.githubusercontent.com/assets/5680096/11528564/61f16856-9909-11e5-8439-fc51035f9789.png">

Vadim-Yelagin:
Looking at implementation of `init`, `observe` and `map` I see that the retain cycle:
(original signal) -> (original signal observers) -> (observer added by map) -> (mapped signal's internal observer) -> (mapped signal's generator disposable) -> (disposable from original signal's observe in map) -> (original signal)

NachoSoto:
Cool I guess that confirms the bug!
Let's keep the conversation in #2585 to avoid duplication.

Vadim-Yelagin:
There might be a way to fix this without affecting https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2585 or breaking the signal contract. There is nothing that says that a signal must be retained by its observers or corresponding disposables. The only thing that must retain the signal is its generator observable. While in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2585 the original generator is not deallocated and the signal might potentially generate events and even terminate (which is the only thing that can break the retain cycle), in case of `Signal.never` the generator does not retain the observer it's provided with.
What I propose is to prevent the retain cycle here by making observing disposable capture the signal weakly. It will allow the original signal (e.g. `Signal.never`) to get deallocated, even though it has observers, since it cannot generate any events anymore.

Vadim-Yelagin:
TL;DR Might be not duplicate, please see https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2603

Issue #2585, opened 2015-12-02
Using map on long-living Signal


Vadim-Yelagin:
I have a `Signal` that can exist throughout the entire run time of the app and never complete. Then I have temporary views that observe that signal and dispose of the corresponding `Disposable` when they deinit. This seems to work fine, but if I use some functions like `map` or `filter` on that signal and then observe the resulting signal, I see that those transformed signals stay alive, and the only way to dispose of them is to destroy the original signal (which does not happen).

This is a simple example of this

	class ViewController: UIViewController {

		let (signal, observer) = Signal<Int, NoError>.pipe()

		deinit {
			observer.sendCompleted()
		}

		var counter = 0
		@IBAction func foo() {
			let currentCounter = counter
			let disposable = signal.map {
				value -> Int in
				print("GOT \(value) IN MAP #\(currentCounter)")
				return value
			}.observeNext {
					value in
					print("GOT \(value) IN OBSERVE #\(currentCounter)")
			}
			observer.sendNext(counter)
			disposable?.dispose()
			observer.sendNext(counter + 1000)
			counter++
		}

	}

If I call `foo` multiple times that's what I see

    GOT 0 IN MAP #0
    GOT 0 IN OBSERVE #0
    GOT 1000 IN MAP #0
    GOT 1 IN MAP #0
    GOT 1 IN MAP #1
    GOT 1 IN OBSERVE #1
    GOT 1001 IN MAP #0
    GOT 1001 IN MAP #1
    GOT 2 IN MAP #0
    GOT 2 IN MAP #1
    GOT 2 IN MAP #2
    GOT 2 IN OBSERVE #2
    GOT 1002 IN MAP #0
    GOT 1002 IN MAP #1
    GOT 1002 IN MAP #2
    GOT 3 IN MAP #0
    GOT 3 IN MAP #1
    GOT 3 IN MAP #2
    GOT 3 IN MAP #3
    GOT 3 IN OBSERVE #3
    GOT 1003 IN MAP #0
    GOT 1003 IN MAP #1
    GOT 1003 IN MAP #2
    GOT 1003 IN MAP #3

Is this a bug? Am I wrong to use `map` in this situation? Am I wrong to expose this long-living `Signal` and should use `SignalProducer` instead?

andersio:
[...]

Issue #2203, opened 2015-07-22
Discourage global signals in the Design Guidelines


neilpa:
Should we add a section to the best practices for using producers instead of signals for "global" event streams. `NSNotificationCenter` being the prime example (#1980, #2200).

@ReactiveCocoa/reactivecocoa thoughts?

neilpa:
I'll do this post 3.0

NachoSoto:
I disagree with this. There is a reason why `NSNotificationCenter` subscriptions are better represented with a producer, for sure, because they're controlled by the observer. But not all event streams are the same so that would be an invalid generalization IMO.

mdiep:
I agree with @NachoSoto. I don't think that's a valid generalization.

PR #2588, opened 2015-12-02
Add test case for issue 2585

Vadim-Yelagin:
SignalSpec map should not keep resulting signal alive indefinitely (fails)

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -453,6 +453,12 @@ class SignalSpec: QuickSpec {
 				observer.sendNext(1)
 				expect(lastValue).to(equal("2"))
 			}
+
+			it("should not keep resulting signal alive indefinitely") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal.never.map { $0 }
+
+				expect(signal).to(beNil())
+			}
 		}
 		
 		
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -459,6 +459,18 @@ class SignalSpec: QuickSpec {
 
 				expect(signal).to(beNil())
 			}
+
+			it("should not keep resulting signal alive indefinitely after observing and disposing") {
+				var disposable: Disposable? = nil
+				weak var signal: Signal<AnyObject, NoError>? = {
+					let signal: Signal<AnyObject, NoError> = Signal.never.map { $0 }
+					disposable = signal.observe(Observer())
+					return signal
+				}()
+				expect(signal).toNot(beNil())
+				disposable?.dispose()
+				expect(signal).to(beNil())
+			}
 		}
 		
 		


Vadim-Yelagin:
This test case is based on "init should not keep signal alive indefinitely"

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +453,12 @@ class SignalSpec: QuickSpec {
 				observer.sendNext(1)
 				expect(lastValue).to(equal("2"))
 			}
+
+			it("should not keep resulting signal alive indefinitely") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal.never.map { $0 }
+
+				expect(signal).to(beNil())
```
NachoSoto: Can you change this to `expect(signal).toEventually(beNil())`?

I'm not sure if `expect` creates a strong reference though. I think it does because of the `@autoclosure`.

NachoSoto: Oh never mind, I tested this:
```
class S { }

it("weak") {
	weak var s = S()

	expect(s).to(beNil()) // passes
	expect(s).toEventually(beNil()) // passes too, of course.
}
```

So this is fine :)

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +453,12 @@ class SignalSpec: QuickSpec {
 				observer.sendNext(1)
 				expect(lastValue).to(equal("2"))
 			}
+
+			it("should not keep resulting signal alive indefinitely") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal.never.map { $0 }
```
NachoSoto: I'd be surprised if this fails! This is even more reduced than I think, good catch :)

Could you write another, similar one that actually subscribes and then disposes the subscription?
That way we verify that fixing this also makes that other test pass :)

Vadim-Yelagin: I'm not sure what exactly to test there. Have both signals retained and test that the transform closure is not called after the mapped signal's observer is disposed?

andersio: Like https://github.com/andersio/ReactiveCocoa/pull/1/files#diff-9f6d020c3d0578997abeca0af04f0e16R457, perhaps?

NachoSoto: No, same thing but adding a subscription:
```swift
var signal: Signal<Signal<AnyObject, NoError>? = Signal.never.map { $0 }
weak var weakSignal = signal

var disposable = signal.observe(Observer())!

signal = nil // now there is only one weak reference to signal
expect(weakSignal).toNot(beNil())

disposable.dispose()
expect(weakSignal).to(beNil()) // signal has no observers nor strong references. Should be deallocated.
```

Vadim-Yelagin: @NachoSoto @andersio Yeah, the thing is, I don't know whether to test for one thing or for a completely opposite one. Why don't we try to deal with this simple case, unless anyone thinks that it really should leak.

andersio: @NachoSoto But what you describe isn't exactly the case of #2585, which is about a running signals that neither deallocates nor emit terminating events.

The issue can also happen on normal signals, if `takeUntil` happens later than a `map`, so that the derived signal of the mapped signals ends while the mapped signal is still active.

Nonetheless, it seems another boundary case that may need a fix.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +453,24 @@ class SignalSpec: QuickSpec {
...
+
+				expect(signal).to(beNil())
+			}
+
+			it("should not keep resulting signal alive indefinitely after observing and disposing") {
```
Vadim-Yelagin: @NachoSoto Here's a bit tweaked version of what you suggested

ReactiveCocoaTests/Swift/SignalSpec.swift
NachoSoto: >  Why don't we try to deal with this simple case, unless anyone thinks that it really should leak.

My goal with adding a second, more complex test is so that once we fix the original, we can also verify that this second one also passes. It's possible that there's a separate leak!

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +453,24 @@ class SignalSpec: QuickSpec {
...
+				var disposable: Disposable? = nil
+				weak var signal: Signal<AnyObject, NoError>? = {
```
NachoSoto: Ooooh good idea.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +453,24 @@ class SignalSpec: QuickSpec {
...
+					let signal: Signal<AnyObject, NoError> = Signal.never.map { $0 }
+					disposable = signal.observe(Observer())
+					return signal
+				}()
+				expect(signal).toNot(beNil())
+				disposable?.dispose()
+				expect(signal).to(beNil())
```
NachoSoto: Looks good!

neilpa:
This is basically "by design" at the moment. See #1516 and #1608 for some backstory.

/cc @jspahrsummers who's spent the most time thinking about this.

Vadim-Yelagin:
Does this approach make sense https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2585#issuecomment-161353878 ?

neilpa:
Probably, although I haven't thought through exactly what that looks like.

NachoSoto:
@neilpa we can argue that we don't have to fix the second one (you should do `takeUntil`) anyway, but the first one is clearly a leak, no?

Vadim-Yelagin:
@NachoSoto I believe they both leave the signals in exactly the same state now.

neilpa:
Yes it's a leak. My point was that this was a conscience decision around signal lifetime so that consumers didn't have to explicitly worry about it. I'm all for revisiting that and attempting to fix it if we can without forcing users to manage lifetimes.

Vadim-Yelagin:
@neilpa You mean the users wouldn't worry because they wouldn't know that their signals implicitly leak? : )
Explicitly worrying about terminating every signal and applying operators to signals only after a `takeUntil` is what the customers have to do now, but apparently we aren't informed about it.

neilpa:
Well, I almost universally recommend using a `SignalProducer` and avoid exposing signals directly.

jspahrsummers:
Since signals are expected to evaluate side effects regardless of whether anyone is listening, they _must_ stay alive until termination to preserve the semantics. This is, as @neilpa says, by design.

I don't mind if it gets changed, but that was the rationale so far.

andersio:
@jspahrsummers The PR #2591 works this around by retaining only the internals of the signal. If the `Signal` is not - or no longer - retained by the external world, the deallocated `Signal`s would *never* be attachable again in the future. This means the event streams can be safely terminated, without a terminating event, when they have no observer eventually.

Speaking in another way, if the event stream is no longer attachable anyway, it would have no reason to keep it alive as a zombie when it has no observer.

This doesn't break the existing semantics: Since if the mapped signal is retained, #2591 would have no effect on it.

Vadim-Yelagin:
@jspahrsummers Do you mean side effects like those inside a map's transform block? Aren't they supposed to be pure functions? Could you please give a practical example of such a side effect that's not a bad design? If you're referring to a situation when a signal is associated to some process that does have side effects by design and uses the signal as a way for observing some events produced by that process, then it's the process that performs the events, not the signal, and the signal does not in any way retain that process and keep it alive by not being deallocated itself.

jspahrsummers:
> Aren't they supposed to be pure functions?

No, as this cannot be guaranteed or enforced by the type system.

Vadim-Yelagin:
This doesn't mean that one can type an arbitrary code that compiles and expect that their assumptions based on nothing but misguided intuition and wild guess will be right.

jspahrsummers:
I find it generally unhelpful to have a condescending attitude toward users of this framework. There are valid reasons to inject side effects, your objections notwithstanding.

Vadim-Yelagin:
@jspahrsummers isn't it the case with `Signal` that injecting any side effect can be always done in `observe`?

neilpa:
But that doesn't preclude the signal generator (including closures passed to operators when creating new signals) from being side-effecting. And if we kill the signal before it terminates it could subtly break a lot of existing code depending on signals running to completion even without observers.

neilpa:
These semantics have been in place since at least RAC 2 because there's no way to enforce the purity of signal operators.

Vadim-Yelagin:
@neilpa Oh, I see. The use case here is that instead of creating some operation that uses a `Signal.pipe` to send events, the operation is started inside of `Signal.init`'s generator closure which returns a disposable that terminates the operation. And the assumption here is that the terminating closure will be called only in case that same generator code sends a terminating event down the observer. I haven't really thought of this because it seems rather contrived, since that generator code is in full control to directly terminate the operation without relying on the disposable being disposed at the exact moment in that same generator code.
Situation with primitives like `map` is somewhat different in the sense that it's reasonable to state that the only purpose of the resulting signal is to be observed, and the only purpose of the generator's implementation is to push events to that signal.
Afaik with the current implementation there is no difference for primitives like `map` whether to free up generator resources in a disposable that the generator returns or in response to a terminating event from the original signal. With these changes, the disposable can be called before the original signal terminates when the resulting events will be impossible to observe. Currently there is no way to know this in the generator.
Basically, this gives a new opportunity for optimization an avoiding leaks, and does not break any opportunities that have already existed. What will break is the assumption that closures like the one supplied to `map` will be called in certain moments (which were not documented anyway and can always be repaired by putting side-effect in `observe`), and the assumption the disposable in `Signal.init` will only be disposed as a direct result of that same generator's code (which can be repaired calling the closure directly without unnecessary use of disposable).

Vadim-Yelagin:
I really don't get this argument about inability to enforce purity.
Neither can you enforce absence of memory leaks and deadlocks.
Moreover, you can document that execution of some code is not guaranteed (because of cool optimizations, yay!), and when a customer asks why their code in `map` is not called, you can advise them to move that code to `observe`, and everyone is happy.
But trying to explain why `Signal.never.map` is a memory leak by design in the readme is something that will make lots of people a bit less enthusiastic about trying the library out.

andersio:
I think in #2591, a reasonable trade-off is made. That is we only guarantee the signal would be alive - so that side effect would still be on-going despite having no observers - if and only if the signal is retained by the user. It also aligns with the ARC semantics of the language.

Vadim-Yelagin:
> These semantics have been in place since at least RAC 2
You mean `RACSignal`? Wasn't it like `SignalProducer`, not like `Signal`? The generator was only called upon subscription, and the disposable was disposed upon un-subscription. `Signal` on the other hand is a thing in itself, executing by itself and disposing of itself and by itself.

Vadim-Yelagin:
@andersio I think that's missing the point.
If the question is if some code can potentially break, then it certainly can. If it's something you can test (e.g. checking if weak var becomes nil) it's something some code could rely on.
If the question is if some behavior will become harder or impossible to achieve, I believe the answer is no.
If the question is if more reasonably looking code will result in unexpected and dangerous behavior, then I believe it's that actually less code will.

Vadim-Yelagin:
Most of the confusion here came from the notion that a signal *is* the process that generates events.
This kind of mixing responsibilities within a single identity leads to confusing the execution of the process with the existence of the signal.
My perspective was that a signal is a primitive through which events can be observed. It's not an actor that can on its own perform arbitrary tasks and "have side effects".

neilpa:
> What will break is the assumption that closures like the one supplied to map will be called in certain moments (which were not documented anyway and can always be repaired by putting side-effect in observe)

This is the crux of the problem and I've not heard a strong enough argument for changing the semantics (even if undocumented). This has the potential to break way too many RAC consumers. Tracking these types of bugs down and simply putting the side-effecting code in `observe` blocks isn't nearly as easy as you make it seem.

Vadim-Yelagin:
@neilpa The argument here is that it isn't clear that `signal.map` will retain resources and perform computations for as long as `signal` does not terminate. In case of `Signal.never` this is never, which makes it *a memory leak by design*, and even *maintainers are surprised to know this*.
Saying "use `SignalProducer` instead" and "always do `takeUntil` first` is kind of fine, but sometimes you are almost forced to use `Signal`, and using `takeUntil` is just too cumbersome.
Take this practical example: you already have an `Action` in some kind of singleton, and every now and then you want to `start` some `SignalProducer` and cancel it whenever that `Action` sends a value (e.g. the value means that something happened that makes already running network requests irrelevant).
The obvious, neat and would-be-correct solution is `producer.takeUntil(action.values.map{_ in ()}).start()`. However, it's not correct, because this mapped signal is stuck to the `action.values` signal, and it keeps doing useless work. And the more times such code is executed, the more of these parasite mapped signal pile up.
And there are no nice alternatives here, really.
`producer.takeUntil(action.values.takeUntil(terminator).map{_ in ()}).start()` is correct, but often managing that "terminator" is a pain, if at all possible.
`producer.takeUntil(SignalProducer(action.values).map{_ in ()}).start()` is probably correct too, but why is there this `public func takeUntil(trigger: Signal<(), NoError>) -> SignalProducer<Value, Error>` in the API then if you can't use it?
Luckily, there is not a lot of situations where you need (or even have) a signal. That's probably the only reason why these leak are not a big issue now. But then there is https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2543 I wonder how many people will begin mapping properties of some singletons all over their codebase (not condescending, nothing wrong with them). How am I to explain my peers in code review why they cannot do obvious and apparently harmless things with the library that I had been promoting?

Vadim-Yelagin:
Essentially, the proposed change here is introducing reference counting instead of manual `free` where that `free` is not a public API but rather an internal concern of the object itself. And the argument against this is that "objects want to live for as long as they want and they have the right to make as many side effects as they wish".

Vadim-Yelagin:
https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2203
I wonder why not discourage any use of `Signal` then? Better yet, make it a private implementation detail for `SignalProducer`.

NachoSoto:
> I wonder why not discourage any use of Signal then? Better yet, make it a private implementation detail for SignalProducer.

Why would we discourage `Signal`? both types play an equally important role, as they represent different semantics in your application.

Vadim-Yelagin:
Because of memory leaks for one. After 7 months of using Swift RAC I still can't think of any situation where I would use a signal instead of a signal producer, except for cases when it's RAC API providing a signal (e.g. in Action). But even in those cases it turns out that the best thing I can do with that signal is to wrap it in a signal producer like in the example I provided above.
And I'm still curious to see any example where using signals is beneficial, and where side-effects put in awkward places play essential role and are hard to move to proper places.
Maybe you could point me at some open-source projects using Swift RAC. I took a look at CarthageKit, and it only has signal producers, no signals.

Vadim-Yelagin:
This whole Signal/SignalProducer separation made no sense to me from the very first look at it, but I just couldn't put my finger on why is that a bad idea.
Then there was https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1980 concluded by a vague sentiment about some things that seem to be signals are in fact producers because they have a side effect of subscribing, and that we shouldn't feel bad about it.
That also inspired https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2203 which was never done, probably because nobody wants to tell these things to potential customers.
Anyway, how do you decide whether to use a signal or a signal producer? There's no strict criteria, only philosophical meditation about signal being alive and whatever not, but still, you know, some things that seem to be signals are not, or else they leak or set your house on fire.
Well, okay, probably `Signal` is a good thing to be a basis for defining operations on signals, and then lifting them to producers, so that we have separation of concerns and neat code, even if we aren't going to use signals directly.
But now there's https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2453 which proves that it is not always possible to do so, and that it leads to nothing but confusion.
Then there's README.md that states in bold that "ReactiveCocoa’s changes here are simple, not easy" regarding Signal/SignalProducer separation. That fact that those weasel words are there to feed the idea about this separation being a good thing is at least suspicious.

Vadim-Yelagin:
The original test case is included in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2603 which also fixes https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2586 (and the original test case in particular) without the need to modify the `Signal` contract (all the side effects that would currently take place will still take place).
Breaking changes that are also supposed to fix https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2585 are proposed in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2591

Issue #2584, opened 2015-12-02
"This application is modifying the autolayout engine from a background thread..."


ton03:
Hi,

I have the ff. versions:
- Swift version `2.1 (swiftlang-700.1.101.6 clang-700.1.76)`,
- Xcode version `7.1.1 (7B1005)`, and
- ReactiveCocoa `v4.0.0-alpha.4`

I understand that I'm using an alpha version of ReactiveCocoa, but I was only planning on using simple functionalities. I just wanted a stable (or stable enough) version for Swift 2.1. If anyone could suggest a version I could use that would fix this issue, it would be very awesome and much appreciated. :)

Anyway, this is the warning I'm getting, and this delays some view changes that should happen when properties in the view model changes

    2015-12-02 11:05:59.028 My_Project [291:24217] This application is modifying the autolayout engine from a background thread, which can lead to engine corruption and weird crashes.  This will cause an exception in a future release.
     Stack:(
        0   CoreFoundation                      0x000000018537cf60 <redacted> + 148
        1   libobjc.A.dylib                     0x0000000199f2ff80 objc_exception_throw + 56
        2   CoreFoundation                      0x000000018537ce90 <redacted> + 0
        3   Foundation                          0x000000018639b2d8 <redacted> + 88
        4   Foundation                          0x0000000186221a1c <redacted> + 56
        5   Foundation                          0x000000018621d5dc <redacted> + 260
        6   UIKit                               0x000000018b173158 <redacted> + 964
        7   UIKit                               0x000000018b17457c <redacted> + 788
        8   UIKit                               0x000000018b243740 <redacted> + 480
        9   UIKit                               0x000000018b240708 <redacted> + 172
        10  UIKit                               0x000000018b240978 <redacted> + 96
        11  UIKit                               0x000000018b240e3c <redacted> + 80
        12  UIKit                               0x000000018b2453d0 <redacted> + 492
        13  UIKit                               0x000000018b24453c <redacted> + 148
        14  UIKit                               0x000000018b24447c <redacted> + 284
        15  UIKit                               0x000000018ad74d40 <redacted> + 444
        16  UIKit                               0x000000018ad74e14 <redacted> + 60
        17  UIKit                               0x000000018ad74f54 <redacted> + 28
        18  UIKit                               0x000000018ad74504 <redacted> + 100
        19  UIKit                               0x000000018a910098 <redacted> + 996
[...]

Issue #2578, opened 2015-11-29
Fix flaky test


NachoSoto:
`RACSequenceExamples.m:58`: `should only evaluate one value per scheduling`

Example: https://travis-ci.org/ReactiveCocoa/ReactiveCocoa/builds/93791324

A temporary change can be disabling the test (but let's not close this issue until it's actually fixed!)

NachoSoto:
Closing this since #2592 seems to have fixed it. Will reopen if it fails again.

PR #2589, opened 2015-12-02
Add test cases and fix for #2515

Vadim-Yelagin:
SignalProducerSpec SignalProducer.buffer should not deadlock in pair when started while sending (deadlocking)

Test case for #2515.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -123,14 +123,6 @@ public struct SignalProducer<Value, Error: ErrorType> {
 	public static func buffer(capacity: Int) -> (SignalProducer, Signal<Value, Error>.Observer) {
 		precondition(capacity >= 0, "Invalid capacity: \(capacity)")
 
-		// This is effectively used as a synchronous mutex, but permitting
-		// limited recursive locking (see below).
-		//
-		// The queue is a "variable" just so we can use its address as the key
-		// and the value for dispatch_queue_set_specific().
-		var queue = dispatch_queue_create("org.reactivecocoa.ReactiveCocoa.SignalProducer.buffer", DISPATCH_QUEUE_SERIAL)
-		dispatch_queue_set_specific(queue, &queue, &queue, nil)
-
 		// Used as an atomic variable so we can remove observers without needing
 		// to run on the queue.
 		let state: Atomic<BufferState<Value, Error>> = Atomic(BufferState())
@@ -139,31 +131,27 @@ public struct SignalProducer<Value, Error: ErrorType> {
 			// Assigned to when replay() is invoked synchronously below.
 			var token: RemovalToken?
 
-			let replay = {
-				let originalState = state.modify { state in
-					var state = state
-					token = state.observers?.insert(observer)
-					return state
-				}
-
-				originalState.values.forEach(observer.sendNext)
-
-				if let terminationEvent = originalState.terminationEvent {
-					observer.action(terminationEvent)
-				}
+			// This lock is used to ensure that each observer receives events
+			// in the same order as they were buffered.
+			// This is required because another thread may push an event while
+			// the buffer is being replayed to the observer.
+			// In that case the lock prevents the observer from receiving that event
+			// before it receives all of the buffered events.
+			let replayLock = NSLock()
+			replayLock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.buffer.replay"
+			replayLock.lock()
+			let originalState = state.modify { state in
+				var state = state
+				token = state.observers?.insert((observer, replayLock))
+				return state
 			}
 
-			// Prevent other threads from sending events while we're replaying,
-			// but don't deadlock if we're replaying in response to a buffer
-			// event observed elsewhere.
-			//
-			// In other words, this permits limited signal recursion for the
-			// specific case of replaying past events.
-			if dispatch_get_specific(&queue) != nil {
-				replay()
-			} else {
-				dispatch_sync(queue, replay)
+			originalState.values.forEach(observer.sendNext)
+
+			if let terminationEvent = originalState.terminationEvent {
+				observer.action(terminationEvent)
 			}
+			replayLock.unlock()
 
 			if let token = token {
 				disposable.addDisposable {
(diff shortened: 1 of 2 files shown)


Vadim-Yelagin:
This is similar to "should not deadlock when started while sending".
It succeeds if the lock "org.reactivecocoa.ReactiveCocoa.SignalProducer.buffer" is removed.
No other test cases fail in that case, which is unfortunate and I'll try to create one that would fail because of events observed out of order.

NachoSoto:
Thanks!

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -373,6 +373,55 @@ class SignalProducerSpec: QuickSpec {
...
+				let (producer, observer) = SignalProducer<Int, NoError>.buffer()
+
+				observer.sendNext(1)
+				observer.sendNext(2)
+
+				var values: [Int] = []
+				producer.startWithNext { value in
+					values.append(value)
+					if value == 1 {
+						// the expected behavior is for the other queue to block until
```
Vadim-Yelagin: I'm not sure if this is the right way to test for concurrency effects or even if RAC supports this kind of concurrent access (I guess that was the idea behind all those locks, but I never considering such usage in my projects).

Vadim-Yelagin:
What's up with Travis? It says "RACSequenceSpec shouldn_t_overflow_the_stack_when_deallocated_on_a_background_queue_UserstravisbuildReactiveCocoaReactiveCocoaReactiveCocoaTestsObjectiveCRACSequenceSpecm_301" fails : /

NachoSoto:
https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2578

neilpa:
We should just disable that test until we have an actual fix for it.

Vadim-Yelagin:
@andersio why? afaik using dispatch_sync on a serial queue is the best practice for serializing access to a resource.

Vadim-Yelagin:
@andersio dispatch_sync shouldn't jump to a different thread, it's simply locking the serial queue. Also, the old implementation already used dispatch_sync and I there were no such problems that I know of.

andersio:
@Vadim-Yelagin It is an optimisation, but not a guarantee.
```
https://developer.apple.com/library/ios/documentation/Performance/Reference/GCD_libdispatch_Ref/#//apple_ref/c/func/dispatch_sync
As an optimization, this function invokes the block on the current thread when possible.
```

Vadim-Yelagin:
Uh oh. I found `Action` currently using `dispatch_sync` and `Atomic` using `OS_SPINLOCK_INIT`.
`SignalProducer.single` uses a `dispatch_semaphore` but that's because it really needs one.
I don't want to resort to using semaphores for serializing access : (

Vadim-Yelagin:
@andersio Probably what Apple meant by that is that if you are on a concurrent queue and you dispatch_sync to the same concurrent queue, it can execute it on a different thread instead of deadlocking. I never heard of dispatch_sync to a serial queue called on the main thread executing the block on a different thread. Probably those issues with non-main-thread errors are caused by the event pushed not from the main queue in the first place. The stack trace in #2584 is rather tarnished but it appears that that was the case there.

andersio:
I believe the reason a queue was used before was because dispatch queue specific is used to elide some cases of signal recursion. In your patch, it is simply used as a lock.

Vadim-Yelagin:
Either way, let's not try to hit two birds with one stone here. There were dispatch_sync and I used them as well, so it shouldn't change anything regarding that matter. Let's deal with the deadlock now, and investigate if there's any non-main-thread issue later.

Vadim-Yelagin:
@andersio what about this kind of lock:

	final class ObjCLock: NSObject {
		func synchronized(@noescape closure: () -> ()) {
			objc_sync_enter(self)
			closure()
			objc_sync_exit(self)
		}
	}

Vadim-Yelagin:
Okay, there's also `Signal` using `NSLock`. I'll replace `dispatch_sync` with that.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -139,33 +131,29 @@ public struct SignalProducer<Value, Error: ErrorType> {
...
-				}
+			// This lock is used to ensure that each observer receives events
+			// in the same order as they were buffered.
+			// This is required because another thread may push an event while
+			// the buffer is being replayed to the observer.
+			// In that case the lock prevents the observer from receiving that event
+			// before it receives all of the buffered events.
+			let replayLock = NSLock()
+			replayLock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.buffer.replay"
+			replayLock.lock()
```
NachoSoto: I wonder if this lock can be replaced with
```swift
synchronized(state) {
  ...
}
```

Vadim-Yelagin: State is per entire buffer. The idea of this fix is to prevent the deadlock by having an individual lock for each observer, and using that lock to serialize sending the events to that observer only.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -373,6 +373,55 @@ class SignalProducerSpec: QuickSpec {
...
+						// the first queue wakes up and receives all the buffered events
+						dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)) {
+							observer.sendNext(3)
+						}
+						usleep(1000)
```
NachoSoto: Can you explain this? We shouldn't have this sort of thing in tests, and instead of `TestScheduler`.

NachoSoto:
(BTW, I didn't list this as a known issue because it's not really an issue, but an existing limitation since `RAC 3.0`(

Vadim-Yelagin:
AFAIK, the limitation is that you cannot push new events while in observer callback, which makes sense.
That limitation is still there with the fix.
The issue here was that it deadlocked when adding a new observer while in another observer callback. It had been known and kind of fixed by having a kind of recursive locking (see "should not deadlock when started while sending"). But it didn't work in nontrivial cases because recursive locking is [never a correct solution](http://inessential.com/2013/09/24/recursive_locks).

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
Vadim-Yelagin: I need to test that another thread will wait until the original one replays all the buffer to the observer. It doesn't seem like `TestScheduler` can do this. Since it doesn't use any other threads, it will probably just deadlock.

Vadim-Yelagin: Anyway, this code here is not correct : (
I can simply remove this test unless there's a way to test this properly.

vadim-fueled:
I've removed the test case with questionable implementation and rebased the changes.

NachoSoto:
This looks good to me! :shirt:

@neilpa could you take another look?

neilpa:
Sorry I missed this before. My one concern is the potential perf fallout of all the extra lock allocations that this entails. `buffer` is the common base for a lot of signals and a lock per observer could get problematic.

I'm wondering if there's an alternative that's more inline with the original implementation that also solves this bug.

NachoSoto:
Or maybe we can add perf tests instead. What do you think @vadim-fueled?

neilpa:
I'm not sure perf tests in the traditional sense capture my concern. I'm more worried that number of locks scales with observers here. Furthermore, the lock isn't really needed beyond initial subscription and push of the set of buffered events.

vadim-fueled:
I don't know if allocating a lock is more expensive than allocating any other object or closure. On the other hand, in concurrent context there might even be a perfo benefit due to less congestion on per-buffer lock. The lock can be deallocated after the buffer replay completes, but I didn't want to prematurely complicate the implementation, would need to perfo test prior to any optimization efforts.

vadim-fueled:
I've rebased the branch and restored it after some breaking changes.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -176,31 +164,36 @@ public struct SignalProducer<Value, Error: ErrorType> {
...
-						// attachments.
-						mutableState.terminationEvent = event
-						mutableState.observers = nil
-					}
-
-					return mutableState
+			// Serialize pushing the events with a lock to prevent
+			// different observers from receiving the events in different order
+			// in case multiple threads push events at the same time.
+			pushLock.lock()
```
azeff: Previous implementation had following comment
```
// Send serially with respect to other senders, and never while
// another thread is in the process of replaying.
```

I don't think NSLock provides fair locking, so this will change original behavior. @NachoSoto @neilpa is it ok?

vadim-fueled: What's unfair about NSLock locking? Sending is still serial, but replaying threads are protected by per-observer locks, not this lock. There are comments describing it too.

vadim-fueled:
Flaky tests again?

mdiep:
I think this can be solved without an additional lock—and without the dispatch queue—by tracking the values that need to be replayed by replaying observers.

`BufferState` would add something like `[Observer: [Value]]`. Replaying would add the observer to the dictionary with an empty array and grab the initial values to replay with `state`. When a value is added to `state`, it would also append the value to each replaying observer array. After sending the initial replay values, the replay would loop, checking to see if there were more values it needed to send, removing them and sending them if there were, and removing self from the dictionary if there aren't.

But I'm not sure that's worth the effort since we plan to remove `buffer` in RAC5 (#2831). Maybe we should either (a) ship something like this PR, without worrying about performance, or (b) just ignore the issue since we plan to remove the method.

Thoughts?

mdiep:
I tried implementing my solution in #2900.

ikesyo:
Superseded by #2900.

PR #2591, opened 2015-12-02
Proposed Fix for Signal Lifetime (#2585, #2588)

andersio:
Related issue: #2585, #2588
Based on an idea from @Vadim-Yelagin (https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2585#issuecomment-161353878).

## Intro
Sticking to the Signal contract, #2585 is in fact only a matter of operator ordering - applying `takeUntil` as soon as possible would mitigate the issue. But I still rolled this patch to propose a special handling in unretained signals.

**TL;DR:** Goal of the PR: Keeping unretained signals (in the user space) without any observer alive requires explicit declaration in advance, or else the signals should be terminated ASAP.

## Breaking changes to the `Signal` contract
*See also: [the updated Signal contract](https://github.com/andersio/ReactiveCocoa/blob/signal-lifetime/Documentation/DesignGuidelines.md)*

The observer block in `Signal.init:`, and the ActionDisposable returned by `Signal.observe:` no longer retains the `Signal` instance. They now retains only the generator disposable and the atomic bag of observers.

In other words, `Signal` is now just a proxy for users to attach observers. The event stream of a signal can outlive the respective `Signal` instance, and should still end only upon a terminating event.

On the other hand, this means we can now know if the user no longer retains the `Signal` in the user space. This allows the assumptions below to be made.

## Assumptions
1. When `Signal` has been deallocated:
It means the signal will have no new observer in the future.
So if the signal has no observer at the same moment, the signal can be safely terminated even without a terminating event.

2. When an observer detaches itself from a signal:
If the resulting number of observers is zero, and if the respective `Signal` instance was deallocated already, the signal can be safely terminated without a terminating event.

**TL;DR:** A signal must either have its respective `Signal` instance being retained, or have at least one observer to be kept alive.

**TL;DR 2:** If a signal with side effects is to be kept alive but expected not to be observed, the user must declare the willingness by retaining the signal. Otherwise, the signal would be terminated right away.

## Pros
Consumers needn't care about how the ordering of operators might affect the signal lifetime, and can expect similar semantics to ARC in the language.

e.g. The following pairs have the same effect between themselves, assuming `infiniteSignal` never sends a terminating event:
```
infiniteSignal.map(transform1).takeUntil(willDeinitSignal) // A
infiniteSignal.takeUntil(willDeinitSignal).map(transform1) // B

infiniteSignal
    .map(transform1).map(transform2).map(transform3)
    .takeUntil(willDeinitSignal) // A
infiniteSignal
    .takeUntil(willDeinitSignal)
    .map(transform1).map(transform2).map(transform3) // B
```
Currently, for cases that has `takeUntil` later in the chain, the `transform*` blocks would still be continuously evaluated upon new values from upstream, even if `willDeinitSignal` is triggered. In other words, these intermediate signals - at least in the case of set B being expected results - are now unexpected leaks.

In the PR, though, as these single-observer intermediate signals were not declared to be keep alive forever, they all would be terminated as soon as the disposal caused by `takeUntil` propagates up or down.

## Cons
It breaks existing code that uses unretained, transformed signals with side effects, but without observing it.

The signals would have to be either: (1) retained; or (2) converted to an observer to the signal. Either way, a new convenience operator `keepAlive` can be introduced as the middle ground. 

## Results
Existing specs all pass on Xcode 7.2 beta 4, except for those test cases which test against "deallocated signals after disposal" - ~~they are removed in this PR~~. The PR also includes a test case from #2588 and an additional one that replicates #2585.

Test cases originally testing for the deallocation of `Signal` upon disposal has been updated to match the new `Signal` lifetime model.

## Note
The implementation relies on the generator disposable to be able to detach itself from its upstream signal, *without* an terminating event received.


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -26,17 +28,25 @@ public final class Signal<Value, Error: ErrorType> {
...
 		sendLock.name = "org.reactivecocoa.ReactiveCocoa.Signal"
 
-		let generatorDisposable = SerialDisposable()
+		generatorDisposable = SerialDisposable()
 
 		/// When set to `true`, the Signal should interrupt as soon as possible.
 		let interrupted = Atomic(false)
 
-		let observer = Observer { event in
+		let observer = Observer { [atomicObservers = self.atomicObservers, generatorDisposable = self.generatorDisposable] event in
```
NachoSoto: Good idea!

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -126,11 +127,14 @@ public final class Signal<Value, Error: ErrorType> {
...
 		if let token = token {
-			return ActionDisposable {
-				self.atomicObservers.modify { observers in
+			return ActionDisposable { [weak self, generatorDisposable = self.generatorDisposable, atomicObservers = self.atomicObservers] in
+				atomicObservers.modify { observers in
 					guard var observers = observers else { return nil }
 
 					observers.removeValueForToken(token)
+					if self == nil && observers.count == 0 {
+						generatorDisposable.dispose()
```
NachoSoto: I think this makes sense

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -139,6 +143,12 @@ public final class Signal<Value, Error: ErrorType> {
 			return nil
 		}
 	}
+	
+	deinit {
+		if atomicObservers.value?.count ?? 0 == 0 {
```
NachoSoto: I'm confused about this. Can you explain why you'd only dispose if there are no observers?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -36,71 +36,6 @@ class SignalSpec: QuickSpec {
...
-				expect(errored).to(beFalsy())
-				expect(signal).toNot(beNil())
-				
-				testScheduler.run()
-				
-				expect(errored).to(beTruthy())
-				expect(signal).to(beNil())
-			}
-
-			it("should deallocate after completing") {
```
NachoSoto: These tests should still be valid, right?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -218,48 +153,57 @@ class SignalSpec: QuickSpec {
...
+
+					signal.observeFailed { _ in error = true }
 				}
 				test()
-				
-				expect(weakSignal).toNot(beNil())
-				
-				testScheduler.run()
+
+				expect(error).to(beFalsy())
```
NachoSoto: You should do `expect(error) == false`

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -218,48 +153,57 @@ class SignalSpec: QuickSpec {
...
 					}
+
+					signal.observeCompleted { completed = true }
 				}
 				test()
 				
-				expect(weakSignal).toNot(beNil())
-
-				testScheduler.run()
+				expect(completed).to(beFalsy())
```
NachoSoto: This is redundant anyway, it should be covered by another test. What this test should cover is the deallocation, which it no longer does?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,9 +400,54 @@ class SignalSpec: QuickSpec {
...
+      it("should not keep resulting signal alive indefinitely") {
+        var counter = 0
+        var lastValue: String?
+
+        let (signal, observer) = Signal<Int, NoError>.pipe()
+        
+        let foo = {
+          let mappedSignal = signal.map { integer -> String in
+            counter++
+            return String(integer + 1)
```
NachoSoto: You can simplify this by just returning `integer + 1` instead of a `String`.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,9 +400,54 @@ class SignalSpec: QuickSpec {
...
 				expect(lastValue).to(equal("2"))
 			}
+
+			it("should not keep resulting signal alive indefinitely") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal.never.map { $0 }
+
+				expect(signal).to(beNil())
+      }
+
+      it("should not keep resulting signal alive indefinitely") {
```
NachoSoto: This isn't testing the lifetime of the signal, only the subscription.

NachoSoto:
I think the changes in the implementation make sense, but I'm worried about the changes in the tests: you've removed critical ones, and modified others removing critical assertions. If those were failing I'm afraid this would be a regression.

Also, could you include the tests in #2588 to check if this actually fixes the issue at hand?

ReactiveCocoa/Swift/Signal.swift
andersio: Having observers means the event stream isn't really ended. `Signal` here is just an interface to attach observers. Say:
```
func foo() {
    someExternalSignal
        .map { String("value: \($0)") }
        .observeNext { print($0) }
}

foo()
```
Upon exiting `foo()`, the "shell" of the mapped signal would have been deallocated. But the event stream should still go on until a terminating event hits, or all the observers are detached from the mapped signal, i.e. an empty bag of observers.

ReactiveCocoaTests/Swift/SignalSpec.swift
andersio: They aren't, since the `Signal` would deallocate right away. `expect(signal).toNot(beNil())` would always fail.

andersio:
@NachoSoto One of the two test cases in #2588 is included, i.e. the `Signal.never` one. The other one is removed, since like a few test cases removed in this PR, it tests based on the previous `Signal` behaviour, which would deallocate only when the event stream is terminated.

andersio:
Could use a little help on the test cases though. I am not particularly good at writing these. :D

ReactiveCocoaTests/Swift/SignalSpec.swift
vadim-fueled: Could we grab the signal's "internals" and check if they are nil?

andersio: @vadim-fueled Good idea. It works. Glad that Xcode 7 introduced `@testable` which grants access to `internal` stuff. :)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -126,11 +127,14 @@ public final class Signal<Value, Error: ErrorType> {
 		}
 
 		if let token = token {
-			return ActionDisposable {
-				self.atomicObservers.modify { observers in
+			return ActionDisposable { [weak self, generatorDisposable = self.generatorDisposable, atomicObservers = self.atomicObservers] in
```
siemensikkema: This should be the same as:
```swift
return ActionDisposable { [weak self, generatorDisposable, atomicObservers] in
```

guidomb:
Could somebody explain me why this PR has been closed? I was disusing the same issue this PR is addressing in #2949. Personally I think the proposed change to the Signal life cycle in this PR should be merged.

mdiep:
> Could somebody explain me why this PR has been closed?

Because the author closed it. 😄

PR #2592, opened 2015-12-03
Attempt to fix flakey test in travis

neilpa:
Maybe using a background queue instead of the main thread scheduler will fix this on travis since the errors always look like

```
failed - expected to eventually match (Stall on main thread)., got <1.0000>
```

cf. #2578


NachoSoto:
Closed to rerun them. Let's see if they pass again!

NachoSoto:
Looks like this does the trick! :tada:

PR #2594, opened 2015-12-03
Make Xcode configuration dependency public

tonyarnold:
The Mac framework is failing to build when including ReactiveCocoa as an Xcode subproject due to missing xcconfigs. I believe the xcconfigs dependency is required to build, therefore I've made it public in this PR.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,2 @@
 github "antitypical/Result" ~> 1.0.0
+github "jspahrsummers/xcconfigs" "ec5753493605deed7358dec5f9260f503d3ed650"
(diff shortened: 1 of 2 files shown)


NachoSoto:
This should stay in `Cartfile.private`. The primary reason is because one may have multiple dependencies that privately depend on `xcconfigs`, and they don't necessarily need to be compatible, since they're not dynamic frameworks.

This shouldn't be necessary because [this is a submodule](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/.gitmodules#L7-L9), so it should be in `Carthage/Checkouts`. Is that not the case?

tonyarnold:
You're absolutely right — I had a corrupted checkout. Ignore this PR :wink:

NachoSoto:
Glad you figured it out :D

PR #2597, opened 2015-12-04
Remove var binding in functions and guards

natestedman:
`guard var` and `{ (var x) in ... }` are being [removed from Swift](https://github.com/apple/swift-evolution/blob/master/proposals/0003-remove-var-parameters-patterns.md), and cause compiler errors with the Swift 2.2 preview available from [swift.org](http://swift.org).

I've replaced all usages with the ["purely mechanical migration"](https://github.com/apple/swift-evolution/blob/master/proposals/0003-remove-var-parameters-patterns.md#impact-on-existing-code) suggestion. It's not the best approach ever, especially stuff like:

```swift
guard let immutableBag = bag else { return nil }
var mutableBag = immutableBag
```

As a replacement for:

```swift
guard var bag = bag else { return nil }
```

So this should probably be reworked into something better. These are in calls to `.modify` - I briefly tried switching to `inout` there, but ran into many test failures and crashes, and my main goal was to make `swift build` work.

This is required for supporting the [Swift Package Manager](https://github.com/apple/swift-package-manager): https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2599.

### DIFF
--- a/ReactiveCocoa/Swift/Disposable.swift
+++ b/ReactiveCocoa/Swift/Disposable.swift
@@ -198,9 +198,10 @@ public final class SerialDisposable: Disposable {
 		}
 
 		set(d) {
-			let oldState = state.modify { (var state) in
-				state.innerDisposable = d
-				return state
+			let oldState = state.modify { state in
+				var mutableState = state
+				mutableState.innerDisposable = d
+				return mutableState
 			}
 
 			oldState.innerDisposable?.dispose()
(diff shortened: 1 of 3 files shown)


kastiglione:
> `{ (var x) in ... }`

Is this being removed? The proposal makes no mentions of "closure" and doesn't show any code examples of `var` being used in closure variables? I'm mainly just curious.

natestedman:
They don't refer to closures specifically in the proposal, just to functions in general. I think that means *all* functions, which would include closures - in the Swift preview download, using any kind of var binding causes compiler errors.

kastiglione:
> in the Swift preview download, using any kind of var binding causes compiler errors

ah thanks, that's what I was wondering.

NachoSoto:
Awesome, thank you for the PR! I'm glad that we're doing this in preparation for 2.2 :)

PR #2598, opened 2015-12-04
Add Foundation imports to files that use Foundation

natestedman:
ReactiveCocoa inherits the Foundation types from the Objective-C headers, I think. However, if those aren't included when building, those files will generate compiler errors for the missing types.

This is required to support the [Swift Package Manager](https://github.com/apple/swift-package-manager): https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2599.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 /// Represents an action that will do some work when executed with a value of
 /// type `Input`, then return zero or more values of type `Output` and/or fail
 /// with an error of type `Error`. If no failure should be possible, NoError can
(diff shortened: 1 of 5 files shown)


NachoSoto:
This shouldn't be necessary because `CLANG_MODULES_AUTOLINK` is set to `YES`.

neilpa:
That setting only comes into play once you've already `import`ed the framework though which isn't the case here. IIRC, this problem stems from Xcode implicitly linking Foundation whenever it builds even if it's not imported. That doesn't happen when you build directly with `swiftc` which I assume the new package manager is doing.

NachoSoto:
> That setting only comes into play once you've already imported the framework which isn't the case here. IIRC, this problem stems from Xcode implicitly linking Foundation whenever it builds even if it's not imported. That doesn't happen when you build directly with swiftc which I assume the new package manager is doing.

Oooh of course, what am I thinking.

This is good then :+1:

PR #2599, opened 2015-12-04
[WIP] Add support for Swift Package Manager

natestedman:
This is super not ready to merge, but I have ReactiveCocoa working with the [Swift Package Manager](https://github.com/apple/swift-package-manager). I've tried to keep all of the changes as atomic as possible, so there are two dependent pull requests:

- Removing `var` binding from `guard` and closure parameters: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2597
- Adding `import Foundation` where types from Foundation are used: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2598

Here's the caveats:

- `swift build` appears to assume that any directory containing `.swift` files is a package to be built, _unless_ the directory is named `Tests`,  _or_ a `Sources` directory is included - in which case, that directory is used as a root, and the `Tests` rule still applies there.

  Therefore, the `ReactiveCocoaTests` directory had to be renamed to `Tests`. There's a [pull request](https://github.com/apple/swift-package-manager/pull/16) open on the package manager to amend this behavior to ignore `*Tests`.

  Additionally, the `Carthage` directory was considered a package. `ReactiveCocoa` and the newly renamed `Tests` have been moved within a `Sources` directory to fix this.
- Currently, `swift build` doesn't support Objective-C. Therefore, this is *only* the Swift parts of ReactiveCocoa. To "ignore" the bridging functions, I moved them to the `Tests` directory. It's... not great.

   This also includes `DynamicProperty`, since it uses an Objective-C superclass - it has been moved to its own file, which is also located in `Tests`.

   The [Swift Package Manager documentation](https://github.com/apple/swift-package-manager#modules) states that:
  > As a rule of thumb: more modules is probably better than less modules. The package manager is designed to make creating both packages and apps with multiple modules as easy as possible.

  So IMO, it would make sense for the Objective-C components, the Swift components, and the bridging components to be independent modules. That way, people writing a pure-Swift app won't need to include the Objective-C components. I think that `DynamicProperty` would stay with the Swift components instead of the bridging components, since the Objective-C superclass is an implementation detail.
- This depends on https://github.com/antitypical/Result/pull/126 being merged and a new version of `Result` being released, since the `Package.swift` currently refers to my fork.

That's it, I think! I made a little "Hello World" project to try it out: https://github.com/natestedman/ReactiveCocoa-Swift-Package-Manager


NachoSoto:
Awesome! This is so excited :D

I'm sad that we're forced to move everything around. I hope future versions of the package manager allows us to specify the paths.

natestedman:
That is indeed planned: https://github.com/apple/swift-package-manager/pull/16#issuecomment-162084875

neilpa:
> I'm sad that we're forced to move everything around. I hope future versions of the package manager allows us to specify the paths.

Personally I prefer the `Sources` and `Tests` names since they're already scoped to a `ReactiveCocoa` directory. The prefixes just make require extra typing and auto-complete tabbing when navigating at the command line.

mdiep:
I'm going to close this for now. I think this is something we can pursue when we get the repos separated (#2807).

PR #2603, opened 2015-12-05
Fix leak of signal with deallocated generator observer

Vadim-Yelagin:
Fix leak of signal with deallocated generator observer by capturing self weakly in disposable from `Signal.observe`.
Fixes https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2586 , test cases included, all pass.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -128,8 +128,8 @@ public final class Signal<Value, Error: ErrorType> {
 		}
 
 		if let token = token {
-			return ActionDisposable {
-				self.atomicObservers.modify { observers in
+			return ActionDisposable { [weak self] in
+				self?.atomicObservers.modify { observers in
 					guard let immutableObservers = observers else { return nil }
 					var mutableObservers = immutableObservers
 
--- a/ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
@@ -1,0 +1,214 @@
+//
+//  SignalLifetimeSpec.swift
+//  ReactiveCocoa
+//
+//  Created by Vadim Yelagin on 2015-12-13.
+//  Copyright (c) 2015 GitHub. All rights reserved.
+//
+
+import Result
+import Nimble
+import Quick
+import ReactiveCocoa
+
+class SignalLifetimeSpec: QuickSpec {
+	override func spec() {
+		describe("init") {
+			var testScheduler: TestScheduler!
+
+			beforeEach {
+				testScheduler = TestScheduler()
+			}
+
+			it("should deallocate") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal { _ in nil }
+
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate even if it has an observer") {
+				weak var signal: Signal<AnyObject, NoError>? = {
+					let signal: Signal<AnyObject, NoError> = Signal { _ in nil }
+					return signal
+				}()
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate even if it has an observer with retained disposable") {
+				var disposable: Disposable? = nil
+				weak var signal: Signal<AnyObject, NoError>? = {
+					let signal: Signal<AnyObject, NoError> = Signal { _ in nil }
+					disposable = signal.observe(Observer())
+					return signal
+				}()
+				expect(signal).to(beNil())
+				disposable?.dispose()
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate after erroring") {
+				weak var signal: Signal<AnyObject, TestError>? = Signal { observer in
+					testScheduler.schedule {
+						observer.sendFailed(TestError.Default)
+					}
+					return nil
+				}
+
+				var errored = false
+
+				signal?.observeFailed { _ in errored = true }
+
+				expect(errored).to(beFalsy())
+				expect(signal).toNot(beNil())
+
+				testScheduler.run()
+
+				expect(errored).to(beTruthy())
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate after completing") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal { observer in
+					testScheduler.schedule {
+						observer.sendCompleted()
+					}
+					return nil
+				}
+
+				var completed = false
+
+				signal?.observeCompleted { completed = true }
+
+				expect(completed).to(beFalsy())
+				expect(signal).toNot(beNil())
+
+				testScheduler.run()
+
+				expect(completed).to(beTruthy())
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate after interrupting") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal { observer in
+					testScheduler.schedule {
+						observer.sendInterrupted()
+					}
+
+					return nil
+				}
+
+				var interrupted = false
+				signal?.observeInterrupted {
+					interrupted = true
+				}
+
+				expect(interrupted).to(beFalsy())
+				expect(signal).toNot(beNil())
+
+				testScheduler.run()
+
+				expect(interrupted).to(beTruthy())
+				expect(signal).to(beNil())
+			}
+		}
+
+		describe("Signal.pipe") {
+			it("should deallocate") {
+				weak var signal = Signal<(), NoError>.pipe().0
+
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate after erroring") {
+				let testScheduler = TestScheduler()
+				weak var weakSignal: Signal<(), TestError>?
+
+				// Use an inner closure to help ARC deallocate things as we
+				// expect.
+				let test: () -> () = {
+					let (signal, observer) = Signal<(), TestError>.pipe()
+					weakSignal = signal
+					testScheduler.schedule {
+						observer.sendFailed(TestError.Default)
+					}
+				}
+				test()
+
+				expect(weakSignal).toNot(beNil())
+
+				testScheduler.run()
+				expect(weakSignal).to(beNil())
+			}
+
+			it("should deallocate after completing") {
+				let testScheduler = TestScheduler()
+				weak var weakSignal: Signal<(), TestError>?
+
+				// Use an inner closure to help ARC deallocate things as we
+				// expect.
+				let test: () -> () = {
+					let (signal, observer) = Signal<(), TestError>.pipe()
+					weakSignal = signal
+					testScheduler.schedule {
+						observer.sendCompleted()
+					}
+				}
+				test()
+
+				expect(weakSignal).toNot(beNil())
+
+				testScheduler.run()
+				expect(weakSignal).to(beNil())
+			}
+
+			it("should deallocate after interrupting") {
+				let testScheduler = TestScheduler()
+				weak var weakSignal: Signal<(), NoError>?
+
+				let test: () -> () = {
+					let (signal, observer) = Signal<(), NoError>.pipe()
+					weakSignal = signal
+
+					testScheduler.schedule {
+						observer.sendInterrupted()
+					}
+				}
+
+				test()
+				expect(weakSignal).toNot(beNil())
+
+				testScheduler.run()
+				expect(weakSignal).to(beNil())
+			}
+		}
+
+		describe("map") {
+			it("should deallocate") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal { _ in nil }.map { $0 }
+
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate even if it has an observer") {
+				weak var signal: Signal<AnyObject, NoError>? = {
+					let signal: Signal<AnyObject, NoError> = Signal { _ in nil }.map { $0 }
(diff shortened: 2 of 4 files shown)


ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +473,33 @@ class SignalSpec: QuickSpec {
...
 				expect(lastValue).to(equal("2"))
 			}
+
+			it("should not keep resulting signal alive indefinitely") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal.never.map { $0 }
+
+				expect(signal).to(beNil())
+			}
+
+			it("should not keep resulting signal alive indefinitely even if it has an observer") {
```
NachoSoto: Sorry but this changes [the semantics of `Signal`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/23dab6c3e69a64fb4d027e29ff90ac6896a7bac8/ReactiveCocoa/Swift/Signal.swift#L15-L16).`

Vadim-Yelagin: @NachoSoto What that means is that the signal is kept alive by its generator observer. If it has no generator observer, then it cannot terminated anymore because there is no way to have new events. In that case there are two possibilities: it either leaks or can deallocate. This PR does exactly the latter. That comment doesn't state anything about the signal being observed or that a disposable from an observation is being retained somewhere (this is what kept it from deallocation). In fact, this test case succeeds even without this PR because it does not retain the disposable. It's the next case that failed.
Here are some relevant lines from the Signal contract (https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/Documentation/DesignGuidelines.md#the-signalproducer-contract):
> The work associated with a Signal does not start or stop when observers are added or removed

> A signal is retained until the underlying observer is released

(underlying observer is generator observer, not observers of the signal)

> Note that it is possible to release a signal before a terminating event has been sent upon it. This should usually be avoided, as it can result in resource leaks, but is sometimes useful to disable termination.

Vadim-Yelagin: > In fact, this test case succeeds

Sorry, that's "should not keep signal alive indefinitely even if it has an observer" that passes, the one without map (which should not make any difference). Anyway, if signal is to be alive until it terminates (even if can never terminate) then "should not keep signal alive indefinitely" (that had been there for a long time) should also fail. The only way for it not to fail is to have a retain cycle inside `Signal.never` itself, which is ridiculous. Btw deallocation of a signal here does not affect (stop) any underlying work (in case it's still being performed, but lost its reference to the generator observer and is not going to send events anymore) because it does not call the generator disposable (which could stop the work) as it would if signal terminated.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +473,33 @@ class SignalSpec: QuickSpec {
...
+				weak var signal: Signal<AnyObject, NoError>? = {
+					let signal: Signal<AnyObject, NoError> = Signal.never.map { $0 }
+					signal.observe(Observer())
+					return signal
+				}()
+				expect(signal).to(beNil())
+			}
+
+			it("should not keep resulting signal alive indefinitely even if it has an observer with retained disposable") {
```
NachoSoto: This test is repeated, right?

NachoSoto:
@neilpa does this look good to you guys?

ReactiveCocoaTests/Swift/SignalSpec.swift
Vadim-Yelagin: This test is different because unlike the previous one it has a disposable from `observe` retained.
Unlike such tests without `map` they both failed before the fix though, because `map` already retains the disposable.
This is how tests were prior to the fix:
init:
✅should not keep signal alive indefinitely
✅should not keep signal alive indefinitely even if it has an observer
❌should not keep signal alive indefinitely even if it has an observer with retained disposable
map:
❌should not keep resulting signal alive indefinitely
❌should not keep resulting signal alive indefinitely even if it has an observer
❌should not keep resulting signal alive indefinitely even if it has an observer with retained disposable
I've added all variations for `map` for completeness, but if you think it's excessive, I can leave only one.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -36,6 +36,26 @@ class SignalSpec: QuickSpec {
 				expect(signal).to(beNil())
 			}
 
+			it("should not keep signal alive indefinitely even if it has an observer") {
+				weak var signal: Signal<AnyObject, NoError>? = {
+					let signal: Signal<AnyObject, NoError> = Signal.never
```
neilpa: Can we make this explicit (e.g. `Signal.init { _ in }`) and tweak the test description to "should deallocate the signal if it never sends events"

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -36,6 +36,26 @@ class SignalSpec: QuickSpec {
...
+					return signal
+				}()
+				expect(signal).to(beNil())
+			}
+
+			it("should not keep signal alive indefinitely even if it has an observer with retained disposable") {
```
neilpa: Can we rephrase this as "should deallocate the signal ..." instead of "should not keep the signal alive indefinitely ...". Same for the rest of the tests below.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -36,6 +36,26 @@ class SignalSpec: QuickSpec {
...
+				var disposable: Disposable? = nil
+				weak var signal: Signal<AnyObject, NoError>? = {
+					let signal: Signal<AnyObject, NoError> = Signal.never
```
neilpa: Can we be explicit here with `Signal.init { _ in }` and in the other tests.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -453,6 +473,33 @@ class SignalSpec: QuickSpec {
 				observer.sendNext(1)
 				expect(lastValue).to(equal("2"))
 			}
+
+			it("should not keep resulting signal alive indefinitely") {
+				weak var signal: Signal<AnyObject, NoError>? = Signal.never.map { $0 }
```
neilpa: Can we split these out into their own set of tests that don't rely on map and instead have a custom transformer

neilpa: Basically, either these should exist for every operator (which is tedious and error prone) or we just test it once with a custom test operator

ReactiveCocoaTests/Swift/SignalSpec.swift
neilpa: This set seems reasonable. Although instead of using `map` as I mention above these can be organized as a set of deallocation tests both with and w/out a transform.

neilpa:
Some minor formatting and organizational comments, but overall this seems like a good fix to me.

Additionally, we should also just remove `Signal.never`. I see no reason to have that as part of the API.

NachoSoto:
I'll take another look at this after the changes @neilpa suggested.

> Additionally, we should also just remove Signal.never. I see no reason to have that as part of the API.

:+1:

Vadim-Yelagin:
@neilpa do you want these tests in a separate source file (e.g. SignalLifetimeSpec.swift), in a separate `describe("??")` in SignalSpec.swift, or something else?

Vadim-Yelagin:
@NachoSoto @neilpa I've made the changes. Not sure it's exactly what you meant though : )

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -128,8 +128,8 @@ public final class Signal<Value, Error: ErrorType> {
 		}
 
 		if let token = token {
-			return ActionDisposable {
-				self.atomicObservers.modify { observers in
+			return ActionDisposable { [weak self] in
```
NachoSoto: Curious: do tests pass with this as `unowned`? I'd rather do that since it defines the relationship more correctly.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -0,0 +1,224 @@
...
+//  Created by Vadim Yelagin on 2015-12-13.
+//  Copyright (c) 2015 GitHub. All rights reserved.
+//
+
+import Result
+import Nimble
+import Quick
+import ReactiveCocoa
+
+class SignalLifetimeSpec: QuickSpec {
```
NachoSoto: I'm :+1: for separating these. Good idea.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -0,0 +1,224 @@
...
+					let signal: Signal<AnyObject, NoError> = Signal { _ in nil }
+					disposable = signal.observe(Observer())
+					return signal
+				}()
+				expect(signal).to(beNil())
+				disposable?.dispose()
+				expect(signal).to(beNil())
+			}
+
+			it("should deallocate after erroring") {
```
NachoSoto: Cool

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -0,0 +1,224 @@
...
+				var completed = false
+
+				signal?.observeCompleted { completed = true }
+
+				expect(completed).to(beFalsy())
+				expect(signal).toNot(beNil())
+
+				testScheduler.run()
+
+				expect(completed).to(beTruthy())
```
NachoSoto: Can you change all these to use the new syntax: `expect(completed) == true`, etc.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -0,0 +1,224 @@
...
+//
+
+import Result
+import Nimble
+import Quick
+import ReactiveCocoa
+
+class SignalLifetimeSpec: QuickSpec {
+	override func spec() {
+		describe("init") {
```
NachoSoto: Maybe change this to `Signal.init`?

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -0,0 +1,224 @@
...
+				expect(signal).to(beNil())
+				disposable?.dispose()
+				expect(signal).to(beNil())
+			}
+		}
+	}
+}
+
+private extension SignalType {
+
```
NachoSoto: Nit: unnecessary space

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -0,0 +1,224 @@
...
+	func testTransform() -> Signal<Value, Error> {
+		return Signal { observer in
+			self.observe(observer.action)
+		}
+	}
+
```
NachoSoto: Nit: unnecessary space

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -0,0 +1,224 @@
...
+			}
+		}
+	}
+}
+
+private extension SignalType {
+
+	func testTransform() -> Signal<Value, Error> {
+		return Signal { observer in
+			self.observe(observer.action)
```
NachoSoto: this should be `return self.observe(observer.action)`

NachoSoto:
Looks good! Just a few final comments on the tests.

ReactiveCocoa/Swift/Signal.swift
Vadim-Yelagin: No, something in ActionSpec.swift crashes. Apparently, disposable can be disposed after the signal is deallocated.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
Vadim-Yelagin: I don't know. There's `describe("init")` in SignalSpec and SignalProducerSpec too.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
Vadim-Yelagin: I've moved them from their original place in SignalSpec.swift as well.

NachoSoto: Okay :)

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
Vadim-Yelagin: There's over 700 of `.to(beTruthy())` and `.to(beFalsy())` in the project. Since I only moved these cases, not written them, I think it would make more sense to either leave this as is, or replace all of those enties in a separate PR.

NachoSoto: I was thinking we can start changing them as we go over the tests, like here. But I understand your reasoning, feel free to leave as is!

neilpa:
Looks good to me too

NachoSoto:
:tada:! Thanks a lot @Vadim-Yelagin!

Issue #2615, opened 2015-12-14
[Feature Request] Atomic.modify for MutableProperty?


andersio:
In some circumstances, one might modify a MutableProperty based on its previous value, e.g. a counter. While so far in my own tests I haven't faced any out of sync situation, it still has a chance to go out of track on paper, since the getter and setter of `MutableProperty.value` is not protected by the same locking attempt. I wonder if a similar offering to `Atomic.modify` could be introduced to `MutableProperty`.

e.g.
```
let oldValue = property.modify { $0 + 1 }
```

NachoSoto:
I think that would be valuable! :)

PR #2617, opened 2015-12-15
Introduce Atomic's `modify`, `swap` and `withValue` to MutableProperty.

andersio:
Related Issue: #2615 

In a nutshell, this PR:

1. introduced `modify(action:)`, `swap(newValue:)` and `withValue(action:)` to `MutableProperty`;
2. new test cases for these new methods;
3. a stylish refactor of `value.getter`;
4. makes `value.setter` use the new `modify(action:)` method; and
5. removed an invalid test case from AtomicSpec (the related overload was removed in https://github.com/ReactiveCocoa/ReactiveCocoa/commit/d83b219cd61d27b9a0e8402ac864de42e85bf657 as mentioned by @ikesyo).

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -90,16 +90,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public var value: Value {
 		get {
 			lock.lock()
-			let value = _value
-			lock.unlock()
-			return value
+			defer { lock.unlock() }
+
+			return _value
 		}
 
 		set {
-			lock.lock()
-			_value = newValue
-			observer.sendNext(newValue)
-			lock.unlock()
+			modify { _ in newValue }
 		}
 	}
 
@@ -116,6 +113,37 @@ public final class MutableProperty<Value>: MutablePropertyType {
 
 		_value = initialValue
 		observer.sendNext(initialValue)
+	}
+
+	/// Atomically replaces the contents of the variable.
+	///
+	/// Returns the old value.
+	public func swap(newValue: Value) -> Value {
+		return modify { _ in newValue }
+	}
+
+	/// Atomically modifies the variable.
+	///
+	/// Returns the old value.
+	public func modify(@noescape action: (Value) throws -> Value) rethrows -> Value {
+		lock.lock()
+		defer { lock.unlock() }
+
+		let oldValue = _value
+		_value = try action(_value)
+		observer.sendNext(_value)
+		return oldValue
+	}
+
+	/// Atomically performs an arbitrary action using the current value of the
+	/// variable.
+	///
+	/// Returns the result of the action.
+	public func withValue<Result>(@noescape action: (Value) throws -> Result) rethrows -> Result {
+		lock.lock()
+		defer { lock.unlock() }
+
+		return try action(_value)
 	}
 
 	deinit {
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -89,10 +89,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// created from the `values` producer.
 	public var value: Value {
 		get {
-			lock.lock()
-			defer { lock.unlock() }
-
-			return _value
+			return withValue { $0 }
 		}
 
 		set {
--- a/ReactiveCocoaTests/Swift/AtomicSpec.swift
+++ b/ReactiveCocoaTests/Swift/AtomicSpec.swift
@@ -35,13 +35,6 @@ class AtomicSpec: QuickSpec {
 			expect(atomic.value) == 2
 		}
 
-		it("should modify the value and return some data") {
-			let (orig, data) = atomic.modify { ($0 + 1, "foobar") }
-			expect(orig) == 1
-			expect(data) == "foobar"
-			expect(atomic.value) == 2
-		}
-
 		it("should perform an action with the value") {
 			let result: Bool = atomic.withValue { $0 == 1 }
 			expect(result) == true
--- a/ReactiveCocoaTests/Swift/PropertySpec.swift
+++ b/ReactiveCocoaTests/Swift/PropertySpec.swift
@@ -79,6 +79,59 @@ class PropertySpec: QuickSpec {
 				expect(signalCompleted) == true
 			}
 
+			it("should modify the value atomically") {
+				let property = MutableProperty(initialPropertyValue)
+
+				expect(property.modify({ _ in subsequentPropertyValue })) == initialPropertyValue
+				expect(property.value) == subsequentPropertyValue
+			}
+
+			it("should modify the value atomically and subsquently send out a Next event with the new value") {
+				let property = MutableProperty(initialPropertyValue)
+				var value: String?
+
+				property.producer.startWithNext {
+					value = $0
+				}
+
+				expect(value) == initialPropertyValue
+				expect(property.modify({ _ in subsequentPropertyValue })) == initialPropertyValue
+
+				expect(property.value) == subsequentPropertyValue
+				expect(value) == subsequentPropertyValue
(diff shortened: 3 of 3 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -90,9 +90,9 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public var value: Value {
 		get {
 			lock.lock()
-			let value = _value
-			lock.unlock()
-			return value
+			defer { lock.unlock() }
```
NachoSoto: +1

ReactiveCocoa/Swift/Property.swift
```diff
@@ -118,6 +118,37 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+	}
+
+	/// Atomically modifies the variable.
+	///
+	/// Returns the old value.
+	public func modify(@noescape action: (Value) throws -> Value) rethrows -> Value {
+		lock.lock()
+		defer { lock.unlock() }
+
+		let oldValue = _value
```
NachoSoto: This is all repeated in `value.set`. Can you change that to use this now? :)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -118,6 +118,37 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+		_value = try action(_value)
+		observer.sendNext(_value)
+		return oldValue
+	}
+
+	/// Atomically performs an arbitrary action using the current value of the
+	/// variable.
+	///
+	/// Returns the result of the action.
+	public func withValue<Result>(@noescape action: (Value) throws -> Result) rethrows -> Result {
```
NachoSoto: Awesome

ReactiveCocoaTests/Swift/AtomicSpec.swift
```diff
@@ -35,13 +35,6 @@ class AtomicSpec: QuickSpec {
 			expect(atomic.value).to(equal(2))
 		}
 
-		it("should modify the value and return some data") {
```
NachoSoto: Can you explain why you removed this? I think this is still relevant.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -79,6 +79,42 @@ class PropertySpec: QuickSpec {
 				expect(signalCompleted).to(beTruthy())
 			}
 
+			it("should swap the value atomically") {
+				let property = MutableProperty(initialPropertyValue)
+
+				expect(property.swap(subsequentPropertyValue)).to(equal(initialPropertyValue))
```
NachoSoto: This can be `expect(property.swap(subsequentePropertyValue) == initialPropertyValue`

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -79,6 +79,42 @@ class PropertySpec: QuickSpec {
...
+				expect(property.value).to(equal(subsequentPropertyValue))
```
NachoSoto: DItto

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -79,6 +79,42 @@ class PropertySpec: QuickSpec {
...
+			}
+
+			it("should modify the value atomically") {
+				let property = MutableProperty(initialPropertyValue)
+
+				expect(property.modify({ _ in subsequentPropertyValue })).to(equal(initialPropertyValue))
```
NachoSoto: DItto

NachoSoto:
Awesome! Thanks for tackling this :100:

ReactiveCocoaTests/Swift/AtomicSpec.swift
ikesyo: The overloaded method tested here was removed in d83b219.

NachoSoto: Ooooh

ReactiveCocoa/Swift/Property.swift
andersio: I think it is fine to repeat outselves here. The setter doesn’t need to do any action on the previous value.

NachoSoto: Still, it's duplicated code, 6 critical lines to be exact.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -101,10 +168,10 @@ class PropertySpec: QuickSpec {
 					property.producer.startWithNext { x in value = x }
 				}
 
-				expect(value).to(equal(0))
+				expect(value) == 0
```
andersio: Hmm. I have overlooked this which is not a new test case introduced in this PR, and applied the new Nimble matchers to it anyway as part of the requested change from @NachoSoto. Please tell in case it has to be reverted.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -90,16 +90,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+
+			return _value
 		}
 
 		set {
-			lock.lock()
-			_value = newValue
-			observer.sendNext(newValue)
-			lock.unlock()
+			modify { _ in newValue }
```
NachoSoto: Perfect

ReactiveCocoaTests/Swift/PropertySpec.swift
NachoSoto: This is good!

NachoSoto:
Looks like there are some conflicts: do you mind rebasing?

ReactiveCocoa/Swift/Property.swift
andersio: `swap` is a syntactic sugar of `modify`, and in terms of `Atomic` it is used quite a few times, e.g. in Signal.swift there is `self.atomicObservers.swap(nil)`. At least it saves you to type `{ _ in }` if the intention is just swapping the value.

NachoSoto: I meant that it's exactly equivalent to the setter:
```swift
property.value = newValue
property.swap(newValue)
```

Because of this redundancy, I would get rid of `swap`.

andersio: But the setter won't return an old value that is read within the same critical section, and without it you will have to use `oldValue = modify { _ in newValue }` instead for the same purpose.

NachoSoto: Ooooh my bad, I missed that detail.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -90,16 +90,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public var value: Value {
 		get {
 			lock.lock()
-			let value = _value
-			lock.unlock()
-			return value
+			defer { lock.unlock() }
+
+			return _value
```
NachoSoto: I'd consider using `withValue` here if we want to remove the duplication and guarantee that reads and writes always go through the same (locking) path: `withValue` and `modify`. What do you think?

This could be `return self.withValue { $0 }`

NachoSoto:
Let me know what you think. Other than that, this is good to go! :shipit:

ReactiveCocoa/Swift/Property.swift
JaviSoto: Using `Result` as a generic type parameter here may be confusing, since `Result` already exists. Maybe `Output`?

ReactiveCocoa/Swift/Property.swift
andersio: It makes sense. :-) Pushed a new commit.

NachoSoto:
Prefect!

PR #2622, opened 2015-12-20
Implement `PropertyType.signal`.

andersio:
Fixes: #2543 

#### Signal.empty
It is a signal that completes immediately without emitting any value. Future observation to this signal would only result in an `Interrupted` event per the `Signal.observe` implementation.

#### ConstantProperty
`ConstantProperty.signal` is just a `Signal.empty`, since there is never a change anyway.

#### MutableProperty
`MutableProperty.signal` is a signal started from the buffering producer at the time the property is initialised.

#### DynamicProperty
`DynamicProperty.signal` and `DynamicProperty.producer` are backed by an underlying `MutableProperty`. The implementation is not thread-safe (as usual...?).

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -10,6 +10,9 @@ public protocol PropertyType {
 	/// A producer for Signals that will send the property's current value,
 	/// followed by all changes over time.
 	var producer: SignalProducer<Value, NoError> { get }
+
+	/// A signal that will send the property's changes over time.
+	var signal: Signal<Value, NoError> { get }
 }
 
 /// A read-only property that allows observation of its changes.
@@ -17,6 +20,8 @@ public struct AnyProperty<Value>: PropertyType {
 
 	private let _value: () -> Value
 	private let _producer: () -> SignalProducer<Value, NoError>
+	private let _signal: () -> Signal<Value, NoError>
+
 
 	public var value: Value {
 		return _value()
@@ -25,11 +30,16 @@ public struct AnyProperty<Value>: PropertyType {
 	public var producer: SignalProducer<Value, NoError> {
 		return _producer()
 	}
+
+	public var signal: Signal<Value, NoError> {
+		return _signal()
+	}
 	
 	/// Initializes a property as a read-only view of the given property.
 	public init<P: PropertyType where P.Value == Value>(_ property: P) {
 		_value = { property.value }
 		_producer = { property.producer }
+		_signal = { property.signal }
 	}
 	
 	/// Initializes a property that first takes on `initialValue`, then each value
@@ -54,11 +64,13 @@ public struct ConstantProperty<Value>: PropertyType {
 
 	public let value: Value
 	public let producer: SignalProducer<Value, NoError>
+	public let signal: Signal<Value, NoError>
 
 	/// Initializes the property to have the given value.
 	public init(_ value: Value) {
 		self.value = value
 		self.producer = SignalProducer(value: value)
+		self.signal = .empty
 	}
 }
 
@@ -80,7 +92,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// Need a recursive lock around `value` to allow recursive access to
 	/// `value`. Note that recursive sets will still deadlock because the
 	/// underlying producer prevents sending recursive events.
-	private let lock = NSRecursiveLock()
+	private let lock: NSRecursiveLock
 	private var _value: Value
 
 	/// The current value of the property.
@@ -97,6 +109,10 @@ public final class MutableProperty<Value>: MutablePropertyType {
 		}
 	}
 
+	/// A signal that will send the property's changes over time,
+	/// then complete when the property has deinitialized.
+	public let signal: Signal<Value, NoError>
+
 	/// A producer for Signals that will send the property's current value,
 	/// followed by all changes over time, then complete when the property has
 	/// deinitialized.
@@ -104,11 +120,18 @@ public final class MutableProperty<Value>: MutablePropertyType {
 
 	/// Initializes the property with the given value to start.
 	public init(_ initialValue: Value) {
-		lock.name = "org.reactivecocoa.ReactiveCocoa.MutableProperty"
-
-		(producer, observer) = SignalProducer<Value, NoError>.buffer(1)
-
 		_value = initialValue
+
+		lock = NSRecursiveLock()
+		lock.name = "org.reactivecocoa.ReactiveCocoa.MutableProperty"
+		(producer, observer) = SignalProducer.buffer(1)
+
+		var extractedSignal: Signal<Value, NoError>?
+		producer.startWithSignal { signal, _ in
+			extractedSignal = signal
+		}
+		signal = extractedSignal!
+
 		observer.sendNext(initialValue)
 	}
 
@@ -160,6 +183,8 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	private weak var object: NSObject?
 	private let keyPath: String
 
+	private var _property: MutableProperty<AnyObject?>?
+
 	/// The current value of the property, as read and written using Key-Value
 	/// Coding.
 	public var value: AnyObject? {
@@ -179,16 +204,11 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// By definition, this only works if the object given to init() is
 	/// KVO-compliant. Most UI controls are not!
 	public var producer: SignalProducer<AnyObject?, NoError> {
-		if let object = object {
-			return object.rac_valuesForKeyPath(keyPath, observer: nil).toSignalProducer()
-				// Errors aren't possible, but the compiler doesn't know that.
-				.flatMapError { error in
-					0 // suppresses implicit return error on fatalError
-					fatalError("Received unexpected error from KVO signal: \(error)")
-				}
-		} else {
-			return .empty
-		}
+		return _property?.producer ?? .empty
+	}
+
+	public var signal: Signal<AnyObject?, NoError> {
+		return _property?.signal ?? .empty
 	}
 
 	/// Initializes a property that will observe and set the given key path of
@@ -196,11 +216,24 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public init(object: NSObject?, keyPath: String) {
 		self.object = object
 		self.keyPath = keyPath
-		
+		self._property = MutableProperty(nil)
+
 		/// DynamicProperty stay alive as long as object is alive.
 		/// This is made possible by strong reference cycles.
 		super.init()
-		object?.rac_willDeallocSignal()?.toSignalProducer().startWithCompleted { self }
+
+		object?.rac_valuesForKeyPath(keyPath, observer: nil)?
+			.toSignalProducer()
+			.start {
+				switch $0 {
+				case let .Next(newValue):
+					self._property?.value = newValue
+				case let .Failed(error):
+					fatalError("Received unexpected error from KVO signal: \(error)")
+				case .Interrupted, .Completed:
+					self._property = nil
+				}
+			}
 	}
 }
 
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -112,11 +112,11 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// A signal that will send the property's changes over time,
 	/// then complete when the property has deinitialized.
 	public lazy var signal: Signal<Value, NoError> = { [unowned self] in
-		var extractedSignal: Signal<Value, NoError>?
+		var extractedSignal: Signal<Value, NoError>!
 		self.producer.startWithSignal { signal, _ in
 			extractedSignal = signal
 		}
-		return extractedSignal!
+		return extractedSignal
 	}()
 
 	/// A producer for Signals that will send the property's current value,
@@ -183,7 +183,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	private weak var object: NSObject?
 	private let keyPath: String
 
-	private var _property: MutableProperty<AnyObject?>?
+	private var property: MutableProperty<AnyObject?>?
 
 	/// The current value of the property, as read and written using Key-Value
 	/// Coding.
@@ -204,11 +204,11 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// By definition, this only works if the object given to init() is
 	/// KVO-compliant. Most UI controls are not!
 	public var producer: SignalProducer<AnyObject?, NoError> {
-		return _property?.producer ?? .empty
+		return property?.producer ?? .empty
 	}
 
 	public var signal: Signal<AnyObject?, NoError> {
-		return _property?.signal ?? .empty
+		return property?.signal ?? .empty
 	}
 
 	/// Initializes a property that will observe and set the given key path of
@@ -216,7 +216,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public init(object: NSObject?, keyPath: String) {
 		self.object = object
 		self.keyPath = keyPath
-		self._property = MutableProperty(nil)
+		self.property = MutableProperty(nil)
 
 		/// DynamicProperty stay alive as long as object is alive.
 		/// This is made possible by strong reference cycles.
@@ -224,14 +224,14 @@ public final class MutableProperty<Value>: MutablePropertyType {
 
 		object?.rac_valuesForKeyPath(keyPath, observer: nil)?
 			.toSignalProducer()
-			.start {
-				switch $0 {
+			.start { event in
+				switch event {
 				case let .Next(newValue):
-					self._property?.value = newValue
+					self.property?.value = newValue
 				case let .Failed(error):
 					fatalError("Received unexpected error from KVO signal: \(error)")
 				case .Interrupted, .Completed:
-					self._property = nil
+					self.property = nil
 				}
 			}
 	}
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -111,7 +111,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
 
 	/// A signal that will send the property's changes over time,
 	/// then complete when the property has deinitialized.
-	public let signal: Signal<Value, NoError>
+	public lazy var signal: Signal<Value, NoError> = { [unowned self] in
+		var extractedSignal: Signal<Value, NoError>?
+		self.producer.startWithSignal { signal, _ in
+			extractedSignal = signal
+		}
+		return extractedSignal!
+	}()
 
 	/// A producer for Signals that will send the property's current value,
 	/// followed by all changes over time, then complete when the property has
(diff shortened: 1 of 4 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -54,11 +64,17 @@ public struct ConstantProperty<Value>: PropertyType {
...
 	public let producer: SignalProducer<Value, NoError>
+	public let signal: Signal<Value, NoError>
 
 	/// Initializes the property to have the given value.
 	public init(_ value: Value) {
 		self.value = value
-		self.producer = SignalProducer(value: value)
+
+		producer = SignalProducer(value: value)
+		signal = Signal { observer in
```
NachoSoto: This should be `Signal.empty`

ReactiveCocoa/Swift/Property.swift
```diff
@@ -89,35 +105,71 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
 
+	/// A signal that will send the property's changes over time,
+	/// then complete when the property has deinitialized.
+	public let signal: Signal<Value, NoError>
+
 	/// A producer for Signals that will send the property's current value,
 	/// followed by all changes over time, then complete when the property has
 	/// deinitialized.
-	public let producer: SignalProducer<Value, NoError>
+	public var producer: SignalProducer<Value, NoError> {
```
NachoSoto: This is clever, but did you consider keeping the producer the source of truth of everything and defining the signal in terms of that, using `skip(1)`?
That way the implementation would be simpler and still threadsafe.

ReactiveCocoa/Swift/Property.swift
andersio: There is only `Signal.never`. :D

ReactiveCocoa/Swift/Property.swift
andersio: It should be still the source of truth. When a producer is being started, it acquires the recursive lock and then send the current value and setup the forwarding within the same critical section. So it should be consistent.

andersio: Yeah, you can extract a `skip(1)` signal before sending the initial value in the initialiser to achieve the same goal.

Anyway, the implementation here is thread-safe if I didn't miss something - the start handler has to acquire the recursive lock to do its work (replay the current value + setup the forwarding to the started signal). It should be slightly simpler too, since now you get one less lock and copy of value (BufferState) in the memory.

ReactiveCocoa/Swift/Property.swift
NachoSoto: Ooh. Then let's add that! :)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -171,11 +226,21 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
-		object?.rac_willDeallocSignal()?.toSignalProducer().startWithCompleted { self }
+
+		object?.rac_valuesForKeyPath(keyPath, observer: nil)?
+			.toSignalProducer()
+			.start {
+				switch $0 {
+				case let .Next(newValue):
+					self._property?.value = newValue
+				default:
+					self._property = nil
```
NachoSoto: This shouldn't ignore errors.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -154,16 +208,17 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
-					0 // suppresses implicit return error on fatalError
-					fatalError("Received unexpected error from KVO signal: \(error)")
-				}
-		} else {
-			return .empty
+		return _property?.producer ?? .empty
+	}
+
+	public var signal: Signal<AnyObject?, NoError> {
+		guard let _property = _property else {
```
NachoSoto: This is duplicated. Why not use the signal from the underlying property?

NachoSoto:
Thanks for tackling this! Can you rebase now that #2617 is merged?

ReactiveCocoa/Swift/Property.swift
andersio: Doesn't `rac_valuesAndChangesForKeyPath` never emit an error? This is what the current implementation in master branch said.

ReactiveCocoa/Swift/Property.swift
andersio: It is because the DynamicProperty can possibly be retained and outlive the underlying object. So in order to have the signal and producer complete at the time the underlying object is gone, the underlying MutableProperty must be deallocated.

ReactiveCocoa/Swift/Property.swift
NachoSoto: Yeah but that can't be guaranteed at compile time so we shouldn't ignore a potential error. That's why the previous implementation used `fatalError`.

ReactiveCocoa/Swift/Property.swift
NachoSoto: Makes sense. Let's add a `Signal.empty` So that this can be just `return _property.signal ??.empty`

ReactiveCocoa/Swift/Property.swift
```diff
@@ -135,6 +187,8 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	private weak var object: NSObject?
 	private let keyPath: String
 
+	private var _property: MutableProperty<AnyObject?>? = MutableProperty<AnyObject?>(nil)
```
NachoSoto: The type here is redundant.

ReactiveCocoa/Swift/Property.swift
NachoSoto: Also, can you use an explicit list of events instead of `default`? If we add a new event this would force us to think about how to handle it, whereas `default` would ignore that fact.

andersio:
Rebased and updated. :-)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -54,11 +64,13 @@ public struct ConstantProperty<Value>: PropertyType {
 
 	public let value: Value
 	public let producer: SignalProducer<Value, NoError>
+	public let signal: Signal<Value, NoError>
 
 	/// Initializes the property to have the given value.
 	public init(_ value: Value) {
 		self.value = value
-		self.producer = SignalProducer(value: value)
+		producer = SignalProducer(value: value)
```
NachoSoto: Can you use `self.* = ...` here for consistency?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -87,6 +87,14 @@ public final class Signal<Value, Error: ErrorType> {
 		return self.init { _ in nil }
 	}
 
+	/// A Signal that completes immediately.
```
NachoSoto: Maybe add "without emitting any values".

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -87,6 +87,14 @@ public final class Signal<Value, Error: ErrorType> {
...
+	public static var empty: Signal {
```
NachoSoto: Awesome. This will need a test though!

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -66,6 +87,20 @@ class PropertySpec: QuickSpec {
 				expect(sentValue) == subsequentPropertyValue
 			}
 
+			it("should yield a signal that sends the all the changes") {
```
NachoSoto: Maybe `that emits subsequent changes`

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -66,6 +87,20 @@ class PropertySpec: QuickSpec {
...
+				let mutableProperty = MutableProperty(initialPropertyValue)
+
+				var sentValue: String?
+
+				mutableProperty.signal.observeNext { value in
+					sentValue = value
+				}
+
+				expect(sentValue).to(beNil())
```
NachoSoto: Good

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -66,6 +87,20 @@ class PropertySpec: QuickSpec {
...
+				mutableProperty.value = subsequentPropertyValue
+				expect(sentValue) == subsequentPropertyValue
```
NachoSoto: It might be worth adding another modification to verify that it keeps emitting changes.
Also, can you add a separate test to verify that it also emits values when the property is set **even though the value didn't change**?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -23,6 +23,27 @@ class PropertySpec: QuickSpec {
...
+
+				var sentValue: String?
+				var signalInterrupted = false
+
+				constantProperty.signal.observe { event in
+					switch event {
+					case let .Next(value):
+						sentValue = value
+					case .Interrupted:
+						signalInterrupted = true
```
NachoSoto: Can you verify `.Completed` as well?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -66,6 +87,20 @@ class PropertySpec: QuickSpec {
 				expect(sentValue) == subsequentPropertyValue
 			}
 
+			it("should yield a signal that sends the all the changes") {
+				let mutableProperty = MutableProperty(initialPropertyValue)
+
+				var sentValue: String?
+
+				mutableProperty.signal.observeNext { value in
```
NachoSoto: Maybe keep this in one line?
```swift
mutableProperty.signal.observeNext { value = $0 }
```

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -79,6 +114,19 @@ class PropertySpec: QuickSpec {
 				expect(signalCompleted) == true
 			}
 
+			it("should complete its signal when deallocated") {
+				var mutableProperty: MutableProperty? = MutableProperty(initialPropertyValue)
+
+				var signalCompleted = false
+
+				mutableProperty?.signal.observeCompleted {
```
NachoSoto: This should explicitly unwrap, since we shouldn't expect it to be `nil` here.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -79,6 +114,19 @@ class PropertySpec: QuickSpec {
...
+					signalCompleted = true
+				}
+
+				mutableProperty = nil
```
NachoSoto: Cool

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -182,7 +230,9 @@ class PropertySpec: QuickSpec {
 					let property = AnyProperty(constantProperty)
 
 					var sentValue: String?
+					var signalSentValue: String?
 					var producerCompleted = false
+					var signalInterrupted = false
```
NachoSoto: Why not check for `completed` instead?

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -287,7 +350,23 @@ class PropertySpec: QuickSpec {
 				expect(values) == [ 0, 1, 2 ]
 			}
 
-			it("should complete when the underlying object deallocates") {
+			it("should yield a signal that sends the changes for the key path of the underlying object") {
+				var values: [Int] = []
+				property.signal.observeNext { value in
+					expect(value).notTo(beNil())
+					values.append((value as? Int) ?? -1)
```
NachoSoto: This should be `as! Int`. Anything other than an `Int` should be considered a failure.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -307,7 +386,28 @@ class PropertySpec: QuickSpec {
 				expect(completed).toEventually(beTruthy())
 				expect(property.value).to(beNil())
 			}
-			
+
+			it("should have a completed signal when the underlying object deallocates") {
```
NachoSoto: Very thorough tests! :+1:

ReactiveCocoa/Swift/Property.swift
```diff
@@ -97,16 +108,27 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+	/// then complete when the property has deinitialized.
+	public let signal: Signal<Value, NoError>
+
 	/// A producer for Signals that will send the property's current value,
 	/// followed by all changes over time, then complete when the property has
 	/// deinitialized.
-	public let producer: SignalProducer<Value, NoError>
+	public var producer: SignalProducer<Value, NoError> {
+		return SignalProducer { observer, disposable in
+			self.withValue {
```
NachoSoto: I'm not sure about this: isn't this an unnecessary ownership? As long as there's a copy of this producer the property will be retained, which changes the semantics.

I still think that the implementation would be more correct by keeping the implementation using `SignalProducer.buffer` and using `.skip(1)` for the signal.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -196,11 +215,25 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+
+		object?.rac_valuesForKeyPath(keyPath, observer: nil)?
+			.toSignalProducer()
+			.start {
+				switch $0 {
+				case let .Next(newValue):
+					self._property?.value = newValue
+				case let .Failed(error):
+					fatalError("Received unexpected error from KVO signal: \(error)")
+				case .Interrupted:
```
NachoSoto: This should be `case .Interrupted, .Completed:`.

ReactiveCocoaTests/Swift/PropertySpec.swift
andersio: `.Completed` cannot be caught.

ReactiveCocoaTests/Swift/PropertySpec.swift
andersio: External observers can never catch a `Completed` event from the signal of a `ConstantProperty`, and doing `signal.observe` on such a signal would result in an `Interrupted` event passed to the `Observer`.

ReactiveCocoaTests/Swift/PropertySpec.swift
andersio: The test case would fail anyway if such a thing happens.

NachoSoto: Only because you chose a magic number that happens to be different than the assertion, but that's a very fragile way of verifying that, and it doesn't correctly communicate what you *expect* in the test.

ReactiveCocoa/Swift/Property.swift
andersio: Ohhhhh. It is true in this regard. The current value is inaccessible after the property is deallocated.

andersio:
Reverted to use `SignalProducer.buffer`. Note that in the initialiser, an extra `Signal` is created in order to work around the initialisation requirement of Swift. Otherwise, `PropertyType.signal` has to be an IUO. :-(

andersio:
The failing test case is from RACSignalSpec, hmm.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -97,6 +109,16 @@ public final class MutableProperty<Value>: MutablePropertyType {
 		}
 	}
 
+	/// A signal that will send the property's changes over time,
+	/// then complete when the property has deinitialized.
+	public lazy var signal: Signal<Value, NoError> = { [unowned self] in
```
NachoSoto: This is fine. Eventually I think we should actually expose this from `SignalProducer`.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -97,6 +109,16 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+		var extractedSignal: Signal<Value, NoError>?
```
NachoSoto: Maybe just make this an `ImplititlyUnwrappedOptional`?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -196,11 +216,24 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	public init(object: NSObject?, keyPath: String) {
 		self.object = object
 		self.keyPath = keyPath
-		
+		self._property = MutableProperty(nil)
```
NachoSoto: Can you just name this `property`? No need for the underscore.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -196,11 +216,24 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+
 		/// DynamicProperty stay alive as long as object is alive.
 		/// This is made possible by strong reference cycles.
 		super.init()
-		object?.rac_willDeallocSignal()?.toSignalProducer().startWithCompleted { self }
+
+		object?.rac_valuesForKeyPath(keyPath, observer: nil)?
+			.toSignalProducer()
+			.start {
+				switch $0 {
```
NachoSoto: Nit: maybe name this `event`?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -87,6 +87,14 @@ public final class Signal<Value, Error: ErrorType> {
 		return self.init { _ in nil }
 	}
 
+	/// A Signal that completes immediately without emitting any value.
+	public static var empty: Signal {
```
NachoSoto: This needs a test.

NachoSoto:
Looks perfect! Only thing missing is a test for `Signal.empty`, and then it's good to go!

ReactiveCocoa/Swift/Property.swift
andersio: It needn't be an IUO (https://github.com/andersio/ReactiveCocoa/commit/2651f6ff98ac77bcf84bd913680ca5da17076a2d#diff-84e9bb9c50af433fe0ae0e9e9bc2469fR129), but I think it makes sense to extract the signal only when it is needed. What do you think?

ReactiveCocoa/Swift/Property.swift
andersio: Maybe a `SignalProducer.startedSignal`?

NachoSoto: Yeah. Feel free to extract that now, but I think we should leave it as `internal` for now.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -23,6 +24,25 @@ class PropertySpec: QuickSpec {
...
+				let constantProperty = ConstantProperty(initialPropertyValue)
+
+				var signalInterrupted = false
+				var hasUnexpectedEventsEmitted = false
+
+				constantProperty.signal.observe { event in
+					switch event {
+					case .Interrupted:
+						signalInterrupted = true
+					case .Next(_), .Failed(_), .Completed:
```
NachoSoto: Ditto: don't need parenthesis

NachoSoto:
Awesome :shipit:

ReactiveCocoa/Swift/Property.swift
andersio: Perhaps just leave it alone then.

RuiAAPeres:
This is pretty nice! GJ @andersio. @NachoSoto could it be merged (I actually have use case for this)?

Issue #2561, opened 2015-11-19
`scan` and initial value


mbrandonw:
I think it's pretty common for reactive frameworks to implement `scan` so that it immediately emits its initial value, e.g. the whole Rx family. This is very useful for setting up initial state. To me it's the FRP way of getting around putting display code in a function that gets called when the view is initialized and then again when things start happening.

Here's an example for a producer, but I would expect it to be the same for signals:

```swift
SignalProducer<Int, NoError>(values: [1, 2, 3]).scan(0, +).start { print($0) }
```

That prints `1, 3, 6` where I would expect `0, 1, 3, 6`.

Thoughts on changing `scan` to have this behavior?

neilpa:
> I think it's pretty common for reactive frameworks to implement scan so that it immediately emits its initial value, e.g. the whole Rx family.

Nope, other Rx libraries [implement `scan` the same way](http://www.introtorx.com/Content/v1.0.10621.0/07_Aggregation.html#Scan).

> That prints `1, 3, 6` where I would expect `0, 1, 3, 6`.

How would this work in the general case? Remember that the type signature for the accumulator is `(U, T) -> U`. In your example both `T` and `U` are of type `Int` so it's easy to think that we can generate an initial `0` value. However, there are many other cases where `T` is another type that isn't default constructible so we have to wait for the first value from the signal.

mbrandonw:
Ah, did not check Rx .NET, but it is indeed true that `scan` works as I described in RxJS and RxJava:

http://reactivex.io/documentation/operators/scan.html

<img width="640" alt="scanseed" src="https://cloud.githubusercontent.com/assets/135203/11279425/9605e87e-8ebe-11e5-84bf-463ae444b09b.png">

Not quite sure I understand your hypothetical. `scan` currently needs an initial value (which would be of type `T` in your terms), so this wouldn't add any more restrictions. 

The general case is to simply emit `initial` value immediately and then the first accumulation after the signal emits for the first time.

NachoSoto:
You can always start emitting that value:

```swift
let producer = SignalProducer<Int, NoError>(values: [1, 2, 3])
SignalProducer<Int, NoError>(value: 0)
    .concat(producer)
    .scan(0, +)
    .start { print($0) }
```

You can simplify this even more by creating a simple `startWith` operator as well.

Personally I don't have a preference, I think if we were to change this some people would want the opposite behavior and they'd need `.skip(1)`.

[...]

PR #2632, opened 2015-12-23
Add no-argument form of combinePrevious(), where previous value is optional

sharplet:
I personally found this operator useful when filtering a stream, such that the first value should always be passed through, but subsequent values should be compared to previous values. For example:

```swift
producer.combinePrevious()
  .filter { previous, next in
    guard let previous = previous else { return true }
    return previous.value != next.value
  }
  .map { $1 }
```

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -990,6 +990,20 @@ extension SignalType {
 		}
 	}
 
+	/// Forwards events from `self` with history, if available: values of the
+	/// returned signal are a tuple whose second value is the current value, and
+	/// whose first member is optionally the previous value. The first time `self`
+	/// sends a value, the previous value will be `nil`.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func combinePrevious() -> Signal<(Value?, Value), Error> {
+		var previous: Value?
+		return map { next in
+			let combined = (previous, next)
+			previous = next
+			return combined
+		}
+	}
+
 	/// Like `scan`, but sends only the final value and then immediately completes.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func reduce<U>(initial: U, _ combine: (U, Value) -> U) -> Signal<U, Error> {
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -597,6 +597,15 @@ extension SignalProducerType {
 		return lift { $0.combinePrevious(initial) }
 	}
 
+	/// Forwards events from `self` with history, if available: values of the
+	/// returned signal are a tuple whose second value is the current value, and
+	/// whose first member is optionally the previous value. The first time `self`
+	/// sends a value, the previous value will be `nil`.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func combinePrevious() -> SignalProducer<(Value?, Value), Error> {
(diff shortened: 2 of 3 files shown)


NachoSoto:
Thanks for the PR! There are 3 issues with this, however:
- One of the `combinePrevious` implementation is not implemented in terms of the other.
- The closure passed to `map` should not have side effects.
- This operator is redundant: the existing `combinePrevious` is more general. You can accomplish the same thing by doing something like `producer.combinePrevious(nil).skip(1)`.

NachoSoto:
There is a similar discussion in #2561.

sharplet:
> You can accomplish the same thing by doing something like `producer.combinePrevious(nil).skip(1)`.

Unfortunately this doesn't compile, as the type of `producer` is initially non-optional, thus `nil` isn't a valid initial value. My first attempt at solving this problem used the existing `combinePrevious(_:)`, and looked like this:

```swift
producer
  .map(Optional.init) // allow nil to be a valid initial value
  .combinePrevious(nil)
  .filter { previous, next in
    guard let next = next else { fatalError() } // not ideal
    guard let previous = previous else { return true }
    return previous.value != next.value
  }
  .map { $1 }
  .ignoreNil() // this is necessary because we had to lift the whole stream into Optional at the start
```

There's a few reasons this isn't ideal, but the main one for me is how much this version obscures the intent, because of the dance around Optional.

> One of the `combinePrevious` implementation is not implemented in terms of the other.

Yeah, perhaps this is actually overloading the name `combinePrevious` with something too different? However it still feels to me like it's in the same spirit, even though the types are different.

> The closure passed to `map` should not have side effects.

Would you mind elaborating on this? Totally get there could be a latent problem here, I'm just having difficulty reasoning about it in my head.

NachoSoto:
> Unfortunately this doesn't compile, as the type of `producer` is initially non-optional

Sorry, I was typing from my phone and I missed that. In any case, you can easily add this as an extension in your project. I'm not in favor of adding an overload to `combinePrevious` with different semantics. The goal for now is to get `RAC 4`'s stable, with a minimum API surface. Since the behavior that you want can be easily created using existing operators I'd say that it doesn't add much value for now.

sharplet:
Thanks for taking the time to review! :bow:

I'll go ahead and close this.

If you get a chance, I'd love to know more about the closures with side effects issue. I had trouble working out if there's anything in the [Design Guidelines](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/Documentation/DesignGuidelines.md) addressing this particular issue.

sharplet:
FWIW, I did some investigation on whether it's a problem for the block to have side effects in this case.  I'm not sure if my reasoning is 100% correct, but I thought it might be useful to throw this out there.
If you have an opportunity to take a look at this @NachoSoto I'd really appreciate it!

Here's a simplified object graph for the `SignalType` version of the operator:

```
Signal (underlying) <--- closure passed to `map()` <-- Signal (combinePrevious)
                      /
       `next` var <--/
```

Because the closure has a reference to the `next` var, if multiple objects gained a reference to the closure then the effect would be shared amongst all of them.

However, I don't think this is a problem in this case, because there's no way to copy a `Signal`.

If the same implementation was implemented for `SignalProducerType`, that would be a problem.  Signal producers are easily copied, and so any effects are shared between the different copies.

```
SignalProducer (underlying) <--- closure passed to `map()` <-- SignalProducer (combinePrevious)
                              /                          ^
               `next` var <--/                            \--- SignalProducer (copy)
```

In this case the `next` var is effectively shared between multiple producers.  This effectively breaks the value semantics.

PR #2442, opened 2015-10-02
Kill "sink", reify "Observer", kill send* free functions

andymatuschak:
When we were introducing teammates to RAC 4, there were two related sticking points that seemed like incidental complexity: the conflation of “sink” and “observer”, and the fact that `sendNext` and friends were free functions (didn’t autocomplete, looked weird, etc).

Here, I’ve done the following:
* `Event` no longer knows about sinks or observers.
* `Observer` is now its own thing akin to `SinkOf` but specific to `Event`s. There is no such thing as a sink anymore.
* `Event.sink()` becomes a convenience initializer for `Observer`.
* `send*` become methods on `Observer`
* Added some convenience overloads where raw observer/sink functions were arguments so that they can continue to be used that way for less line noise.

If you folks like this approach, I’ll update docs etc.

PR #2634, opened 2015-12-25
Replace most Signal<T, E>.Observer typealias usages with plain Observer<T, E>

ikesyo:
That was useful before #2442 is merged.


NachoSoto:
I actually think this is semantically more accurate, and it helps users, as they don't need to find this external type themselves.

PR #2639, opened 2016-01-02
Added SignalProducer.replayLazily for multicasting

NachoSoto:
Motivation
====================

I expect this to be a controversial one, so let me provide some context: multicasting has always been an important feature.
Notably, *Rx* contains multiple operators ([replay and cache](http://reactivex.io/documentation/operators/replay.html), [multicast](http://reactivex.io/documentation/operators/publish.html), etc) for this. However, we've always been very cautious about including this in *ReactiveCocoa* **because it's very easy to misuse**. Specifically, the most common pattern that I see is this:
```java
final BehaviorSubject<Integer> subject = BehaviorSubject.create();
final Observable<Integer> observable = subject.getObservable();

// ... somewhere else in the code:

// This is safe because it's backed by a `BehaviorSubject`.
final int lastNumber = observable.toBlocking().first();
```

I've seen this **so many times**. It seems fine at first, but it's very important to realize that the comment "this is safe" is actually a lie: we have no compile time guarantees that it will not deadlock, or block the current thread for several seconds.
This is precisely why `PropertyType` was born: not only does it make this pattern a lot easier, but it provides compile-time guarantees and better semantics to implement this sort of thing.

This is exactly why I and others have felt that A: we don't need multicasting, and B: if we did add it, it would be used instead of using `PropertyType`.
However, I've ran into several scenarios where *replaying* was necessary. To accomplish it, I've been using `MutableProperty` (I even [suggested this in StackOverflow](http://stackoverflow.com/questions/33946571/how-do-i-observe-a-signal-and-immediately-receive-a-next-event-if-it-has-alrea/34010425#34010425)), so I thought it was time to provide a first-party solution to this.

Details
====================

I've been putting a lot of thought into this, whether it really belongs in RAC's codebase, the semantics that it should have, etc. And this is the result!
First let me explain what the semantics are (all of this is tested in this commit):

- Invoking the operator has no side effects, i.e. it's *lazy*. This is important to maintain consistency with the rest of `SignalProducer` semantics: only `start` can have side effects.
- Upon the first call to `start`, the underlying `SignalProducer* is started.
- Values are buffered to all subsequent subscribers, from that point on.
- This is slightly weird because the `SignalProducer` no longer has *value semantics* (its buffered values are shared). But this is already how things like `SignalProducer.buffer` or `PropertyType.producer` work: they also have reference semantics, so I think it's fine. It just means that starting the producer *can* have side effects, but most importantly, different observers will see different versions of the producer depending on when they `start` it.
- When the producer *fails*, the `error` is automatically forwarded to any subsequent observer.
- When the producer *completes*, new observers will see the last `n` values emitted, and then complete.
- The subscription to the underlying `SignalProducer` will stay alive even if the last observer disposes of it. This is important, because there might be new observers later on.
- If there are no more subscriptions to the replayed `SignalProducer`, and it goes out of scope, the underlying producer is disposed.
- The result of this operator **is another `SignalProducer`**: this is important because, as explained above, each observer will see a different version of the truth.
- This operator **is only implemented for `SignalProducer`**: this is also very important (and something that Rx can't do). it wouldn't make sense to multicast a `Signal`!

So notice that this is exactly the same semantics as `SignalProducer.buffer`. It allows you to multicast elements from a single source.
Thanks to this, its implementation is actually 100% based on `SignalProducer.buffer`!

I tried explaining these points as concisely as possible as part of the *dosctring*, but let me know if it could be improved.

Use cases
====================

- As an optimization: one can use this operator to avoid doing the same work multiple times. I needed this myself to get these semantics [in my rendering framework](https://github.com/NachoSoto/AsyncImageView/blob/master/AsyncImageView/Renderers/MulticastedRenderer.swift), and using an operator would be a lot easier.
- In cases where the same producer is needed multiple times as part of a complicated operator chain. For example, being able to do things like `combineLatest(p.filter(f1), p.filter(f2))` without incurring in 2 subscriptions to the same `SignalProducer`.

Questions
====================

- I'd like @jspahrsummers to take a look at this if possible. He actually taught me a lot of what I explained here regarding the common cases that lead to misusing *multicasting*. **I think** I've managed to come up with an implementation that minimizes the room for errors, and the *docstring* explains when `PropertyType` would be preferred, but I'd like to hear his thoughts, and anybody else's, regarding this controversial operator.
- I'm also interested in hearing other people's ideas for cases where they wished *multicasting* were easier, and make sure that these are the right semantics we want.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1167,6 +1167,8 @@ extension SignalProducerType {
 
 	/// Creates a new `SignalProducer` that will multicast values emitted by
 	/// the underlying producer, up to `capacity`.
+	/// This means that all clients of this `SignalProducer` will see the same version
+	/// of the emitted values/errors.
 	///
 	/// The underlying `SignalProducer` will not be started until `self` is started
 	/// for the first time. When subscribing to this producer, all previous values
@@ -1175,6 +1177,13 @@ extension SignalProducerType {
 	/// If you find yourself needing *the current value* (the last buffered value)
 	/// you should consider using `PropertyType` instead, which, unlike this operator,
 	/// will guarantee at compile time that there's always a buffered value.
+	/// This operator is not recommended in most cases, as it will introduce an implicit
+	/// relationship between the original client and the rest, so consider alternatives
+	/// like `PropertyType`, `SignalProducer.buffer`, or representing your stream using 
+	/// a `Signal` instead.
+	///
+	/// This operator is only recommended when you absolutely need to introduce
+	/// a layer of caching in front of another `SignalProducer`.
 	///
 	/// This operator has the same semantics as `SignalProducer.buffer`.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1187,8 +1187,9 @@ extension SignalProducerType {
 		let lock = NSLock()
 		lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.replayLazily"
 
-		// This will go "out of scope" when the return `SignalProducer` is deallocated.
+		// This will go "out of scope" when the returned `SignalProducer` goes out of scope.
 		// This lets us know when we're supposed to dispose the underlying producer.
+		// This is necessary because `struct`s don't have `deinit`.
 		let token = NSObject()
 
 		return SignalProducer { observer, disposable in
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1230,6 +1230,6 @@ private extension NSObject {
 			.mapError { error in
 				fatalError("Unexpected error: \(error)")
 				()
-		}
+			}
 	}
 }
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -121,7 +121,7 @@ public struct SignalProducer<Value, Error: ErrorType> {
 	/// will not add any further events. This _does not_ count against the
 	/// value capacity so no buffered values will be dropped on termination.
 	public static func buffer(capacity: Int = Int.max) -> (SignalProducer, Signal<Value, Error>.Observer) {
-		precondition(capacity >= 0)
+		precondition(capacity >= 0, "Invalid capacity: \(capacity)")
 
 		// This is effectively used as a synchronous mutex, but permitting
 		// limited recursive locking (see below).
@@ -1035,7 +1035,6 @@ public func zip<S: SequenceType, Value, Error where S.Generator.Element == Signa
 	return .empty
 }
 
-
 extension SignalProducerType {
 	/// Repeats `self` a total of `count` times. Repeating `1` times results in
 	/// an equivalent signal producer.
@@ -1164,5 +1163,73 @@ extension SignalProducerType {
 	@warn_unused_result(message="Did you forget to check the result?")
 	public func wait() -> Result<(), Error> {
 		return then(SignalProducer<(), Error>(value: ())).last() ?? .Success(())
+	}
+
+	/// Creates a new `SignalProducer` that will multicast values emitted by
+	/// the underlying producer, up to `capacity`.
+	///
+	/// The underlying `SignalProducer` will not be started until `self` is started
+	/// for the first time. When subscribing to this producer, all previous values
+	/// (up to `capacity`) will be emitted, followed by any new values.
+	///
+	/// If you find yourself needing *the current value* (the last buffered value)
+	/// you should consider using `PropertyType` instead, which, unlike this operator,
+	/// will guarantee at compile time that there's always a buffered value.
+	///
+	/// This operator has the same semantics as `SignalProducer.buffer`.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func replayLazily(capacity: Int = Int.max) -> SignalProducer<Value, Error> {
+		precondition(capacity >= 0, "Invalid capacity: \(capacity)")
+
+		var producer: SignalProducer<Value, Error>?
+		var producerObserver: SignalProducer<Value, Error>.ProducedSignal.Observer?
+
+		let lock = NSLock()
+		lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.replayLazily"
+
+		// This will go "out of scope" when the return `SignalProducer` is deallocated.
+		// This lets us know when we're supposed to dispose the underlying producer.
+		let token = NSObject()
+
+		return SignalProducer { observer, disposable in
+			let initializedProducer: SignalProducer<Value, Error>
+			let initializedObserver: SignalProducer<Value, Error>.ProducedSignal.Observer
+			let shouldStartUnderlyingProducer: Bool
+
+			lock.lock()
+			if let producer = producer, producerObserver = producerObserver {
+				(initializedProducer, initializedObserver) = (producer, producerObserver)
+				shouldStartUnderlyingProducer = false
+			} else {
+				let (producerTemp, observerTemp) = SignalProducer<Value, Error>.buffer(capacity)
+
+				(producer, producerObserver) = (producerTemp, observerTemp)
+				(initializedProducer, initializedObserver) = (producerTemp, observerTemp)
+				shouldStartUnderlyingProducer = true
+			}
+			lock.unlock()
+
+			// subscribe `observer` before starting the underlying producer.
+			disposable += initializedProducer.start(observer)
+
+			if shouldStartUnderlyingProducer {
+				self.producer
+					.takeUntil(token.willDeallocSignal)
+					.start(initializedObserver)
+			}
+		}
+	}
+}
+
+private extension NSObject {
+	var willDeallocSignal: SignalProducer<(), NoError> {
+		return self
+			.rac_willDeallocSignal()
+			.toSignalProducer()
+			.map { _ in () }
+			.mapError { error in
+				fatalError("Unexpected error: \(error)")
+				()
+		}
 	}
 }
--- a/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerSpec.swift
@@ -1861,6 +1861,244 @@ class SignalProducerSpec: QuickSpec {
 				expect(started) == false
 			}
 		}
+
+		describe("replayLazily") {
+			var producer: SignalProducer<Int, TestError>!
+			var observer: SignalProducer<Int, TestError>.ProducedSignal.Observer!
+
+			var replayedProducer: SignalProducer<Int, TestError>!
+
+			beforeEach {
+				let (producerTemp, observerTemp) = SignalProducer<Int, TestError>.buffer(0)
+				producer = producerTemp
+				observer = observerTemp
+
+				replayedProducer = producer.replayLazily(2)
+			}
+
+			context("subscribing to underlying producer") {
+				it("emits new values") {
+					var last: Int?
+
+					replayedProducer.startWithNext { last = $0 }
+					expect(last).to(beNil())
+
+					observer.sendNext(1)
+					expect(last) == 1
+
+					observer.sendNext(2)
+					expect(last) == 2
+				}
+
+				it("emits errors") {
+					var error: TestError?
+
+					replayedProducer.startWithFailed { error = $0 }
+					expect(error).to(beNil())
+
+					observer.sendFailed(.Default)
+					expect(error) == TestError.Default
+				}
+			}
+
+			context("buffers past values") {
+				it("emits last value upon subscription") {
+					let disposable = replayedProducer
+						.start()
+
+					observer.sendNext(1)
+					disposable.dispose()
+
+					var last: Int?
+
+					replayedProducer
+						.startWithNext { last = $0 }
+					expect(last) == 1
+				}
+
+				it("emits previous failure upon subscription") {
+					let disposable = replayedProducer
+						.start()
+
+					observer.sendFailed(.Default)
+					disposable.dispose()
+
+					var error: TestError?
+
+					replayedProducer
+						.startWithFailed { error = $0 }
+					expect(error) == TestError.Default
+				}
+
+				it("emits last n values upon subscription") {
+					var disposable = replayedProducer
+						.start()
+
+					observer.sendNext(1)
+					observer.sendNext(2)
+					observer.sendNext(3)
+					observer.sendNext(4)
+					disposable.dispose()
+
+					var values: [Int] = []
+
+					disposable = replayedProducer
+						.startWithNext { values.append($0) }
+					expect(values) == [ 3, 4 ]
+
+					observer.sendNext(5)
+					expect(values) == [ 3, 4, 5 ]
+
+					disposable.dispose()
+					values = []
+
+					replayedProducer
+						.startWithNext { values.append($0) }
+					expect(values) == [ 4, 5 ]
+				}
+			}
+
+			context("starting underying producer") {
+				it("starts lazily") {
+					var started = false
+
+					let producer = SignalProducer<Int, NoError>(value: 0)
+						.on(started: { started = true })
+					expect(started) == false
+
+					let replayedProducer = producer
+						.replayLazily(1)
+					expect(started) == false
+
+					replayedProducer.start()
+					expect(started) == true
+				}
+
+				it("shares a single subscription") {
+					var startedTimes = 0
+
+					let producer = SignalProducer<Int, NoError>.never
+						.on(started: { startedTimes++ })
+					expect(startedTimes) == 0
+
+					let replayedProducer = producer
+						.replayLazily(1)
+					expect(startedTimes) == 0
+
+					replayedProducer.start()
+					expect(startedTimes) == 1
+
+					replayedProducer.start()
+					expect(startedTimes) == 1
+				}
+
+				it("does not start multiple times when subscribing multiple times") {
+					var startedTimes = 0
+
+					let producer = SignalProducer<Int, NoError>(value: 0)
+						.on(started: { startedTimes++ })
+
+					let replayedProducer = producer
+						.replayLazily(1)
+
+					expect(startedTimes) == 0
+					replayedProducer.start().dispose()
+					expect(startedTimes) == 1
+					replayedProducer.start().dispose()
+					expect(startedTimes) == 1
+				}
+
+				it("does not start again if it finished") {
+					var startedTimes = 0
+
+					let producer = SignalProducer<Int, NoError>.empty
+						.on(started: { startedTimes++ })
+					expect(startedTimes) == 0
+
+					let replayedProducer = producer
+						.replayLazily(1)
+					expect(startedTimes) == 0
+
+					replayedProducer.start()
+					expect(startedTimes) == 1
+
+					replayedProducer.start()
+					expect(startedTimes) == 1
+				}
+			}
+
+			context("lifetime") {
+				it("does not dispose underlying subscription if the replayed producer is still in memory") {
+					var disposed = false
+
+					let producer = SignalProducer<Int, NoError>.never
+						.on(disposed: { disposed = true })
+
+					let replayedProducer = producer
+						.replayLazily(1)
+
+					expect(disposed) == false
+					let disposable = replayedProducer.start()
+					expect(disposed) == false
+
+					disposable.dispose()
+					expect(disposed) == false
+				}
+
+				it("disposes underlying producer when the producer is deallocated") {
+					var disposed = false
+
+					let producer = SignalProducer<Int, NoError>.never
+						.on(disposed: { disposed = true })
+
+					var replayedProducer = ImplicitlyUnwrappedOptional(producer.replayLazily(1))
+
+					expect(disposed) == false
+					let disposable = replayedProducer.start()
+					expect(disposed) == false
+
+					disposable.dispose()
+					expect(disposed) == false
+
+					replayedProducer = nil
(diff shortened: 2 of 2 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1165,4 +1164,54 @@ extension SignalProducerType {
...
+	/// for the first time. When subscribing to this producer, all previous values
+	/// (up to `capacity`) will be emitted, followed by any new values.
+	///
+	/// If you find yourself needing *the current value* (the last buffered value)
+	/// you should consider using `PropertyType` instead, which, unlike this operator,
+	/// will guarantee at compile time that there's always a buffered value.
+	///
+	/// This operator has the same semantics as `SignalProducer.buffer`.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func replayLazily(capacity: Int = Int.max) -> SignalProducer<Value, Error> {
```
JaviSoto: Is that a sensible default? With a "noisy" producer, this could inadvertently use way too much memory. Not sure I have a better suggestion though. I think using 1 may be a common case, but I don't think that that would pass the "principle of least surprise" test for the API...

JaviSoto:
I love this 👍 I hope the idea I gave you online helps you make that test pass

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1165,4 +1164,54 @@ extension SignalProducerType {
 	public func wait() -> Result<(), Error> {
 		return then(SignalProducer<(), Error>(value: ())).last() ?? .Success(())
 	}
+
+	/// Creates a new `SignalProducer` that will multicast values emitted by
```
JaviSoto: What do you think about explaining here what multitasking entails?

NachoSoto:
@JaviSoto thanks! that fixed the last test :dancer:

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Good point. I just followed `SignalProducer.buffer`'s maximum, but I agree, I think we should remove this maximum for the reason that you mention. It's probably common for people to not think about this capacity and leave the default, wasting memory.

I opened #2640 to track this.

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Good point.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1165,4 +1164,72 @@ extension SignalProducerType {
...
+	public func replayLazily(capacity: Int = Int.max) -> SignalProducer<Value, Error> {
+		precondition(capacity >= 0, "Invalid capacity: \(capacity)")
+
+		var producer: SignalProducer<Value, Error>?
+		var producerObserver: SignalProducer<Value, Error>.ProducedSignal.Observer?
+
+		let lock = NSLock()
+		lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.replayLazily"
+
+		// This will go "out of scope" when the return `SignalProducer` is deallocated.
```
JaviSoto: Minor: returnED producer?
Also the struct is not really de allocated, maybe say go out of scope? And to explain better why the need of this trick maybe say that structs don't have deinit, so you're using this object that's retained by the closure (and therefore by the producer) as a way to know?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1165,4 +1164,72 @@ extension SignalProducerType {
...
+				shouldStartUnderlyingProducer = true
+			}
+			lock.unlock()
+
+			// subscribe `observer` before starting the underlying producer.
+			disposable += initializedProducer.start(observer)
+
+			if shouldStartUnderlyingProducer {
+				self.producer
+					.takeUntil(token.willDeallocSignal)
```
JaviSoto: Hmmm, this retains token??

NachoSoto: It doesn't, it retains the signal which doesn't hold a reference to the token (otherwise this wouldn't work).

JaviSoto: Oh I think I see it now

ReactiveCocoa/Swift/SignalProducer.swift
JaviSoto: Maybe also add another paragraph discouraging the use of those operator unless it's strictly necessary, explaining it may introduce caching issues, etc?

JaviSoto:
I'm +1 on this one, granted that we clearly state the disadvantages that using this may have. I think it's still really useful in many other cases though

aschuch:
Just a quick question that came to my mind when reading the docs.
 
Would you recommend to use `replayLazily` in case I would find myself needing the last value OR the last error? Or would something like `AnyProperty<Result<MyValue, NSError>>` be a better choice here?

jspahrsummers:
Couldn't you do this with something like:

```swift
SignalProducer
  .buffer(1)
  .concat(/* the real work */)
  .take(1)
```

We omitted multicasting originally because "bad patterns should be hard." I agree that there are valid use cases for it, but I think it's much better to explicitly express them with the types and the operators that RAC already has.

However, you all are now using RAC much more than I am! I'll defer to your judgment. :blush:

NachoSoto:
@jspahrsummers thanks for your input! :bow: 

I'm not sure how that would work. `SignalProducer.buffer` wouldn't buffer the `concat`ed values, only those sent to the returned `Observer`.

> but I think it's much better to explicitly express them with the types and the operators that RAC already has.

Totally, I'm going to make it even more obvious in the doc that `PropertyType` should be used in most cases, and explain in which cases this is acceptable, etc. There is however no existing way to do what this operator offers (unless you mistyped your example and there would be a simpler way by composing `buffer`?).

I'd also love to hear what @andymatuschak thinks if he has time to take a quick look 🏃

jspahrsummers:
Oops, yeah, I fucked that up a bit.

It's more like this (excuse my rustiness):

```swift
let (buffer, bufferSink) = SignalProducer.buffer(1)

buffer
  .concat(theRealWork
    .on(event: bufferSink.send))
  .take(1)
```

NachoSoto:
Right, that's a similar approach to how I implemented the operator: leveraging `SignalProducer.buffer`. The advantage to having its own operator is that we formalize the semantics and cover them with tests.

andymatuschak:
Hello yes! :phone: 

I spent some time reflecting on this issue. I think your approach here is a solid one in terms of implementing this operator. But my feelings about this operator are connecting to an ongoing internal dialogue I'm having around "just how powerful should we try to make these signals?"

We can expand the possible domain of signal expression to encompass all programs and idioms; every expansion seems valuable because it makes certain compositions possible, but every expansion introduces new terminology and expands the conceptual surface.

I'm worried that these signal producers are "doing a lot." They're "fat." Starting one might do a lot, or it might do nothing. Keeping one around might mean a really sizable memory footprint… or it might not. You can't tell by looking at the signal as a client—it's totally innocuous. Its surface hides the potential complexity of the caching and laziness hiding underneath, and of course that's the whole point! But it also means that e.g. you can't voluntarily flush the cache, or cause the side effects to be re-performed, or know ahead of time whether starting the signal would require side effects…

This is clearly a matter of taste, but I'm starting to develop a significant aversion to such "fat" signals. They're these little mysteries orbs that travel through my program, and it's often very difficult to tell what surprises they carry beneath their surfaces. In such cases, I'd rather pass around an object (a "controller" even? perish the thought!), with a name that suggests its behavior, and possibly with some more explicit API controlling its behavior. Yes, it's less composable, but until the debugging story for signal graphs becomes less nightmarish, it's more maintainable in my book. Maybe the object uses a signal as an implementation detail, but at least it decorates the signal with some more explicit semantics.

All that is to say: if you're sympathetic to my aversion, this is probably not a great pattern to encourage in the first-class set of operators.

---

More controversial post-script while we're talking style: I'm starting to believe that significantly chaining signal producers (not signals) across multiple modules is a smell: subscription effects compound, swirl around each other, get lost in a haze, and pop out the other end in an inscrutable, shapeless, semanticless blob.

The most successful signal producer uses I've seen are ones where the producer is pretty directly passed from its creation to being started.

---

Even more controversial post-script: I think laziness delivered via signal producers is overrated and leads to more bugs and complexity than benefits. Most uses of signal producers (aside from `Property`) would probably be more maintainably and idiomatically expressed eagerly.

This is a pretty common scenario in RAC-land: a table view cell's reuse causes a signal producer to be started causing some effect to take place halfway across the program. This is a nightmare scenario in my book. Tree-like control flow, please~~~

kastiglione:
Thanks for the write up Andy!

mdiep:
@andymatuschak Your ideas are intriguing to me and I wish to subscribe to your newsletter.

NachoSoto:
@andymatuschak thanks a lot for taking your time in writing that! As we've spoken in person, I'm excited to continue exploring patterns and paradigms. RAC 4 is a step forward, but I agree that we're still following a road, and not at all at the end of the path!

Most of what you point out, however, are intrinsic characteristics of `SignalProducer`, not this operator, right?

> it's totally innocuous. Its surface hides the potential complexity of the caching and laziness hiding underneath, and of course that's the whole point!

That's the "downside" of `SignalProducer`s: they abstract (i.e., hide), side effects. 

>  But it also means that e.g. you can't voluntarily flush the cache, or cause the side effects to be re-performed, or know ahead of time whether starting the signal producer would require side effects this time…

That is true. This is the same in other `Rx` implementations though. The benefit of that, however, is that we can treat this multicasted operator like a regular `SignalProducer`. We could introduce a new type to enable those things (and maybe we should one day!), but for the moment I'd say that it's more complexity than we need. Certainly this is the simplest version of multicasting, and it's possible that there are use cases for a more fine-controlled version. I'm just trying to solve the simplest case :)

> All that is to say: if you're sympathetic to my aversion, this is probably not a great pattern to encourage in the first-class set of operators.

Thanks! That's exactly the kind of feedback I was looking for. I want to make sure we really want to add this (what I argue is that it's a needed operator, but that doesn't necessarily mean that we must include it). However, I don't think any of the reasons you exposed are specific to this operator, but rather inherent features of `SignalProducer`s in general.

I'd love to continue this more "philosophical" debate for future versions of RAC, but in the interest of releasing a stable, production ready, and completely useful version of Reactive Cocoa 4, I'd love it if we could focus on that goal, and discuss whether this operator makes sense **today** in the context of RAC 4. Once we get there, I'd love to explore deeper your *tree-like* idea for future versions of RAC, though.

andymatuschak:
Hm! I think I disagree that this is all intrinsic to `SignalProducer`. The problem varies! If I can rephrase/summarize, this is my thinking:

* `SignalProducer`s can hide a lot of semantics under their surface.
* It's worse when they hide more complex semantics.
* `replayLazily` adds pretty complex semantics.
* Maybe we want to encourage people to make "simpler" signal producers by leaving this one out.

I don't feel all that strongly about it, though.

NachoSoto:
I see what you're saying, but:
> `replayLazily` adds pretty complex semantics.

This operator doesn't actually add **new** semantics to RAC 4: it's exactly the same semantics as `PropertyType.producer` and `SignalProducer.buffer`. It's just a 3rd way to create such producer: from an existing one, rather than creating a new one (with `SignalProducer.buffer`) or getting it from a `PropertyType`.

andymatuschak:
Ah, I was unclear; I'm not saying it adds new semantics to RAC 4. I'm saying that by composing a signal producer with `replayLazily`, you add pretty complex semantics _to that signal producer_, though the operator looks innocuous.

But no, it's not totally out of line with the rest of RAC 4. I guess I just think of it as emphasizing the bad parts? Anyway, please don't consider these thoughts blocking!

NachoSoto:
> I guess I just think of it as emphasizing the bad parts? 

Yeah, for sure! That's what I worry too.

One of the ways that RAC differs from Rx is that it tries to stay away from the *easy* and instead enforce good patterns and practices, which I love. Having used both for a while, I think there's no absolute truth, though, and this is just a balance we have to find. This is my attempt to bringing RAC closer to "the real world needs", even if it means sacrificing some of its purity. In my opinion this is something we need to do very carefully to keep RAC great, but I think this is one case where, by not having this operator, we're artificially making the framework less useful in real-world scenarios compared to Rx.

kastiglione:
@andymatuschak Would you mind explaining what you mean by "tree-like control flow"? It's the one piece of your write up that I don't think I understand.

andymatuschak:
I'll try!

The difference between trees and graphs is that each node in a tree has at most one parent, right? So one of the important differences between value types and reference types, for instance, is that the former _always_ form trees, while the latter often form graphs (but sometimes form trees).

That's a visual way of thinking about a lot of the things we like about value types. There's lots of different graphs we might draw. If we consider control flow: code's only getting called from one other component. If we consider dependencies: no other element of the system can depend on the state of children in a tree. If we consider dataflow: there can be no confusion about where truth lies. It's nice if these graphs look like each other.

Of course, in interactive systems, you can't always have unidirectional flow. For instance, a button needs to propagate tap events backwards. But you can structure your objects in trees so rumors still flow backwards along the tree. From [my 2014 WWDC talk](https://developer.apple.com/videos/play/wwdc2014-229/) with Colin:

screen shot 2016-01-09 at 8 20 34 am

The advantage of this structure is that effects stay local. A parent can receive a message and direct it back down a different branch of the tree, or pass it up, or swallow it, but it stays in control of the effects in its domain.

Backing up to my observation about signals: they often create diagonal lines in these graphs. You expose some signal as a property, pass it somewhere else, and now effects stop being local. You get this:

screen shot 2016-01-09 at 8 30 36 am

That's not _always_ a problem. It's just a heuristic, a smell, a factor that contributes to the complexity of program execution.

NachoSoto:
Going back to topic, I want to give a real example from my app that's currently impossible (well, it's [possible by misusing `PropertyType`](https://github.com/NachoSoto/AsyncImageView/blob/master/AsyncImageView/Renderers/MulticastedRenderer.swift), but that's what I explain in this PR: there should be a simpler way) to accomplish without this operator:
tv
Notice the background in each of the cells on the left side, as well as the app background. The background of the app is derived from the selected tournament, and the background of that cell is composited from the same image (by adding a gradient). This is the source image:
london
The download for this image is represented with a simple `SignalProducer`. Then, to accomplish what I described I create 2 other `SignalProducer`s that compose the first one to perform the image processing. Because I don't want each of those 2 operators to download the image, **I need multicasting**.

This is just **one example**, but I'm sure there are many more out there. We use several multicasting operators in *Khan Academy*'s Android app (`RxJava`) as well, so I really think it would considerably increase the value of `ReactiveCocoa 4`, despite the downsides that we've talked about.

liscio:
I feel that multicasting (or some form of "one producer to many consumers that want the same result at the same time") is a huge missing aspect in RAC right now. 

I also have long-running `SignalProducer`s for which I want to push results to many consumers without invoking the side-effects again. For instance, I build a lot of long-running processing operations for which the results can be utilized by other `SignalProducer` instances. But I don't want to necessarily re-run them to get the result in each sub-task that depends on them.

Many times it's been recommended that `startWithSignal` is a good way to achieve this one-to-many feeding of results to multiple consumers, but it's awkward as heck in practice—even worse to those new to the framework.

So I give this a :+1: however I would personally like to see it be called simply, `buffer`. Intuitively speaking, I think it makes sense—"give me a buffer of *this signal*" as opposed to a buffer that is to be manually driven. Frankly, I actually tried calling this "imaginary method" in hopes that it existed a few times…

kastiglione:
Back in RAC v1 there was `+[RACSignal let:]` ([source](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/4ec0e490d23acefe8568cd144b5b4016626fe59b/ReactiveCocoaFramework/ReactiveCocoa/RACSignal%2BOperations.m#L1016-L1036)). Would something like that work for the uses cases you need `replayLazily`, @NachoSoto?

e.g.:

```swift
producer.share { sharedProducer in
    // subscribe to `sharedProducer` multiple times but knowing the effects happen once,
    // and that the resulting signals are identical.
}
```

The advantage ostensibly being that the shared signal doesn't escape this scope.

NachoSoto:
I wasn't familiar with `+[RACSignal let:]`, but that does seem like it has similar semantics, except for the "not escaping current scope" part. That certainly seems useful, but more limited, and wouldn't be viable for the use cases I outlined above since subscription needs to happen elsewhere.

mdiep:
We've definitely used `replayLazily` from RAC 2 in GitHub Desktop quite a bit. Unfortunately, I haven't been able to do as much in Swift yet, so I've been hesitant to speak up. But based on that usage I would expect it be a valuable addition here for the use cases outlined above.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1165,4 +1164,72 @@ extension SignalProducerType {
...
+private extension NSObject {
+	var willDeallocSignal: SignalProducer<(), NoError> {
+		return self
+			.rac_willDeallocSignal()
+			.toSignalProducer()
+			.map { _ in () }
+			.mapError { error in
+				fatalError("Unexpected error: \(error)")
+				()
+		}
```
mdiep: Indentation is off here.

NachoSoto: This is Xcode's default 🙄. I normally change this, but I missed this one, thanks! I'll fix.

NachoSoto:
So what are everyone's thoughts regarding including/not including this after all the provided context?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1165,4 +1164,72 @@ extension SignalProducerType {
...
+
+private extension NSObject {
+	var willDeallocSignal: SignalProducer<(), NoError> {
+		return self
+			.rac_willDeallocSignal()
+			.toSignalProducer()
+			.map { _ in () }
+			.mapError { error in
+				fatalError("Unexpected error: \(error)")
+				()
```
mdiep: Is the `()` needed here to provide a return value? Could it be done in the type signature for the block instead?

NachoSoto: I'll try.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1165,4 +1164,72 @@ extension SignalProducerType {
...
+			let initializedProducer: SignalProducer<Value, Error>
+			let initializedObserver: SignalProducer<Value, Error>.ProducedSignal.Observer
+			let shouldStartUnderlyingProducer: Bool
+
+			lock.lock()
+			if let producer = producer, producerObserver = producerObserver {
+				(initializedProducer, initializedObserver) = (producer, producerObserver)
+				shouldStartUnderlyingProducer = false
+			} else {
+				let (producerTemp, observerTemp) = SignalProducer<Value, Error>.buffer(capacity)
```
mdiep: Can we get rid of these temporary variables and assign directly to `producer` and `producerObserver`?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1861,6 +1861,244 @@ class SignalProducerSpec: QuickSpec {
...
+		describe("replayLazily") {
+			var producer: SignalProducer<Int, TestError>!
+			var observer: SignalProducer<Int, TestError>.ProducedSignal.Observer!
+
+			var replayedProducer: SignalProducer<Int, TestError>!
+
+			beforeEach {
+				let (producerTemp, observerTemp) = SignalProducer<Int, TestError>.buffer(0)
+				producer = producerTemp
+				observer = observerTemp
```
mdiep: Same question here: can the temp variables be removed

mdiep:
I left a few minor notes. It also looks like a couple comments on the documentation may not be resolved.

@ReactiveCocoa/ReactiveCocoa Does anyone object to merging this? I think we're in agreement that we wish it wasn't necessary. But it seems like a needed addition for RAC 4.

andymatuschak:
(to be clear: I don’t seriously object! I’m just too hesitant to be the merger)

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: No, because `producer` and `producerObserver` are `Optional`s, so Swift can't infer it :P

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
NachoSoto: Same reason :P

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Nah this doesn't work :/

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Fixed.

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Done.

NachoSoto:
> I left a few minor notes. It also looks like a couple comments on the documentation may not be resolved.

Thanks @mdiep! I just addressed everything.

> I think we're in agreement that we wish it wasn't necessary. But it seems like a needed addition for `RAC 4`.

I think that's a great way to put it! :)

Let me know what you think of the new docstring. I tried to list the alternatives, and be very explicit about the downsides of this operator.

ReactiveCocoa/Swift/SignalProducer.swift
mdiep: :weary:

mdiep:
Looks good to me. :+1:

I'll merge tomorrow unless someone objects before then.

jspahrsummers:
just kidding, merge it, whatever

mdiep:
@jspahrsummers I think you've mistaken this project for [this one](https://rxcpp.codeplex.com).

mdiep:
:dancers:

NachoSoto:
:tada: Thanks everyone! :D 
yes

PR #2641, opened 2016-01-03
Require that Disposable is a Class

Adlai-Holler:
I think conceptually this makes good sense, and it also enables users to treat disposables as `AnyObject` which opens up interesting possibilities like making a weak set of disposables using `NSMapTable.weakToStrongObjectsMapTable: Disposable -> NSNull` etc.


NachoSoto:
I think this makes sense, I can't imagine a `DisposableType` with value semantics.

NachoSoto:
I'm going to merge since it's certainly valid for our current `Disposable`s. We can revert if needed, but I think this is the right way to go.

*(Failing test is unrelated)*

PR #2646, opened 2016-01-07
PropertyType Composition

natestedman:
This adds these functions and extensions for composing multiple `PropertyType` values:

- `PropertyType.map(:)`
- `PropertyType.combineLatestWith(:)`
- `PropertyType.zipWith(:)`
- `combineLatest(...)`
- `zip(...)`

I'm not sure how useful `zip` is with properties, but I simulate `combineLatest` and `map` with `initialValue:producer:` a lot, and thought it would be useful to have as a real thing. Unlike just [making PropertyType a SignalProducerType](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2543), which would allow those operators to be used instead, composing properties directly guarantees that there is a `value` when the composed property is created.

This means that some of the `SignalProducer` operators can't be implemented on `PropertyType` - for example, `filter` won't work, because if the initial value is filtered out, the property cannot be created. So, for now, I stuck to `map` and the combining operators.

I haven't added tests for the `combineLatest` and `zip` functions yet, just the extensions. If this is something that people are interested in merging, I will add those and rebase.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -320,3 +320,82 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
 public func <~ <Destination: MutablePropertyType, Source: PropertyType where Source.Value == Destination.Value>(destinationProperty: Destination, sourceProperty: Source) -> Disposable {
 	return destinationProperty <~ sourceProperty.producer
 }
+
+/// Combines the values of all the given properties, in the manner described by
+/// `combineLatestWith`.
+@warn_unused_result
+public func combineLatest<A: PropertyType, B: PropertyType>(a: A, _ b: B) -> AnyProperty<(A.Value, B.Value)> {
+	return a.combineLatestWith(b)
+}
+
+/// Combines the values of all the given properties, in the manner described by
+/// `combineLatestWith`.
+@warn_unused_result
+public func combineLatest<A: PropertyType, B: PropertyType, C: PropertyType>(a: A, _ b: B, _ c: C) -> AnyProperty<(A.Value, B.Value, C.Value)> {
+	return combineLatest(a, b)
+		.combineLatestWith(c)
+		.map(repack)
+}
+
+/// Combines the values of all the given properties, in the manner described by
+/// `combineLatestWith`.
+@warn_unused_result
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -15,6 +15,37 @@ public protocol PropertyType {
...
+		)
+	}
+
+	/// Zips elements of two properties into pairs. The elements of any Nth pair
+	/// are the Nth elements of the two input properties.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func zipWith<Other: PropertyType>(otherProperty: Other) -> AnyProperty<(Value, Other.Value)> {
+		return AnyProperty(
+			initialValue: (value, otherProperty.value),
+			producer: producer.zipWith(otherProperty.producer).skip(1)
```
NachoSoto: This logic of `skip(1)` specifically is repeated in all the operators. But in general, I think all of these could be defined in terms of a `lift` function that takes a `PropertyType` and a `SignalProducer<T, NoError> -> SignalProducer<R, NoError>` to create the derived property. What do you think?

NachoSoto:
I like this :+1:

robgrant:
I just wrote my own implementation of `map` for `PropertyType` so :+1: to this.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -15,6 +15,37 @@ public protocol PropertyType {
 	var signal: Signal<Value, NoError> { get }
 }
 
+extension PropertyType {
+	/// Maps each value in the property to a new value.
+	@warn_unused_result
+	public func map<U>(transform: Value -> U) -> AnyProperty<U> {
+		return AnyProperty(
```
NachoSoto: Cool. I still think we should extract this to a `lift` function to avoid having each operator repeat the `return AnyProperty(...)` dance.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -15,6 +15,37 @@ public protocol PropertyType {
...
+			initialValue: transform(value),
+			signal: signal.map(transform)
```
andersio: `MutableProperty.producer` should be more appropriate here, shouldn't it? Since there is no umbrella lock here (and you can't access `withValue(action:)` here either), the signal could have pushed a new value before the `AnyProperty` eventually subscribed to the signal.

andersio: I suppose you pick the `signal` overload because replaying twice the current value doesn't seem a good thing. As an alternative, we may consider moving `withValue(action:)` from `MutablePropertyType` to `PropertyType`, so these compositing methods can be synchronised as needed.

PR #2647, opened 2016-01-07
Added latestValues

gkaimakas:
Added a function that returns a new producer that emits the latest N values from the original producer

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -377,7 +377,7 @@ extension SignalProducerType {
 	/// removing the earliest value. If intermediateResults is false then the new producer will begin emitting
 	/// only when it has reached capacity.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
-	public func latestValues(capacity:Int, intermediateResults: Bool = true) -> SignalProducer<[Value], Error> {
+	public func latestValues(capacity:Int, intermediateResults: Bool = false) -> SignalProducer<[Value], Error> {
 		var array: [Value] = []
 		return self.map {
 			value in
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -370,6 +370,30 @@ extension SignalProducerType {
 	public func startWithInterrupted(interrupted: () -> ()) -> Disposable {
 		return start(Observer(interrupted: interrupted))
 	}
+	
+	/// Creates a new producer that emits an array that contains the latest N values that were emitted 
+	/// by the original producer as specified in 'capacity'. If intermediateResults is true the
+	/// the new producer will emit an array with each new value until it reaches capacity and begins 
+	/// removing the earliest value. If intermediateResults is false then the new producer will begin emitting
+	/// only when it has reached capacity.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func latestValues(capacity:Int, intermediateResults: Bool = true) -> SignalProducer<[Value], Error> {
+		var array: [Value] = []
+		return self.map {
+			value in
+
+			array.append(value)
+
+			if array.count >= capacity {
+				array.removeFirst(array.count - capacity)
+			}
+
+			return array
+		}
+			.filter {
+				(intermediateResults) ? true : $0.count == capacity
+		}
+	}
 
 	/// Lifts an unary Signal operator to operate upon SignalProducers instead.
 	///


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -370,6 +370,30 @@ extension SignalProducerType {
...
+	
+	/// Creates a new producer that emits an array that contains the latest N values that were emitted 
+	/// by the original producer as specified in 'capacity'. If intermediateResults is true the
+	/// the new producer will emit an array with each new value until it reaches capacity and begins 
+	/// removing the earliest value. If intermediateResults is false then the new producer will begin emitting
+	/// only when it has reached capacity.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func latestValues(capacity:Int, intermediateResults: Bool = true) -> SignalProducer<[Value], Error> {
+		var array: [Value] = []
+		return self.map {
```
NachoSoto: `map` shouldn't be a stateful function.

NachoSoto:
Thanks! :star: but this is essentially the same as #2639. We're having a conversation there about the implications of such operator :)

gkaimakas:
I read the conversation there and can't see how this is the same as the topic there. What I did was combine some already available functions to be able to get the latest values that were emitted by the producer.

I did not want to multicast anything, just to have access on these values at once. I was trying to apply FFT to some data from a peripheral and that't how I came up with that function

NachoSoto:
Ooooh my bad. I didn't realize the returned type changed to `SignalProducer<[Value], Error>`.

In that case, this operator should be implemented with `scan` instead.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -370,6 +370,30 @@ extension SignalProducerType {
...
 		return start(Observer(interrupted: interrupted))
 	}
+	
+	/// Creates a new producer that emits an array that contains the latest N values that were emitted 
+	/// by the original producer as specified in 'capacity'. If intermediateResults is true the
+	/// the new producer will emit an array with each new value until it reaches capacity and begins 
+	/// removing the earliest value. If intermediateResults is false then the new producer will begin emitting
+	/// only when it has reached capacity.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func latestValues(capacity:Int, intermediateResults: Bool = true) -> SignalProducer<[Value], Error> {
```
NachoSoto: Having a method behavior depend on a boolean parameter is a smell: it tells you that the method is actually doing 2 completely different things!
Instead, I would settle on one behavior, that allows users to get the other one by means of using another operator.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -370,6 +370,30 @@ extension SignalProducerType {
...
 		return start(Observer(interrupted: interrupted))
 	}
+	
+	/// Creates a new producer that emits an array that contains the latest N values that were emitted 
+	/// by the original producer as specified in 'capacity'. If intermediateResults is true the
+	/// the new producer will emit an array with each new value until it reaches capacity and begins 
+	/// removing the earliest value. If intermediateResults is false then the new producer will begin emitting
+	/// only when it has reached capacity.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func latestValues(capacity:Int, intermediateResults: Bool = false) -> SignalProducer<[Value], Error> {
```
NachoSoto: Having a method behavior depend on a boolean parameter is a smell: it tells you that the method is actually doing 2 completely different things!
Instead, I would settle on one behavior, that allows users to get the other one by means of using another operator.

NachoSoto:
My apologies @gkaimakas, I misread the operator! I see how this can be useful. Since this can be accomplished using `scan` I don't think we should add it to the `RAC 4` API, though. We're trying to keep the API surface to a minimum, at least until we can release a stable version. I'd recommend opening a PR to [Rex](https://github.com/neilpa/Rex) though, which has additions to the RAC API :)

gkaimakas:
I think that scan can't provide the same functionality. Scan only provides the previous value and the result is a tuple. 

I tried using scan and all I got was some convoluted mess that I couldn't figure out how to flatten in an array

ReactiveCocoa/Swift/SignalProducer.swift
gkaimakas: You are probably right on this one

NachoSoto:
For sure, `scan` **is** different, but what I'm saying is that these semantics can be achieved in terms of `scan` and `map`.

gkaimakas:
The exact semantics cannot be achieved with the use of `scan` and `map` of either a combination of the two. Even if you could overcome the complexities that arise from the tuples and create an array of values that array's size will be a power of two only. And the code that will be required would be to complex.

PR #2649, opened 2016-01-09
Fix typo on BasicOperators.md

JuIioCesar:


NachoSoto:
Your change looks a lot better to me too (as a non-native speaker), but turns out the existing version is correct American English according to the APA: http://blog.apastyle.org/apastyle/2011/08/punctuating-around-quotation-marks.html

Thanks for the PR though!

bencochran:
I, for one, fully support a [change in the American style](http://www.slate.com/articles/life/the_good_word/2011/05/the_rise_of_logical_punctuation.html) (though I agree that this is probably best left as-is for now)

NachoSoto:
Me too, I find "this kind of stuff," very silly :/

kastiglione:
> the existing version is correct American English

"correct" lol

Personally, I could care less what various sites or authorities say. Let's go with changes like this if we all like them.

JuIioCesar:
+1

NachoSoto:
Agreed, and merged! :)

Thanks everyone.

PR #2650, opened 2016-01-10
Update Documentation for QueueScheduler(:,name:)

Adlai-Holler:
This method's behavior does not currently match its documentation. I believe a better fix would be to change the behavior to match the documentation i.e. target the global concurrent queue. If the community agrees, just say so and I'll put in a PR for that.

### DIFF
--- a/ReactiveCocoa/Swift/Scheduler.swift
+++ b/ReactiveCocoa/Swift/Scheduler.swift
@@ -118,8 +118,8 @@ public final class QueueScheduler: DateSchedulerType {
 		return NSDate()
 	}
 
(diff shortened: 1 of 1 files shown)


jspahrsummers:
Schedulers are serial by design. Allowing concurrent queues without a private serial queue in between would violate this assumption.

liscio:
Yeah, the behavior has always been serial, and the existing "target the global queue" documentation is my bad—I didn't update it when the API was cleaned up. 

Even worse: I fixed the "old" method signature to be more specific while ignoring the new API documentation. No :cookie: for me.

Adlai-Holler:
So is this pull request ready-to-merge then?

PR #2659, opened 2016-01-13
Property enhancements

andersio:
##### Non-breaking Changes (For API consumers)

1. `withValue(action:)` is now available on all `PropertyType` implementations.

2. `modify(action:)` and `swap(value:)` are now available on all `MutablePropertyType` implementations.

##### Breaking Changes (affecting only `PropertyType` and `MutablePropertyType` implementers)

1. New `PropertyType` requirement: `withValue(action:)`.
Meanwhile, `value` and `signal` are no longer required. Default implementations are provided for them instead.

2. New `MutablePropertyType` requirement: `modify(action:)`
Meanwhile, `value` is no longer required. Default implementations are provided for `value` and also `swap(value:)` instead.

-
This PR was originated from the attempt to introduce `withValue` to `PropertyType`. The only obstacle is `AnyProperty` given the current limitations of Swift's generic system. Specifically, since `@noescape` and `rethrows` are not supported in type identifier, closures cannot be used in this case.

Consequently, this PR refactored `AnyProperty` to use type erased wrappers for serving its read-only view. But more importantly, since now `withValue` is available on PropertyType, this means the atomicity of `withValue` - if enforced by an implementation, e.g. MutableProperty - is also available in the wrapping `AnyProperty` and during property composition.

For example, for `map` (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2646), we can now instead do:
```
func map<U>(transform: Value -> U) -> AnyProperty<U> {
    return withValue { value in
        return AnyProperty(initialValue: transform(value),
            signal: signal.map(transform))
    }
}
```
and semantically it is always consistent across threads, since (if `self` is a MutableProperty) the closure that constructs the `AnyProperty` is executed before the recursive lock is released.

Note that atomicity is defined to be implementation dependent, since there is probably no way to enforce atomicity in `DynamicProperty`.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -4,42 +4,44 @@ import Foundation
 public protocol PropertyType {
 	typealias Value
 
-	/// The current value of the property.
-	var value: Value { get }
-
 	/// A producer for Signals that will send the property's current value,
 	/// followed by all changes over time.
 	var producer: SignalProducer<Value, NoError> { get }
 
+	/// Synchronously execute the action with the current value of the property.
+	///
+	/// - Note: Implementations are encouraged to enforce atomicity.
+	func withValue<Result>(@noescape action: (Value) throws -> Result) rethrows -> Result
+}
+
+/// Default implementations of PropertyType.
+extension PropertyType {
+	/// The current value of the property.
+	public var value: Value {
+		return withValue { $0 }
+	}
+
 	/// A signal that will send the property's changes over time.
-	var signal: Signal<Value, NoError> { get }
+	public var signal: Signal<Value, NoError> {
+		var extractedSignal: Signal<Value, NoError>!
+		producer.startWithSignal { signal, _ in
+			extractedSignal = signal
+		}
+		return extractedSignal
+	}
 }
 
 /// A read-only property that allows observation of its changes.
 public struct AnyProperty<Value>: PropertyType {
-
-	private let _value: () -> Value
-	private let _producer: () -> SignalProducer<Value, NoError>
-	private let _signal: () -> Signal<Value, NoError>
-
-
-	public var value: Value {
-		return _value()
-	}
+	private let box: _AnyPropertyBoxBase<Value>
 
 	public var producer: SignalProducer<Value, NoError> {
-		return _producer()
+		return box.producer
 	}
 
-	public var signal: Signal<Value, NoError> {
-		return _signal()
-	}
-	
 	/// Initializes a property as a read-only view of the given property.
 	public init<P: PropertyType where P.Value == Value>(_ property: P) {
-		_value = { property.value }
-		_producer = { property.producer }
-		_signal = { property.signal }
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -299,3 +313,33 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
 public func <~ <Destination: MutablePropertyType, Source: PropertyType where Source.Value == Destination.Value>(destinationProperty: Destination, sourceProperty: Source) -> Disposable {
 	return destinationProperty <~ sourceProperty.producer
 }
+
+
+/// Type erased wrapper for composable PropertyType.
+private class _AnyPropertyBoxBase<Value>: PropertyType {
```
NachoSoto: This should be `final`.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -299,3 +313,33 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
...
+	var producer: SignalProducer<Value, NoError> {
+		fatalError()
```
NachoSoto: I don't like this: now we can't guarantee at compile time that this wouldn't crash. Why do you need this case class anyway?

neilpa:
I'm not sure this warrants breaking API compatibility (even if it is just for `PropertyType` implementers) given that we're at an RC point for v4.

ReactiveCocoa/Swift/Property.swift
andersio: The type erased wrapper is to workaround the lack of way to legally define a non-escaping and `rethrows` stored closure in Swift 2's type system, as far as I tried. This prohibits the `withValue` method from being wrapped in a closure. As a matter of fact, this pattern is used in the Swift STL to workaround Swift 2's lacking existential. (https://github.com/apple/swift/blob/master/stdlib/public/core/ExistentialCollection.swift.gyb).

There are alternatives that can be considered though, e.g.
1. an escaping, non-throwing version of `withValue` and `modify`, specifically for property composition
2. a private `AnyProperty` initialiser that assumes a buffering producer.
2. wait for Swift 3.0 which hopefully comes with partially/fully conformed existentials. ;-)

andersio:
A little background on this PR. It is born to make
```
func map<U>(transform: Value -> U) -> AnyProperty<U> {
        return AnyProperty(initialValue: transform(value),
            signal: signal.map(transform))
}
```
consistently valid.

It is logical to use a signal which does NOT replay the current value. But there is a race condition between `PropertyType.value` and `PropertyType.signal` in the case of the wrapped property being a MutableProperty, and `PropertyType` does not expose any kind of lock-with-action, aka `withValue`. Therefore, this PR aims to provide this functionality.

But as said in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2646#discussion_r49162957, it can be replaced with `producer` and still be consistent after all, since the current value would always be replayed by the producer. So this PR might not be important enough to get merged, I have to admit.

PR #2660, opened 2016-01-13
[WIP] User Interface Recipes

liscio:
Please note that this document is *very much* a work in progress. Some things that I've left to do:

- [ ] Remove the dependency on https://github.com/neilpa/Rex
- [ ] Produce an actual example project. (Much of the below code was typed "raw" into my text editor, and likely won't compile.)
- [ ] Get feedback from people who have been doing this much longer
- [ ] Fix the TODOs

There really isn't any sort of "canonical document" for creating UI code using ReactiveCocoa, and that's probably due to the fact that it's difficult to find a single approach or "recipe" that applies to all scenarios. In my own work on [Capo](http://capoapp.com) and [FuzzMeasure](http://fuzzmeasure.com) over the past 6 months I have found so many different scenarios and examples that forced me to repeatedly question what I was doing. 

Just when I think that I have an approach figured out, I find another handful of new challenges when I move to another part of the UI to work on. For instance, when I built a lot of iOS UI code I found that it was very easy to build static UIs for `UITableViewCell`s and use `ViewModel`s to ensure that I encapsulate the presentation behavior.

Once I moved to editable controls for the [Neptune](http://capoapp.com/neptune) UI I faced a whole new set of challenges when trying to get values propagated from my ViewModel down into my audio engine, and encapsulating a live-updating visualization (that wasn't needlessly updating the screen while playback wasn't active, also!)

I feel that this collection of "Recipes" for various UI patterns would be extremely helpful to people that are new to ReactiveCocoa, and folks that might be looking for ideas when discovering new challenges that they haven't faced before.

### DIFF
--- a/Documentation/UserInterfaceRecipes.md
+++ b/Documentation/UserInterfaceRecipes.md
@@ -1,0 +1,128 @@
+# User Interface Recipes
+
+### ViewModel / Presenter / ViewData Objects
+
+Regardless of what you call them, or how much responsibility they should (or should not) have in your project, generally these objects contain a 1:1 mapping of properties to the controls on your screen. The type of properties you use (`MutableProperty`, `ConstantProperty`, or `AnyProperty`) depends on your specific needs.
+
+For example, consider a UI that displays both a slider and a text field that represent a magnitude. The text field specifies its values in dBFS (decibels, full scale). A common approach is to expose these values in your ViewModel using only a single `MutableProperty`:
+
+``` swift
+final class VolumeViewModel {
+	let volume: MutableProperty<Double>
+	
+	init(initialVolume volume: Double) {
+		volume = MutableProperty(volume)
+	}
+}
+```
+
+Now, consider the accompanying view controller:
+
+``` swift
+import Foundation
+import AppKit
+import ReactiveCocoa
+import Rex
+
+final class VolumeViewController : NSViewController {
+	@IBOutlet var slider: NSSlider!
+	@IBOutlet var textField: NSTextField!
+	
+	let viewModel: MutableProperty<VolumeViewModel?>(nil)
+	
+	func viewDidLoad() {
+		super.viewDidLoad()
+		
+		let parseVolume: Double -> String = { /* ... */ }
+		let formatVolume: String -> Double = { /* ... */ }
+		
+		// Create the source signal producers
+		let volumeProducer = viewModel.flatMap(.Latest) { $0?.volume.producer ?? .empty }
+		let decibelStringProducer = volumeProducer.map { 20 * log10( $0 ) }.map(formatVolume)
+		
+		// Populate the controls
+		volumeProducer.observeOn(UIScheduler()).startWithNext { [slider] in slider.value = $0 }
+		decibelStringProducer.observeOn(UIScheduler()).startWithNext { [textField] in textField.stringValue = $0 }
+		
+		// Update the underlying ViewModel on UI changes
+		textField.rex_stringValues.map(parseVolume).startwithNext { [model] in model?.value.volume.value = $0 }
+		slider.rex_doubleValues.startWithNext { [model] in model?.value.volume.value = $0 }
+	}
+}
+```
+
+#### `SignalProducer`s of `SignalProducer`s
+
+You'll notice that I declare the `viewModel` as a `MutableProperty` rather than just the `VolumeViewModel`, and its individual producers must be "unwrapped":
+
+``` swift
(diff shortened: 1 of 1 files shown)


Documentation/UserInterfaceRecipes.md
```diff
@@ -0,0 +1,128 @@
+# User Interface Recipes
+
+### ViewModel / Presenter / ViewData Objects
+
+Regardless of what you call them, or how much responsibility they should (or should not) have in your project, generally these objects contain a 1:1 mapping of properties to the controls on your screen. The type of properties you use (`MutableProperty`, `ConstantProperty`, or `AnyProperty`) depends on your specific needs.
+
+For example, consider a UI that displays both a slider and a text field that represent a magnitude. The text field specifies its values in dBFS (decibels, full scale). A common approach is to expose these values in your ViewModel using only a single `MutableProperty`:
+
+``` swift
+final class VolumeViewModel {
```
NachoSoto: I'm not keen at all on the word "view model". The term is overloaded and it has too much baggage.
To avoid confusion, how about just `VolumeData`, or if you want, `VolumeViewData`?

NachoSoto:
Awesome idea! :heart_eyes: Thanks for getting this started. I'll take a closer look in a few days :)

Documentation/UserInterfaceRecipes.md
liscio: Yep, I dislike the term as well. 

I've switched to `ViewData` as the default naming for much of my own code and have even kicked around the use of `Presenter` in some spots. I only used `ViewModel` due to its recognition from other frameworks and the "MVVM Craze" though I suppose that will die down over time.

(The joys of a one-man coding team—schizophrenic/inconsistent naming conventions!)

I tend to agree with the overall sentiment from [this article](http://khanlou.com/2015/12/mvvm-is-not-very-good/) so I'd be glad to help spread a better approach to figuring this all out.

Any war stories you and others can share in terms of usage/architecture will be helpful.

RuiAAPeres:
This looks really nice @liscio. ❤️ I also started [this](https://github.com/RuiAAPeres/RACNest).

Documentation/UserInterfaceRecipes.md
```diff
@@ -0,0 +1,136 @@
+# User Interface Recipes
+
+### ViewData (aka ViewModel, Presenter) Objects
```
mdiep: I think we should default to `ViewModel` here since that's been the preferred RAC terminology. (See also: https://github.com/ReactiveCocoa/ReactiveViewModel)

Documentation/UserInterfaceRecipes.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### "Too Much Behavior" in the `VolumeViewController`?
+
+The above `ViewController`, while still compact, still carries an awful lot of logic with it. This isn't immediately obvious with a simple model as above, but you'll notice that `parseVolume` and `formatVolume` exist inside the view controller as well. To keep your ViewControllers "as dumb as possible" and achieve a more testable UI, you can push these responsibilities into the view data class and maintain that 1:1 control mapping:
+
+``` swift
+final class VolumeViewData {
+	// "Outlet" Property / SignalProducer
+	let volume: MutableProperty<Double>
+	let volumeStringInDecibels: SignalProducer<Double, NoError>
```
mdiep: You should be able to make `volumeStringInDecibels` a property as well.

mdiep:
@liscio Is this something you're still interested in pursuing?

liscio:
@mdiep Certainly, as time permits. 

Thanks for all the feedback & suggestions so far!

Documentation/UserInterfaceRecipes.md
liscio: It sounds like there's a `ViewData` vs `ViewModel` (and `Presenter`) discussion that needs to be resolved at some point.

And is that ReactiveViewModel repo even considered useful anymore? Especially considering that the Objective-C API will be disappearing soon in the RAC 5 release?

liscio:
Sorry this has sat stale for so long. There are a few things that are holding it up:

1. https://github.com/neilpa/Rex being merged into the RAC organization. This will make it easier to pick up some of the UI bits and pieces that Rex exposes. Of course this doesn't stop me from writing about the approach, but it probably shouldn't be merged until that happens.

2. My open PR in Rex with "bindings" code for UI work. As I was writing the documentation above I was in the middle of coding another project that used a different approach than what I describe here. I don't want to jump the gun and prescribe anything approaching a "right way" to build UI until I feel confident that I have an approach that works. :)

PR #2667, opened 2016-01-19
Use Dispatch Semaphores in Atomic Instead of Spin Locks

### LINKED ISSUE #2619: Atomic: SPINLOCK is not safe in iOS
https://twitter.com/steipete/status/676851647042203648

I'll refrain from making any comments (I'll just say... if it's illegal on iOS, why is it a public API???)

Apparently [`pthread` mutexes are faster now](https://twitter.com/catfish_man/status/676852111527706624). We should create a benchmark to compare, and try to convert `Atomic` to this (also `RACCompoundDisposable` and `RACSerialDisposable`?)

Adlai-Holler:
Resolves #2619. I chose dispatch semaphores based on the discussion in that thread, plus the following tests I did with 16 dispatch_asyncs on global concurrent queue, 100000 small arithmetic operations per block in iOS 9.2 simulator:

- OSSpinLock average: 0.153, relative standard deviation: 1.338%
- dispatch semaphore: average: 0.547, relative standard deviation: 3.032%
- NSLock: average: 4.484, relative standard deviation: 1.039%
- pthread mutex: average: 4.550, relative standard deviation: 3.291%,

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -10,7 +10,7 @@ import Foundation
 
 /// An atomic variable.
 public final class Atomic<Value> {
-	private var spinLock = OS_SPINLOCK_INIT
+	private let semaphore = dispatch_semaphore_create(1)
 	private var _value: Value
 	
 	/// Atomically gets or sets the value of the variable.
@@ -30,11 +30,11 @@ public final class Atomic<Value> {
 	}
 	
 	private func lock() {
-		OSSpinLockLock(&spinLock)
+		dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER)
 	}
 	
 	private func unlock() {
-		OSSpinLockUnlock(&spinLock)
+		dispatch_semaphore_signal(semaphore)
 	}
 	
 	/// Atomically replaces the contents of the variable.


liscio:
Thanks for attempting this. Please also include the test you created so we can run it on varied hardware, and expand upon it. Specifically I'd like to see what sort of "valley" of performance there is on physical iOS devices, as well as a beefy Mac Pro.

Also, does this "resolve" #2619? It doesn't address either of the ObjC classes that also use `OSSpinlock` (`RACCompoundDisposable` and `RACSerialDisposable`.) Either we create a new issue to separate the (sadly not as straightforward) work that needs to be done there, or prepare to re-open and correct the title if/when this is merged so those don't get lost.

Finally, if we do settle on `dispatch_semaphore` being the locking mechanism of choice here (and it _should_ suffice for RAC's `Atomic` class for a certain class of protected values) I suggest that we add documentation in the class to explain the decision.

NachoSoto:
Awesome!

@Adlai-Holler did you happen to run the same tests using the old version of `Atomic` to compare?

jspahrsummers:
I think we should prefer pthread mutexes unless we run into an actual performance problem. Although `dispatch_semaphore` should avoid the problems of a spin lock, it cannot donate priority like other mutexes.

This means that if a UI-related (high QoS) thread is waiting on the lock, and a low-priority thread is currently doing work, the kernel does not have enough information to increase the priority of the latter to make it finish more quickly.

Adlai-Holler:
Superseded by #2670 .

PR #2670, opened 2016-01-19
Replace OSSpinLock with pthread_mutex

### LINKED ISSUE #2619: Atomic: SPINLOCK is not safe in iOS

Adlai-Holler:
This supersedes #2667, resolves #2619.

Here is the performance testing I did. https://gist.github.com/Adlai-Holler/f811809e57d06a6e7a2a 16 concurrent blocks on the default-qos global queue, each running  100,000 iterations of "lock, read, increment, set, read, decrement, set, unlock".

I added assertions into `Atomic` when locking/unlocking because in optimized builds Swift assertions are not evaluated so I figure why not have the extra information available.

### DIFF
--- a/ReactiveCocoa/Objective-C/RACCompoundDisposable.m
+++ b/ReactiveCocoa/Objective-C/RACCompoundDisposable.m
@@ -78,7 +78,7 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 	return [[self alloc] initWithDisposables:disposables];
 }
 
-- (id)init {
+- (instancetype)init {
 	self = [super init];
 	if (self == nil) return nil;
 
@@ -88,7 +88,7 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 	return self;
 }
 
-- (id)initWithDisposables:(NSArray *)otherDisposables {
+- (instancetype)initWithDisposables:(NSArray *)otherDisposables {
 	self = [self init];
 	if (self == nil) return nil;
 
@@ -112,7 +112,7 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 	return self;
 }
 
-- (id)initWithBlock:(void (^)(void))block {
+- (instancetype)initWithBlock:(void (^)(void))block {
 	RACDisposable *disposable = [RACDisposable disposableWithBlock:block];
 	return [self initWithDisposables:@[ disposable ]];
 }
--- a/ReactiveCocoa/Objective-C/RACCompoundDisposable.m
+++ b/ReactiveCocoa/Objective-C/RACCompoundDisposable.m
@@ -8,7 +8,7 @@
 
 #import "RACCompoundDisposable.h"
 #import "RACCompoundDisposableProvider.h"
-#import <libkern/OSAtomic.h>
+#import <pthread/pthread.h>
 
 // The number of child disposables for which space will be reserved directly in
 // `RACCompoundDisposable`.
@@ -30,7 +30,7 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 
 @interface RACCompoundDisposable () {
 	// Used for synchronization.
-	OSSpinLock _spinLock;
+	pthread_mutex_t _mutex;
 
 	#if RACCompoundDisposableInlineCount
 	// A fast array to the first N of the receiver's disposables.
@@ -38,19 +38,19 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 	// Once this is full, `_disposables` will be created and used for additional
 	// disposables.
 	//
-	// This array should only be manipulated while _spinLock is held.
+	// This array should only be manipulated while _mutex is held.
 	RACDisposable *_inlineDisposables[RACCompoundDisposableInlineCount];
 	#endif
 
 	// Contains the receiver's disposables.
 	//
-	// This array should only be manipulated while _spinLock is held. If
+	// This array should only be manipulated while _mutex is held. If
 	// `_disposed` is YES, this may be NULL.
 	CFMutableArrayRef _disposables;
 
 	// Whether the receiver has already been disposed.
 	//
-	// This ivar should only be accessed while _spinLock is held.
+	// This ivar should only be accessed while _mutex is held.
 	BOOL _disposed;
 }
 
@@ -61,9 +61,9 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 #pragma mark Properties
 
 - (BOOL)isDisposed {
-	OSSpinLockLock(&_spinLock);
+	pthread_mutex_lock(&_mutex);
 	BOOL disposed = _disposed;
-	OSSpinLockUnlock(&_spinLock);
+	pthread_mutex_unlock(&_mutex);
 
 	return disposed;
 }
@@ -76,6 +76,16 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 
 + (instancetype)compoundDisposableWithDisposables:(NSArray *)disposables {
 	return [[self alloc] initWithDisposables:disposables];
+}
+
+- (id)init {
+	self = [super init];
+	if (self == nil) return nil;
+
+	int result = pthread_mutex_init(&_mutex, NULL);
+	NSCAssert(0 == result, @"Failed to initialize mutex with error %d.", result);
+
+	return self;
 }
 
 - (id)initWithDisposables:(NSArray *)otherDisposables {
@@ -118,6 +128,9 @@ static CFMutableArrayRef RACCreateDisposablesArray(void) {
 		CFRelease(_disposables);
 		_disposables = NULL;
 	}
+
+	int result = pthread_mutex_destroy(&_mutex);
+	NSCAssert(0 == result, @"Failed to destroy mutex with error %d.", result);
 }
 
 #pragma mark Addition and Removal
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
I added 0af5eb7 because I believe the prior behavior is unsafe and to be consistent with the behavior in RACCompoundDisposable.

ReactiveCocoa/Objective-C/RACCompoundDisposable.m
```diff
@@ -78,6 +78,16 @@ + (instancetype)compoundDisposableWithDisposables:(NSArray *)disposables {
 	return [[self alloc] initWithDisposables:disposables];
 }
 
+- (id)init {
```
NachoSoto: Could you change these to `instancetype`?

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -27,14 +27,23 @@ public final class Atomic<Value> {
 	/// Initializes the variable with the given initial value.
 	public init(_ value: Value) {
 		_value = value
+		let result = pthread_mutex_init(&mutex, nil)
+		assert(0 == result, "Failed to initialize mutex with error \(result).")
```
NachoSoto: This can just be `assert(result == 0, ...)`. No need for the defensive ordering since `Swift` will warn you if you accidentally make it an assignment :)

Adlai-Holler: Haha I guess I think it's more readable that way – "assert 0" is the essence of the statement. I'll switch it up.

Adlai-Holler:
Changes made! Is this ready to go?

ReactiveCocoa/Objective-C/RACCompoundDisposable.m
```diff
@@ -118,6 +128,9 @@ - (void)dealloc {
 		CFRelease(_disposables);
 		_disposables = NULL;
 	}
+
+	int result = pthread_mutex_destroy(&_mutex);
```
NachoSoto: Nit: make this `const int`

ReactiveCocoa/Objective-C/RACCompoundDisposable.m
```diff
@@ -78,7 +78,17 @@ + (instancetype)compoundDisposableWithDisposables:(NSArray *)disposables {
 	return [[self alloc] initWithDisposables:disposables];
 }
 
-- (id)initWithDisposables:(NSArray *)otherDisposables {
+- (instancetype)init {
+	self = [super init];
+	if (self == nil) return nil;
+
+	int result = pthread_mutex_init(&_mutex, NULL);
```
NachoSoto: Ditto

ReactiveCocoa/Objective-C/RACSerialDisposable.m
```diff
@@ -32,15 +29,19 @@ @implementation RACSerialDisposable
 #pragma mark Properties
 
 - (BOOL)isDisposed {
-	return _disposed;
+	pthread_mutex_lock(&_mutex);
+	BOOL disposed = _disposed;
```
NachoSoto: Ditto

ReactiveCocoa/Objective-C/RACSerialDisposable.m
```diff
@@ -57,7 +58,17 @@ + (instancetype)serialDisposableWithDisposable:(RACDisposable *)disposable {
 	return serialDisposable;
 }
 
-- (id)initWithBlock:(void (^)(void))block {
+- (instancetype)init {
+	self = [super init];
+	if (self == nil) return nil;
+
+	int result = pthread_mutex_init(&_mutex, NULL);
```
NachoSoto: Ditto

NachoSoto:
This looks good to me. I'll leave the decision to someone else with more knowledge about locks though.

Adlai-Holler:
Thoughts @liscio @jspahrsummers ?

Adlai-Holler:
426f7f5a8afe411df794742a1f077c9d

liscio:
@Adlai-Holler I've been traveling, and wasn't able to reply sooner.

I feel that the "benchmark" is hardly pressing on anything, really. I'd be surprised if the "basic operations" didn't get completely canceled out by a basic optimization pass, to be honest…

Furthermore, a better benchmark would be to stress RAC itself (or even just the `Atomic` class) with this construct in place. Sadly, a bigger undertaking than anyone appears to be able/willing to do.

So I'd suggest holding off on shipping this until someone is able to give a solid thumbs-up on a large-ish RAC code base with this patch applied, or we have a really good torture test for signals in place.

#### Why am I being cautious?

I implemented almost this exact same patchset (https://github.com/ReactiveCocoa/ReactiveCocoa/compare/master...liscio:pthread_mutex) and RAC's unit tests would sometimes hang on me in certain spots in the Mac test suite.

So at the _very least_ we should try and get reasonable hardware coverage of RAC's unit tests on this branch. I'll try and build / run it on a few boxes and let you know what I find.

cortices:
I'm getting a crash on OSSpinLockLockSlow on iOS 9.3 most recent beta (13E5200d). Is this possibly because the OSSpinLock API has been closed now or something? I'll try to build with the pthread_mutex patch.
screen shot 2016-02-12 at 12 05 05

cortices:
The error was fixed by replacing OSSpinLock with pthread_mutex in Atomic.swift. It is live in https://github.com/cortices/ReactiveCocoa/tree/swift2.2, I will submit a pull request for a swift 2.2 branch.

Adlai-Holler:
@liscio What's the status here?

liscio:
@Adlai-Holler on the two Macs I tried (13" rMBP, 8-core 3GHz Mac Pro), all the unit tests seemed to work fine.

Unfortunately I've not yet been able to test this branch against a larger code base. If you're able, maybe try building Carthage yourself and see how it runs with and without this patch applied. That's all I can think of for a large-ish Mac-side code base.

Beyond testing suggestion, I think it makes sense to merge this for no other reason than there are clearly spinlock issues (as evidenced above) coming down the pipeline soon in iOS 9.3. The code itself looks fine.

Adlai-Holler:
Ping

RuiAAPeres:
Given what was discussed on #2619, what do you guys think, should we move forward with this change? @kballard, what do you think?  cc @mdiep @ikesyo

mdiep:
This is on my to-do list as I make my way through unmerged RAC PRs. Hoping to get to it soon.

ReactiveCocoa/Objective-C/RACSerialDisposable.m
```diff
@@ -32,15 +29,17 @@ @implementation RACSerialDisposable
 #pragma mark Properties
 
 - (BOOL)isDisposed {
-	return _disposed;
```
dmcrodrigues: Nice catch 👍

dmcrodrigues:
This looks great 👍 However, I would like to arguing if we couldn't be using a `dispatch_queue` to provide the mutual exclusion mechanism. Here's some pros and cons.

**Pros**
- *Portability* - libpthread is a standard in unix systems but is not natively available in all platforms which can be relevant if we want to make RAC cross-platform (#2807). libdispatch has the advantage that will be included in Swift;
- *Safety* - a `dispatch_queue` provides guarantees that we never forget to pair a lock call with an unlock call;
- *Performance* - some operations may be performed asynchronously without locking the caller until the execution has completed, e.g. `-removeDisposable:`. Another possibility are libdispatch's barriers which can be used to perform reads concurrently without locking anyone until there's a write operation.

**Cons**
- *Performance* - there's some obvious performance penalty in creating the blocks (they're first class citizens) to be executed in the queue;
- *Complexity* - using barriers is a little more complex but I think we could set up some convenience methods to help with this.

mdiep:
Clearly [something needs to be done on iOS](https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20151214/000321.html).

I'm not a lock expert by any means. This change definitely looks correct, but I'm not sure how to appropriately evaluate its affect on performance. But I don't think that matters, at least on iOS—avoiding livelocks is worth a performance hit.

So I think there are 2 realistic options:

1. Merge this as-is
2. Conditionalize the code to use a different lock on OS X and iOS.

I don't think (2) is worth the complexity or the potential confusion of different performance characteristics. pthread mutexes do seem to be the [best](http://perpendiculo.us/2009/09/synchronized-nslock-pthread-osspinlock-showdown-done-right/) [option](https://www.mikeash.com/pyblog/friday-qa-2015-02-06-locks-thread-safety-and-swift.html)>

So I think we ought to merge this as-is.

@ReactiveCocoa/reactivecocoa Could anyone else share insight or an opinion on this? I'll feel more comfortable if I can get confirmation that what I said makes sense.

Adlai-Holler:
I don't think we should use dispatch queues for now because of the cost of allocating dispatch blocks and because locks are an extremely simple and well-known tool throughout CS. In the future if the change is justified then you migrate to dispatch blocks.

kballard:
Using spinlocks on OS X isn't necessarily safe either. From what I understand, the fanless MacBook has the same behavior as iOS.

liscio:
I am 👍 for the change as-is. As I said above, the limited testing I was able to perform all went fine. I just wish we had a better mechanism to profile and assert the behavior of the fix.

The sooner we get this in, the sooner we'll be able to push on it harder with larger code bases. It's clearly a problem to leave the spinlock in there, so let's move forward already!

ikesyo:
> 1. Merge this as-is

👍

mdiep:
✨ 

Thanks for the PR, @Adlai-Holler! Sorry it took so long to get to it.

Adlai-Holler:
Lol no worries, very glad to see it shipped!

PR #2672, opened 2016-01-20
Remove default capacity of SignalProducer.buffer() and SignalProducer.replayLazily()

### LINKED ISSUE #2640: Remove SignalProducer.buffer capacity Int.max default
Pointed out by @JaviSoto in #2639:
I think we should remove this maximum. It's probably common for people to not think about this capacity and leave the default, wasting memory.

This would be a breaking change, so it would be great to get in for 4.0.

What does everyone think?

mdiep:
Fixes #2640.


neilpa:
:+1:

NachoSoto:
Thank you @mdiep!

PR #2651, opened 2016-01-10
Added missing UIKit extensions to tvOS target

NachoSoto:

PR #2673, opened 2016-01-21
Actually added several UIKit extensions to the tvOS target

NachoSoto:
This is a follow up to #2651. Did I mention that I hate `Xcode`? Even though I added the files to the `tvOS` target, and they're imported in `ReactiveCocoa.h`, these extensions aren't found in `RC 2`.
Why the hell is this not an error, for importing a non-modular header? I swear I understand Xcode less and less every day.

Anyway, with this change, NOW their headers are `public`...


NachoSoto:
I updated the release notes to link to this.

mdiep:
Oh Xcode. :disappointed:

PR #2689, opened 2016-01-28
Actually fix tvOS headers

NachoSoto:
Follow up to #2651 and #2673. I forgot to actually update `ReactiveCocoa.h`... why is this in THREE places, [Xcode](https://twitter.com/NachoSoto/status/692785182525382656)?

The hilarious part is that `TARGET_OS_IPHONE` is defined for `tvOS`, which [Apple closed as "works as intended"](http://www.openradar.me/22639221). I changed that to `TARGET_OS_IOS`. Because it makes sense that `tvOS` is an iPhone.


neilpa:
The never ending struggle

Issue #2085, opened 2015-06-08
Incorporate Objective-C generics into RACSignal?


NachoSoto:
Somebody had to ask, and I couldn't help myself :yum: 

This will probably be great for bridging between Swift signals and Objective-C :)

```
RACSignal<NSData *> *signal = toRACSignal(signalProducer)
```

akl32:
A worrisome quote from the new iBooks volume about Obj-C / Swift 2 interop:

“NOTE

Aside from than these Foundation collection classes, Objective-C lightweight generics are ignored by Swift. Any other types using lightweight generics are imported into Swift as if they were unparameterized.”

Excerpt From: Apple Inc. “Using Swift with Cocoa and Objective-C (Swift 2 Prerelease).” iBooks. https://itun.es/us/utTW7.l

"These foundation classes" are referring to Array, Set, etc.

NachoSoto:
![are you kidding me](https://cloud.githubusercontent.com/assets/685609/8045306/3d581196-0de7-11e5-85fd-a20e54ebf27d.jpg)

neilpa:
Dreams crushed

zintus:
It's still makes sense for objc world

rpowelll:
I guess it all depends on whether the Objective-C API is something we want to continue updating, or whether it's heading towards being in 'maintenance mode'.

mciuba:
Some people claim that generics can also be used with custom classes:

https://twitter.com/arekholko/status/608187695219769344?t=1&cn=cmVjb3NfbmV0d29ya19kaWdlc3RfdHJpZ2dlcmVk&sig=ab9d7b57fa692e82ff0030e42b7e87271fb3a2a1&al=1&refsrc=email&iid=551eb3f36fa34135830793ff66a79d98&autoactions=1433854217&uid=1136915245&nid=244+594

neilpa:
@mciuba They do work with custom classes, our complaint is that they are ignored when bridged to Swift.

jspahrsummers:
This could be useful for the extant Objective-C code, but it might be too confusing when it doesn't bridge to Swift.

LightD:
Would it really be too confusing? from Swift perspective nothing has changed.. but for Obj-C code it would make a lot of difference! I don't see a reason why this shouldn't be a feature in the Obj-C code, am I wrong? :confounded:

jspahrsummers:
[...]

PR #2696, opened 2016-01-30
Allow RACCommand to specify generic input type

erichoracek:
Inputs to `RACCommand` to do not suffer from the same issues as `RACSignal` when attempting to adopt lightweight generics as discussed in #2085. As such, adding them is as simple as declaring a `<__covariant InputType>` in the interface.

In projects that are still using RAC 2.X, this will be helpful to ensure that commands are executed with an input of the correct type. It also helps to catch programmer error when refactoring commands.

I was unable to find a way to expose this annotated type to the compiler when bridging a `RACCommand` to `Action` in `ObjectiveCBridging.swift`. Open to any suggestions on how this might be possible.

### DIFF
--- a/ReactiveCocoa/Objective-C/NSControl+RACCommandSupport.h
+++ b/ReactiveCocoa/Objective-C/NSControl+RACCommandSupport.h
@@ -8,7 +8,7 @@
 
 #import <Cocoa/Cocoa.h>
 
-#import "RACCommand.h"
+@class RACCommand<__contravariant InputType>;
 
 @interface NSControl (RACCommandSupport)
 
--- a/ReactiveCocoa/Objective-C/NSControl+RACCommandSupport.h
+++ b/ReactiveCocoa/Objective-C/NSControl+RACCommandSupport.h
@@ -8,7 +8,7 @@
 
 #import <Cocoa/Cocoa.h>
 
-@class RACCommand;
+#import "RACCommand.h"
 
 @interface NSControl (RACCommandSupport)
 
@@ -17,6 +17,6 @@
 /// to the command's `canExecute`.
 ///
 /// Note: this will reset the control's target and action.
-@property (nonatomic, strong) RACCommand *rac_command;
+@property (nonatomic, strong) RACCommand<__kindof NSControl *> *rac_command;
 
 @end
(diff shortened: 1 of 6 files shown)


NachoSoto:
This is great, thank you!

As for bridging this type back to Swift, I don't think it's possible right now to inherit the generic type for `Action`.

robb:
LGTM

erichoracek:
Seems that if the [Importing Objective-C Lightweight Generics](https://github.com/apple/swift-evolution/blob/master/proposals/0057-importing-objc-generics.md) proposal is accepted we may be able to supply the generic input to the bridged `Action` in `ObjectiveCBridging.swift`!

Regardless, is there anything preventing this from being merged as-is? Thanks!

ReactiveCocoa/Objective-C/RACCommand.h
```diff
@@ -24,7 +24,7 @@ extern NSString * const RACUnderlyingCommandErrorKey;
 
 /// A command is a signal triggered in response to some action, typically
 /// UI-related.
-@interface RACCommand : NSObject
+@interface RACCommand<__covariant InputType> : NSObject
```
mdiep: Shouldn't `InputType` be contravariant?

You've specified that `NSControl.rac_command` should be `RACCommand<__kindof NSControl *>`, but I should be able to assign a `RACCommand<NSObject *>` to that property. That command needs to be able to take an `NSControl`, but it needn't accept only `NSControl`s. Conversely, it wouldn't be right to assign a `RACCommand<NSButton *>` to it, because `NSControl` can't guarantee that an `NSButton` will be sent.

In other words, `RACCommand` inputs behave like function parameters, which are contravariant.

erichoracek: @mdiep Great point! I was unaware that multiple variance keywords were even available. After reading up on this the suggested change definitely makes sense. Strange that this is available in Obj-C but not in Swift. Thanks!

ReactiveCocoa/Objective-C/NSControl+RACCommandSupport.h
```diff
@@ -8,7 +8,7 @@
 
 #import <Cocoa/Cocoa.h>
 
-@class RACCommand;
+#import "RACCommand.h"
```
ikesyo: It seems that we can write as `@class RACCommand<__contravariant InputType>;` here.

mdiep:
Thanks! And sorry for the delay.

In the future, you should make sure that you leave a comment after you push so that I get a notification. I think that's why I missed this.

ikesyo:
Should this be backported to `2.x-maintenance`?

ikesyo:
> Should this be backported to `2.x-maintenance`?

Done in b4a166f.

PR #2701, opened 2016-02-01
Signal lifetime: added test to verify that observe handlers are deallocated

### LINKED ISSUE #2698: Is there any test case to test the lifetime of the observing closures?
For example

signal.observeNext { x in }

The closure { x in } should not be retained after the signal terminates.

NachoSoto:
Brought up by @CosynPa, I realized that it's a pretty important test to have.
Fixes #2698.

The test is a bit verbose, but I wanted to make it very exhaustive.

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
@@ -210,6 +210,67 @@ class SignalLifetimeSpec: QuickSpec {
 				expect(signal).to(beNil())
 			}
 		}
+
+		describe("observe") {
+			var signal: Signal<Int, TestError>!
+			var observer: Signal<Int, TestError>.Observer!
+
+			var token: NSObject? = nil
+			weak var weakToken: NSObject?
+
+			func expectTokenNotDeallocated() {
+				expect(weakToken).toNot(beNil())
+			}
+
+			func expectTokenDeallocated() {
+				expect(weakToken).to(beNil())
+			}
+
+			beforeEach {
+				let (signalTemp, observerTemp) = Signal<Int, TestError>.pipe()
+				signal = signalTemp
+				observer = observerTemp
+
+				token = NSObject()
+				weakToken = token
+
+				signal.observe { [token = token] _ in
+					_ = token!.description
+				}
+			}
+
+			it("should deallocate observe handler when signal completes") {
+				expectTokenNotDeallocated()
+
+				observer.sendNext(1)
+				expectTokenNotDeallocated()
+
+				token = nil
+				expectTokenNotDeallocated()
+
+				observer.sendNext(2)
+				expectTokenNotDeallocated()
+
+				observer.sendCompleted()
+				expectTokenDeallocated()
+			}
+
+			it("should deallocate observe handler when signal fails") {
+				expectTokenNotDeallocated()
+
+				observer.sendNext(1)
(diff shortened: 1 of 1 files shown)


ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -210,6 +210,91 @@ class SignalLifetimeSpec: QuickSpec {
...
+			var observer: Signal<Int, TestError>.Observer!
+
+			var tokenWasDeallocated: Bool!
+
+			var token: Token? = nil
+			weak var weakToken: Token?
+
+			func expectTokenNotDeallocated() {
+				expect(tokenWasDeallocated) == false
+				expect(weakToken).toNot(beNil())
```
mdiep: These two tests seem redundant. Shouldn't it be enough to just test a weak reference? In which case, we could remove `Token` and use an object from Foundation or Swift that supports weak references?

NachoSoto: Yeah I guess you're right. I wanted to be as exhaustive as possible but I can't think of any reason why just relying on `weak` wouldn't be enough.

NachoSoto: Changed.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -210,6 +210,71 @@ class SignalLifetimeSpec: QuickSpec {
 				expect(signal).to(beNil())
 			}
 		}
+
+		describe("observe") {
+			var signal: Signal<Int, TestError>!
+			var observer: Signal<Int, TestError>.Observer!
+
+			var tokenWasDeallocated: Bool!
```
mdiep: This is no longer used for anything and can be removed.

mdiep:
Just the one note. This is a great test to have. :clap:

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
NachoSoto: Good catch!

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -210,6 +210,69 @@ class SignalLifetimeSpec: QuickSpec {
...
+			func expectTokenDeallocated() {
+				expect(weakToken).to(beNil())
+			}
+
+			beforeEach {
+				let (signalTemp, observerTemp) = Signal<Int, TestError>.pipe()
+				signal = signalTemp
+				observer = observerTemp
+
+				tokenWasDeallocated = false
```
mdiep: This is still in here too, which is causing the CI failure. I should have called this out specifically in my last comment—sorry.

NachoSoto: OMG that's what I get for not even compiling my change 😰

NachoSoto:
Bump @mdiep :)

mdiep:
Ha! I never saw the new commit. :relaxed:

NachoSoto:
> Ha! I never saw the new commit. :relaxed:

Yeah `GitHub` is missing that feature! I like that Phabricator does send notifications when that happens.

PR #2715, opened 2016-02-08
Add `.uniqueValues` to SignalProducer

nixterrimus:
Adds a `uniqueValues` function to SignalProducer so that only values that are unique across the entire stream will be sent.

## Rationale

A common operation that I've needed for a recent project is to take only the unique values that come out of a SignalProducer.  In this project I have a data source which returns snapshots are regular intervals, however what I want is only the items that have never been seen before.  I've needed this in several places and so I've generalized it into a pull request.

## Quick Demo

quickdemo

## Work Done

- [x] Code for `uniqueValues`
- [x] Inline documentation
- [x] Tests for `uniqueValues`

## Todo

- [x] Note that this causes values to be retained
- [x] Make accessing and mutating set thread safe
- [x] Fix so that each call to `start` gets its own Set
- [x] Add lifted Signal version

## Feedback

- Is this something that makes sense in the ReactiveCocoa proper?
- Is `uniqueValues` a good name for the operation?
- Is there anything else I can do to make this code feel at home in RAC?

## Thanks

I've been using ReactiveCocoa for the last 6 months and it's changed the way I write applications.  This is my first contribution and I want to say thanks for maintaining a project which has been very useful to me.  🎉

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1039,23 +1039,22 @@ extension SignalType where Value: Hashable {
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func uniqueValues() -> Signal<Value, Error> {
 		return Signal { observer in
-			var set = Set<Value>()
-			let lock = NSLock()
-			lock.name = "org.reactivecocoa.ReactiveCocoa.Signal.uniqueValues"
+			let seenValues = Atomic<Set<Value>>(Set<Value>())
 			
 			return self
 				.filter { value in
-					lock.lock()
-					let containsValue = !set.contains(value)
-					lock.unlock()
-					return containsValue
+					return seenValues.withValue { set in
+						!set.contains(value)
+					}
 				}
 				.observe { event in
 					switch event {
 					case let .Next(value):
-						lock.lock()
-						set.insert(value)
-						lock.unlock()
+						seenValues.modify { set in
+							var mutableSet = set
+							mutableSet.insert(value)
+							return mutableSet
+						}
 						fallthrough
 					default:
 						observer.action(event)
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1031,6 +1031,40 @@ extension SignalType {
 	}
 }
 
+extension SignalType where Value: Hashable {
+	/// Forwards only those values from `self` that are unique across the set of
+	/// all values that have been seen
+	/// Note: This causes values that are forwarded to be retained to check for
+	/// uniquness
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func uniqueValues() -> Signal<Value, Error> {
+		return Signal { observer in
+			var set = Set<Value>()
+			let lock = NSLock()
+			lock.name = "org.reactivecocoa.ReactiveCocoa.Signal.uniqueValues"
+			
+			return self
+				.filter { value in
+					lock.lock()
+					let containsValue = !set.contains(value)
+					lock.unlock()
+					return containsValue
+				}
+				.observe { event in
+					switch event {
+					case let .Next(value):
+						lock.lock()
+						set.insert(value)
+						lock.unlock()
+						fallthrough
+					default:
+						observer.action(event)
+					}
+			}
+		}
+	}
+}
+
 private struct ThrottleState<Value> {
 	var previousDate: NSDate? = nil
 	var pendingValue: Value? = nil
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1042,11 +1042,7 @@ extension SignalType where Value: Hashable {
 			let seenValues = Atomic<Set<Value>>(Set<Value>())
 			
 			return self
-				.filter { value in
-					return seenValues.withValue { set in
-						!set.contains(value)
-					}
-				}
+				.filter { value in return seenValues.withValue { !$0.contains(value) } }
 				.observe { event in
 					switch event {
 					case let .Next(value):
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1033,9 +1033,9 @@ extension SignalType {
 
 extension SignalType where Value: Hashable {
 	/// Forwards only those values from `self` that are unique across the set of
-	/// all values that have been seen
+	/// all values that have been seen.
 	/// Note: This causes values that are forwarded to be retained to check for
-	/// uniquness
+	/// uniquness.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func uniqueValues() -> Signal<Value, Error> {
 		return Signal { observer in
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1039,7 +1039,7 @@ extension SignalType where Value: Hashable {
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func uniqueValues() -> Signal<Value, Error> {
 		return Signal { observer in
-			let seenValues = Atomic<Set<Value>>(Set<Value>())
+			let seenValues: Atomic<Set<Value>> = Atomic([])
 			
 			return self
 				.filter { value in return seenValues.withValue { !$0.contains(value) } }
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -777,6 +777,8 @@ extension SignalProducerType where Value: Equatable {
 extension SignalProducer where Value: Hashable {
 	/// Forwards only those values from `self` that are unique across the set of
 	/// all values that have been seen
+	/// Note: This causes values that are forwarded to be retained to check for
+	/// uniquness
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func uniqueValues() -> SignalProducer<Value, Error> {
 		let producer = SignalProducer.init { observer, disposable in
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -783,13 +783,20 @@ extension SignalProducer where Value: Hashable {
 	public func uniqueValues() -> SignalProducer<Value, Error> {
 		let producer = SignalProducer.init { observer, disposable in
 			var set = Set<Value>()
+			let lock = NSLock()
+			lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.uniqueValues"
 			
 			self
 				.filter { value in
-					return !set.contains(value)
+					lock.lock()
+					let containsValue = !set.contains(value)
+					lock.unlock()
+					return containsValue
 				}
 				.on( next: { value in
+					lock.lock()
 					set.insert(value)
+					lock.unlock()
 				})
 				.start(observer)
 		}
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -774,34 +774,14 @@ extension SignalProducerType where Value: Equatable {
 	}
 }
 
-extension SignalProducer where Value: Hashable {
+extension SignalProducerType where Value: Hashable {
 	/// Forwards only those values from `self` that are unique across the set of
 	/// all values that have been seen
 	/// Note: This causes values that are forwarded to be retained to check for
 	/// uniquness
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func uniqueValues() -> SignalProducer<Value, Error> {
-		let producer = SignalProducer.init { observer, disposable in
-			var set = Set<Value>()
-			let lock = NSLock()
-			lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.uniqueValues"
-			
-			self
-				.filter { value in
-					lock.lock()
-					let containsValue = !set.contains(value)
-					lock.unlock()
-					return containsValue
-				}
-				.on( next: { value in
-					lock.lock()
-					set.insert(value)
-					lock.unlock()
-				})
-				.start(observer)
-		}
-		
-		return producer
+		return lift { $0.uniqueValues() }
 	}
 }
 
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -779,15 +779,20 @@ extension SignalProducer where Value: Hashable {
 	/// all values that have been seen
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
 	public func uniqueValues() -> SignalProducer<Value, Error> {
-		var set = Set<Value>()
+		let producer = SignalProducer.init { observer, disposable in
+			var set = Set<Value>()
+			
+			self
+				.filter { value in
+					return !set.contains(value)
+				}
+				.on( next: { value in
+					set.insert(value)
+				})
+				.start(observer)
+		}
 		
-		return self
-			.filter { value in
-				return !set.contains(value)
-			}
-			.on( next: { value in
-				set.insert(value)
-			})
+		return producer
 	}
 }
 
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -774,6 +774,22 @@ extension SignalProducerType where Value: Equatable {
 	}
 }
 
+extension SignalProducer where Value: Hashable {
+	/// Forwards only those values from `self` that are unique across the set of
+	/// all values that have been seen
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func uniqueValues() -> SignalProducer<Value, Error> {
+		var set = Set<Value>()
+		
+		return self
+			.filter { value in
+				return !set.contains(value)
+			}
+			.on( next: { value in
+				set.insert(value)
+			})
+	}
+}
 
 /// Creates a repeating timer of the given interval, with a reasonable
 /// default leeway, sending updates on the given scheduler.
--- a/ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
@@ -258,7 +258,7 @@ class SignalProducerLiftingSpec: QuickSpec {
 				expect(values) == [ "a", "cc", "d" ]
 			}
 		}
-		
+
 		describe("skipWhile") {
 			var producer: SignalProducer<Int, NoError>!
 			var observer: Signal<Int, NoError>.Observer!
--- a/ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
@@ -258,6 +258,31 @@ class SignalProducerLiftingSpec: QuickSpec {
 				expect(values) == [ "a", "cc", "d" ]
 			}
 		}
+		
+		describe("uniqueValues", {
+			it("should skip values that have been already seen"){
+				let (baseProducer, observer) = SignalProducer<String, NoError>.buffer(1)
+				let producer = baseProducer.uniqueValues()
+				
+				var values: [String] = []
+				producer.startWithNext { values.append($0) }
+				
+				observer.sendNext("a")
+				expect(values) == [ "a" ]
+				
+				observer.sendNext("b")
+				expect(values) == [ "a", "b" ]
+				
+				observer.sendNext("a")
+				expect(values) == [ "a", "b" ]
+				
+				observer.sendNext("b")
+				expect(values) == [ "a", "b" ]
+				
+				observer.sendNext("c")
+				expect(values) == [ "a", "b", "c" ]
+			}
+		})
 
 		describe("skipWhile") {
 			var producer: SignalProducer<Int, NoError>!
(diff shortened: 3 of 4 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -774,6 +774,22 @@ extension SignalProducerType where Value: Equatable {
 	}
 }
 
+extension SignalProducer where Value: Hashable {
+	/// Forwards only those values from `self` that are unique across the set of
+	/// all values that have been seen
```
mdiep: I would explicitly call out that this retains all values that are sent on `self`. It's obvious once you think about it, but the consequences can be large, so it deserves to be documented.

mdiep:
I've written and used this operator once before in RAC2. But RAC3/4 are intentionally slimmed down, so I'm not sure whether it's something we'd include.

RuiAAPeres:
@mdiep 

I like a lot the idea of keeping RAC's core small and focused. This is something that we generally don't see in the iOS community. 👍

On the other hand tho, this might create a bad feeling when someone is trying to contribute to the project, because there is a high chance that your PR (normally with a new operator) will be left out.* I would recommend creating a place where these extras would be added to (  @neilpa 's Rex could be a possibility).

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -774,6 +774,22 @@ extension SignalProducerType where Value: Equatable {
...
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func uniqueValues() -> SignalProducer<Value, Error> {
+		var set = Set<Value>()
```
NachoSoto: `Set` is not thread safe.

ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
```diff
@@ -258,6 +258,31 @@ class SignalProducerLiftingSpec: QuickSpec {
 				expect(values) == [ "a", "cc", "d" ]
 			}
 		}
+		
+		describe("uniqueValues", {
+			it("should skip values that have been already seen"){
```
NachoSoto: Test looks good :+1:

NachoSoto:
I'm open to start increasing the API surface now that 4.0 is out!
I'm not sure about this operator **being on `SignalProducer`**, though: correct me if I'm wrong, but the way this is implemented, the returned producer no longer has value semantics, it has reference semantics, meaning that multiple calls to `start` would share the same set?

This can be fixed by implementing the operator using `SignalProducer.init` instead.

Once that's fixed (and tested) I'm :+1: for adding this!
Also, I think the name is perfect.

Oh, and one last thing, could you create a lifted `Signal` version too?

liscio:
To second what @mdiep said above regarding the fact that "the consequences can be large," I wonder whether instead you should be aiming to achieve the same thing in a different way. Perhaps it could be achieved using other operators, or maybe this operator needs a parameter to specify the upper bound on the set.

Suppose the stream contains `Int`s or `String`s. The storage required is (theoretically) unbounded and it's quite easy to miss that fact. This line of thinking—I presume—is what led to the `Int.Max` default value of `buffer()` being removed.

Adding to what @RuiAAPeres said above, there could be a case made for a "dumping ground" of potentially useful operators like this. They don't need to live in the framework, but it might be useful in the form of demonstrating how to add your own domain-specific operators. https://github.com/neilpa/Rex could be just the place.

Now, commenting on what you said originally:

> In this project I have a data source which returns snapshots are regular intervals, however what I want is only the items that have never been seen before.

This sounds like it could either be a major design flaw, or brilliant. What kind of snapshots are we talking about, exactly? The contents of a text document? A table view data source?

nixterrimus:
Thanks for the feedback!

## Todo

- [ ] Note that this causes values to be retained
- [ ] Note that this is not thread safe
- [ ] Fix so that each call to `start` gets its own Set
- [ ] Add lifted Signal version
- [ ] Add Signal version tests

NachoSoto:
>  Note that this is not thread safe

Instead of that, let's actually make it thread safe!

>  Fix so that each call to start gets its own Set

:+1: 

>  Add Signal version tests

This isn't necessary, we only need to test one version: ideally `Signal`s. The `SignalProducer` version will implicitly be tested because `lift` is already covered with tests.

nixterrimus:
@liscio 

> This sounds like it could either be a major design flaw, or brilliant. What kind of snapshots are we talking about, exactly? The contents of a text document? A table view data source?

These snapshots are a set of ids for items, so the first time I start the signal I might get [1,2,3]; the next time I get values I'd get [1,2,3,4] but what I really want is just the 4.

I think that this operator is generally useful because if a user is polling against a data source they might want to get only the new values as the data source keeps changing.

For example if I was building a "live" rss reader I could set a refresh interval and then use `uniqueValues` to get only new entries.  Even though the underlying mechanism is polling, as a consumer I get the benefits of the signal yielding values as if I was tuned in live.

Good point on the unbounded storage.  I think this is worth thinking about and I'd love more feedback on approaches to handling it.

nixterrimus:
@NachoSoto

I've addressed the feedback; Thanks for the feedback.

This PR has been a fun way to dig deeper into RAC.  Let me know if anything else looks out of place, I moved the operator to SignalType and lifted it for SignalProducerType.

I'm still learning about thread safety, does an NSLock seem a sensible approach here?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1031,6 +1031,40 @@ extension SignalType {
...
 
+extension SignalType where Value: Hashable {
+	/// Forwards only those values from `self` that are unique across the set of
+	/// all values that have been seen
+	/// Note: This causes values that are forwarded to be retained to check for
+	/// uniquness
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func uniqueValues() -> Signal<Value, Error> {
+		return Signal { observer in
+			var set = Set<Value>()
```
NachoSoto: Using `NSLock` is fine, but I think it would be simpler to use `Atomic<Set<Value>>` instead of implementing an adhoc locking mechanism here. Then you can use `Atomic.modify` to `insert` and `Atomic.withValue` to check if it `contains` it :)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1031,6 +1031,40 @@ extension SignalType {
 	}
 }
 
+extension SignalType where Value: Hashable {
+	/// Forwards only those values from `self` that are unique across the set of
+	/// all values that have been seen
```
NachoSoto: Nit: can you finish these sentences with a `.`?

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -626,7 +626,34 @@ class SignalSpec: QuickSpec {
...
+				expect(values) == [ "a", "b" ]
+				
+				observer.sendNext("a")
+				expect(values) == [ "a", "b" ]
+				
+				observer.sendNext("b")
+				expect(values) == [ "a", "b" ]
+				
+				observer.sendNext("c")
+				expect(values) == [ "a", "b", "c"]
```
NachoSoto: Nit: space before `]`

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -626,7 +626,34 @@ class SignalSpec: QuickSpec {
...
+			}
```
NachoSoto: Maybe to be safe, can you call `observer.sendCompleted()` and verify that `values` has not changed?

NachoSoto:
> This PR has been a fun way to dig deeper into RAC

Glad to hear! I'm excited to have this in `RAC 4.1`, I think it's a good addition and something that `Rx` users will welcome.

> I'm still learning about thread safety, does an NSLock seem a sensible approach here?
That works, but like mentioned I think using `Atomic` would be simpler and less error prone since it's very well tested.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1031,6 +1031,39 @@ extension SignalType {
...
 
+extension SignalType where Value: Hashable {
+	/// Forwards only those values from `self` that are unique across the set of
+	/// all values that have been seen.
+	/// Note: This causes values that are forwarded to be retained to check for
+	/// uniquness.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func uniqueValues() -> Signal<Value, Error> {
+		return Signal { observer in
+			let seenValues = Atomic<Set<Value>>(Set<Value>())
```
NachoSoto: The type is redundant here :) You can just do:
```swift
let seenValues = Atomic<Set<Value>>([])
```

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1031,6 +1031,39 @@ extension SignalType {
...
+			
+			return self
+				.filter { value in
+					return seenValues.withValue { set in
```
NachoSoto: Awesome. Maybe just inline this whole thing?
```swift
.filter { value in return seenValues.withValue { !$0.contains(value) } }
```

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1031,6 +1031,39 @@ extension SignalType {
...
+						!set.contains(value)
+					}
+				}
+				.observe { event in
+					switch event {
+					case let .Next(value):
+						seenValues.modify { set in
```
NachoSoto: You can just do `var set in`, I think?

NachoSoto:
Just 2 last nits, but :ship: it!

What do others think? I'm on board, but I want to make sure I hear other people's thoughts /cc @ReactiveCocoa/reactivecocoa.

If I don't hear back I'll merge by tomorrow :100:

andymatuschak:
My bias is that complex behaviors like the RSS-unread-ness example @nixterrimus cites above should be build explicitly with purpose-specific objects, and that we should discourage unbounded-size, complexly-behaving entities from occupying the same type as straightforward "values changing over time." 

It would be nice if a programmer could have some kind of reasonable "best practice" expectations of an instance of the `SignalProducer` type; if there is no such expectation, hanging onto one becomes a bit like hanging onto an arbitrary `id`: the consequences could be literally anything. Yes, I understand that the API surface _allows_ users to make signals of arbitrary complexity, but there's a difference between allowing and encouraging, etc etc etc.

That said, I recognize that my philosophies here differ from others, and I'm happy to defer the people who actually maintain this framework instead of just soapboxing. ;)

NachoSoto:
I was actually about to say that I think the canonical version of `uniqueValues` should be like this:
```swift
extension SignalType {
	public fund uniqueValues<Key: Hashable>(mapping: Value -> Key) -> Signal<Value, Error> {
		return Signal { observer in
			let seenValues = Atomic<Set<Key>>([])
			
			return self
				.filter { value in return seenValues.withValue { !$0.contains(mapping(value)) } }
				.observe { event in
					switch event {
					case let .Next(value):
						seenValues.modify { var set in
							set.insert(mapping(value))
							return set
						}
						fallthrough
					default:
						observer.action(event)
					}
			}
		}
	}
}
```

This is equivalent to `Rx`'s `distinct` implementation. We can then have an overload that uses the `identity` function as a default (`where Value: Hashable`), but that would have the advantage of making it explicit that only a part of the `Value` emitted is kept in memory.

> should be build explicitly with purpose-specific objects

Can you describe what you mean by that @andymatuschak?

> but there's a difference between allowing and encouraging, etc etc etc.

I mean, the same could be said for `skipRepeats()`, but we can't argue that's not a useful operator. I agree this is implicitly keeping more memory, but I think anyone using this operator would be aware of that, in the same way that anyone keeping around an `NSCache` object understands the consequences.

ReactiveCocoa/Swift/Signal.swift
nixterrimus: I thought that `var [name] in` was correct syntax as well, but it looks like its not valid.

screen shot 2016-02-18 at 8 57 57 pm

ReactiveCocoa/Swift/Signal.swift
nixterrimus: Great feedback!

andymatuschak:
> > should be build explicitly with purpose-specific objects

> Can you describe what you mean by that @andymatuschak?

Hm, I've given it a shot below, but I struggled to do it because I'm not sure this RSS reader example makes sense. In the case of an RSS reader polling an endpoint and sending values to a UI, you really want `skipRepeats()`, not perpetual-uniqueness. Is there some better example of where `uniqueValues()` actually makes sense?

Anyway, I'm basically suggesting that you'd be better off implementing the uniquing logic in a type with clear names, like this:
```swift
class NewsItemStore {
  // Called repeatedly with new data from polling.
  var newsItems: [NewsItem]
  // Called when there's a uniquely new set of news items. I don't know when/why it would make sense to use this.
  let uniqueNewsItemsSignal: Signal<[NewsItem], NoError>
}
```

> > but there's a difference between allowing and encouraging, etc etc etc.

> I mean, the same could be said for skipRepeats(), but we can't argue that's not a useful operator.

`skipRepeats()` has a constant memory footprint; this operator has an _unbounded_ memory footprint—totally different beasts.

>  I think anyone using this operator would be aware of that, in the same way that anyone keeping around an NSCache object understands the consequences.

The danger is not in the component which _calls_ `uniqueValues()`, but rather in the component which _consumes_ the resultant `Signal`, which has to deal with unbounded possible consequences this signal could be hiding. Taking this observation to `NSCache`: yes, an object which has an `NSCache`-typed reference to an `NSCache` understands the consequences; an object with an `id`-typed reference _can't_ understand the consequences. Which is much of why we usually avoid `id`-typed interfaces.

mdiep:
> Is there some better example of where uniqueValues() actually makes sense?

GitHub Desktop uses it to calculate recent branches:

recent-branches

We read the branch names from git's reflog. But there may be repeats and we're interested in the 5 most recent unique values. So we use a combo of `uniqueValues` + `take` to get them. That's in RAC2, but should translate directly.

andymatuschak:
That's a cool example, @mdiep!

In that situation, I'd suggest that state is being introduced to unique across values over time when time is not actually meaningful. i.e. the signal pipeline looks like:

```
reflogCommandOutput: Signal<String>
-> reflogEntryStrings: Signal<String> // i.e. for each reflog invocation, emits an event for each line
-> reflogEntries: Signal<ReflogEntry> // parse the lines...
-> uniqueReflogEntries: Signal<ReflogEntry> // isn't this gonna emit only the _first_ unique value? I guess it implicitly relies on the fact that git reflog emits newest first...
-> firstFiveUniqueReflogEntries: Signal<ReflogEntry>
```

But time is only actually meaningful for the first signal. The other signals are used as streams more than as signals. State wouldn't be necessary at all if the pipeline were instead arranged like this:

```
reflogCommandOutput: Signal<String>
-> reflogEntryStrings: Signal<[String]> // i.e. for each reflog invocation, emit a list of the lines
-> reflogEntries: Signal<[ReflogEntry]> // .map{$0.map}
-> uniqueReflogEntries: Signal<[ReflogEntry]> // no longer requires state
-> firstFiveUniqueReflogEntries: Signal<[ReflogEntry]> // this stage feels silly now; it just maps a range subscript
```

Have I interpreted this correctly?

mdiep:
That's mostly correct. The actual pipeline is like this:

```
reflogCommandOutput: Signal<String>
-> reflogEntryStrings: Signal<RefLogEntry> // i.e. for each reflog invocation, emits an event for each line
-> branchNameForReflogEntry: Signal<BranchName?> // parse the lines...
-> ignoreNil // drop lines that aren't relevant
-> uniqueBranchNames: Signal<BranchName>
-> take(5): Signal<BranchName>
```

But I don't think this affects your point.

The advantage of the first pipeline is that it avoids unnecessary work. The earlier signals can be interrupted as soon as 5 unique values are returned. In this case, it's probably not that much work: a couple thousand memory allocations and regular expression matching.

But that's the trouble with signals of arrays as I see it. `uniqueValues` does add some complexity to the signal chain, and consumers of the signal can't know what those are. But we're in that situation anyway, and using signals of arrays only increases the memory consumption of the signal.

andymatuschak:
Great point, @mdiep! What's interesting about this cascade is that it's mixing time-oriented reactive programming (i.e. the `reflogCommandOutput` signal—presumably it polls or something?) with stream-based dataflow programming (the rest of the chain, presumably `flatMap`'d from the command output). Subscriptions are ephemeral in the latter use case and long-lasting in the former.

In this instance, `uniqueValues` provides value in the `RACStream` case more than the `RACSignal` case. And it's less scary there, too: my concern about it comes from its invisible unbounded-size state, but that doesn't matter (as much) in an ephemeral context—only in the long-lasting subscription context.

RAC 2 very explicitly tried to be for both use cases—it even had a little type hierarchy about them. RAC 4 seems to have explicitly focused on the `RACSignal` case. Do we still want to promote the same tools for solving both kinds of problems?

If so, can we shape the API to make clear that certain signals are intended to be ephemeral? I don't know what that looks like, exactly, but I worry about overloading `Signal` too much. Overreaching abstractions can lead to sometimes dangerous—but often just frustrating–interfaces.

As always, feel free to ignore my hand-wringing and just merge the thing. :peanut:

mdiep:
> (i.e. the reflogCommandOutput signal—presumably it polls or something?)

It shells out to `git` and returning the result. It's a one-time thing that I should have made into a `SignalProducer`.

What you may find interesting is that the end result is captured in a property. Our actual structure around this isn't the cleanest right now, but it's something like this simplified RAC2:

```objc
RAC(self, recentBranches) = [[[[[[RACSignal
    merge:@[
        self.HEADChangedSignal,
        self.applicationBecameActiveSignal,
    ]]
    mapReplace:self.repository]
    map:^(GHRepository *repository) {
        return [[repository recentBranches] collect]; // the signal we were discussing above
    }]
    switchToLatest]
    distinctUntilChanged]
    startWith:@[]];
```

It observes certain conditions under which it should change. When we have reason to believe it changed, we run the command, and store the result in a property. This avoids passing the signal around, which I believe is what you've advocated, and it's what I try to do when possible.

> What's interesting about this cascade is that it's mixing time-oriented reactive programming with stream-based dataflow programming (the rest of the chain, presumably flatMap'd from the command output). Subscriptions are ephemeral in the latter use case and long-lasting in the former.

I'm a little confused about the terminology here. I think you're distinguishing between "the value of this thing over time" vs. "a value for this thing". I don't really see those as separate things, but more as two sides of the same thing. I'd love to hear more about what you see as the differences, particularly as it relates to the API.

I see both sides of that as time-oriented. A long-lasting signal is a view of a thing that's independent of time. (I wish I had a better word than _thing_ to use in that case, but _value_ is the only thing I can think of and it's wrong.) An ephemeral signal is a single value that arrives at unknown time and allows our code to be independent of time—like a future/promise.

It's natural to want to mix these, I think. In the example above, the `recentBranches` is defined independent of time. But it also depends on asynchronous commands, so writing it that way almost depends on the ability to mix ephemeral and long-lasting signals.

Like I said, I'd love to hear more of your thoughts around this. I'm not sure I'm talking about the same things.

> If so, can we shape the API to make clear that certain signals are intended to be ephemeral? I don't know what that looks like, exactly, but I worry about overloading Signal too much.

That's a really interesting question. How does this differ from the current `Signal`/`SignalProducer` division? I guess `SignalProducer`s could be renamed to and made to return something other than `Signal` to emphasize the difference between always-on hot signals and run-once results of producers? I'm not sure whether that really addresses it.

> As always, feel free to ignore my hand-wringing and just merge the thing

I think we've arrived at the same conclusion as the discussion around `replayLazily`—this is a good fit for RAC as it stands today, but something to think about for the future design.

So long as merging won't stop the discussion, I'm all :+1:. :smile:

andymatuschak:
> This avoids passing the signal around, which I believe is what you've advocated, and it's what I try to do when possible.

Yeah, I'm basically "computing lifetimes" of the constituent signals as I read this code (which is itself a hard thing to do, and it's difficult to know from here what the consequences of those lifetimes are); high locality makes this much easier and limits potential damage from errors.

> I see both sides of that as time-oriented. A long-lasting signal is a view of a thing that's independent of time. (I wish I had a better word than thing to use in that case, but value is the only thing I can think of and it's wrong.) An ephemeral signal is a single value that arrives at unknown time and allows our code to be independent of time—like a future/promise.

I see the ephemeral thing as a more convenient API for doing very traditional constant-memory stream processing. It's the same problem which `InputStreamBuffer`-type APIs solve in traditional I/O libraries (e.g. I wanna read this file in 1024 bytes at a time…)

I don't know what the API impact is. One principle I have is: _I'm wary of all signals which capture mutable state that isn't represented in the signal's streamed data_. I become more wary as that signal's lifetime becomes longer or less determinate, and as it escapes module boundaries. If we come to believe that traditional constant-memory stream processing type stuff needs a bunch of state, I'd love to be able to see the boundaries clearly.

Not-very-thought-through straw-man: maybe all the stream processing stuff is best represented as manipulations of `GeneratorType`s. That's basically what `RACStream` was, after all…

RuiAAPeres:
@mdiep should we merge this?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1031,6 +1031,35 @@ extension SignalType {
 	}
 }
 
+extension SignalType where Value: Hashable {
+	/// Forwards only those values from `self` that are unique across the set of
+	/// all values that have been seen.
+	/// Note: This causes values that are forwarded to be retained to check for
+	/// uniquness.
```
mdiep: Small typo here: should be `uniqueness`.

mdiep:
@nixterrimus Sorry for the long delay in this! I got sidetracked by the discussion and forgot to come back to the actual PR. ☺️ 

Is there any chance you'd be willing to update this to take [@nachosoto's suggestion](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2715#issuecomment-185912504)?

I think we should merge this. It is a useful operator as things stand today, and its memory imprint is less than `collect`. Adding a version that takes a keying function will enable people to further improve the memory usage.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1031,6 +1031,35 @@ extension SignalType {
...
+			return self
+				.filter { value in return seenValues.withValue { !$0.contains(value) } }
+				.observe { event in
+					switch event {
+					case let .Next(value):
+						seenValues.modify { set in
+							var mutableSet = set
+							mutableSet.insert(value)
+							return mutableSet
+						}
```
mdiep: I think there's a race condition here between the `filter` and `observe`. Can't a repeat of a value could make it past the `filter` before the first occurrence makes it into `observe`?

I think we need to filter and modify the seen values in a single atomic operation.

mdiep:
Superseded by #2884.

PR #2718, opened 2016-02-11
make skipRepeats not capture previous nexts

trifonov-ivan:
Original implementation of skipRepeats holds strong reference of previous next because of combinePrevious internal state. In case of objects pipeline it may be significant memory issue

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -724,16 +724,16 @@ extension SignalType {
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func skipRepeats(isRepeat: (Value, Value) -> Bool) -> Signal<Value, Error> {
 		return self
-			.map(Optional.init)
-			.combinePrevious(nil)
-			.filter { a, b in
-				if let a = a, b = b where isRepeat(a, b) {
-					return false
-				} else {
-					return true
+			.scan((nil, false)) { (accumulated: (Value?, Bool), next: Value) -> (Value?, Bool) in
+				switch accumulated.0 {
+				case .None:
+					return (next, false)
+				case let .Some(prev):
+					return isRepeat(prev, next) ? (prev, true) : (Optional.Some(next), false)
 				}
-			}
-			.map { $0.1! }
+			}.filter { $0.1 == false }
+			.map { $0.0 }
+			.ignoreNil()
 	}
 
 	/// Does not forward any values from `self` until `predicate` returns false,
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -724,15 +724,18 @@ extension SignalType {
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func skipRepeats(isRepeat: (Value, Value) -> Bool) -> Signal<Value, Error> {
 		return self
-			.scan((nil, false)) { (accumulated: (Value?, Bool), next: Value) -> (Value?, Bool) in
+			.scan((nil, false)) { (accumulated: (Value?, Bool), next: Value) -> (value: Value?, repeated: Bool) in
 				switch accumulated.0 {
 				case .None:
 					return (next, false)
-				case let .Some(prev):
-					return isRepeat(prev, next) ? (prev, true) : (Optional.Some(next), false)
+				case let .Some(prev) where isRepeat(prev, next):
+					return (prev, true)
+				case .Some:
+					return (Optional.Some(next), false)
 				}
-			}.filter { $0.1 == false }
-			.map { $0.0 }
+			}
+			.filter { !$0.repeated }
+			.map { $0.value }
 			.ignoreNil()
 	}
 
--- a/ReactiveCocoaTests/Swift/SignalSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalSpec.swift
@@ -625,6 +625,40 @@ class SignalSpec: QuickSpec {
 				observer.sendNext("d")
 				expect(values) == [ "a", "cc", "d" ]
 			}
+
+			it("should not store strong reference to previously passed items") {
+				var disposedItems: [Bool] = []
+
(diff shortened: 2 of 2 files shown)


ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -625,6 +625,40 @@ class SignalSpec: QuickSpec {
 				observer.sendNext("d")
 				expect(values) == [ "a", "cc", "d" ]
 			}
+
+			it("should not store strong reference to previously passed items") {
+				var disposedItems: [Bool] = []
+
+				struct Item {
```
NachoSoto: It doesn't make sense to talk about "strong references" with a struct, which has value semantics. I would change this to a `class`.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -625,6 +625,40 @@ class SignalSpec: QuickSpec {
...
+					let payload: Bool
+					let disposable: ScopedDisposable
+				}
+
+				func item(payload: Bool) -> Item {
+					return Item(
+						payload: payload,
+						disposable: ScopedDisposable(ActionDisposable { disposedItems.append(payload) })
```
NachoSoto: Nice idea!

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -625,6 +625,40 @@ class SignalSpec: QuickSpec {
...
+				expect(disposedItems) == [ true ]
+
+				observer.sendNext(item(false))
+				expect(disposedItems) == [ true, false ]
+
+				observer.sendNext(item(true))
+				expect(disposedItems) == [ true, false, false ]
+
+				observer.sendCompleted()
+				expect(disposedItems) == [ true, false, false, true ]
```
NachoSoto: Awesome!

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -724,16 +724,16 @@ extension SignalType {
...
+			.scan((nil, false)) { (accumulated: (Value?, Bool), next: Value) -> (Value?, Bool) in
+				switch accumulated.0 {
+				case .None:
+					return (next, false)
+				case let .Some(prev):
+					return isRepeat(prev, next) ? (prev, true) : (Optional.Some(next), false)
 				}
-			}
-			.map { $0.1! }
+			}.filter { $0.1 == false }
```
NachoSoto: This `filter` should go in its own line

NachoSoto:
Nice find! :star2:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -724,16 +724,16 @@ extension SignalType {
...
-			.filter { a, b in
-				if let a = a, b = b where isRepeat(a, b) {
-					return false
-				} else {
-					return true
+			.scan((nil, false)) { (accumulated: (Value?, Bool), next: Value) -> (Value?, Bool) in
+				switch accumulated.0 {
+				case .None:
+					return (next, false)
+				case let .Some(prev):
```
NachoSoto: I think this can be clearer if you separate both cases in separate `case`s:
```swift
case .None:
  return (next, false)
case let .Some(prev) where isRepeat(prev, next):
  return (prev, true)
default: 
  return (.Some(next), false)
}
```

ReactiveCocoa/Swift/Signal.swift
NachoSoto: Note that I gave names to the tuple values above. That would make this clearer:
```swift
.filter { !$0.repeated }
```

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -724,16 +724,16 @@ extension SignalType {
...
+					return isRepeat(prev, next) ? (prev, true) : (Optional.Some(next), false)
 				}
-			}
-			.map { $0.1! }
+			}.filter { $0.1 == false }
+			.map { $0.0 }
```
NachoSoto: And this:
```swift
.map { $0.value }
```

ReactiveCocoaTests/Swift/SignalSpec.swift
trifonov-ivan: we may talk about strong refs in context of number of copies - structure hold strong ref to internal ScopedDisposable - and since someone holds copy of structure - this disposable will not be disposed.
We may change it to class and move this logic to deinit - but disposed items is only scope variable and it will be harder to handle insertion in deinit.

ReactiveCocoaTests/Swift/SignalSpec.swift
trifonov-ivan: According to combinePrevious in implementation of skipRepeats - it hold previous item.
In original implementation - all disposedItems steps has one less element.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -724,16 +724,19 @@ extension SignalType {
...
 	public func skipRepeats(isRepeat: (Value, Value) -> Bool) -> Signal<Value, Error> {
 		return self
-			.map(Optional.init)
-			.combinePrevious(nil)
-			.filter { a, b in
-				if let a = a, b = b where isRepeat(a, b) {
-					return false
-				} else {
-					return true
+			.scan((nil, false)) { (accumulated: (Value?, Bool), next: Value) -> (value: Value?, repeated: Bool) in
```
NachoSoto: Maybe remove the unnecessary types here?
```swift
.scan((nil, false)) { accumulated, next -> (value: Value?, repeated: Bool) in
```

ReactiveCocoaTests/Swift/SignalSpec.swift
NachoSoto: No you can keep the implementation as is, I just think it would be clearer if the container type (`Item`) had reference semantics itself.

NachoSoto:
Merging this, thank you @trifonov-ivan! 
clapping

PR #2723, opened 2016-02-12
Add variableTimer.

inamiy:
This pull request adds `variableTimer()` feature which allows to change `timeInterval` over time using `nextInterval: NSTimeInterval -> NSTimeInterval?` closure.
This is useful when one wants to slow down timer as time passes, 
and it also allows timer itself to automatically `sendCompleted` by returning nil in `nextInterval`.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -804,6 +804,59 @@ public func timer(interval: NSTimeInterval, onScheduler scheduler: DateScheduler
 	}
 }
 
+/// Creates a repeating timer of the variable interval starting with `nextInterval(0)`,
+/// with a reasonable default leeway, sending updates on the given scheduler.
+///
+/// This timer will complete when `nextInterval` returns nil.
+@warn_unused_result(message="Did you forget to call `start` on the producer?")
+public func variableTimer(onScheduler scheduler: DateSchedulerType, nextInterval: NSTimeInterval -> NSTimeInterval?) -> SignalProducer<NSDate, NoError> {
+	guard let interval = nextInterval(0) else {
+		return .empty
+	}
+
+	// Apple's "Power Efficiency Guide for Mac Apps" recommends a leeway of
+	// at least 10% of the timer interval.
+	return variableTimer(onScheduler: scheduler, withLeeway: interval * 0.1, nextInterval: nextInterval)
+}
+
+/// Creates a repeating timer of the variable interval starting with `nextInterval(0)`,
+/// sending updates on the given scheduler.
+///
+/// This timer will complete when `nextInterval` returns nil.
+@warn_unused_result(message="Did you forget to call `start` on the producer?")
+public func variableTimer(onScheduler scheduler: DateSchedulerType, withLeeway leeway: NSTimeInterval, nextInterval: NSTimeInterval -> NSTimeInterval?) -> SignalProducer<NSDate, NoError> {
+
+	func startTimer(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType, withLeeway leeway: NSTimeInterval, nextInterval: NSTimeInterval -> NSTimeInterval?, startDate: NSDate, observer: Observer<NSDate, NoError>, serialDisposable: SerialDisposable) -> Disposable {
+		return timer(interval, onScheduler: scheduler, withLeeway: leeway)
+			.startWithNext { [unowned serialDisposable] date in
+				observer.sendNext(date)
+
+				let elapsedTime = date.timeIntervalSinceDate(startDate)
+				if let preferredInterval = nextInterval(elapsedTime) {
+					if preferredInterval != interval {
+						serialDisposable.innerDisposable = startTimer(preferredInterval, onScheduler: scheduler, withLeeway: leeway, nextInterval: nextInterval, startDate: startDate, observer: observer, serialDisposable: serialDisposable)
+					}
+				} else {
+					observer.sendCompleted()
+				}
+			}
+	}
+
+	return SignalProducer { observer, disposable in
+		guard let interval = nextInterval(0) else {
+			observer.sendCompleted()
+			return
+		}
+
+		let startDate = scheduler.currentDate
+		let serialDisposable = SerialDisposable()
+
+		serialDisposable.innerDisposable = startTimer(interval, onScheduler: scheduler, withLeeway: leeway, nextInterval: nextInterval, startDate: startDate, observer: observer, serialDisposable: serialDisposable)
+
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -804,6 +804,59 @@ public func timer(interval: NSTimeInterval, onScheduler scheduler: DateScheduler
 	}
 }
 
+/// Creates a repeating timer of the variable interval starting with `nextInterval(0)`,
+/// with a reasonable default leeway, sending updates on the given scheduler.
+///
+/// This timer will complete when `nextInterval` returns nil.
+@warn_unused_result(message="Did you forget to call `start` on the producer?")
+public func variableTimer(onScheduler scheduler: DateSchedulerType, nextInterval: NSTimeInterval -> NSTimeInterval?) -> SignalProducer<NSDate, NoError> {
```
NachoSoto: Interesting operator!

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -804,6 +804,59 @@ public func timer(interval: NSTimeInterval, onScheduler scheduler: DateScheduler
 	}
 }
 
+/// Creates a repeating timer of the variable interval starting with `nextInterval(0)`,
+/// with a reasonable default leeway, sending updates on the given scheduler.
```
NachoSoto: > With a reasonable leeway

I would parametrize that like `timer` does

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -804,6 +804,59 @@ public func timer(interval: NSTimeInterval, onScheduler scheduler: DateScheduler
...
+///
+/// This timer will complete when `nextInterval` returns nil.
```
NachoSoto: I don't think it's worth introducing yet another way of completing a producer :-1:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -804,6 +804,59 @@ public func timer(interval: NSTimeInterval, onScheduler scheduler: DateScheduler
...
+/// Creates a repeating timer of the variable interval starting with `nextInterval(0)`,
+/// sending updates on the given scheduler.
+///
+/// This timer will complete when `nextInterval` returns nil.
+@warn_unused_result(message="Did you forget to call `start` on the producer?")
+public func variableTimer(onScheduler scheduler: DateSchedulerType, withLeeway leeway: NSTimeInterval, nextInterval: NSTimeInterval -> NSTimeInterval?) -> SignalProducer<NSDate, NoError> {
+
+	func startTimer(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType, withLeeway leeway: NSTimeInterval, nextInterval: NSTimeInterval -> NSTimeInterval?, startDate: NSDate, observer: Observer<NSDate, NoError>, serialDisposable: SerialDisposable) -> Disposable {
+		return timer(interval, onScheduler: scheduler, withLeeway: leeway)
+			.startWithNext { [unowned serialDisposable] date in
```
NachoSoto: Why `unowned serialDisposable`?

NachoSoto:
I added a few comments on the implementation.
I'm also curious to hear what others think about this operator!

ReactiveCocoa/Swift/SignalProducer.swift
inamiy: Oops, it's `@noescape` closure so it's actually not needed :sweat:

ReactiveCocoa/Swift/SignalProducer.swift
inamiy: There is also `public func variableTimer(onScheduler scheduler: DateSchedulerType, withLeeway leeway: NSTimeInterval, nextInterval: NSTimeInterval -> NSTimeInterval?) -> SignalProducer<NSDate, NoError>` function, so `leeway` is already parametrized.

But I'm wondering if we should accept fixed leeway for variable timer intervals. Maybe replacing to `leewayFactor` (default = 0.1) will be better.

ReactiveCocoa/Swift/SignalProducer.swift
inamiy: Is there any other good way of completing timer at some point?
I find `timeoutWithError` or manual disposal a bit cumbersome...

inamiy:
@NachoSoto 
Thank you for review! 
I would also like to hear from others and will improve code once this feature is good to go.

ReactiveCocoa/Swift/SignalProducer.swift
iv-mexx: > But I'm wondering if we should accept fixed leeway for variable timer intervals. Maybe replacing to leewayFactor (default = 0.1) will be better.

That might be a good idea (for `timer` as well). 
I'm wondering if thats also the reason that there are two `timer` functions, one with a default leeway and one with leeway as a parameter instead of using a default value for the second one and saving the variant with default leeway? 
By changing it to `leewayFactor`, it could be changed to 

```public func timer(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType, withLeeway leewayFactor: Float = 0.1) -> SignalProducer<NSDate, NoError> {```

and same for `variableTimer` I think.

File: ReactiveCocoa/Swift/SignalProducer.swift (hunk 3 above)
Comment (iv-mexx) [inline ReactiveCocoa/Swift/SignalProducer.swift]: Hmm, how about a `variableTimer` operator: 

```swift
extension SignalType where Value: NSTimeInterval {
  public func variableTimer(onScheduler scheduler: DateSchedulerType) -> SignalProducer<NSDate, NoError> {
    // Time-Interval is received as next-event
    // Termination of timer as usual by sending `Completed`, `Interrupted` oder `Error` to the timer
  }
}
```

just brainstorming here...

ReactiveCocoa/Swift/SignalProducer.swift
inamiy: @iv-mexx I agree :+1:

ReactiveCocoa/Swift/SignalProducer.swift
inamiy: @iv-mexx 
Sounds interesting! But things will get more complicated because this `SignalType` will emit timeIntervals with time-dependency. There might be a better solution from this approach, though ;)

filblue:
My 5 cents - don't feel like this operator belongs to the core. This is some kind of non-trivially implemented easing utility built on RAC which unlikely is going to bring any profit to the majority of users.

inamiy:
@filblue Thanks for your feedback.

This proposal seems not much interesting to many developers, so I will close now.

PR #2732, opened 2016-02-18
Replace increment operator ++ with += 1

cxa:
because of ++ is deprecated in Swift 2.2 and will be removed in Swift 3


NachoSoto:
Thank you!

PR #2739, opened 2016-02-24
Add inclusive `takeWhile` on `Signal` and `SignalProducer`

Rupert-RR:
Sometimes we want to be able to listen to a signal until a value conforms to certain criteria but we want to include this value in the stream before completing.

This PR adds an optional parameter to the `takeWhile` methods on `Signal` and `SignalProducer` to allow us to specify that we want to include the `Value` which fails `predicate`, while (hopefully) maintaining backward compatibility and defaulting to the original non-inclusive behaviour.

Let me know if this is good and if anything should be done differently.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -827,10 +827,13 @@ extension SignalType {
 	/// Forwards any values from `self` until `predicate` returns false,
 	/// at which point the returned signal will complete.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func takeWhile(predicate: Value -> Bool) -> Signal<Value, Error> {
+	public func takeWhile(inclusive: Bool = false, _ predicate: Value -> Bool) -> Signal<Value, Error> {
 		return Signal { observer in
 			return self.observe { event in
 				if case let .Next(value) = event where !predicate(value) {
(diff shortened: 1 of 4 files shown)


NachoSoto:
I'm pretty :-1: on having methods with implementations that depend on an input boolean. I'd much rather have a separate name for this operator, but in that case, what would you call it?

Rupert-RR:
All I can think of is the very uninspired `takeUntilInclusive`!

However, while it might make sense to expose separate public methods, surely it doesn't make sense to duplicate functionality in private implementations?

mdiep:
> I'd much rather have a separate name for this operator, but in that case, what would you call it?

`takeThrough`?

sharplet:
How about `takeUpto`?

mdiep:
The Swift stdlib uses `upTo` to mean exclusive and `through` to mean inclusive. So I think `upTo` is a bad idea.

mdiep:
Closing this since it's stale.

Rupert-RR:
Stale in what sense?

RuiAAPeres:
@Rupert-RR you opened this PR on the 24th of February.

mdiep:
@Rupert-RR There are merge conflicts and unaddressed feedback. Update the PR and we'll still consider it!

PR #2747, opened 2016-03-03
rac_notifications shouldn't strongly capture the observed object

andymatuschak:
Other Observer-pattern-style APIs (e.g. `rac_valuesForKey`) weakly reference the observed object; I was surprised to find that `rac_notifications` extended its lifetime.

As an aside (not covered in this revision), I was surprised to find that the subscription wasn’t terminated when the observed object deallocated. I guess that’s because the API supports `AnyObject`, and we can’t observe any arbitrary object for deallocation? We could make a `Signal`-returning variant of this method—with a required `NSObject` observed object argument—which could automatically terminate the underlying effects correctly.

### DIFF
--- a/ReactiveCocoaTests/Swift/FoundationExtensionsSpec.swift
+++ b/ReactiveCocoaTests/Swift/FoundationExtensionsSpec.swift
@@ -14,9 +14,9 @@ import ReactiveCocoa
 class FoundationExtensionsSpec: QuickSpec {
 	override func spec() {
 		describe("NSNotificationCenter.rac_notifications") {
+			let center = NSNotificationCenter.defaultCenter()
 
 			it("should send notifications on the producer") {
-				let center = NSNotificationCenter.defaultCenter()
 				let producer = center.rac_notifications("rac_notifications_test")
 
 				var notif: NSNotification? = nil
@@ -34,6 +34,21 @@ class FoundationExtensionsSpec: QuickSpec {
 				center.postNotificationName("rac_notifications_test", object: nil)
 				expect(notif).to(beNil())
 			}
+
+			it("should send Interrupted when the observed object is freed") {
+				var observedObject: AnyObject? = NSObject()
+				let producer = center.rac_notifications(object: observedObject)
+				observedObject = nil
+
+				var interrupted = false
+				let disposable = producer.startWithInterrupted {
+					interrupted = true
+				}
+				expect(interrupted).to(beTrue())
+
+				disposable.dispose()
+			}
+
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


NachoSoto:
> I guess that’s because the API supports `AnyObject`

I think the only reason for that is because `NSNotificationCenter`'s API takes `(id)object` instead of `NSObject`, but I bet internally they do assume it's an `id<NSObject>` (`NSObjectProtocol`), like most other `Foundation` APIs. So that goes to say, I think it would be safe (though technically a breaking change) to change `rac_notifications` to take `object: NSObjectProtocol` instead. That way we can observe its lifetime.

ReactiveCocoaTests/Swift/FoundationExtensionsSpec.swift
```diff
@@ -34,6 +34,21 @@ class FoundationExtensionsSpec: QuickSpec {
...
+			it("should send Interrupted when the observed object is freed") {
+				var observedObject: AnyObject? = NSObject()
+				let producer = center.rac_notifications(object: observedObject)
+				observedObject = nil
+
+				var interrupted = false
+				let disposable = producer.startWithInterrupted {
+					interrupted = true
+				}
+				expect(interrupted).to(beTrue())
```
NachoSoto: Awesome

NachoSoto:
Awesome, thanks for doing this @andymatuschak! This is already an improvement, but I do wonder, like we spoke, if adding `takeUntil(object.rac_willDeallocSignal())` to the returned `SignalProducer` would make the semantics better. I would personally vote for that.

andymatuschak:
I'd definitely love to do that if everyone's on-board with restricting the `object` argument to `NSObject` (can't be `NSObjectProtocol` because `willDeallocSignal` is implemented as an extension on `NSObject`).

---

I'd separately love to push on the idea of these kinds of APIs returning `SignalProducer`s instead of `Signal`s—it seems to me that there's a crystal-clear lifetime semantic (i.e. `removeObserver` if the observed object or the `Signal` itself `deinit`), and I don't see why the client should ever care about the implementation details.

andymatuschak:
cc @ReactiveCocoa/reactivecocoa on API design discussion above

mdiep:
> I'd definitely love to do that if everyone's on-board with restricting the object argument to NSObject

I'm definitely on-board with that. :+1: The only question in my mind is whether we should (1) ship this as a 4.X, (2) ship it as 5.0, or (3) branch for 5.0 and include it there.

andymatuschak:
If we're trying to be semver we should branch for 5

robb:
How about giving the new one a different name and deprecating the old one? That way we wouldn't have to branch out for 5.

mdiep:
Or we could merge this as-is and release it as 5.0 now with the Swift 2.2 changes. :relaxed:

ikesyo:
Superseded by #2859 (and #2854).

PR #2750, opened 2016-03-04
Add ScopedCompositeDisposable

victorlin:
When I was developing apps with ReactiveCocoa 4, I try to save disposable in a property of the controller, would be like this

```Swift
class MyViewController: UIViewController {
    var bindingDisposable: CompositeDisposable?
    let imageLoader = ImageLoader()

    @IBOutlet weak var myImage: UIImageView!
    @IBOutlet weak var myLabel: UILabel!

    func bindViewModel(viewModel: MyViewModel) {
        self.bindingDisposable?.dispose()
        self.bindingDisposable = CompositeDisposable()
        self.bindingDisposable += (DynamicProperty(object: myImage, keyPath: "image") <~ viewModel.image.producer
            .loadForBinding { [unowned self] url in
                return self.imageLoader.loadImage(url)
            }
        )
        self.bindingDisposable += (DynamicProperty(object: myLabel, keyPath: "text") <~ viewModel.text)
    }
}
```

so that you can dispose all the bindings easily. However, there is a problem in the code shown above, that is if the `loadForBinding` method was called after `MyViewController` got destroyed, as the `unowned self` will be nil, that will raise accessing `self` as a nil error, and end up crashing the app.

To solve the problem, I was trying to use `ScopedDisposable` along with `CompositeDisposable` like this

```Swift
self.bindingDisposable = ScopedDisposable(disposable)
disposable += (DynamicProperty(object: myLabel, keyPath: "text") <~ viewModel.text)
...
```

But it's pretty verbose, so I think it would be nice to have a `ScopedCompositeDisposable`, with that, it would be way easier to manage all binding lifecycle with the `UIViewController`, given the example, just need to change the `bindingDisposable` type to `ScopedCompositeDisposable`, and it's all done.

### DIFF
--- a/ReactiveCocoa/Swift/Disposable.swift
+++ b/ReactiveCocoa/Swift/Disposable.swift
@@ -175,33 +175,30 @@ public final class ScopedDisposable: Disposable {
 	}
 }
 
+
 /// A disposable that, upon deinitialization, will automatically dispose of
-/// any number of other disposables.
-public final class ScopedCompositeDisposable: Disposable {
-    private let scoppedDisposable = ScopedDisposable(CompositeDisposable())
-	private var compositeDisposable: CompositeDisposable {
-		return scoppedDisposable.innerDisposable as! CompositeDisposable
-	}
-
-	public init() {}
-    
-    public var disposed: Bool {
-        return scoppedDisposable.disposed
-    }
+/// another disposable.
+public final class GenericScopedDisposable<DisposableType: Disposable>: Disposable {
+	/// The disposable which will be disposed when the ScopedDisposable
+	/// deinitializes.
+	public let innerDisposable: DisposableType
 	
-    public func dispose() {
-        scoppedDisposable.dispose()
-    }
-	
-	/// Adds the given disposable to the list, then returns a handle which can
-	/// be used to opaquely remove the disposable later (if desired).
-	public func addDisposable(d: Disposable?) -> CompositeDisposable.DisposableHandle {
-		return compositeDisposable.addDisposable(d)
+	public var disposed: Bool {
+		return innerDisposable.disposed
 	}
-
-	/// Adds an ActionDisposable to the list.
-	public func addDisposable(action: () -> ()) -> CompositeDisposable.DisposableHandle {
-		return addDisposable(ActionDisposable(action: action))
+	
+	/// Initializes the receiver to dispose of the argument upon
+	/// deinitialization.
+	public init(_ disposable: DisposableType) {
+		innerDisposable = disposable
+	}
+	
+	deinit {
+		dispose()
+	}
+	
+	public func dispose() {
+		innerDisposable.dispose()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


NachoSoto:
Hmm I'm not sure this warrants creating a new type. That's the power of composition: you can create all possible permutations of N types without needing to introduce a type for each one. This one is just creating a redundant type for one of those permutations, right?

victorlin:
@NachoSoto Yes, I know it's always nice to do composition instead of providing all possible compositions, however this case it's a really good sugar from syntax perspective to have for promoting the right way of managing disposables.

victorlin:
@NachoSoto but yet it makes sense to keep the core slime, I am actually thinking about to create a new project that extends ReactiveCocoa for some practical stuff, like this `ScopedCompositeDisposable`, and also the `loadForBinding` method (it basically loads the resource, suppress error and transform the return value into AnyObject?, to make it super easy to bind with `DynamicProperty`). We have these stuff in our project, I think it would be nice if I open source it so that people don't need to build their own stuff again and again.

NachoSoto:
I'm just saying that you can accomplish the same thing without introducing a new type. What we can do is parametrize `ScopedDisposable` over the inner `Disposable` type so that you can do this with just an initializer in an extension.

victorlin:
@NachoSoto okay, I see, that's a good idea. Will try to modify it to work that way

victorlin:
@NachoSoto I tried it like this

```Swift
/// A disposable that, upon deinitialization, will automatically dispose of
/// another disposable.
public final class GenericScopedDisposable<DisposableType: Disposable>: Disposable {
	/// The disposable which will be disposed when the ScopedDisposable
	/// deinitializes.
	public let innerDisposable: DisposableType
	
	public var disposed: Bool {
		return innerDisposable.disposed
	}
	
	/// Initializes the receiver to dispose of the argument upon
	/// deinitialization.
	public init(_ disposable: DisposableType) {
		innerDisposable = disposable
	}
	
	deinit {
		dispose()
	}
	
	public func dispose() {
		innerDisposable.dispose()
	}
}

/// Adds the right-hand-side disposable to the left-hand-side
/// `ScopedDisposable<CompositeDisposable>`.
///
///     disposable += producer
///         .filter { ... }
///         .map    { ... }
///         .start(observer)
///
public func +=<T: CompositeDisposable>(lhs: GenericScopedDisposable<T>, rhs: Disposable?) -> CompositeDisposable.DisposableHandle {
	return lhs.innerDisposable += rhs
}
```

I found myself seems not possible to break `ScopedDisposable` backward compatibility, as I cannot do something like

```
let disposable = ScopedDisposable()
```

as it asks you to provide the type, any idea how can I work around this? it seems there is no `default type parameter` for generics in Swift?

mdiep:
@victorlin We're currently making plans for RAC 5. If you're interested in pursing this, we could make a breaking change to make disposables generic for RAC 5. But that would take some work—I'm not sure what the APIs would look like or how that might affect the RAC codebase.

mdiep:
I'm going to close this because we don't want to add another class to handle this.

If you're interested in pursuing generic disposables, we'd be very interested in that!

victorlin:
@mdiep I am kind of interested in making disposable more generic, is there timeline for RAC 5 yet?

mdiep:
I'm hoping to ship 4.3 in the next release and target 5.0 for a month or two from now. But it will depend on how long the work takes. If this isn't done for RAC 5, we can try it for RAC 6. (I'm think that it doesn't make sense to wait to ship RAC 5 until Swift 3 is ready.)

PR #2752, opened 2016-03-05
Add a withCurrent option to timer

CosynPa:
I found it's common to start a timer with current time immediately. But implementing this with operators is difficult, since there is no `merge` operator for `SignalProducer`. So I add a withCurrent option.


CosynPa:
Can be accomplished by the concat operator.

PR #2759, opened 2016-03-20
New Relic compatibility

soxjke:
### Abstract

Fixed NewRelic compatibility

### Root cause

NewRelicAgent asserts when _cmd in profiled method is not equal it's name. rac_signalForSelector: introduces "rac_alias_" prefix for swizzled selectors, and it causes NewRelic agent to assert.

### Fix

Synchronously sets back an original implementation for selector while invoking

### Affected areas

3d-party race conditions on runtime calls, performance of method invocation due to additional synchronization against NSObjectRACSignalForSelector() performance gap

### Development testing done

Tested on unit tests and smoke on project used rac_signalForSelector: simultaneously with NewRelicAgent

### DIFF
--- a/ReactiveCocoa/NSObject+RACSelectorSignal.m
+++ b/ReactiveCocoa/NSObject+RACSelectorSignal.m
@@ -40,14 +40,19 @@ static NSMutableSet *swizzledClasses() {
 @implementation NSObject (RACSelectorSignal)
 
 static BOOL RACForwardInvocation(id self, NSInvocation *invocation) {
-	SEL aliasSelector = RACAliasForSelector(invocation.selector);
+    SEL originalSelector = invocation.selector;
+	SEL aliasSelector = RACAliasForSelector(originalSelector);
 	RACSubject *subject = objc_getAssociatedObject(self, aliasSelector);
 
 	Class class = object_getClass(invocation.target);
 	BOOL respondsToAlias = [class instancesRespondToSelector:aliasSelector];
 	if (respondsToAlias) {
-		invocation.selector = aliasSelector;
-		[invocation invoke];
+        @synchronized(self) {
+            IMP implementation = method_getImplementation(class_getInstanceMethod(class, aliasSelector));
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/NSObject+RACSelectorSignal.m
```diff
@@ -40,14 +40,19 @@
 @implementation NSObject (RACSelectorSignal)
 
 static BOOL RACForwardInvocation(id self, NSInvocation *invocation) {
-	SEL aliasSelector = RACAliasForSelector(invocation.selector);
+    SEL originalSelector = invocation.selector;
```
NachoSoto: Indentation here is wrong.

ReactiveCocoa/NSObject+RACSelectorSignal.m
```diff
@@ -40,14 +40,19 @@
...
 	Class class = object_getClass(invocation.target);
 	BOOL respondsToAlias = [class instancesRespondToSelector:aliasSelector];
 	if (respondsToAlias) {
-		invocation.selector = aliasSelector;
-		[invocation invoke];
+        @synchronized(self) {
+            IMP implementation = method_getImplementation(class_getInstanceMethod(class, aliasSelector));
+            method_setImplementation(class_getInstanceMethod(class, originalSelector), implementation);
+            [invocation invoke];
+            method_setImplementation(class_getInstanceMethod(class, originalSelector), _objc_msgForward);
```
NachoSoto: I wonder what the performance implication of this 2 extra calls is.

mdiep: Yeah, I think there should be some benchmarking to understand the performance implications before merging.

mdiep:
> NewRelicAgent asserts when cmd in profiled method is not equal it's name. rac_signalForSelector: introduces "rac_alias" prefix for swizzled selectors, and it causes NewRelic agent to assert.

Maybe New Relic should remove this assert?

mdiep:
Closing this because the feedback hasn't been addressed.

Issue #2751, opened 2016-03-04
Xcode 7.3 beta 5 miscompiles `SignalProducerType.combineLatestWith` on ARM64


jckarter:
I hate to incovenience you all, but we have a bug in the 7.3 betas that causes us to miscompile the currying of `Signal.combineLatestWith` here:

```swift
	public func combineLatestWith<U>(otherProducer: SignalProducer<U, Error>) -> SignalProducer<(Value, U), Error> {
		return liftRight(Signal.combineLatestWith)(otherProducer)
	}
```

I'm investigating a fix right now. In case we don't fix it in time for 7.3 final, I recommend hand-inlining the `liftRight` combinator here as a workaround, 

```swift
	public func combineLatestWith<U>(otherProducer: SignalProducer<U, Error>) -> SignalProducer<(Value, U), Error> {	
		return SignalProducer { observer, outerDisposable in
			self.startWithSignal { signal, disposable in
				outerDisposable.addDisposable(disposable)
					
				otherProducer.startWithSignal { otherSignal, otherDisposable in
					outerDisposable.addDisposable(otherDisposable)
						
					signal.combineLatestWith(otherSignal).observe(observer)
				}
			}
		}
	}
```

The bug on our side is rdar://problem/24960559, for reference.

NachoSoto:
I can't even begin to express how excited I am to see a Swift contributor open an issue, not only explaining the bug but offering a work around :heart_eyes: 

If it's not fixed by the GM we can temporarily make this change in #2684, since we'll likely need to change other things for `Swift 2.2`.

Thanks a lot @jckarter :smile:

mdiep:
> I can't even begin to express how excited I am to see a Swift contributor open an issue, not only explaining the bug but offering a workaround

Yes! Thank you!!! :heart_eyes:

calebd:
@jckarter Is there an issue on bugs.swift.org that we can follow as well?

jckarter:
Not that I know of. The radar unfortunately contains a private project I can't post publicly.

NachoSoto:
@calebd reports that this is still an issue on the final 7.3. I'm finishing the download now, and I'll make a branch with 2.2 changes (probably including this _hopefully_ temporary workaround) and try to make a release by EOD today.

NachoSoto:
[...]

Issue #141, opened 2012-11-27
Why is a separate version of EXTKeyPathCoding used for OS X?


jwilling:
This is causing problems with CocoaPods. https://github.com/CocoaPods/Specs/pull/804

PR #2760, opened 2016-03-21
Swift 2.2 support

### LINKED ISSUE #2684: Swift 2.2 support
I've started working on Swift 2.2 support (which we'll probably add in `RAC 4.1`). Right now I'm blocked by [this issue](https://github.com/Quick/Nimble/pull/243) in Nimble though.

NachoSoto:
Fixes #2684 and #2751.

All tests pass except for `_sequenceWithHeadBlock_tailBlock___behaves_like_a_sequence__RACSequenceExamples__should_fold_right` in `RACSequenceSpec.m:90`. It might be related to 2fc59b7bcc32455815dcd9938a7362b59cc360af (I made that change because having `qck_itBehavesLike` inside of `qck_afterEach` is illegal now), but I'm not sure if that changed the semantics. Either way, all Swift tests do pass.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -134,7 +134,7 @@ public final class Action<Input, Output, Error: ErrorType> {
 public final class CocoaAction: NSObject {
 	/// The selector that a caller should invoke upon a CocoaAction in order to
 	/// execute it.
-	public static let selector: Selector = "execute:"
+	public static let selector: Selector = #selector(CocoaAction.execute(_:))
 
 	/// Whether the action is enabled.
 	///
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode7.1
+osx_image: xcode7.3
 before_install: true
 install: true
 git:
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,3 +1,3 @@
 github "jspahrsummers/xcconfigs" "ec5753493605deed7358dec5f9260f503d3ed650"
-github "Quick/Quick" ~> 0.8
-github "Quick/Nimble" ~> 3.1
+github "Quick/Quick" ~> 0.9.1
+github "Quick/Nimble" ~> 3.2
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
-github "Quick/Nimble" "v3.1.0"
-github "Quick/Quick" "v0.8.0"
+github "Quick/Nimble" "v3.2.0"
+github "Quick/Quick" "v0.9.1"
 github "antitypical/Result" "1.0.2"
 github "jspahrsummers/xcconfigs" "ec5753493605deed7358dec5f9260f503d3ed650"
--- a/ReactiveCocoa/Swift/Event.swift
+++ b/ReactiveCocoa/Swift/Event.swift
@@ -131,9 +131,9 @@ extension Event: CustomStringConvertible {
 /// Event protocol for constraining signal extensions
 public protocol EventType {
 	// The value type of an event.
-	typealias Value
+	associatedtype Value
 	/// The error type of an event. If errors aren't possible then `NoError` can be used.
-	typealias Error: ErrorType
+	associatedtype Error: ErrorType
 	/// Extracts the event from the receiver.
 	var event: Event<Value, Error> { get }
 }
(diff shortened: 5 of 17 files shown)


ReactiveCocoa/Swift/Action.swift
```diff
@@ -134,7 +134,7 @@ public final class Action<Input, Output, Error: ErrorType> {
 public final class CocoaAction: NSObject {
 	/// The selector that a caller should invoke upon a CocoaAction in order to
 	/// execute it.
-	public static let selector: Selector = "execute:"
+	public static let selector: Selector = #selector(CocoaAction.execute(_:))
```
mdiep: Are the changes like these required for Swift 2.2? (Are they errors or just warnings?) They're breaking changes with Swift 2.1 / Xcode 7.2.1, so we'll need to release these as RAC 5.0 to follow semver.

(My Xcode download hasn't finished yet.)

NachoSoto: > Are the changes like these required for Swift 2.2? (Are they errors or just warnings?) 
Most of them are not, they're just warnings.

> They're breaking changes with Swift 2.1 / Xcode 7.2.1.
Oh right, good point... we need to make these changes at some point anyway, so I'd vote for not staying behind. There's already a major version available for `Swift 2.1`.

> so we'll need to release these as RAC 5.0 to follow semver.
That's absolutely true, but I don't think it'll be productive to release a new major version of RAC for every Swift version, which given that it's not ABI stable, is by definition always a breaking change. I'd vote for following the same semantic versioning as `Swift`: since this is supposedly just a minor change, I'd say we release this as `ReactiveCocoa 4.1`, and reserve `ReactiveCocoa 5.0` for `Swift 3.0`. Not that there's anything wrong with upgrading to a new major version, but I think it's really confusing. When it comes to keeping backwards compatibility I just think that we don't have enough time to deal with all the pains involved in that, so I'd strongly suggest looking forward :)

NachoSoto:
Like I mentioned, there's [only one failing test](https://travis-ci.org/ReactiveCocoa/ReactiveCocoa/builds/117592674) due to some `Quick` change. I'd love if anyone more familiar with `Objective-C`'s `Quick` can take a quick look, otherwise I'd say that we can safely release this since it doesn't look like a regression, but rather a failure due to `Quick`'s interface (see https://github.com/ReactiveCocoa/ReactiveCocoa/commit/2fc59b7bcc32455815dcd9938a7362b59cc360af).

ReactiveCocoa/Swift/Action.swift
NachoSoto: What do you think @mdiep?

mdiep: Sorry for the delay… I was traveling all day yesterday (with the whole family) and am just catching up on things.

I'm hesitant to give up on semver, but given the way Swift/Xcode work it's not very practical, as you mentioned. I can be okay with breaking changes in minor RAC versions for minor Swift versions. 

But I'd **really** like us to make releases like RAC 4.1 compatible with both Swift 2.1 and Swift 2.2 if possible. Then we could make RAC 4.2 compatible with Swift 2.2 only. Upgrading Xcode versions is a major hassle because _everything breaks_. If we can be compatible with both for a single minor release, I think that will significantly ease the pain for our consumers. (Benefitting both the frameworks and the apps that I maintain.)

But ultimately, you're the one doing the work, so it depends on what you're willing to do.

mdiep: The current release does build with Swift 2.2, so I'm :+1: to be 2.2+ only in the next release.

I do think it should be 5.0 if we're dropping Swift 2.1 compatibility, but I'll defer to you.

mdiep:
:ship:

mdiep:
We should probably see if we can bump `Result` before releasing this.

NachoSoto:
Yup, I'll do as soon as #141 is merged and a new release created.

ReactiveCocoa/Swift/Action.swift
NachoSoto: > The current release does build with Swift 2.2, so I'm :+1: to be 2.2+ only in the next release.

Yeah, that was my thinking too :)

As for the version, I'm torn. In the end it doesn't really matter. In general I'd love if as a community we could all agree on a pattern to follow with this so that we're all on the same page. I'd argue that following the same minor/major change semantics as Swift would be the clearest.

mdiep:
:metal:

MikeSilvis:
@NachoSoto it looks like Result is merged. When are you going to make a new cocoapods release?

Thanks,
Mike

Issue #2761, opened 2016-03-22
ReactiveCocoa2.5  build faild in Xcode Version 7.3 (7D175)


Sdoy:
![3f344c2f-8dbe-4597-b07a-548a0b72c42b](https://cloud.githubusercontent.com/assets/10302939/13940970/93497b58-f01d-11e5-9211-a96927537365.png)

iworkcloud:
I am also seeking solutions

iworkcloud:
Do not put the pre-compiled headers can be

lacklock:
@iworkcloud how?

lacklock:
after  upgraded to 7.3  i have the same problem .

shenhualxt:
I am also seeking solutions

lacklock:
https://forums.developer.apple.com/thread/38934

jjksam:
Can I just replace all the __weak NSObject * to id to solve this ? Not sure about that, but that can make the problem pass the compilation.

lacklock:
@jjksam yes,it's worked,juet delete __weak.

undefined-zola:
http://stackoverflow.com/questions/4728225/is-it-ok-to-remove-prefix-pch-file-from-the-xcode-project

remove pre-compiled headers

lacklock:
@undefined-zola ths , i get it

NagiYan:
edit like this
#if __has_feature(objc_arc)
- (RACSignal *)rac_valuesForKeyPath:(NSString *)keyPath observer:(__weak NSObject *)observer;
#else
- (RACSignal *)rac_valuesForKeyPath:(NSString *)keyPath observer:(NSObject *)observer;
#endif

kernel-lj:
delete the cocoapods location .h file which have __weak, it is a temporary  solutions，the best way is waiting 3rd party libraries update!

Sdoy:
agree @kernel-lj ,the best way is waiting 3rd party libraries update,  because  not sure if it will cause memory leaks after  replace all the __weak NSObject * or use other ways to stop the Xcode complain

willonboy:
@NagiYan thank you !

WithdrwnKing:
why i can't unlock NSObject+RACPropertySubscribing.h.

gcox:
You can also avoid this by not importing ReactiveCocoa in your precompiled header.

gcox:
If removing from your pch file is too painful, you can do this:

```
[...]

PR #2763, opened 2016-03-22
fix compile error in xcode7.3

xinzhengzhang:
#2761

### DIFF
--- a/ReactiveCocoa/NSObject+RACPropertySubscribing.m
+++ b/ReactiveCocoa/NSObject+RACPropertySubscribing.m
@@ -21,7 +21,7 @@
 
 @implementation NSObject (RACPropertySubscribing)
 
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/NSObject+RACPropertySubscribing.m
```diff
@@ -31,7 +31,7 @@ - (RACSignal *)rac_valuesForKeyPath:(NSString *)keyPath observer:(__weak NSObjec
 		setNameWithFormat:@"RACObserve(%@, %@)", self.rac_description, keyPath];
 }
 
-- (RACSignal *)rac_valuesAndChangesForKeyPath:(NSString *)keyPath options:(NSKeyValueObservingOptions)options observer:(__weak NSObject *)weakObserver {
```
NachoSoto: This changes the semantics of this method, so even though it might compile now, this is certainly not the right fix.
The error in #2761 says "in file using manual reference counting", but `ReactiveCocoa` only supports `RAC`.

PR #2764, opened 2016-03-22
Drop `SignalProducer.buffer(1)` in MutableProperty

andersio:
Preceded by @olegshnitko's #2744.

Goals:

Drop the use of `SignalProducer.buffer(1)` in `MutableProperty` without breaking the perceived semantic, and consequently allow signal recursion in MutableProperty.

Caveats:

1. The source of truth of a `MutableProperty` is now a pair of get-set closures which capture an underlying storage. The getter is retained until all generated producers and the property itself have been released.
2. Less locks in `MutableProperty`!
3. Included the signal recursion test case from #2744.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -83,6 +83,15 @@ public protocol MutablePropertyType: class, PropertyType {
 	var value: Value { get set }
 }
 
+/// The state of a `MutableProperty`.
+private final class MutablePropertyState<Value> {
+	var value: Value
+
+	init(_ value: Value) {
+		self.value = value
+	}
+}
+
 /// A mutable property of type `Value` that allows observation of its changes.
 ///
 /// Instances of this class are thread-safe.
@@ -94,7 +103,10 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// `value`. Note that recursive sets will still deadlock because the
 	/// underlying producer prevents sending recursive events.
 	private let lock: NSRecursiveLock
-	private var _value: Value
+
+	/// The underlying storage of the property. It could outlive the
+  /// property if the producer is being retained.
+	private let state: MutablePropertyState<Value>
 
 	/// The current value of the property.
 	///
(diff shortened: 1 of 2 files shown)


NachoSoto:
This is cool, I'll review in a bit, but one question for now: why does this have to depend on #2670? I don't see how the `Atomic` implementation needs to change for this, and I'd rather keep both things separated.

NachoSoto:
Also I thought the goal was to get rid of `SignalProducer.buffer` :)

andersio:
Since `Atomic` is already a reference type and #2670 would likely get a green light, I thought it would be great to reuse the code by just introducing a recursive option in the initialiser.

An alternative implementation would be keeping the `NSRecursiveLock` and a private `Box<Value>` class to decouple the lifetime of the value storage from the `MutableProperty`.

andersio:
This PR has been cleaned up and rebased. :)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -83,6 +83,15 @@ public protocol MutablePropertyType: class, PropertyType {
 	var value: Value { get set }
 }
 
+/// The state of a `MutableProperty`.
+private final class MutablePropertyState<Value> {
```
NachoSoto: Why the extra level of indirection, rather than using `Value` directly?

andersio: It is necessary so that the `MutableProperty.producer` can still access the latest value after the deallocation of the `MutableProperty`. If we let the producer to capture a `Value`, since `Value` is not guaranteed to be a reference type, the producer might end up to have only a snapshot of the value instead of the latest, thus breaking the semantic.

andersio:
I have replaced the wrapper with a getter and a setter closure, which is more concise and kind of inline with the `AnyProperty` implementation.

P.S. there is a failing case in RACSequenceSpec, causing all the crosses above.

andersio:
@NachoSoto May you please take a look on at PR? The flaky test from RAC2 somehow passed again.  😄

ReactiveCocoa/Swift/Property.swift
```diff
@@ -94,7 +94,13 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
 	/// underlying producer prevents sending recursive events.
 	private let lock: NSRecursiveLock
-	private var _value: Value
+
+	/// The getter of the underlying storage, which may outlive the property
+	/// if a returned producer is being retained.
+	private let getter: () -> Value
+
+	/// The setter of the underlying storage.
+	private let setter: Value -> Void
```
dmcrodrigues: To promote code consistency (generally speaking), I would suggest changing the type to:

```swift
private let setter: Value -> ()
```

ReactiveCocoa/Swift/Property.swift
```diff
@@ -112,28 +118,36 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
+	}
 
 	/// Initializes the property with the given value to start.
 	public init(_ initialValue: Value) {
-		_value = initialValue
+		var value = initialValue
 
 		lock = NSRecursiveLock()
-		lock.name = "org.reactivecocoa.ReactiveCocoa.MutableProperty"
+		getter = { value }
```
RuiAAPeres: Is there a reason to drop the lock's name?

RuiAAPeres:
LGTM. Checked both tests with our current implementation:

1. `should yield a producer which emits the latest value and complete even if the property is deallocated` ✅
2. `should not deadlock on recursive ABA observation` ❌

This PR fixes the recursive deadlock (2nd test).

ReactiveCocoa/Swift/Property.swift
andersio: No. I will add it back ASAP. Thanks for reminding me.

ReactiveCocoa/Swift/Property.swift
andersio: I am following the convention of the Swift corelibs and stdlib of using `Void` instead of `()` for function types.

RuiAAPeres:
If the goal is to remove `buffer` (due to not be aligned with the `SignalProducer` semantics #2744 ) I am up for it. 👍

mdiep:
Thanks for the pull request! ✨

This looks good to me. 👍

PR #2766, opened 2016-03-22
Add SignalProducer equivalents for UIKit RACSignal extensions

natestedman:
This pull request adds `SignalProducer` equivalents for `RACSignal` extensions to `UI`-prefixed classes. There are a few exceptions:

- `UIActionSheet` - this class is deprecated.
- `UIAlertView` - this class is deprecated.
- `UIImagePickerController` - I feel like this whole thing deserves a better Swift API than just sending a dictionary of `String` to `AnyObject`. I think this is better handled by a dedicated µframework, especially once the Swift Package Manager goes mainstream and those become very easy to use and consume.

I've added protocol types for each signal-providing category message that is wrapped. The `SignalProducer` implementations are then written as extensions of those protocols. This allows classes providing a similar signal to reuse the same implementation (e.g. `prepareForReuse`), and, more importantly, allows `SignalProducer` `Value` types to be `Self`. For example, a simple implementation of the `UIGestureRecognizer` extension would be:

```swift
extension UIGestureRecognizer {
	/// Returns a signal producer that sends the receiver when its gesture occurs.
	public var gestureProducer: SignalProducer<Self, NoError> {
		return rac_gestureSignal().toSignalProducer()
			.demoteErrors()
			.map { $0 as? Self }
			.ignoreNil()
	}
}
```

That's not valid Swift right now, but it's important to have the actual type of the gesture recognizer, not just `UIGestureRecognizer`, so that subclass properties and methods can be accessed. Instead, using a protocol:

```swift
public protocol GestureSignalProvidingType {
	func rac_gestureSignal() -> RACSignal!
}

extension UIGestureRecognizer: GestureSignalProvidingType {}

extension GestureSignalProvidingType {
	/// Returns a signal producer that sends the receiver when its gesture occurs.
	public var gestureProducer: SignalProducer<Self, NoError> {
		return rac_gestureSignal().toSignalProducer()
			.demoteErrors()
			.map { $0 as? Self }
			.ignoreNil()
	}
}
```

Works correctly - `panGestureRecognizer.gestureProducer` is of type `SignalProducer<UIPanGestureRecognizer, NoError>`.

----

Right now there aren't any tests for this, because the version of `Nimble` referenced by `master` won't build in Xcode 7.3 - once that is fixed, I can add tests for these producers.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducerExtensions.swift
+++ b/ReactiveCocoa/Swift/SignalProducerExtensions.swift
@@ -1,0 +1,21 @@
+//
+//  SignalProducerExtensions.swift
+//  ReactiveCocoa
+//
+//  Created by Nate Stedman on 3/22/16.
+//  Copyright © 2016 GitHub. All rights reserved.
+//
+
+import enum Result.NoError
+
+extension SignalProducer {
+	func demoteErrors() -> SignalProducer<Value, NoError> {
+		return flatMapError { _ in SignalProducer<Value, NoError>.empty }
+	}
+}
+
+extension SignalProducer {
+	func constrainToType<T>(type: T.Type) -> SignalProducer<T, Error> {
+		return map({ $0 as? T }).ignoreNil()
+	}
+}
--- a/ReactiveCocoa/Swift/UIKitExtensions.swift
+++ b/ReactiveCocoa/Swift/UIKitExtensions.swift
@@ -66,3 +66,20 @@ extension TextSignalProvidingType {
 			.map { $0 as? String }
 	}
 }
+
+// MARK: - gestureProducer
+public protocol GestureSignalProvidingType {
+	func rac_gestureSignal() -> RACSignal!
+}
+
+extension UIGestureRecognizer: GestureSignalProvidingType {}
+
+extension GestureSignalProvidingType {
+	/// Returns a signal producer that sends the receiver when its gesture occurs.
+	public var gestureProducer: SignalProducer<Self, NoError> {
+		return rac_gestureSignal().toSignalProducer()
+			.demoteErrors()
+			.map { $0 as? Self }
+			.ignoreNil()
+	}
+}
--- a/ReactiveCocoa/Swift/UIKitExtensions.swift
+++ b/ReactiveCocoa/Swift/UIKitExtensions.swift
@@ -27,3 +27,21 @@ extension PrepareForReuseSignalProvidingType {
 			.map { _ in () }
 	}
 }
+
+// MARK: - producerForControlEvents
+public protocol SignalForControlEventsProvidingType {
+	func rac_signalForControlEvents(controlEvents: UIControlEvents) -> RACSignal!
+}
+
+extension UIControl: SignalForControlEventsProvidingType {}
+
+extension SignalForControlEventsProvidingType {
+	/// Creates a signal producer that sends the sender of the control event
+	/// whenever one of the control events is triggered.
+	public func producerForControlEvents(controlEvents: UIControlEvents) -> SignalProducer<Self, NoError> {
+		return rac_signalForControlEvents(controlEvents).toSignalProducer()
+			.demoteErrors()
+			.map { $0 as? Self }
+			.ignoreNil()
+	}
+}
--- a/ReactiveCocoa/Swift/UIKitExtensions.swift
+++ b/ReactiveCocoa/Swift/UIKitExtensions.swift
@@ -1,0 +1,29 @@
+//
+//  UIKitExtensions.swift
+//  ReactiveCocoa
+//
+//  Created by Nate Stedman on 3/22/16.
+//  Copyright © 2016 GitHub. All rights reserved.
+//
+
+import UIKit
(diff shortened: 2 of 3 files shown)


ReactiveCocoa/Swift/SignalProducerExtensions.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+//  Created by Nate Stedman on 3/22/16.
+//  Copyright © 2016 GitHub. All rights reserved.
+//
+
+import enum Result.NoError
+
+extension SignalProducer {
+	func demoteErrors() -> SignalProducer<Value, NoError> {
+		return flatMapError { _ in SignalProducer<Value, NoError>.empty }
```
JaviSoto: I think if there happens to be an error, we should either assert or at least print to the console

JaviSoto: Or perhaps rename the function to `ignoreErrors`, for consistency with `ignoreNil`

ReactiveCocoa/Swift/UIKitExtensions.swift
```diff
@@ -0,0 +1,85 @@
...
+extension TextSignalProvidingType {
+	/// Creates and returns a signal producer for the text of the receiver,
+	/// starting with the current text.
+	///
+	/// For underlying behavior and potential side-effects, see the
+	/// documentation for the receiver's implementation of `rac_textSignal`.
+	var textProducer: SignalProducer<String?, NoError> {
+		return rac_textSignal().toSignalProducer()
+			.demoteErrors()
+			.map { $0 as? String }
```
JaviSoto: Why `as?`? Perhaps this should be `.map { $0 as! String? }` ?

ReactiveCocoa/Swift/UIKitExtensions.swift
```diff
@@ -0,0 +1,85 @@
...
+}
+
+extension UIGestureRecognizer: GestureSignalProvidingType {}
+
+extension GestureSignalProvidingType {
+	/// Returns a signal producer that sends the receiver when its gesture occurs.
+	public var gestureProducer: SignalProducer<Self, NoError> {
+		return rac_gestureSignal().toSignalProducer()
+			.demoteErrors()
+			.map { $0 as? Self }
```
JaviSoto: Also perhaps `as! Self`?

JaviSoto:
This is fantastic! I think now that we have stable RAC releases, I'm in favor of slowly incorporating these missing pieces back into RAC. Thoughts?

ReactiveCocoa/Swift/UIKitExtensions.swift
natestedman: In these cases, I was following the [style guide](https://github.com/github/swift-style-guide#avoid-using-force-unwrapping-of-optionals), which is sort of linked in the [contributing](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/CONTRIBUTING.md) guidelines -  the link is to the Objective-C version, but figured that implied to follow the Swift one as well.

RuiAAPeres:
@JaviSoto I agree, but as, discussed previously, couldn't we put these in @neilpa 's Rex? Maybe make it official and move it to the ReactiveCocoa organization?

ReactiveCocoa/Swift/SignalProducerExtensions.swift
NachoSoto: Yeah. I think we can incorporate some of the extensions [here](https://gist.github.com/NachoSoto/25b0003c1b15d7d11a8d).

ReactiveCocoa/Swift/UIKitExtensions.swift
```diff
@@ -0,0 +1,85 @@
...
+
+extension UIControl: SignalForControlEventsProvidingType {}
+
+extension SignalForControlEventsProvidingType {
+	/// Creates a signal producer that sends the sender of the control event
+	/// whenever one of the control events is triggered.
+	public func producerForControlEvents(controlEvents: UIControlEvents) -> SignalProducer<Self, NoError> {
+		return rac_signalForControlEvents(controlEvents).toSignalProducer()
+			.demoteErrors()
+			.map { $0 as? Self }
```
NachoSoto: This shouldn't be `as?`

NachoSoto:
Cool! Excited to start expanding our API. I think we should outline a plan for what we want to add exactly. 

That's assuming we do want to add these. Like @RuiAAPeres said, there's also the option of keeping these separate, which I would advocate for too, especially since we'll want to make `RAC` Linux compatible, I'd suggest maybe keeping `ReactiveCocoa` pure and decoupled from `Foundation` / `UIKit`.

neilpa:
> I'd suggest maybe keeping ReactiveCocoa pure and decoupled from Foundation / UIKit.

`Foundation` is safe since there's the Linux port of that. But we definitely want to be careful about the Cocoa dependencies going forward.

natestedman:
Cool, makes sense to build these integration bridges in microframeworks!

PR #2777, opened 2016-04-05
[Objective-C] Fix -[RACSequence foldRightWithStart:reduce:]

ikesyo:
With this change, the failing test noted in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2760#issuecomment-199539387 passes. That was actually broken. :sweat_smile:

### DIFF
--- a/ReactiveCocoa/Objective-C/RACSequence.m
+++ b/ReactiveCocoa/Objective-C/RACSequence.m
@@ -219,7 +219,11 @@
 	if (self.head == nil) return start;
 	
 	RACSequence *rest = [RACSequence sequenceWithHeadBlock:^{
-		return [self.tail foldRightWithStart:start reduce:reduce];
+		if (self.tail) {
+			return [self.tail foldRightWithStart:start reduce:reduce];
+		} else {
+			return start;
+		}
 	} tailBlock:nil];
 	
 	return reduce(self.head, rest);
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,21 @@
 language: objective-c
 osx_image: xcode7.3
-before_install:
-  - brew update
-  - brew uninstall xctool && brew install --HEAD xctool
+before_install: true
 install: true
 git:
   submodules: false
-script: script/cibuild ReactiveCocoa-Mac ReactiveCocoa-iOS
+before_script:
+  - git submodule update --init --recursive
+script:
+  - set -o pipefail
+  - xcodebuild test -workspace "$TRAVIS_XCODE_WORKSPACE" -scheme "$TRAVIS_XCODE_SCHEME" -sdk "$XCODE_SDK" | xcpretty
+xcode_workspace: ReactiveCocoa.xcworkspace
+matrix:
+  include:
+    - xcode_scheme: ReactiveCocoa-Mac
+      env: XCODE_SDK=macosx
+    - xcode_scheme: ReactiveCocoa-iOS
+      env: XCODE_SDK=iphonesimulator
 notifications:
   email: false
   slack:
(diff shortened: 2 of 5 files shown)


ikesyo:
This is really annoying... https://travis-ci.org/ReactiveCocoa/ReactiveCocoa/builds/120938289#L7292

> script/cibuild: line 110:  3207 Abort trap: 6           xctool -workspace "$XCWORKSPACE" $XCTOOL_OPTIONS "$@" 2>&1

ikesyo:
ping

ReactiveCocoa/Objective-C/RACSequence.m
```diff
@@ -219,7 +219,11 @@ - (id)foldRightWithStart:(id)start reduce:(id (^)(id, RACSequence *))reduce {
 	if (self.head == nil) return start;
 	
 	RACSequence *rest = [RACSequence sequenceWithHeadBlock:^{
-		return [self.tail foldRightWithStart:start reduce:reduce];
+		if (self.tail) {
+			return [self.tail foldRightWithStart:start reduce:reduce];
+		} else {
+			return start;
+		}
```
mdiep: I'm a bit confused about how `RACSequence` is supposed to work.

The documentation for `-[RACSequence tail]` says this:

```
/// The first object in the sequence, or nil if the sequence is empty.
///
/// Subclasses must provide an implementation of this method.
@property (nonatomic, strong, readonly) id head;
```

That makes me think that the existing code here is correct. If `self.head` is non-nil (we check above), then `self.tail` should also be non-nil. Only if `self.head` is nil should `self.tail` be nil (an empty sequence).

`-[RACUnarySequence tail]` is hard-coded to return `nil`, however. Changing it to return `RACSequence.empty` does make this test pass. However, another test fails:

```
	qck_it(@"should use the values from the head and tail blocks", ^{
		expect(sequence.head).to(equal(@0));
		expect(sequence.tail.head).to(equal(@1));
		expect(sequence.tail.tail).to(beNil());
	});
```

Since I don't think we want to change the behavior at this point, I suppose this fix is correct. However, it'd be nice to correct/clarify the `-[RACSequence tail]` documentation. Would you be up for doing that here?

mdiep:
I suppose this should also be ported to the `2.x-maintenance` branch after this is merged?

ikesyo:
> I suppose this should also be ported to the `2.x-maintenance` branch after this is merged?

I think so too.

ReactiveCocoa/Objective-C/RACSequence.m
ikesyo: > However, it'd be nice to correct/clarify the `-[RACSequence tail]` documentation. Would you be up for doing that here?

How does it look like?

```diff
-/// All but the first object in the sequence, or nil if the sequence is empty.
+/// All but the first object in the sequence, or nil if the sequence is empty or
+/// contains only one object.
 ///
 /// Subclasses must provide an implementation of this method.
 @property (nonatomic, strong, readonly) RACSequence *tail;
```

mdiep: How about this:

> All but the first object in the sequence, or nil if there are no other objects.

ikesyo: Sounds good, thank you!

ikesyo:
Thanks for the review, I updated the `-tail` documentation. :boom:

mdiep:
👍

mdiep:
Which of us is going to backport this?

ikesyo:
> Which of us is going to backport this?

I already have a branch for that. :hand:

mdiep:
💖

PR #2780, opened 2016-04-06
Inject side effect on next and failed

ajjnix:
Side effect on Next and Failed is very popular. It's more convenient than `on` if need only one status

example:

`producer.on(next: bar.foo))`

or

`producer.doNext(bar.foo)`


ikesyo:
Thanks for the pull request! Because `on` has default arguments, you can use it with omitting unnecessary arguments as you wrote: `producer.on(next: bar.foo)`. Closing.

Issue #1124, opened 2014-02-20
Memory leak in RACSubject


Strilanc:
The following code will cause memory to increase without bound:

    RACSubject* s1 = [RACSubject new];
    RACSubject* s2 = [RACSubject new];
    while (true) {
        @autoreleasepool {
            [[s1 subscribe:s2] dispose];
        }
    }

The issue is that every time a RACSubject is used as a subscriber, `didSubscribeWithDisposable:` is called with the disposable and that disposable is placed into an array (to be disposed if the Subject completes). But those disposable are not removed from the list upon being disposed.

So if you have a use case where many streams can temporarily 'drive' a RACSubject, such as a server value being written by currently connected clients, the list of disposables will increase without bound.

I'm not sure how to fix this problem. Give RACDisposable an 'also do this if/when disposed' method, to register the removal on? Incrementally scan the list, removing already disposed items when new ones are added? A `didDisposeSubscription:` to go with `didSubscribeWithDisposable:`?

RuiAAPeres:
Closing due to inactivity.

ikesyo:
There is the ongoing pull request for this: #2782. Reopening.

RuiAAPeres:
@ikesyo 👍

PR #2782, opened 2016-04-07
dispose subscriber when completed

tzongw:
try fix Memory leak in RACSubject #1124


mdiep:
😳 

Thanks for the PR! Could you add a unit test that verifies that the leak is fixed?

mdiep:
This just needs a test if anyone wants to contribute!

mdiep:
Superseded by #2845.

PR #2783, opened 2016-04-07
public RACDelegateProxy.h

SKIT1:


SKIT1:
custom delegate proxy property to the UI component

SKIT1:
I think there is a need to expose RACDelegateProxy. h header file.For example, extending the delegate proxy.

struct AssociationKey {
	static var delegateProxy: UInt8 = 0
}

extension UISearchBar {
	private func RACUseDelegateProxy(_self: UISearchBar) {
		if let delegate = _self.delegate as? RACDelegateProxy {
			if delegate == _self.rac_delegateProxy {
                
				return
			}
		}
		_self.rac_delegateProxy?.rac_proxiedDelegate = _self.delegate
		_self.delegate = _self.rac_delegateProxy as? UISearchBarDelegate
	}

	public var rac_delegateProxy: RACDelegateProxy? {

		return lazyAssociatedProperty(self, key: &AssociationKey.delegateProxy, factory: { () -> RACDelegateProxy in
			return RACDelegateProxy(withProtocol: UISearchBarDelegate.self)
		})
	}

	public func rac_textSignal() -> SignalProducer<String?, NSError> {
		let triggerSignal = toVoidSignal(self.rac_willDeallocSignal().asSignal())

		let signal = SignalProducer<AnyObject?, NSError> { [unowned self] observer, _ in
			observer.sendNext(self)
			observer.sendCompleted()
		}.concat((self.rac_delegateProxy!.rac_signalForSelector(#selector(UISearchBarDelegate.searchBar(_:textDidChange:))).toSignalProducer())).map({ (next) -> String? in
			if let tuple = next as? RACTuple {
				guard let searchBar = tuple.first as? UISearchBar else {
					return nil
				}
				return searchBar.text
			} else {
				guard let searchBar = next as? UISearchBar else {
					return nil
				}
				return searchBar.text
			}
		}).takeUntil(triggerSignal)

		RACUseDelegateProxy(self)

		return signal
	}
}

mdiep:
Did you decide that this wasn't needed?

SKIT1:
Needed

PR #2788, opened 2016-04-10
Add Property map operator

sdduursma:
I need a way to map the current value of a property and all subsequent values to a new read-only property.

To achieve this, currently one has to map the current value and the signal/signal producer separately:
```
let name = MutableProperty("")
let transform = { !$0.isEmpty }
let nameValid = AnyProperty(initialValue: transform(name.value), signal: name.signal.map(transform))
```
This seems like a lot of code for what essentially is mapping a property.

Alternatively one can bind the signal producer of `name` to `nameValid`:
```
let name = MutableProperty("")
let nameValid = MutableProperty(false)
nameValid <~ name.producer.map { !$0.isEmpty }
```
The downside to this approach is that I have to supply an initial value for `nameValid` even though it will be overridden immediately after by the binding operator.
Also `nameValid` needs to be a `MutableProperty`.

I think it would be useful to introduce a `map` operator for properties.

To-dos:
- [x] Add unit tests.
- [x] Add inline documentation.
- [x] Squash WIP commits.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -60,6 +60,17 @@ public struct AnyProperty<Value>: PropertyType {
 	}
 }
 
+extension PropertyType {
+	/// Maps the current value and all subsequent values to a new value.
+	public func map<U>(transform: Value -> U) -> AnyProperty<U> {
+		let mappedProducer = SignalProducer<U, NoError> { observer, disposable in
+			disposable += ActionDisposable { self }
+			disposable += self.producer.map(transform).start(observer)
+		}
+		return AnyProperty(initialValue: transform(value), producer: mappedProducer)
+	}
+}
+
 /// A property that never changes.
 public struct ConstantProperty<Value>: PropertyType {
 
--- a/ReactiveCocoaTests/Swift/PropertySpec.swift
+++ b/ReactiveCocoaTests/Swift/PropertySpec.swift
@@ -520,6 +520,20 @@ class PropertySpec: QuickSpec {
 			}
 		}
 
+		describe("map") {
+			it("should transform the current value and all subsequent values") {
+				let property = MutableProperty(1)
+				let mappedProperty = property
+					.map { $0 + 1 }
+					.map { $0 + 2 }
+
+				expect(mappedProperty.value) == 4
+
+				property.value = 2
+				expect(mappedProperty.value) == 5
+			}
+		}
+
 		describe("binding") {
 			describe("from a Signal") {
 				it("should update the property with values sent from the signal") {


filblue:
`map`-ed `Signal` captures the original one. Most surely that should be the same for `PropertyType`.

Chain constructs like `let p1 = p0.map { $0.location }.map { $0.latitude }` are not safe with the proposed implementation, unless you explicitly capture `p0` for the whole `p1`'s lifecycle. 

Here's how we implemented it in our project:

```
extension SignalProducerType where Error == NoError {

    static func capture<P: PropertyType where P.Value == Value>(property: P) -> SignalProducer<Value, NoError> {
            disposable += ActionDisposable { property }
            disposable += property.producer.start(observer)
        }
    }
}

extension PropertyType {

    func map<U>(transform: Value -> U) -> AnyProperty<U> {
        return AnyProperty<U>(
            initialValue: transform(self.value),
            producer: SignalProducer.capture(self).skip(1).map(transform))
    }
}
```

mdiep:
I like it. 👍 

Unit tests should definitely cover the concern raised by @filblue.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -60,6 +60,12 @@ public struct AnyProperty<Value>: PropertyType {
 	}
 }
 
+extension PropertyType {
+	public func map<U>(transform: Value -> U) -> AnyProperty<U> {
+		return AnyProperty(initialValue: transform(value), signal: signal.map(transform))
```
JaviSoto: Hmmm is this thread safe? It seems to be it could be possible to skip a value?

sdduursma:
Thanks for the feedback @filblue. I've updated the PR capture the original property. Also in the unit test I'm chaining two `map`s the cover that issue.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -60,6 +60,17 @@ public struct AnyProperty<Value>: PropertyType {
...
 }
 
+/// Maps the current value and all subsequent values to a new value.
+extension PropertyType {
+	public func map<U>(transform: Value -> U) -> AnyProperty<U> {
+		let signalProducer = SignalProducer<Value, NoError> { observer, disposable in
+			disposable += ActionDisposable { self }
+			disposable += self.producer.start(observer)
+		}
+		return AnyProperty(initialValue: transform(value), producer: signalProducer.map(transform))
```
filblue: `transform` is applied twice to the same value, because `property.producer` sends its current value on subscription. Seems like it can be solved by observing `signal` instead of `producer` on line 68.

sdduursma:
@filblue I've updated the PR to observe `signal` instead of producer as you suggested.

ReactiveCocoa/Swift/Property.swift
sdduursma: I'm not sure why this would not be thread safe. What do you mean by "skip a value"?

andersio:
While it makes sense in terms of semantics, using `signal` currently could have a chance of race condition (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2646#discussion_r49162957). Specifically, the subscription to the signal happens after the getter of the value (`withValue(_:)`) released the lock, that means the setter (`modify(:_)`) could possibly write and push out a new value before the subscription happens, causing lost events.

Until `@noescape` can be used in type declarations so that `withValue` can be introduced to `PropertyType` clearly, i.e. Swift 3, I think this operator should use `producer`.

sdduursma:
@andersio So we're ok then with `transform` being applied twice to the same value as @filblue has pointed out until Swift 3?

andersio:
IMO it would be fine for the standard set of Property impl's. It happens before the `map` operator returns, and the extra event cannot be observed externally.

sdduursma:
@andersio @filblue  I've added a commit that goes back to using `producer` instead of `signal`.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -60,6 +60,17 @@ public struct AnyProperty<Value>: PropertyType {
 	}
 }
 
+/// Maps the current value and all subsequent values to a new value.
```
mdiep: This documentation string should go on the method, not on the extension. ☺️

mdiep:
Other than the small documentation issue, this looks good to me. 👍 Fix that up and I'll merge!

sdduursma:
@mdiep Fixed the documentation issue and squashed the commits!

mdiep:
Thanks for the contribution! ✨

PR #2792, opened 2016-04-12
Add sampleWith method

victorlin:
This pull request adds a new `sampleWith` method which acts exactly the same as `sampleOn`, except it pass along with the next event value from the given sampler.

## Rationale

I found myself very frequently need this method when I was developing a game-ish app. The basic scenario is that I have a clock signal looks like

`Signal<NSDate, NoError>`

It's basically emitted from `CADisplayLink` or other signal for syncing with display

and I have some properties of game objects, say like position

```Swift
class Player {
    let position: AnyProperty<CGPoint>
    /// ...
}
```

To display the game object position, I need to sample the game object's position when I see a next event emit from `CADisplayLink`. And also, very often, I need to know the current game clock from that `CADisplayLink` signal.

Another case is sometimes, I want to do some async operations for a game object along with another signal. In this case, I tried to use `combineLatestWith`

```Swift
focusObject.producer
    .combineLatestWith(activeEffect)
    .flatMap { (object, effect) in
        // .... apply effect on object
    }
```

but the problem in this case is, I want `activeEffect` to only affect current `focusObject`, but not a new object just becomes the focus. With `combineLatestWith`, when `focusObject` is changed, it also emits a next event, will end up also consuming the `activeEffect`.

To solve the problem I can use `sampleWith`.

```Swift
focusObject.producer
    .sampleWith(activeEffect)
    .flatMap { (object, effect) in
        // .... apply effect on object
    }
```

I don't care how `focusObject` changes, I only care if the `activeEffect` changes, then it should sample the object and apply on it.

There are also lots of cases like this, the scenario is pretty much like, I only want to sample the signal when another signal emits, but I also need the value from the another signal.

## TODOs

- [x] Add documents for `sampleWith`

## Discussion

Not sure why `sampleOn` was designed to not to carry any value, but it kinda limits what you can do with it as if you need the value from the sampler. I tried to implement this with compositions of different operations as simple as possible, but still not as easy to just modify `sampleOn` method, to make it carry out the next value, and make `sampleOn` just dropping the value. Maybe I missed something, is there actually an easy way to make `sampleOn` carrying the sampler value? If it's the case, not sure is it worthwhile to add this operation. What do you guys think about this?

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -551,17 +551,9 @@ private struct SampleState<Value> {
 }
 
 extension SignalType {
-	/// Forwards the latest value from `signal` whenever `sampler` sends a Next
-	/// event.
-	///
-	/// If `sampler` fires before a value has been observed on `signal`, nothing
-	/// happens.
-	///
-	/// Returns a signal that will send values from `signal`, sampled (possibly
-	/// multiple times) by `sampler`, then complete once both input signals have
-	/// completed, or interrupt if either input signal is interrupted.
+	// TODO: fix the doc here
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func sampleOn(sampler: Signal<(), NoError>) -> Signal<Value, Error> {
+	public func sampleWith<T>(sampler: Signal<T, NoError>) -> Signal<(Value, T), Error> {
 		return Signal { observer in
 			let state = Atomic(SampleState<Value>())
 			let disposable = CompositeDisposable()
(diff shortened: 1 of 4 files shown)


mdiep:
I'll think about it some more, but my initial impression is that `sampleWith` makes a lot of sense.

mdiep:
@ReactiveCocoa/reactivecocoa Any objections to adding a `sampleWith` method that behaves like `sampleOn` but sends a tuple with the value from each signal?

mdiep:
@victorlin Do you want to fill in the documentation? It doesn't sound like anyone objects, so let's finish this up! 😄

victorlin:
will do

victorlin:
@mdiep will do it tonight

victorlin:
@mdiep done

mdiep:
This is great! Thanks @victorlin!

Issue #2787, opened 2016-04-10
UIControl rac_signalForControlEvents will cause "memory leak"


vagase:
If subscribe and dispose the signal returned by UIControl's `rac_signalForControlEvents` over and over again, we can find that there are many `RACCompounedDisposable` objects are allocated and not released via Instruments.

The problem is: (UIControl+RACSignalSupport.m L28 - L35)

```
[self.rac_deallocDisposable addDisposable:[RACDisposable disposableWithBlock:^{
	[subscriber sendCompleted];
}]];

return [RACDisposable disposableWithBlock:^{
	@strongify(self);
	[self removeTarget:subscriber action:@selector(sendNext:) forControlEvents:controlEvents];
}];
```

There are two RACDisposable are created while subscribing, the first one is added to  `self.rac_deallocDisposable`, the second one is returned to callee. When the callee disposes the second RACDisposable object, it just remove the target and action. But the first RACDisposable still be retained by `self.rac_deallocDisposable`, not got released.

mdiep:
Good catch! Would you be interested in opening a PR to fix test?

RuiAAPeres:
@vagase would you be able to provide a PR for this issue? Or it's something you would prefer us to do?

vagase:
Ok, I'd like to provide a PR soon.

PR #2797, opened 2016-04-15
fix memory leak caused by unremoved disposable object

vagase:
Fix "Memory leak" caused by unremoved disposable object in UIControl, after calling `rac_signalForControlEvents` and `dispose` over and over again.

See https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2787 for more detail.


mdiep:
Thanks for the pull request!

PR #2800, opened 2016-04-16
Fix memory leak on `lift`

ikesyo:
Another approach for the issue described in #2758.


mdiep:
Nice. It seems like this is probably a better approach.

@trifonov-ivan Does this address your concerns about `lift`?

trifonov-ivan:
@mdiep looks like so. sendInterrupted usage is nice for it, i think) Thank you. @ikesyo for fix!

mdiep:
Thanks @ikesyo!

PR #2811, opened 2016-04-21
Use a plain class for dealloc signal instead of NSObject.rac_willDeallocSignal

ikesyo:
This should be a more portable implementation.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1217,7 +1217,7 @@ extension SignalProducerType {
 		// This will go "out of scope" when the returned `SignalProducer` goes out of scope.
 		// This lets us know when we're supposed to dispose the underlying producer.
 		// This is necessary because `struct`s don't have `deinit`.
-		let tracker = Tracker()
+		let token = DeallocationToken()
 
 		return SignalProducer { observer, disposable in
 			let initializedProducer: SignalProducer<Value, Error>
@@ -1242,14 +1242,14 @@ extension SignalProducerType {
 
 			if shouldStartUnderlyingProducer {
 				self.producer
-					.takeUntil(tracker.deallocSignal)
+					.takeUntil(token.deallocSignal)
 					.start(initializedObserver)
 			}
 		}
 	}
 }
 
-private final class Tracker {
+private final class DeallocationToken {
 	let (deallocSignal, observer) = Signal<(), NoError>.pipe()
 
 	deinit {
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1242,22 +1242,17 @@ extension SignalProducerType {
...
-			.rac_willDeallocSignal()
-			.toSignalProducer()
-			.map { _ in () }
-			.mapError { error in
-				fatalError("Unexpected error: \(error)")
-				()
-			}
+private final class Tracker {
+	let (deallocSignal, observer) = Signal<(), NoError>.pipe()
+
```
RuiAAPeres: The observer should be private.

RuiAAPeres:
I love this! ❤️

ReactiveCocoa/Swift/SignalProducer.swift
ikesyo: Since the class itself is private, the properties are also private (we only have _file private_ and don't have real _scoped private_ in Swift 2.2).

RuiAAPeres: They are not private to the `SignalProducer`. Otherwise you couldn't do this `tracker.deallocSignal`.

ikesyo: `private` modifier in Swift 2.2 means the symbol is visible within the current file, not within the current scope. So adding `private` to `observer` property have no effect. This should be changed in Swift 3 by the proposal: https://github.com/apple/swift-evolution/blob/master/proposals/0025-scoped-access-level.md

RuiAAPeres: You are right, disregard my comments!

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1242,22 +1242,17 @@ extension SignalProducerType {
...
-	var willDeallocSignal: SignalProducer<(), NoError> {
-		return self
-			.rac_willDeallocSignal()
-			.toSignalProducer()
-			.map { _ in () }
-			.mapError { error in
-				fatalError("Unexpected error: \(error)")
-				()
-			}
+private final class Tracker {
```
NachoSoto: Maybe call this `DeallocationToken`?

NachoSoto:
I'm fine with this, it's definitely simpler in this case. However, the solution is now less general. My idea was that we could eventually make that `willDeallocSignal` public so that it can be used on **any instance**. This can only be used to track when an instance of this particular class goes out of context.

ikesyo:
This is important to support SwiftPM and Linux in future, in the environments we can't use Objective-C implementations (`NSObject.rac_willDeallocSignal` in this case).

> My idea was that we could eventually make that `NSObject.willDeallocSignal` public so that it can be used on **any `NSObject` instance**.

I filed an issue: #2813.

RuiAAPeres:
@ikesyo once it's exposed, it makes sense to remove the one provided by Rex (cc @neilpa )

mdiep:
👍

PR #2814, opened 2016-04-22
Extract DynamicProperty into its own file

ikesyo:
This makes it easy to support SwiftPM (for excluding the files which need the Objective-C implementations).


mdiep:
👍

Issue #2815, opened 2016-04-22
Progressive collect operator


startupthekid:
I opened [a question on SO](http://stackoverflow.com/questions/36783941/reactivecocoa-4-progressive-collect-operator) about this but it could potentially be an incredibly useful operator to have, allowing for the ability to composite stream values into array chunks that work can be processed on.

Function contract:

- aggregates values into an array, similar to `collect`
- values that pass the filter are appended to the previous list
- values that fail the filter are push onto the next sequence and previous sequence is sent

Example usage would be:

```swift
someProducer.aggregateWhile { $0 == someValue }

let (signal, observer) = Signal<Int, NoError>.pipe()

observer.sendNext(5) // no output
observer.sendNext(5) // no output
observer.sendNext(42) // outputs [5, 5] then [42]
observer.sendNext(5)  // outputs [5]
observer.sendCompleted()

signal.aggregateWhile { $ == 5 }.observeNext { print($0) }

// Output: 
// [5, 5]
// [42]
// [5]

```

Essentially what I want is given a value stream of say, 1, 2, 3, 4, 5, 6, the ability to arbitrarily chunk that data based on a filter condition. Depending on my filter the output could look like `[1, 2]`, `[3, 4, 5, 6]` or `[1]`, `[2, 3, 4, 5, 6]`, whatever you want really.

What do you guys think? Something worth taking a look at?

mdiep:
What sort of thing are you using this for?

startupthekid:
@mdiep I want to use it to perform operations on lists of `CMSampleBuffer` objects that are being read in via an `AVAssetReader`. In my particular use case, I need to aggregate a certain number of buffers, that range from 2 to 4 seconds where each frame in that list returns positive for face detection (via `CoreImage`), then crop the each frame to the respective face, and write that list as a separate video file. So for example, if a video file has 4 instances where a face is detected for 2 - 4 seconds, I'd generate four smaller videos from the original asset. 

The reason I'd like a progressive aggregator that sends chunks on a conditional basis is as you can imagine, reading in videos is memory intensive and if you have about 250+ frames in memory at once, you can crash the application and being able to build up a list of frames as they come in and then immediately process that list/write it to file would help to keep my memory imprint low.

mdiep:
What condition would you use in that case? Would a fixed-size buffer like #2817 meet your need?

startupthekid:
[...]

PR #2817, opened 2016-04-23
accumulate values

ajjnix:
This is can use for accumulate values count. For example I used this code for accumulate locations in trip

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -938,6 +938,20 @@ extension SignalType {
 			return disposable
 		}
 	}
+	
+	/// Accumulate values in array. Drain it with propagation after reaching specified size
+	func accumulate(size: Int) -> Signal<[Value], Error> {
+		var values: [Value] = []
+		func next(value: Value) {
+			if values.count >= size {
+				values.removeAll()
+			}
+			values.append(value)
+		}
+		return on(next: next)
+			.filter { _ in values.count < size }
+			.map { _ -> [Value] in return values }
+	}
 
 	/// Applies `operation` to values from `self` with `Success`ful results
 	/// forwarded on the returned signal and `Failure`s sent as `Failed` events.
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -205,6 +205,20 @@ public struct SignalProducer<Value, Error: ErrorType> {
 
 		return (producer, bufferingObserver)
 	}
+	
+	/// Accumulate values in array. Drain it with propagation after reaching specified size
+	func accumulate(size: Int) -> SignalProducer<[Value], Error> {
+		var values: [Value] = []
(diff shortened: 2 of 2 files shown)


mdiep:
Can you provide some more detail about what you'd use this for?

ajjnix:
Yeah of course. I need observe change location (with complex business logic), after I should get trip from point locations and send it to server. So I accumulate point locations to trip in the required size and send it onward. After that I need continue accumulate NEW trip.

```Swift
        location().accumulate(tripSize).observeNext { [unowned self] locations in
            self.tripObserver.sendNext(locations)
        }
```

mdiep:
I think this makes sense as an operator, but I think it should be called `collect` to match the terminology of the existing operator.

dmcrodrigues:
This operator can be helpful but I would like to suggest a small change to make it more flexible, instead of providing a size to collect the values, I rather prefer having a predicate to control how values should be collected. This predicate can receive the `values` already collected along with the `next` value that may be collected.

```swift
(values: [Value], next: Value) -> Bool
```

With this flexibility, I think we can accomplish both this one as #2815. For this particular case we can use something like: 

```swift
{ values, _ in values.count < someSize }
```

@mdiep @ajjnix @startupthekid I'm attaching an implementation that cover both of them. What do you guys think? I don't think its possible to achieve this with a combination of the existing operators but I can be wrong. I also have some tests if this goes forward.

```swift
public func collect(predicate: (values: [Value], next: Value) -> Bool) -> Signal<[Value], Error> {
    return Signal { observer in
        var values: [Value] = []

        func collectNextValue(value: Value) -> Bool {
            if predicate(values: values, next: value) {
                values.append(value)
                return true
            }
            return false
        }

        func pushCollectedValues() {
            if !values.isEmpty {
                observer.sendNext(values)
                values.removeAll()
            }
        }

            switch event {
            case let .Next(value):
                if !collectNextValue(value) {
                    sendCollectedValues()
                    // Retry collecting the next value without all values collected...
                    if !collectNextValue(value) {
                        // Value cannot be collected...
                        observer.sendNext([value])
                    }
                }

            case .Completed:
                // Push all remaining values collected
                sendCollectedValues()
                observer.sendCompleted()

            case let .Failed(error):
                observer.sendFailed(error)
            case .Interrupted:
                observer.sendInterrupted()
            }
        }
    }
}
```

mdiep:
That's a great idea. 👍 I was going to suggest we add both a count-based and a predicate-based version to cover this and #2815. Want to open a PR?

(But I'd expect the predicate to return `true` when the values should be sent—not when they should be collected.)

dmcrodrigues:
Sure 👍  But do you think they should be completely independent (in terms of implementation) of each other? Or can we use the predicate-based version as the base to provide a count-based version?

About the predicate makes sense 👍

mdiep:
> Or can we use the predicate-based version as the base to provide a count-based version?

I think both the count-based and the existing version can be based on the predicate version.

ikesyo:
Superseded by #2835.

PR #2827, opened 2016-04-26
Omit unnecessary `self`s for consistency

ikesyo:


NachoSoto:
I'm :-1: on this, I think explicit-ness is a big win with `self`. Getting rid of it only hurts readability IMO.

ikesyo:
So should we add `self.`s to the entire codebase where those are not necessarily used, instead? Oppositely I think removing unnecessary `self`s makes it clear that whether the `self` is used intentionally, e.g. initializing a property with the same name argument on a initializer, capturing `self` in a closure (this is useful for detecting a retain cycle).

ikesyo:
Anyway I'm withdrawing this. Thanks for the feedback!

iv-mexx:
I personally would be 👍 on this for the reasons mentioned in the rejection of [SE-0009 “Require self for accessing instance members”](http://ericasadun.com/2016/01/06/the-swift-evolution-proposal-se-0009-rejection/). 

But yeah, it seems that is a polarising issue 😃

PR #2570, opened 2015-11-24
Implement #2564: ActionType for protocol extension

cradnovich:
Implemented the `ActionType` protocol for [Issue 2564](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2564) similarly to `SignalType` and `SignalProducerType`.

PR #2828, opened 2016-04-26
ActionType redux

### LINKED ISSUE #2564: ActionType for protocol extension?
I'm looking for a way to extend `Action` with protocol extension like we can with `SignalProducerType` / `SignalType`. Something like the following:

```
extension Action where Error == NoError {
  func exec(input: Input) -> SignalProducer<Ouput, Error> { 
    apply(input).flatMapError { _ in SignalProducer<Input, NoError>.empty }
  }
}
```
But it won't work since AFAIK, swift doesn't support constraining by type parameters (but only through abstract types). 

How should I go about doing this? Should I implement custom `protocol` like `ActionType` for wrapping `Action` class?

ikesyo:
This supersedes @cradnovich's work of #2570. Resolves #2564.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -129,6 +129,26 @@ public final class Action<Input, Output, Error: ErrorType> {
 	}
 }
 
+public protocol ActionType {
+	/// The type of argument to apply the action to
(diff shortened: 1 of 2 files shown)


RuiAAPeres:
👍

PR #2834, opened 2016-04-27
Convinience Initializer

RuiAAPeres:
I have a particular case where the the `CocoaAction` sits on the ViewModel and the VM is injected into the `UIViewController`. The `CocoaAction`'s input is the `UIViewController` itself. ( the[ FB SDK asks for a `UIViewController` for login)](https://developers.facebook.com/docs/facebook-login/ios/permissions). Their example:

```objc
- (void)viewDidLoad {
  [super viewDidLoad];
    if ([FBSDKAccessToken currentAccessToken]) {
       // TODO:Token is already available.
    }
}
// ....
FBSDKLoginManager *loginManager = [[FBSDKLoginManager alloc] init];
[loginManager logInWithReadPermissions:@[@"email"]
                    fromViewController:self
                               handler:^(FBSDKLoginManagerLoginResult *result, NSError *error) {
  //TODO: process error or result
 }];
```
I am guessing other people might have similar needs.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -218,6 +218,14 @@ public final class CocoaAction: NSObject {
 	public convenience init<Input, Output, Error>(_ action: Action<Input, Output, Error>, input: Input) {
 		self.init(action, { _ in input })
 	}
+	
+	/// Initializes a Cocoa action that will invoke the given Action
+	/// with the input provided on `execute`. The provided input should have
+	/// the same type as the `Action`'s `Input` parameter.
+    public convenience init<Input, Output, Error>(_ action: Action<Input, Output, Error>) {
+		let identity: AnyObject? -> Input = { $0 as! Input }
+		self.init(action, identity)
+	}
 
 	deinit {
 		disposable.dispose()
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -222,7 +222,7 @@ public final class CocoaAction: NSObject {
 	/// Initializes a Cocoa action that will invoke the given Action
 	/// with the input provided on `execute`. The provided input should have
 	/// the same type as the `Action`'s `Input` parameter.
-    public convenience init<Input, Output, Error>(_ action: Action<Input, Output, Error>) {
+	public convenience init<Input, Output, Error>(_ action: Action<Input, Output, Error>) {
 		let identity: AnyObject? -> Input = { $0 as! Input }
 		self.init(action, identity)
 	}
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/Swift/Action.swift
```diff
@@ -218,6 +218,14 @@ public final class CocoaAction: NSObject {
 	public convenience init<Input, Output, Error>(_ action: Action<Input, Output, Error>, input: Input) {
 		self.init(action, { _ in input })
 	}
+	
+	/// Initializes a Cocoa action that will invoke the given Action
+	/// with the input provided on `execute`. The provided input should have
+	/// the same type as the `Action`'s `Input` parameter.
+    public convenience init<Input, Output, Error>(_ action: Action<Input, Output, Error>) {
```
mdiep: Indentation is off here. ☺️

mdiep:
Is there a reason you don't use `unsafeCocoaAction` property on your `Action`?

RuiAAPeres:
@mdiep I thought about this and in my case I am exposing a `CocoaAction` and not an `Action` from my ViewModel.

RuiAAPeres:
There is also the other side of question, where the `{ $0 as! Input }` is already being used.  The initializer just makes this a bit more convenient.

mdiep:
I'm not understanding. 😕 Could you try explaining again what you're using this for?

I'm not trying to be difficult; I just want to make sure I understand. 😄

RuiAAPeres:
> I'm not trying to be difficult; I just want to make sure I understand. 😄

You are being thorough and I like that. 👍

> Is there a reason you don't use unsafeCocoaAction property on your Action?

I don't expose an `Action` in my ViewModel, but a `CocoaAction`.

> There is also the other side of question, where the { $0 as! Input } is already being used.

This was very cryptic from my side. 😞  What I mean is that `unsafeCocoaAction` is currently already doing that (`CocoaAction(self) { $0 as! Input }`), having a convenient initializer, just makes it more accessible. 

Of course there is the matter of safety, which the method clearly expresses as being `unsafe` , while the `init` doesn't. My opinion on this, is that from a usage point of View, it offers the same semantics that currently exist in an `Action`:

1. Create an Object that will do some work
2. Execute it with the input it needs

RuiAAPeres:
I am closing this PR. 

During my commute I realized two things:

1. I should actually expose an `Action` on the ViewModel. Because I am exposing a `CocoaAction`, I was trying to somehow manipulate the `CocoaAction` semantics, to fit my need. Bottom line: Instead of going along with the framework and design guidelines I was fighting them. This is wrong. 
2. There is a reason why the computed property is prefixed with `unsafe`. We shouldn't create/expose things that are inherently  prone to runtime crashes, ever. If we do, then they should be marked with big red letters.

PR #2835, opened 2016-04-28
Extend `collect` with a count-based and a predicate-based version

### LINKED ISSUE #2815: Progressive collect operator

dmcrodrigues:
This supersedes #2817 and resolves #2815.

This PR extends `collect` with two new operators:
- yield an array of values until it reaches a certain count
  - `collect(count: Int)`
- yield an array of values based on a predicate
  - `collect(predicate: (values: [Value], next: Value) -> Bool)`

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -301,36 +301,122 @@ extension SignalType {
 	}
 }
 
-public enum CollectStrategy {
-	case Inclusive
-	case Exclusive
+/// A reference type which wraps an array to auxiliate the collection of values
+/// for `collect` operator.
+private final class CollectState<Value> {
+	var values: [Value] = []
+
+	/// Collects a new value.
+	func append(value: Value) {
+		values.append(value)
+	}
+
+	/// Check if there are any items remaining.
+	///
+	/// - Note: This method also checks if there weren't collected any values 
+	/// and, in that case, it means an empty array should be sent as the result
+	/// of collect.
+	var isEmpty: Bool {
+		/// We use capacity being zero to determine if we haven't collected any 
+		/// value since we're keeping the capacity of the array to avoid 
+		/// unnecessary and expensive allocations). This also guarantees 
+		/// retro-compatibility around the original `collect()` operator.
+		return values.isEmpty && values.capacity > 0
+	}
+
+	/// Removes all values previously collected if any.
+	func flush() {
+		// Minor optimization to avoid consecutive allocations. Can
+		// be useful for sequences of regular or similar size and to
+		// track if any value was ever collected.
+		values.removeAll(keepCapacity: true)
+	}
 }
+
 
 extension SignalType {
 
-	/// Returns a signal that will yield an array of values when `self` completes.
-	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func collect() -> Signal<[Value], Error> {
-		return collect(.Inclusive) { _ in false }
-	}
-
-	/// Returns a signal that will yield an array of values until it reaches a certain
-	/// count.
+	/// Returns a signal that will yield an array of values until it reaches a 
+	/// certain count.
 	///
 	/// When the count is reached the array is sent and the signal starts over 
 	/// yielding a new array of values.
+	///
+	/// - Precondition: `count` should be greater than zero.
+	///
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
-	public func collect(count: Int) -> Signal<[Value], Error> {
+	public func collect(count count: Int) -> Signal<[Value], Error> {
 		precondition(count > 0)
-		return collect(.Inclusive) { values, _ in values.count == count }
+		return collect { values in values.count == count }
 	}
 
-	/// Returns a signal that will yield an array of values based on a predicate.
+	/// Returns a signal that will yield an array of values based on a predicate
+	/// which matches the values collected.
 	///
-	/// The predicate should return `true` when the values should be sent and `false`
-	/// when the values should be collected. The predicate receives the `values`
-	/// already collected and the next value that should or should not be
-	/// collected.
+	/// - parameter predicate: Predicate to match when values should be sent
+	/// (returning `true`) or alternatively when they should be collected (where
+	/// it should return `false`). The predicate receives the `values` already 
+	/// collected which will already include the next value. By default, uses a
+	/// predicate that collects all the values until `self` completes.
+	///
+	/// #### Example
+	///
+	///     let (signal, observer) = Signal<Int, NoError>.pipe()
+	///
+	///     signal
+	///         .collect { values in values.reduce(0, combine: +) <= 8 }
+	///         .observeNext { print($0) }
+	///
+	///     observer.sendNext(1)
+	///     observer.sendNext(3)
+	///     observer.sendNext(4)
+	///     observer.sendNext(7)
+	///     observer.sendNext(1)
+	///     observer.sendNext(5)
+	///     observer.sendNext(6)
+	///     observer.sendCompleted()
+	///
+	///     // Output:
+	///     // [1, 3, 4]
+	///     // [7, 1]
+	///     // [5, 6]
+	///
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func collect(predicate: (values: [Value]) -> Bool = { _ in false }) -> Signal<[Value], Error> {
+		return Signal { observer in
+			let state = CollectState<Value>()
+
+			return self.observe { event in
+				switch event {
+				case let .Next(value):
+					state.append(value)
+					if predicate(values: state.values) {
+						observer.sendNext(state.values)
+						state.flush()
+					}
+				case .Completed:
+					if !state.isEmpty {
+						observer.sendNext(state.values)
+					}
+					observer.sendCompleted()
+				case let .Failed(error):
+					observer.sendFailed(error)
+				case .Interrupted:
+					observer.sendInterrupted()
+				}
+			}
+		}
+	}
+
+	/// Returns a signal that will yield an array of values based on a predicate
+	/// which matches the values collected and the next value.
+	///
+	/// - parameter predicate: Predicate to match when values should be sent
+	/// (returning `true`) or alternatively when they should be collected (where
+	/// it should return `false`). The predicate receives both the `values` 
+	/// already collected as the next value that may be collected into the 
+	/// current sequence or, otherwise, trigger the flush of the current and
+	/// starting a new one with the next value included.
 	///
 	/// #### Example
 	///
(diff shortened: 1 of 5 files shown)


RuiAAPeres:
As a suggestion I already shared with @dmcrodrigues. We could probably define the current `collect()` as:

```swift
collect(predicate: (values: [Value], next: Value) -> Bool = { _ in true } ) -> -> SignalProducer<[Value], Error>
```

RuiAAPeres:
I think we should use a single operator to address both #2817 and #2815 as it keeps things simpler (simple != easy).

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -299,24 +299,94 @@ extension SignalType {
...
+	///     observer.sendNext(6)
+	///     observer.sendCompleted()
+	///
+	///     // Output:
+	///     // [1, 1]
+	///     // [7]
+	///     // [7]
+	///     // [5, 6]
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func collect(predicate: (values: [Value], next: Value) -> Bool) -> Signal<[Value], Error> {
```
mdiep: After thinking some more, I think it probably doesn't make sense to send `next` as an individual value. Instead, it could be included in `values`:

```swift
public fund collect(predicate: values: [Value] -> Bool) -> Signal<[Value], Error>
```

The `collect(count:)` above, for instance, checks `values.count == count`, but it's only going to be sent after another value should come in. If you pass 5 as the count, it's only going to send after the 6th value arrives (we should add a test for that). If the value was passed as part of `values`, it would correctly send after the 5th value.

dmcrodrigues:
@mdiep I spent some time analyzing this and thinking in alternatives because you're totally right this should not buffer any element more time than it should but at the same time I feel that this operator can (and should) be extremely flexible to handle any use-case, so I would prefer keeping the `next` as an individual value to give a chance to opt-in for their inclusion or exclusion. For the `collect(count:)` it will be irrelevant but I'm trying to have a broader scope for the operator.

I finally come up with a new signature for the operator:

```swift
enum CollectStrategy {
	case Inclusive
	case Exclusive
}

func collect(strategy: CollectStrategy, predicate: (values: [Value], next: Value) -> Bool) -> Signal<[Value], Error>
```

I was a little worried about adding a new type and more complexity around the operator but I think it has the benefit of not making any assumptions about the behaviour that should be contemplated by the operator.

In practice it works the same way, it yields an array of values while the predicate is not verified. The only difference is with the strategy adopted:
- **Inclusive** - predicate is matched after including the new value which guarantees that the element is always included;
- **Exclusive** - predicate is matched before including the next value to guarantee that the element is always excluded from the current array and included in the next one.

We can adopt a default strategy but this makes it totally configurable and flexible. And there's also the benefit of simplifying the operator's implementation which was not very readable. What do you think?

RuiAAPeres:
Two things regarding this:

1. I like it. As it caters for different needs that might show up. I also find it very elegant. 👍
2. The default value, will depend a lot in what each person finds intuitive. For me it's the `.Inclusive`.   

@dmcrodrigues were you able to address this (test wise):

> If you pass 5 as the count, it's only going to send after the 6th value arrives (we should add a test for that)

dmcrodrigues:
> If you pass 5 as the count, it's only going to send after the 6th value arrives (we should add a test for that)

@RuiAAPeres yes, I now have a test to specifically check this and is passing using the brand-new `.Inclusive` 🍏 

About the default value, `.Inclusive` seems a good (and maybe natural) option.

mdiep:
I need to think about this a bit. I'll try to follow up tonight or tomorrow with some thoughts.

dmcrodrigues:
I pushed the implementation using a strategy which may help to visualize the change. For now, I have not adopted a default value and I kept `collect()` and `collect(count: Int)` as convenience operators.

I also have done small changes in an attempt to optimize the operator internally (avoid consecutive memory allocations around the values array), and to guarantee the backward compatibility sending an empty array only when there were not collected any values.

The documentation is not updated yet to reflect this change.

RuiAAPeres:
@dmcrodrigues could you also add your operator to the Playground? ✨

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -299,24 +299,105 @@ extension SignalType {
...
+	///     observer.sendNext(5)
+	///     observer.sendNext(6)
+	///     observer.sendCompleted()
+	///
+	///     // Output:
+	///     // [1, 1]
+	///     // [7]
+	///     // [7, 5, 6]
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func collect(strategy: CollectStrategy, predicate: (values: [Value], next: Value) -> Bool) -> Signal<[Value], Error> {
```
mdiep: Okay, I think we should split this into two methods:

```swift
// inclusive version
public func collect(predicate: (values: [Value]) -> Bool) -> Signal<[Value], Error>

// exclusive version
public func collect(predicate: (values: [Value], next: Value) -> Bool) -> Signal<[Value], Error> 
```

The existing implementation here treats `values` different depending on whether it's inclusive or exclusive. Iff it's inclusive `value.contains(next)`. I think that sort of subtlety is a mistake, but I understand why you did it. If it's inclusive, you want to that to be true.

So I think we should split it into two different APIs. That would address all my reservations about this.

ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
```diff
@@ -496,6 +496,114 @@ class SignalProducerLiftingSpec: QuickSpec {
...
+				let (original, observer) = SignalProducer<Int, NoError>.buffer(1)
+
+				let producer = original.collect(3)
+
+				var expectedValues = [
+					[1, 2, 3],
+					[4, 5, 6],
+					[7, 8, 9],
+					[0]
+				]
```
mdiep: As written, this test doesn't make any guarantees about the timing of the values—whether this was written as `inclusive` or `exclusive`. Can you rewrite the test to match the style of many of the other RAC tests, where the timing of each value can be verified?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -299,24 +299,105 @@ extension SignalType {
...
+		return collect(.Inclusive) { _ in false }
+	}
+
+	/// Returns a signal that will yield an array of values until it reaches a certain
+	/// count.
+	///
+	/// When the count is reached the array is sent and the signal starts over 
+	/// yielding a new array of values.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func collect(count: Int) -> Signal<[Value], Error> {
```
mdiep: I think we should make the `count` parameter a named parameter.

mdiep:
Sorry for the delay in reviewing! I left a few notes about the API that I think will improve things.

dmcrodrigues:
@mdiep I followed your recommendation and now we have two different APIs:

- `collect(predicate: (values: [Value]) -> Bool = { _ in false })`
  - This one guarantees the backward compatibility with `collect()`
- `collect(predicate: (values: [Value], next: Value) -> Bool)`

They are independent of each other but I'm still using `CollectState` to reuse code common to both.

@RuiAAPeres Sure, I'll wait for feedback about this changes and then add it.

ReactiveCocoa/Swift/Signal.swift
dmcrodrigues: Done 👍

ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
dmcrodrigues: I deleted that test which I think was redundant since there was another doing practically the same and that one now have guarantees about the timing of the values.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -299,24 +299,168 @@ extension SignalType {
...
-	/// Returns a signal that will yield an array of values when `self` completes.
+
+	/// Returns a signal that will yield an array of values until it reaches a 
+	/// certain count.
+	///
+	/// When the count is reached the array is sent and the signal starts over 
+	/// yielding a new array of values.
+	///
+	/// - Precondition: `count` should be greater than zero.
+	///
```
mdiep: Can you also note that any remaining values will be sent when the signal completes? (i.e. the last array may not have `count` items.)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -299,24 +299,168 @@ extension SignalType {
...
+	///     observer.sendNext(6)
+	///     observer.sendCompleted()
+	///
+	///     // Output:
+	///     // [1, 3, 4]
+	///     // [7, 1]
+	///     // [5, 6]
+	///
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func collect(predicate: (values: [Value]) -> Bool = { _ in false }) -> Signal<[Value], Error> {
```
mdiep: I think it might be better to use a separate overloaded method for `collect()`, rather than a default value here.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -299,24 +299,168 @@ extension SignalType {
...
+
+	/// Returns a signal that will yield an array of values based on a predicate
+	/// which matches the values collected and the next value.
+	///
+	/// - parameter predicate: Predicate to match when values should be sent
+	/// (returning `true`) or alternatively when they should be collected (where
+	/// it should return `false`). The predicate receives both the `values` 
+	/// already collected as the next value that may be collected into the 
+	/// current sequence or, otherwise, trigger the flush of the current and
+	/// starting a new one with the next value included.
```
mdiep: This is a little confusing and could use some cleanup.

Maybe something like:

> Predicate to match when values should be sent (returning `true`) or alternatively when they should be collected (where it should return `false`). The most recent value (`next`) is not included in `values` and will be the start of the next array of values if the predicate returns false.

mdiep:
This is really great! 👏 I'm very excited about how it turned out.

There a few minor issues to address, then we can merge this!

dmcrodrigues:
@mdiep I followed all of your suggestions. I really like having this flexibility, let's hope it can be useful to others. Let me know if I should squash commits before merging. And thank you for your suggestions and guidance 👍 

@RuiAAPeres I updated the playground, can you check if everything is good? ✨ Since we have multiple versions of `collect` I don't know if I've followed the recommended approach (for documenting).

RuiAAPeres:
@dmcrodrigues could you just add in the playground's comments, the words: "inclusive" and "exclusive". I think they will help.

dmcrodrigues:
@RuiAAPeres great suggestion! Done ✅

mdiep:
Thanks for the contribution!

PR #2836, opened 2016-04-28
Restore var binding

ikesyo:
We can use that without any concerns as of https://github.com/apple/swift-evolution/blob/master/proposals/0003-remove-var-parameters.md#alternatives-considered.


mdiep:
👍

Issue #2535, opened 2015-11-04
[RFC] DynamicProperty enhancements


NachoSoto:
@andymatuschak and I were exploring ways to use `DynamicProperty` more, instead of doing ad-hoc subscriptions.

The particular example we were working with is the typical (in RAC 2 terms) `RAC(imageView, image) = signal`. Turns out that’s awfully hard to do with `RAC 4`: there’s no way to create an extension of `UIImageView` that provides a `MutableProperty` that allows us to do 2 things:

- Guarantee that subscription of a binded producer (or signal) happens on the main thread.
- Change the type of the `DynamicProperty` to `UIImage`.

#### More specifically:
- It’s currently impossible to compose a `MutableProperty`. But even if it were possible (with something like `AnyMutableProperty`), it’s impossible to "make changes” to a signal or producer that gets binded to it.
- `DynamicProperty` is implemented with `Value` `AnyObject?`.

#### Suggested improvements:
- Create some sort of `QueueDispatchingMutableProperty` (see an example below) that dispatches setting the value to a particular queue.
- Change `DynamicProperty` to be generic over `Value`, and doing `as! Value`: this would require removing its `@objc` qualifier, which is a breaking change, but I really can't imagine why you'd want to use `DynamicProperty` from `Obj-C`.

I feel very strongly about the second point, I think that would be a positive change. But I'm not sure at all about the first one. Ideally, instead of something so specific, `MutablePropertyType` could be more easily composed to introduce modifications like this one, or like #2431.

Of course we could just implement this in our own codebase, but I'm thinking about how people can accomplish this using `RAC 4`:

```swift
let someImageProducer: SignalProducer<UIImage?, NoError>
someImageView.imageProperty <~ someImageProducer
```

And ideally this is as easy to implement as this:

```swift
extension UIImageView {
	public var imageProperty: MutablePropertyType<UIImage?> {
		return QueueDispatchingMutableProperty(
			property: DynamicProperty<UIImage?>(object: self, keyPath: "image"),
			dispatchQueue: dispatch_get_main_queue()
		)
	}
}
```

Any thoughts on this appreciated! :star: :heart: 

---------------
Example of something like `QueueDispatchingMutableProperty`:
```swift
public final class QueueDispatchingMutableProperty<Value>: MutablePropertyType {
	private let _valueGetter: () -> Value
	private let _valueSetter: Value -> ()
	private let _producer: () -> SignalProducer<Value, NoError>
[...]

PR #2838, opened 2016-04-28
Make DynamicProperty generic and automatically support bridgeable value types

sharplet:
This PR is based on #2551, and is an attempt at implementing part 2 of the `DynamicProperty` enhancements described in #2535.

As far as I understand, there are two different sets of types that are considered "representable in Objective-C": those that conform to `AnyObject` (implicitly including all reference types), and those that conform to `_ObjectiveCBridgeable`. The main problem in making `DynamicProperty` generic over a single `Value` was how to constrain it in such a way as to _unify_ these two protocols, which are mutually exclusive.

The simplest way I could imagine was to introduce an `ObjectiveCRepresentable` protocol, and extend the other two protocols like so:

```swift
extension AnyObject: ObjectiveCRepresentable {
  // implementation
}

extension _ObjectiveCBridgeable: ObjectiveCRepresentable {
  // implementation
}
```

Then `DynamicProperty` could be declared like so:

```swift
public final class DynamicProperty<Value: ObjectiveCRepresentable>
```

But unfortunately this isn't possible in Swift, because protocol extension are disallowed from having inheritance clauses.

The one way I discovered to unify two protocols is method overloading. For example:

```swift
func doStuff<Object: AnyObject>(object: Object) {
  // do object stuff
}

func doStuff<Object: _ObjectiveCBridgeable>(object: Object) {
  // do bridging stuff
}

// Allows you to invoke the "same" function with values conforming to either protocol...
doStuff(1)
doStuff(MyClass())
```

So ultimately I made the designated initialiser for `DynamicProperty` private, and then exposed two convenience initialisers, overloaded firstly for types conforming to `AnyObject`, and secondly for types conforming to `_ObjectiveCBridgeable`.

Let me know what you think!


ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -1,29 +1,38 @@
...
+
 /// Wraps a `dynamic` property, or one defined in Objective-C, using Key-Value
 /// Coding and Key-Value Observing.
 ///
 /// Use this class only as a last resort! `MutableProperty` is generally better
 /// unless KVC/KVO is required by the API you're using (for example,
 /// `NSOperation`).
-@objc public final class DynamicProperty: RACDynamicPropertySuperclass, MutablePropertyType {
-	public typealias Value = AnyObject?
+public final class DynamicProperty<Object>: MutablePropertyType {
```
NachoSoto: `Object` implies reference semantics, but this type isn't constrained to anything. Maybe `Value` would be better? Then the `Value` `typealias` is probably not needed.

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -1,29 +1,38 @@
...
+	private var property: MutableProperty<Object?>?
 
 	/// The current value of the property, as read and written using Key-Value
 	/// Coding.
-	public var value: AnyObject? {
-		@objc(rac_value) get {
-			return object?.valueForKeyPath(keyPath)
+	public var value: Object? {
+		get {
+			return object?.valueForKeyPath(keyPath).map(representation.extractValue(fromRepresentation:))
```
NachoSoto: Long line. Maybe separate this like operators?
```swift
return object?
    .valueForKeyPath(keyPath)
    .map(representation.extractValue(fromRepresentation:))
```

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -66,3 +75,69 @@ import enum Result.NoError
...
+private struct BridgeableRepresentation<Value: _ObjectiveCBridgeable>: ObjectiveCRepresenting {
+	func extractValue(fromRepresentation representation: AnyObject) -> Value {
+		let object = representation as! Value._ObjectiveCType
+		var result: Value?
+		Value._forceBridgeFromObjectiveC(object, result: &result)
+		return result!
+	}
+
+	func represent(value value: Value) -> AnyObject {
+		return value._bridgeToObjectiveC()
```
NachoSoto: Neat idea!

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -66,3 +75,69 @@ import enum Result.NoError
...
+	func represent(value value: Value) -> AnyObject {
+		return value
+	}
+}
+
+/// Represents values in Objective-C indirectly, via bridging.
+private struct BridgeableRepresentation<Value: _ObjectiveCBridgeable>: ObjectiveCRepresenting {
+	func extractValue(fromRepresentation representation: AnyObject) -> Value {
+		let object = representation as! Value._ObjectiveCType
+		var result: Value?
```
NachoSoto: Maybe define this implicitly unwrapped instead?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -244,6 +244,19 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
 }
 
 
+public func <~ <P: MutablePropertyType, S: SignalType where P.Value == S.Value?, S.Error == NoError>(property: P, signal: S) -> Disposable {
+	return property <~ signal.map { Optional($0) }
```
NachoSoto: You can do `.map(Optional.Some)` here.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -244,6 +244,19 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
...
+}
+
+public func <~ <P: MutablePropertyType, S: SignalProducerType where P.Value == S.Value?, S.Error == NoError>(property: P, producer: S) -> Disposable {
+	return property <~ producer.map { Optional($0) }
```
NachoSoto: Ditto

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -510,14 +510,20 @@ class PropertySpec: QuickSpec {
...
+				weak var dynamicProperty: DynamicProperty<Int>? = property
 				
 				property = nil
 				expect(dynamicProperty).toNot(beNil())
 				
 				object = nil
 				expect(dynamicProperty).to(beNil())
 			}
+
+			it("should support un-bridged reference types") {
```
NachoSoto: :+1:

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -672,10 +678,53 @@ class PropertySpec: QuickSpec {
...
+
+				afterEach {
+					object = nil
+				}
+
+				it("should bridge values sent on a signal to Objective-C") {
+					let (signal, observer) = Signal<Int, NoError>.pipe()
+					property <~ signal
+					observer.sendNext(1)
+					expect(object.rac_value).to(equal(1))
```
NachoSoto: Can you use the modern syntax? `expect(object.rac_value) == 1`

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -672,10 +678,53 @@ class PropertySpec: QuickSpec {
...
+				}
+
+				it("should bridge values sent on a signal producer to Objective-C") {
+					let producer = SignalProducer<Int, NoError>(value: 1)
+					property <~ producer
+					expect(object.rac_value).to(equal(1))
```
NachoSoto: Ditto

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -672,10 +678,53 @@ class PropertySpec: QuickSpec {
...
+				}
+
+				it("should bridge values from a source property to Objective-C") {
+					let source = MutableProperty(1)
+					property <~ source
+					expect(object.rac_value).to(equal(1))
```
NachoSoto: Ditto

NachoSoto:
Wohooo, I'm all in for this 🎉

ReactiveCocoa/Swift/DynamicProperty.swift
sharplet: I've just rebased and pushed two rewritten commits that no longer use `Object`. The main reason I used something other than `Value` was this code:

```swift
return object?.valueForKeyPath(keyPath).map { $0 as! Value }
```

In that situation, `Value` reads as ambiguous: is this the optional or non-optional `Value`? In practice it's the optional `Value?`, which shadows the generic parameter `Value`.

In the new version I used `SwiftValue` so that I could refer to the type parameter explicitly (as opposed to the typealias), and then ultimately was able to remove that in favour of the implicitly defined `Value` typealias. Let me know what you think.

ReactiveCocoa/Swift/DynamicProperty.swift
sharplet: I tried that, but unfortunately it doesn't compile. An `inout Value!` can't be passed to a parameter of type `inout Value?`.

sharplet:
@NachoSoto I've rebased both https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2551 and this PR with your suggested style changes, thanks! Just waiting for the tests.

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -66,3 +73,69 @@ import enum Result.NoError
...
+		return representation as! Value
+	}
+
+	func represent(value value: Value) -> AnyObject {
+		return value
+	}
+}
+
+/// Represents values in Objective-C indirectly, via bridging.
+private struct BridgeableRepresentation<Value: _ObjectiveCBridgeable>: ObjectiveCRepresenting {
```
mdiep: Could this be an extension on `_ObjectiveCBridgeable` instead of a `struct`?

mdiep:
I don't think we're quite ready for `master` to be 5.0. There are a few more open PRs (that I'm trying to work through) that could still target 4.x. Should we (1) leave this PR open until we're ready to commit to 5.0 or (2) retarget this to a 5.0 branch?

ReactiveCocoa/Swift/DynamicProperty.swift
sharplet: Good idea, I'll try this out tomorrow. I believe `AnyRepresentation` will still be necessary because of the associated type.

sharplet: Actually, this is the issue I discussed in the PR description. It's not possible to add protocol conformance in a protocol extension, so that's why I introduced a concrete type here.

RuiAAPeres:
@mdiep I would go with the second option (retarget this to a 5.0 branch).

mdiep:
I created a `RAC5` branch if you'd like to retarget this @sharplet

ikesyo:
> (2) retarget this to a 5.0 branch?

I'm for this too.

As seen in some pull requests (https://github.com/antitypical/Result/pull/149, https://github.com/antitypical/Result/pull/156, https://github.com/Quick/Quick/pull/510, https://github.com/Quick/Nimble/pull/277), to keep compatibility with both Swift 2.2 and Swift 3 should be a huge pain. So it would be better to have `5.0-development` branch or something and support Swift 3 (and SwiftPM) in that branch, I think.

mdiep:
> As seen in some pull requests (antitypical/Result#149, antitypical/Result#156, Quick/Quick#510, Quick/Nimble#277), to keep compatibility with both Swift 2.2 and Swift 3 should be a huge pain. So it would be better to have 5.0-development branch or something RAC5 branch and support Swift 3 (and SwiftPM) in that branch, I think.

That sounds fine to me. I think we should aim to keep simultaneous Swift 2.2/3 support for the final release of RAC 4.

sharplet:
@mdiep I've opened #2852.

PR #2839, opened 2016-04-29
Debug operator

### LINKED ISSUE #2833: Debug events
RxSwift provides a quite handy mechanism for debugging events via the `debug` operator ( more info [here](https://github.com/ReactiveX/RxSwift/blob/77d72dcaee908bcf61baf88080661e748e367557/Documentation/GettingStarted.md#debugging)). There is a `resourceCount` as well, that could be interesting, although I think it would force us to [scatter some macros](https://github.com/ReactiveX/RxSwift/search?utf8=%E2%9C%93&q=resourceCount) around the code base.

I am up for doing the `debug()`, but I would like your opinion on the `resourceCount`). 

Thanks @scotteg for the reference!

RuiAAPeres:
Closes #2833

### DIFF
--- a/Documentation/DebuggingTechniques.md
+++ b/Documentation/DebuggingTechniques.md
@@ -13,6 +13,7 @@ In both cases errors are related to incorrect assumptions about type. Such issue
 
 Below is an example of type-error scenario:
 
+```
 ```swift
 SignalProducer<Int, NoError>(value:42)
     .on(next: { answer in
@@ -38,10 +39,37 @@ let disposable = sideEffectProducer.startWithCompleted {
 
 The code above will not compile too, but with the error `error: cannot convert value of type '(_) -> _' to expected argument type '(Int -> ())?'` on definition of `on` closure. This gives enough of information to locate unexpected `return _` since `on` closure should not have any return value.
 
+#### Binding `DynamicProperty` with `<~` operator
+
+Using the `<~` operator to bind a `Signal` or a `SignalProducer` to a `DynamicProperty` can result in unexpected compiler errors. 
+
+Below is an example of this scenario:
+
+```swift
+let label = UILabel()
+let property = MutableProperty<String>("")
+
+DynamicProperty(object: label, keyPath: "text") <~ property.producer
+```
+
+This will often result in a compiler error: 
+
+> error: binary operator '<~' cannot be applied to operands of type 'DynamicProperty' and 'SignalProducer<String, NoError>'
+DynamicProperty(object: label, keyPath: "text") <~ property.producer
+
+The reason is a limitation in the swift type checker - A `DynamicProperty` always has a type of `AnyObject?`, but the `<~` operator requires the values of both sides to have the same type, so the right side value would have to be `AnyObject?` as well, but usually a more concrete type is used (in this example `String`).
+
+Usually, the fix is as easy as adding a `.map{ $0 }`.
+
+```swift
+DynamicProperty(object: label, keyPath: "text") <~ property.producer.map { $0 }
+```
+
+This allows the type checker to infer that `String` can be converted to `AnyProperty?` and thus, the binding succeeds.
+
 #### Debugging event streams
 
 As mentioned in the README, stream debugging can be quite difficut and tedious, so we provide the `debug` operator. In its  simplest form:
-
 
 ```swift
 let searchString = textField.rac_textSignal()
@@ -57,7 +85,6 @@ The biggest problem with this approach, is that it will continue to ouput in Rel
 1. Comment out the operator: `//.debug()`. This is the simpleste approach, but it's error prone, since you will eventually forget to do this.
 2. Use the `EventLogger` protocol and manipulate the output as you see fit. This is the recommended approach.
 
-
 Let's see how we could leverage the `EventLogger` protocol, so we don't print in Release mode:
 
 ```swift
@@ -69,7 +96,6 @@ final class MyLogger: EventLogger {
         #endif
     }
 }
-
 ```
 
 You would then:
--- a/Documentation/DebuggingTechniques.md
+++ b/Documentation/DebuggingTechniques.md
@@ -84,10 +84,10 @@ The biggest problem with this approach, is that it will continue to ouput in Rel
 1. Comment out the operator: `//.logEvents()`. This is the simpleste approach, but it's error prone, since you will eventually forget to do this.
 2. Pass your own function and manipulate the output as you see fit. This is the recommended approach.
 
-Let's see how we could leverage the `EventLogger` protocol, so we don't print in Release mode:
+Let's see how this would look like if we didn't want to print in Release mode:
 
 ```swift
-func debugLog(event: String, fileName: String, functionName: String, lineNumber: Int) {
+func debugLog(identifier: String, event: String, fileName: String, functionName: String, lineNumber: Int) {
    // Don't forget to set up the DEBUG symbol (http://stackoverflow.com/a/24112024/491239)
    #if DEBUG
       print(event)
@@ -98,8 +98,6 @@ func debugLog(event: String, fileName: String, functionName: String, lineNumber:
 You would then:
 
 ```swift
-let logger = MyLogger()
-
 let searchString = textField.rac_textSignal()
     .toSignalProducer()
     .map { text in text as! String }
@@ -107,7 +105,7 @@ let searchString = textField.rac_textSignal()
     .logEvents(logger: debugLog)
 ```
 
-Finally we also provide the `identifier` parameter. This is useful when you are debugging multiple streams and you don't want to get lost:
+We also provide the `identifier` parameter. This is useful when you are debugging multiple streams and you don't want to get lost:
 
 ```swift
 let searchString = textField.rac_textSignal()
@@ -115,6 +113,16 @@ let searchString = textField.rac_textSignal()
     .map { text in text as! String }
     .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
     .logEvents("✨My awesome stream ✨")
+```
+
+There also cases, specially with [hot signals][[Signals]], when there is simply too much output. For those, you can specify which events you are interested in:
+
+```swift
+let searchString = textField.rac_textSignal()
+    .toSignalProducer()
+    .map { text in text as! String }
+    .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
+    .logEvents(events:[.Disposed])
 ```
 
 
--- a/Documentation/DebuggingTechniques.md
+++ b/Documentation/DebuggingTechniques.md
@@ -13,7 +13,7 @@ In both cases errors are related to incorrect assumptions about type. Such issue
 
 Below is an example of type-error scenario:
 
-```
+```swift
 SignalProducer<Int, NoError>(value:42)
     .on(next: { answer in
         return _
@@ -24,9 +24,9 @@ SignalProducer<Int, NoError>(value:42)
 ```
 
 The code above will not compile with the following error on a `print` call `error: ambiguous reference to member 'print'
-print("Completed.")` To find the actual source of errors signal chains need to be broken apart. Add explicit definitions of closure types on each of the steps:
+print("Completed.")`. To find the actual compile error, the chain needs to be broken apart. Add explicit definitions of closure types on each of the steps:
 
-```
+```swift
 let initialProducer = SignalProducer<Int, NoError>.init(value:42)
 let sideEffectProducer = initialProducer.on(next: { (answer: Int) in
     return _
@@ -37,3 +37,61 @@ let disposable = sideEffectProducer.startWithCompleted {
 ```
 
 The code above will not compile too, but with the error `error: cannot convert value of type '(_) -> _' to expected argument type '(Int -> ())?'` on definition of `on` closure. This gives enough of information to locate unexpected `return _` since `on` closure should not have any return value.
+
+#### Debugging event streams
+
+As mentioned in the README, stream debugging can be quite difficut and tedious, so we provide the `debug` operator. In its  simplest form:
+
+
+```swift
+let searchString = textField.rac_textSignal()
+    .toSignalProducer()
+    .map { text in text as! String }
+    .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
+    .debug()
+```
+
+This will print to the standard output the events. For most use cases, this is enough and will greatly help you understand your flow. 
+The biggest problem with this approach, is that it will continue to ouput in Release mode. This leaves with you with two options:
+
+1. Comment out the operator: `//.debug()`. This is the simpleste approach, but it's error prone, since you will eventually forget to do this.
+2. Use the `EventLogger` protocol and manipulate the output as you see fit. This is the recommended approach.
+
+
+Let's see how we could leverage the `EventLogger` protocol, so we don't print in Release mode:
+
+```swift
+final class MyLogger: EventLogger {
+    func logEvent(event: String) {
+        // Don't forget to set up the DEBUG symbol (http://stackoverflow.com/a/24112024/491239)
+        #if DEBUG
+            print(event)
+        #endif
+    }
+}
+
+```
+
+You would then:
+
+```swift
+let logger = MyLogger()
+
+let searchString = textField.rac_textSignal()
+    .toSignalProducer()
+    .map { text in text as! String }
+    .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
+    .debug(logger: logger)
+```
+
+Finally we also provide the `identifier` parameter. This is useful when you are debugging multiple streams and you don't want to get lost:
+
+```swift
+let searchString = textField.rac_textSignal()
+    .toSignalProducer()
+    .map { text in text as! String }
+    .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
+    .debug("✨My awesome stream ✨")
+```
+
+
--- a/Documentation/DebuggingTechniques.md
+++ b/Documentation/DebuggingTechniques.md
@@ -112,7 +112,7 @@ let searchString = textField.rac_textSignal()
     .toSignalProducer()
     .map { text in text as! String }
     .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
-    .logEvents("✨My awesome stream ✨")
+    .logEvents(identifier: "✨My awesome stream ✨")
 ```
 
 There also cases, specially with [hot signals][[Signals]], when there is simply too much output. For those, you can specify which events you are interested in:
@@ -122,7 +122,7 @@ let searchString = textField.rac_textSignal()
     .toSignalProducer()
     .map { text in text as! String }
     .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
-    .logEvents(events:[.Disposed])
+    .logEvents(events:[.Disposed]) // This will happen when the `UITextField` is released
 ```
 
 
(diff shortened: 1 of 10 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1267,3 +1267,12 @@ private final class DeallocationToken {
 		observer.sendCompleted()
 	}
 }
+
+extension SignalProducerType where Value: Equatable {
+	/// Logs all events that the receiver sends.
+	/// By default, it will print to the standard output.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func debug() -> SignalProducer<Value, Error> {
```
iv-mexx: Here, the parameters `identifier` and `logger` are missing.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1267,3 +1267,12 @@ private final class DeallocationToken {
 		observer.sendCompleted()
 	}
 }
+
+extension SignalProducerType where Value: Equatable {
```
iv-mexx: I think it should not be necessary to constraint `Value` to conform to `Equatable`, or is there?

JaviSoto:
Super excited to have this kind of operator in RAC :)

I would recommend a name more along the lines of `logEvents()`, `debug()` is kind of an overloaded word.
This is the extension method I wrote in the Fabric app. It is tightly coupled with the logging framework we use (XCGLogger), and I love the way you abstracted that in your implementation, but I think it would be nice to allow for more customization: https://gist.github.com/JaviSoto/cff1da9049ebdd3ae92e98c5d45b7f15

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1277,3 +1277,35 @@ extension SignalType where Error == NoError {
 		}
 	}
 }
+
+/// Used with the `debug` operator to customize 
+/// the `Event` string.
+///
+/// This is useful, when you want to do more than
+/// just printing to the standard output.
+public protocol EventLogger {
```
mdiep: I'm not sure I see utility of `EventLogger` when you could just as easily write a `Event<Value, Error> -> ()` with `on(event:)`.

mdiep:
> I would recommend a name more along the lines of `logEvents()`, `debug()` is kind of an overloaded word.

👍 to that.

Documentation/DebuggingTechniques.md
```diff
@@ -14,6 +14,7 @@ In both cases errors are related to incorrect assumptions about type. Such issue
 Below is an example of type-error scenario:
 
 ```
```
rcgary: I think you need to delete this line :)

RuiAAPeres:
> I would recommend a name more along the lines of logEvents(), debug() is kind of an overloaded word.

👍

@JaviSoto I like your approach with the enum + sets! ✨

> but I think it would be nice to allow for more customization

So I will add the following: 

1. `fileName: String = #file`
2. `functionName: String = #function`
3. `lineNumber: Int = #line`

Do you reckon we should go with the Set as well?

@mdiep the protocol opens room for further development, a `func` in this case would work as well, but I feel the protocol is more appropriate in this case.

RuiAAPeres:
Couple of things regarding 4038989: 

1. Moved the operator (utility types + global function) to a different file. This file could be used for other debugging operators.
2. Added @JaviSoto suggestion with the event enum, plus `#filename`, `#functionaName`, `#lineNumber`.

JaviSoto:
The reason why I think the eventTypes filter parameter ends up being necessary is that some signals are very noisy, and sometimes you don't care about values, just start / disposal for example. And if you're adding this operator in more than one place in the app, it becomes even more noisy (and for this reason as well, being able to provide some sort of name or identifier is useful)

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,82 @@
...
+//
+
+import Foundation
+
+/// Used with the `debug` operator to customize
+/// the `Event` string.
+///
+/// This is useful, when you want to do more than
+/// just printing to the standard output.
+public protocol EventLogger {
```
mdiep: Since we don't have a clear use case in mind, I think we should remove `EventLogger` in the initial implementation. It adds complexity and it's much harder to remove API than to add it. (Removing is a breaking change; adding is a non-breaking change in this case.)

There's nothing that prevents anyone from implementation the behavior that this would allow.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,82 @@
...
+}
+
+extension EventLogger {
+	func logEvent(event: String, fileName: String = #file, functionName: String = #function, lineNumber: Int = #line) {
+		print("\(event) fileName: \(fileName), functionaName: \(functionName), lineNumber: \(lineNumber)")
+	}
+}
+
+public enum LoggingEvent: String {
+	case Started, Next, Completed, Failed, Terminated, Disposed, Interrupted
```
mdiep: Would it make sense to make this an inner `enum` on `Signal` and `SignalProducer`? Then `Started` could be removed as a value for `Signal`.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,82 @@
...
+		}
+		
+		let failed: (Self.Error -> Void)? = createEventLog(.Failed, events: events, logEvent: logEvent)
+		let completed: OptionalClosure = createEventLog(.Completed, events: events, logEvent: logEvent)
+		let interrupted: OptionalClosure = createEventLog(.Interrupted, events: events, logEvent: logEvent)
+		let terminated: OptionalClosure = createEventLog(.Terminated, events: events, logEvent: logEvent)
+		let disposed: OptionalClosure = createEventLog(.Disposed, events: events, logEvent: logEvent)
+		let next: (Self.Value -> Void)? = createEventLog(.Next, events: events, logEvent: logEvent)
+
+		return self.on(failed: failed, completed: completed, interrupted: interrupted, terminated: terminated, disposed: disposed, next: next)
```
mdiep: It might be clearer to write this like:

```swift
return on(
    failed: createEventLog(.Failed, events: events, logEvent: logEvent),
    completed: createEventLog(.Completed, events: events, logEvent: logEvent),
    …
)
```

I think that would remove the need for the type annotations on the local variables.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,82 @@
...
+	/// Logs all events that the receiver sends.
+	/// By default, it will print to the standard output.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func logEvents(identifier: String = "", events: Set<LoggingEvent> = LoggingEvent.allEvents, logger: EventLogger = Logger(), fileName: String = #file, functionName: String = #function, lineNumber: Int = #line) -> Signal<Value, Error> {
+		
+		let logEvent: String -> Void = { event in
+			logger.logEvent("[\(identifier)] \(event)", fileName: fileName, functionName: functionName, lineNumber: lineNumber)
+		}
+		
+		let failed: (Self.Error -> Void)? = createEventLog(.Failed, events: events, logEvent: logEvent)
```
mdiep: You could probably move `createEventLog` inside the methods so that they can close over `events` and `logEvent`:

```swift
func log<T>(event: LoggingEvent) -> (T -> Void) {
    return events.contains(event) ? { logEvent("\(event.rawValue) \($0)") } : nil    
}
return on(
    failed: log(.Failed),
    …
)
```

RuiAAPeres: Doing that like so, will work well for `Next` and `Error. But it will do the following for all the others:

```swift
Completed ()
Interrupted ()
```

ReactiveCocoa/Swift/EventLogger.swift
RuiAAPeres: We are repeating a lot of stuff with that approach. 👎

ReactiveCocoa/Swift/EventLogger.swift
RuiAAPeres: I am ok with going with the function, but I am not completely sold. I find it more idiomatic, to with a protocol approach on this.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,66 @@
...
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func logEvents(identifier: String = "", events: Set<LoggingEvent> = LoggingEvent.allEvents, fileName: String = #file, functionName: String = #function, lineNumber: Int = #line, logger: EventLogger = printLog) -> Signal<Value, Error> {
+		
+		let logEvent: String -> Void = { event in
+			logger(event: "[\(identifier)] \(event)", fileName: fileName, functionName: functionName, lineNumber: lineNumber)
+		}
+		
+		func log<T>(event: LoggingEvent) -> (T -> Void)? {
+			return events.contains(event) ? { logEvent("\(event.rawValue) \($0)") } : nil
+		}
```
mdiep: I think you could add a `Void`-specific `log` function here and below like you had originally to avoid the `()` in the output that you mentioned.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,66 @@
...
+		return [.Started, .Next, .Completed, .Failed, .Terminated, .Disposed, .Interrupted]
+	}()
+}
+
+private func printLog(event: String, fileName: String, functionName: String, lineNumber: Int) -> Void {
+	print("\(event) fileName: \(fileName), functionaName: \(functionName), lineNumber: \(lineNumber)")
+}
+
+public typealias EventLogger = (event: String, fileName: String, functionName: String, lineNumber: Int) -> Void
+typealias OptionalClosure = (Void -> Void)?
```
mdiep: This `typealias` isn't needed anymore. 🎉

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,66 @@
...
+		func log<T>(event: LoggingEvent) -> (T -> Void)? {
+			return events.contains(event) ? { logEvent("\(event.rawValue) \($0)") } : nil
+		}
+
+		return self.on(failed: log(.Failed),
+		               completed: log(.Completed),
+		               interrupted: log(.Interrupted),
+		               terminated: log(.Terminated),
+		               disposed: log(.Disposed),
+		               next: log(.Next))
```
mdiep: It'd be a better match to the RAC style to either put these all on one line or to arrange them like so:

```swift
return self.on(
    failed: log(.Failed),
    completed: log(.Completed),
    interrupted: log(.Interrupted),
    terminated: log(.Terminated),
    disposed: log(.Disposed),
    next: log(.Next)
)
```

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+
+import Foundation
+
+public enum LoggingEvent: String {
+	case Started, Next, Completed, Failed, Terminated, Disposed, Interrupted
+	
+	public static let allEvents: Set<LoggingEvent> = {
+		return [.Started, .Next, .Completed, .Failed, .Terminated, .Disposed, .Interrupted]
+	}()
```
mdiep: Gonna restart the discussion about separate `LoggingEvent` `enum`s for `Signal` and `SignalProducer` since the previous one was collapsed.

You mentioned repeated code. The cases would be repeated, and `SignalProducer.logEvents` wouldn't be able to `lift, but I don't think it'd be much more code.

I'm concerned that people will expected `Started` to work on `Signal`s when it doesn't. Separate types would let use enforce correctness with the types.

What do you think? (I'm not trying to force you; just trying to have a discussion. :relaxed:)

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,66 @@
...
+}
+
+private func printLog(event: String, fileName: String, functionName: String, lineNumber: Int) -> Void {
+	print("\(event) fileName: \(fileName), functionaName: \(functionName), lineNumber: \(lineNumber)")
+}
+
+public typealias EventLogger = (event: String, fileName: String, functionName: String, lineNumber: Int) -> Void
```
mdiep: Would it make sense to pass a `LoggingEvent` and a generic `Value?` here instead of a `String`?

Then the default `printLog` could combine them like you did in the inner `log` functions, but custom log functions would be a bit more flexible.

RuiAAPeres: How would you define that? Generics in typeesalias are only available in Swift 3. https://github.com/apple/swift-evolution/blob/master/proposals/0048-generic-typealias.md

ReactiveCocoa/Swift/EventLogger.swift
RuiAAPeres: > You mentioned repeated code.

You will have 2 enums, where the only difference is one case, plus you won't be able to lift. 😕

ReactiveCocoa/Swift/EventLogger.swift
RuiAAPeres: It will need a difference name, I don't think func overload works in nested functions.

mdiep: I think a different name would be fine.

ReactiveCocoa/Swift/EventLogger.swift
mdiep: I think that'd be a win. `lift` isn't really any simpler than calling `on` directly.

ReactiveCocoa/Swift/EventLogger.swift
mdiep: Interesting. I hadn't considered that. Should we drop the `typealias` then? It could be reintroduced with Swift 3 support.

RuiAAPeres: The code without the typealias is unreadable. I understand the goal is to make this as flexible as possible, but I feel that from a code maintenance perspective it's becoming worst.

RuiAAPeres: > Would it make sense to pass a LoggingEvent and a generic Value? here instead of a String?

the event is fine, the problem is with generic value, because we are using the `printLog` as a default value in the `logEvents` func.

RuiAAPeres: I think we should see if this is necessity, before making the code even more complex.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,88 @@
...
+			return events.contains(event) ? { logEvent("\(event.rawValue) \($0)") } : nil
+		}
+		
+		return self.on(
+			failed: logValue(.Failed),
+			completed: log(.Completed),
+			interrupted: log(.Interrupted),
+			terminated: log(.Terminated),
+			disposed: log(.Disposed),
+			next: logValue(.Next))
```
mdiep: Indentation is off a level here.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,88 @@
...
+		}
+		
+		return self.on(
+			started: log(.Started),
+			failed: logValue(.Failed),
+			completed: log(.Completed),
+			interrupted: log(.Interrupted),
+			terminated: log(.Terminated),
+			disposed: log(.Disposed),
+			next: logValue(.Next))
```
mdiep: Here too.

mdiep:
Just needs a couple indentation fixes now. Otherwise it looks good to me. 👍 ✨

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,88 @@
...
+	public static let allEvents: Set<SignalProducerLoggingEvent> = {
+		return [.Started, .Next, .Completed, .Failed, .Terminated, .Disposed, .Interrupted]
+	}()
+}
+
+private func printLog(identifier: String, event: String, fileName: String, functionName: String, lineNumber: Int) -> Void {
+	print("[\(identifier)] \(event) fileName: \(fileName), functionaName: \(functionName), lineNumber: \(lineNumber)")
+}
+
+public typealias EventLogger = (identifier: String, event: String, fileName: String, functionName: String, lineNumber: Int) -> Void
```
JaviSoto: It seems to me that this is as much of a type (that we must maintain backwards compatibility with) than an actual protocol with a method, and I think a protocol is a lot easier to work with.

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,88 @@
...
+
+public enum SignalProducerLoggingEvent: String {
+	case Started, Next, Completed, Failed, Terminated, Disposed, Interrupted
+	
+	public static let allEvents: Set<SignalProducerLoggingEvent> = {
+		return [.Started, .Next, .Completed, .Failed, .Terminated, .Disposed, .Interrupted]
+	}()
+}
+
+private func printLog(identifier: String, event: String, fileName: String, functionName: String, lineNumber: Int) -> Void {
```
JaviSoto: Maybe call this something along the lines of "default event log"?

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,88 @@
...
+	print("[\(identifier)] \(event) fileName: \(fileName), functionaName: \(functionName), lineNumber: \(lineNumber)")
+}
+
+public typealias EventLogger = (identifier: String, event: String, fileName: String, functionName: String, lineNumber: Int) -> Void
+
+extension SignalType {
+	/// Logs all events that the receiver sends.
+	/// By default, it will print to the standard output.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func logEvents(identifier: String = "", events: Set<SignalLoggingEvent> = SignalLoggingEvent.allEvents, fileName: String = #file, functionName: String = #function, lineNumber: Int = #line, logger: EventLogger = printLog) -> Signal<Value, Error> {
```
JaviSoto: What do you think about making identifier an explicit parameter name? That way callers have to say:

logEvents(identifier: "foo")

And not just

logEvents("foo")

ReactiveCocoa/Swift/EventLogger.swift
```diff
@@ -0,0 +1,88 @@
...
+		
+		let logEvent: String -> Void = { event in
+			logger(identifier: identifier, event: event, fileName: fileName, functionName: functionName, lineNumber: lineNumber)
+		}
+		
+		func log(event: SignalLoggingEvent) -> (Void -> Void)? {
```
JaviSoto: Neat!!

JaviSoto:
Just a few small nits, but I love this :) thanks a lot!!!

ReactiveCocoa/Swift/EventLogger.swift
RuiAAPeres: It's rendering correctly on my side. 😕 
screen shot 2016-05-02 at 12 21 49

ReactiveCocoa/Swift/EventLogger.swift
RuiAAPeres: The function approach gives you a bit more flexibility. I am fine with both approaches

ReactiveCocoa/Swift/EventLogger.swift
mdiep: Oops. My mistake. Not sure what I was thinking. 🙊

mdiep:
Looks good to me. 👍

PR #2845, opened 2016-05-02
Fix memory leak in RACSubject

### LINKED ISSUE #1124: Memory leak in RACSubject

ikesyo:
Supersedes #2782 and should fix #1124.

### DIFF
--- a/ReactiveCocoa/Objective-C/RACSubject.m
+++ b/ReactiveCocoa/Objective-C/RACSubject.m
@@ -62,7 +62,7 @@
 		[subscribers addObject:subscriber];
 	}
 	
-	return [RACDisposable disposableWithBlock:^{
+	[disposable addDisposable:[RACDisposable disposableWithBlock:^{
 		@synchronized (subscribers) {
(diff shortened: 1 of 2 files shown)


mdiep:
Thanks @ikesyo! ✨

PR #2850, opened 2016-05-03
Playground subscription behaviour examples

iv-mexx:
These examples illustrate the different behaviours at subscription between `Signal` and `SignalProducer`

The `Signal` example produces the following output:

```
--- Subscription ---

Subscriber 1 subscribes to the signal
Send value `10` on the signal
Subscriber 1 received 10
Subscriber 2 subscribes to the signal
Send value `20` on the signal
Subscriber 1 received 20
Subscriber 2 received 20
```

The `SignalProducer` example produces the following output: 

```
--- Subscription ---

Subscriber 1 subscribes to producer
New subscription, starting operation
Subscriber 1 received 1
Subscriber 1 received 2
Subscriber 2 subscribes to producer
New subscription, starting operation
Subscriber 2 received 1
Subscriber 2 received 2
```

### DIFF
--- a/ReactiveCocoa.playground/Pages/Signal.xcplaygroundpage/Contents.swift
+++ b/ReactiveCocoa.playground/Pages/Signal.xcplaygroundpage/Contents.swift
@@ -43,6 +43,34 @@ import Foundation
  */
 
 /*:
+ ### `Subscription`
+ A Signal represents and event stream that is already "in progress", sometimes also called "hot". This means, that a subscriber may miss events that have been sent before the subscription.
+ Furthermore, the subscription to a signal does not trigger any side effects
+ */
+scopedExample("Subscription") {
+    // Signal.pipe is a way to manually control a signal. the returned observer can be used to send values to the signal
+    let (signal, observer) = Signal<Int, NoError>.pipe()
+    
+    let subscriber1 = Observer<Int, NoError>(next: { print("Subscriber 1 received \($0)") })
+    let subscriber2 = Observer<Int, NoError>(next: { print("Subscriber 2 received \($0)") })
+    
+    print("Subscriber 1 subscribes to the signal")
+    signal.observe(subscriber1)
+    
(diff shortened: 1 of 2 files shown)


ReactiveCocoa.playground/Pages/Signal.xcplaygroundpage/Contents.swift
```diff
@@ -43,6 +43,34 @@ import Foundation
...
+    print("Subscriber 2 subscribes to the signal")
+    // Notice how nothing happens at this moment, i.e. subscriber2 does not receive the previously sent value
+    signal.observe(subscriber2)
+    
+    print("Send value `20` on the signal")
+    // Notice that now, subscriber1 and subscriber2 will receive the value
+    observer.sendNext(20)
+}
+
+/*:
```
RuiAAPeres: Couldn't we use the logEvents operator to automate some of this?

iv-mexx: Hm, we could use logEvent to the same effect:
```swift
let (signal, observer) = Signal<Int, NoError>.pipe()

print("Subscriber 1 subscribes to the signal")
signal.logEvents(identifier: "Subscriber 1")
print("Send value `10` on the signal")
// subscriber1 will receive the value
observer.sendNext(10)

print("Subscriber 2 subscribes to the signal")
signal.logEvents(identifier: "Subscriber 2")

print("Send value `20` on the signal")
// Notice that now, subscriber1 and subscriber2 will receive the value
observer.sendNext(20)
```

But I'm not sure if this is better from a documentation point of view. 
Having explicit subscribers and explicitly adding them to the signal might be more clear about whats going on (for a beginner).

RuiAAPeres:
Makes sense to be explicit (the Playground's goal is to help beginners understanding ReactiveCocoa's behaviour)

PR #2854, opened 2016-05-03
rac_notifications update redux

ikesyo:
Supersedes #2747.

I've added `takeUntil` to the returned producer only when the given object is `NSObject`. The specs are updated accordingly.

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,3 +1,3 @@
 github "jspahrsummers/xcconfigs" ~> 0.9
 github "Quick/Quick" ~> 0.9.2
-github "Quick/Nimble" ~> 4.0
+github "Quick/Nimble" ~> 4.0.1
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,4 +1,4 @@
-github "Quick/Nimble" "v4.0.0"
+github "Quick/Nimble" "v4.0.1"
 github "Quick/Quick" "v0.9.2"
 github "antitypical/Result" "2.0.0"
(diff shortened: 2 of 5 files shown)


RuiAAPeres:
@ikesyo aren't we making the Swift part (RAC 4.0) be dependent on the old objc API, with this:

```swift
+private extension NSObject {
 +	var willDeallocProducer: SignalProducer<(), NoError> {
 +		return rac_willDeallocSignal()
 +			.toSignalProducer()
 +			.map { _ in () }
 +			.mapError { error in
 +				fatalError("Unexpected error: \(error)")
 +				()
 +		}
 +	}
 +}
```

More specifically the `rac_willDeallocSignal`?

RuiAAPeres:
My idea when we would separate this into different repos:

1. Swift
2. Objc 
3. Objc Bridges

The `FoundationExtensions.swift` would stay with Swift repo.

ikesyo:
@RuiAAPeres That's a good point... so how do you think about the discussion of adding `takeUntil` to the returned signal in #2747? As discussed, I think semantically it would be better to add that.

mdiep:
I don't think the `takeUntil:` approach can be implemented without using Obj-C since it depends on associated objects.

ikesyo:
So it seems that giving up the `takeUntil` approach is the direction we would go in at the moment. I would factor out the inlining `liftRight` in `SignalProducer.takeUntil` (d6c2ba1) in a separate pull request.

ikesyo:
Superseded by #2858 and #2859.

PR #2858, opened 2016-05-04
Inline liftRight in SignalProducer.takeUntil

ikesyo:
This is originally done in #2854. See also #2751.


mdiep:
👍

mdiep:
Did you try this with Xcode 7.3.1 to see if it's fixed?

ikesyo:
I've tried and this and #2751 are fixed in Xcode 7.3.1. But I think we should wait for Travis to support Xcode 7.3.1 in their OS X image.

PR #2859, opened 2016-05-04
rac_notifications update redux

ikesyo:
Supersedes #2747 and #2854.

Gave up the `takeUntil` approach (which is originally suggested in #2747) as discussed in #2854.


mdiep:
👍 Thanks!

andymatuschak:
Late to the party, but thank you for seeing this through @ikesyo!

Issue #2831, opened 2016-04-27
Remove SignalProducer.buffer()


mdiep:
As discussed in #2744.

This is a breaking change, so it will need to be part of the 5.0 release.

ikesyo:
We should also refactor `SignalProducer.replayLazily` not to use `buffer` or consider removing it as well.

filblue:
+ for removing `replayLazily`

mdiep:
This was done in #3028.

PR #2860, opened 2016-05-04
Try to avoid using `SignalProducer.buffer`

ikesyo:
Because that will be removed in RAC 5. `SignalProducer.replayLazily` isn't modified here.


RuiAAPeres:
This is tied to #2831

mdiep:
✨

PR #2862, opened 2016-05-05
DynamicProperty operators tweaks

ikesyo:
- Use `SignalType` / `SignalProducerType` and generic constraints for `DynamicProperty` operators
- Factor out the operators into `DynamicProperty.swift`


mdiep:
👍

PR #2864, opened 2016-05-05
[Travis] Make builds a bit faster

ikesyo:
- Disable static analysis during build
- Don't generate debug symbols

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ script:
       -sdk "$XCODE_SDK"
       -configuration Release
       ENABLE_TESTABILITY=YES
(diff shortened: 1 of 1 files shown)


dmcrodrigues:
This won't help the build times, on the contrary, but shouldn't we also build `tvOS` and `watchOS` targets? And probably create a test target for `tvOS`?

mdiep:
Thanks @ikesyo!

> but shouldn't we also build tvOS and watchOS targets? And probably create a test target for tvOS?

Yeah, those would be great to add to the Travis matrix!

dmcrodrigues:
Cool, I can help with that.

PR #2865, opened 2016-05-05
Style fixes for Playground

ikesyo:


ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1409,9 +1409,9 @@ class SignalProducerSpec: QuickSpec {
 			}
 
 			describe("disposal") {
-				var completeOuter: (Void -> Void)!
-				var disposeOuter: (Void -> Void)!
-				var execute: (FlattenStrategy -> Void)!
+				var completeOuter: (() -> ())!
+				var disposeOuter: (() -> ())!
+				var execute: (FlattenStrategy -> ())!
```
RuiAAPeres: I have to be honest, I hate `()` instead of `Void`, it makes it much more difficult to read. 👎

ikesyo: I prefer `Void` too, but that was previously rejected in #
2059.

RuiAAPeres: That issue was open almost one year ago, maybe we could re-open this discussion?

mdiep: Let's just be consistent in this PR. I don't have a strong opinion and would be open to changing to `Void`. (Particularly if we're going to match the Swift stdlib style.) But let's do that consistently across the codebase if we're going to do it.

RuiAAPeres: 👍

mdiep:
✨

ikesyo:
Thanks @RuiAAPeres and @mdiep. I'll open a separate PR for `Void` return types.

PR #2878, opened 2016-05-06
Resolved a couple of missing newlines that are contrary to the specified project coding conventions.

geoffspears:
Added GCC_WARN_ABOUT_MISSING_NEWLINE to warn for this deviation moving forward.


mdiep:
I love this! It's always great when the compiler does our work for us.

Unfortunately, CI fails because it's trying to apply the warning to a file from one of our test dependencies. Could you try to fix that up?

geoffspears:
I submitted a PR to the Nimble project to resolve the missing new line which was the source of the test build failure. I'll follow up in this thread with any status updates.

mdiep:
My larger concern is that the error shouldn't apply to our dependencies. It should only apply to the RAC codebase.

geoffspears:
Unfortunately, it seems that when the missing newline is in an imported header files that this issue arises. The absence of newlines in the source of the dependencies doesn't have this same undesirable effect.

mdiep:
I wonder if the error would go away if we used `@import`?

Otherwise I don't think we can use the compiler flag. But I'd still take the addition of the newlines!

geoffspears:
Switching to use @import didn't have any benefit when the headers without newlines were included. I reverted the compiler flag so that we can at least move forward with the trivial fix to add the new lines.

mdiep:
Thanks!

Issue #2807, opened 2016-04-20
Split Objective-C and Swift APIs going forward


tonyarnold:
Hey guys, look you know I ❤️ you, and I :heart: ReactiveCocoa, but lately I've been seeing someone else. The library is small, it's simple, and while it's not as robust or mature as ReactiveCocoa is, the straightforward API makes it a bit lovely to work with. I'm not going to [name names](https://github.com/ReactiveKit/ReactiveKit)… we're both bigger than that. 

Correct me if I'm wrong, but maintaining both an Objective-C and Swift API for RAC is a bit of a burden, right? Perhaps going forward it's worth thinking about slimming ReactiveCocoa down to just support Swift. I'm sure you guys feel the pull of writing things in Swift as much as I do 😀 

Obviously there's a lot of people reliant on the Objective-C APIs (myself included, for some rather large projects) but if there's support for this idea, splitting the language support up into separate repos/projects might be a good first step?

What do you think?

iv-mexx:
👍 for splitting the ObjC and Swift APIs. 

One thing to keep in mind though is the interoperability between both. 
Theres not only projects that use the ObjC API and projects that use the Swift API, but also projects that use both. 
I myself have at least one big project that uses both APIs and relies heavily on [ObjectiveCBridging.swift](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/Swift/ObjectiveCBridging.swift)

I guess for me it would work to have 3 repos/projects: 

* ReactiveCocoa Swift
* ReactiveCocoa ObjC
* ReacitveCocoa Swift <-> ObjC Bridge

carlossless:
Well, to offer a thirsty rando counter question: What's the benefit of splitting the two?
Not to be dismissive, but it seems like that would impose a bunch of busy work (both for maintainers and framework consumers), with no currently tangible gain.

neilpa:
So the core Swift library can support non-apple platforms where we won't have Obj-C/Cocoa.

carlossless:
@neilpa that's all the convincing I need.

ikesyo:
> So the core Swift library can support non-apple platforms that don't have Obj-C/Cocoa.

In SwiftPM, we can exclude some source files in `Package.swift` and can use `#if SWIFT_PACKAGE` for conditional compilation, so this is not necessarily needed for that purpose.

dmcrodrigues:
[...]

PR #2880, opened 2016-05-06
[RFC] Add type-safe method lifting with weak reference semantics

sharplet:
This is a potential improvement I'd like to submit for review as part of #2535.

---

### Overview: `Method` and `MethodType`

This change introduces a new `MethodType` protocol and concrete `Method` class. A `Method` encapsulates a weak reference to an object, and a function that accepts a reference to the object, some input, and returns the method's output.

When a `Method` is lifted using a signal or signal producer of input, it returns a signal or signal producer (respectively) of output from the results of calling the function with the input.

If the weakly-referenced object is deallocated, the lifted signal or signal producer will send an interrupted event. When the Objective-C runtime is present, this is achieved by observing the object's deallocation at runtime. There is a fallback implementation for native Swift objects: even though it's impossible to be notified when a weakly-referenced Swift object is deallocated, we can wait until the next input value is sent and lazily interrupt the signal if the result of the method call is `nil`.

---

### Also: Even better bindings?

The changes as they are right now are not the end goal: I'd also like to add support for the `<~` operator, and create a unifying `Bindable` protocol that both `MethodType` and `MutablePropertyType` would conform to. This would mean that the same set of overloads would work for both kinds of bindable things. Just spitballing here, but this might look something like:

```swift
func <~ <Target: Bindable>(target: Target, signal: Signal<Target.Input, NoError>) -> Disposable
func <~ <Target: Bindable>(target: Target, producer: SignalProducer<Target.Input, NoError>) -> Disposable
// etc.
```

And then in use, you might end up with something like this:

```swift
RAC(refreshControl, UIRefreshControl.endRefreshing) <~ feedController.feedRefresh
RAC(tableView, UITableView.reloadData) <~ feedController.feedRefreshed
```

Also, `Method` solves the "I shouldn't have to set up KVO to bind to a property" problem:

```swift
extension UIImageView {
  var rac_setImage: Method<UIImageView, UIImage, Void> {
    return Method(object: self) { $0.image = $1 }
  }
}

// ...
imageView.rac_setImage <~ viewModel.image
```

---

I plan to push up more comments fleshing these other features out.

Would love to hear your thoughts!

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -18,6 +18,10 @@
 		4A3F079F1CDD089B00E568D3 /* MethodSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F079A1CDD07EE00E568D3 /* MethodSpec.swift */; };
 		4A3F07A01CDD089C00E568D3 /* MethodSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F079A1CDD07EE00E568D3 /* MethodSpec.swift */; };
 		4A3F07A11CDD089D00E568D3 /* MethodSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F079A1CDD07EE00E568D3 /* MethodSpec.swift */; };
+		4A3F07A31CE00C1200E568D3 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F07A21CE00C1200E568D3 /* Bindable.swift */; };
+		4A3F07A41CE00C1200E568D3 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F07A21CE00C1200E568D3 /* Bindable.swift */; };
+		4A3F07A51CE00C1200E568D3 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F07A21CE00C1200E568D3 /* Bindable.swift */; };
+		4A3F07A61CE00C1200E568D3 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F07A21CE00C1200E568D3 /* Bindable.swift */; };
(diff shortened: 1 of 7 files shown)


mdiep:
> When the Objective-C runtime is present, this is achieved by observing the object's deallocation at runtime. There is a fallback implementation for native Swift objects: even though it's impossible to be notified when a weakly-referenced Swift object is deallocated, we can wait until the next input value is sent and lazily interrupt the signal if the result of the method call is nil.

My initial thought is that I'm hesitant to add features that more-or-less depend on Objective-C right now. Given our plan to split this repo up (#2807), I don't think it's a great time to evaluate the potential of a feature like this.

sharplet:
Thanks for the heads-up about the repo splitting, I wasn't aware! Do you think it would be worth adding a note about the roadmap to CONTRIBUTING.md, or perhaps even in an issue or pull request template?

> My initial thought is that I'm hesitant to add features that more-or-less depend on Objective-C right now.

FWIW, I still think there's some value to the idea even without Objective-C—I kind of see `lift()` as a safer version of `map()` that adds the notion of object lifetime. UI bindings that can be torn down automatically (without explicit `Disposable` juggling) are then an obvious use-case of the feature.

sharplet:
Just FYI, I've pushed a version of the `Bindable` protocol I mentioned. That should complete all of the functionality I described.

mdiep:
Thanks for the pull request, but I don't think this is a good fit for the RAC core. Unfortunately, I don't really have any concrete feedback for you.

But I will say that your last example, `imageView.rac_setImage <~ viewModel.image`, is something that we would use `MutableProperty` for.

sharplet:
Ok, thanks for considering it. There was a problem this was solving for me, what's the best forum to discuss solving the problem in a different way? I could perhaps raise an issue with a more concrete example?

mdiep:
> There was a problem this was solving for me, what's the best forum to discuss solving the problem in a different way? I could perhaps raise an issue with a more concrete example?

💯 this.

A better understanding of the issue might lead to (a) reconsideration of this PR or (b) a better solution. 😄

sharplet:
I opened an issue: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2946.

PR #2881, opened 2016-05-07
Change void return types to using Void

ikesyo:
Let's use `Void` over `()` for return types. This should make the codebase more readable. And also using `()` over `Void` for an empty parameter list makes sense to me. See the followings for the details:

- Originally motivated by http://ericasadun.com/2015/05/11/swift-vs-void/
- The original PR: #2059
- The recent discussion: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2865#discussion_r62208760


RuiAAPeres:
⭐

mdiep:
💖

Issue #2790, opened 2016-04-11
Moving Rex to the ReactiveCocoa org


RuiAAPeres:
This has been [suggested informally before](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2766#issuecomment-200044265), let's try to formalize it. 

### Why?

1. **Discoverability**. [Rex](https://github.com/neilpa/Rex) is not well known in the community. You can see that by some of the pull requests/issues open, where the reply is "This would suit Rex better" or "Check Rex".  By making it part of the ReactiveCocoa org, people will easily find it, (by navigating to the [Organization root](https://github.com/ReactiveCocoa)) and by linking to it in the README.  
2. **Credibility**. When adding a new dependency to a project, one usually checks who the author is, the support it will receive and how well maintained it is. Having the ReactiveCocoa's name behind it , will help. Of course, I vouch for @neilpa's skills and for Rex's quality, not only that, I am aware of the work he has done both here (ReactiveCocoa main repo) and Rex. My guess, is that most users wouldn't know that. 
3. **Expansion**. ReactiveCocoa focus a lot in making sure its core is not polluted with unrelated operators/features. In one hand this is great, because it doesn't clutter the API and keeps the dependency small, but on the other a ton of awesome operators are left out. A big group that hasn't been receiving attention is the UI one. Although ReactiveCocoa does offer them, the user needs to bridge them from the old objective-c code base (RACSignal to SignalProducer/Signal). Rex already has a quite good catalogue that would definitely benefit ReactiveCocoa org. This also ties up to the staleness in this Repo. Since it's in a good place (with 4.1 release) I think it's time to start moving forward (with new operators and first class citizen UI binds).
4. **Easier to manage**.  @neilpa has been doing a great job reviewing and merging pull requests, but this would drastically improve, by sharing the burden with the rest of the ReactiveCocoa members.

### Next steps

Well, of course both @neilpa and the rest of the team need to agree on this and move the repo's ownership to ReactiveCocoa org. Regarding urls, [Github seems to do all the heavy lifting](https://help.github.com/articles/transferring-a-repository/#redirects-and-git-remotes).

czechboy0:
👍

neilpa:
I'm in favor of moving Rex over to the ReactiveCocoa org. It started as a personal playground but morphed into something more useful. I also don't use RAC in my day job anymore giving ownership to the community makes sense.

[...]

PR #2883, opened 2016-05-07
CocoaAction separation

RuiAAPeres:
Since we want a "pure" repo for the Swift implementation and CocoaAction uses KVO, it should be separated.

### DIFF
--- a/ReactiveCocoa/Swift/Association.swift
+++ b/ReactiveCocoa/Swift/Association.swift
@@ -3,7 +3,7 @@ import Foundation
 /// On first use attaches the object returned from `initial` to the `host` object using
 /// `key` via `objc_setAssociatedObject`. On subsequent usage, returns said object via
 /// `objc_getAssociatedObject`.
-public func associatedObject<Host: AnyObject, T: AnyObject>(host: Host, key: UnsafePointer<()>, @noescape initial: Host -> T) -> T {
+func associatedObject<Host: AnyObject, T: AnyObject>(host: Host, key: UnsafePointer<()>, @noescape initial: Host -> T) -> T {
 	var value = objc_getAssociatedObject(host, key) as? T
 	if value == nil {
 		value = initial(host)
--- a/ReactiveCocoa/Swift/Association.swift
+++ b/ReactiveCocoa/Swift/Association.swift
@@ -1,0 +1,13 @@
+import Foundation
+
+/// On first use attaches the object returned from `initial` to the `host` object using
+/// `key` via `objc_setAssociatedObject`. On subsequent usage, returns said object via
+/// `objc_getAssociatedObject`.
+public func associatedObject<Host: AnyObject, T: AnyObject>(host: Host, key: UnsafePointer<()>, @noescape initial: Host -> T) -> T {
+	var value = objc_getAssociatedObject(host, key) as? T
+	if value == nil {
+		value = initial(host)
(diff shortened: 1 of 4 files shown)


RuiAAPeres:
Since you can't have a stored property in an extension, should we use associated properties for this? I could also aim this PR to the 5.0 branch and remove this from the API.

RuiAAPeres:
The issue with the tests, is that since we are not holding strongly to the `CocoaAction`, it get's out of scope and the tests fail. This dramatically change the semantics, holding it strongly fixes the issue.

ReactiveCocoa/Swift/CocoaAction.swift
```diff
@@ -0,0 +1,89 @@
...
+		_execute(input)
+	}
+	
+	public override class func automaticallyNotifiesObserversForKey(key: String) -> Bool {
+		return false
+	}
+}
+
+extension Action {
+
```
mdiep: This empty line should be removed to adhere to the RAC style guide.

mdiep:
> We can't have a stored property in an extension, should we use associated properties for this?

Yeah, I think so. 👍 

Thanks for working on this! ⚡

RuiAAPeres:
@mdiep What do you think about [this](https://github.com/neilpa/Rex/blob/master/Source/Foundation/Association.swift#L70#L77)? When we move Rex to this org, it could use that specific method, so we would remove duplication then.

mdiep:
I think including that is a fine idea. 👍

ikesyo:
This seems to be ready now.

ReactiveCocoa/Swift/Association.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+/// On first use attaches the object returned from `initial` to the `host` object using
+/// `key` via `objc_setAssociatedObject`. On subsequent usage, returns said object via
+/// `objc_getAssociatedObject`.
+public func associatedObject<Host: AnyObject, T: AnyObject>(host: Host, key: UnsafePointer<()>, @noescape initial: Host -> T) -> T {
```
mdiep: I think we should make this `internal`.

mdiep:
Just the one note. ☺️

RuiAAPeres:
If we make it internal, we won't be able to use it once Rex is in this org (or even before that). Are you ok with this duplication?

mdiep:
> If we make it internal, we won't be able to use it once Rex is in this org (or even before that). Are you ok with this duplication?

Yes. Although my thinking is that we should move the Rex code into this repo as part of the split. Let's make it `internal` for now and discuss the rest later. 😄

RuiAAPeres:
> Although my thinking is that we should move the Rex code into this repo as part of the split.

You mean, adding Rex to the Swift + Foundation Extension?

mdiep:
https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2790#issuecomment-218942290 😄

RuiAAPeres:
@mdiep I misunderstood that bit I thought:

1. Swift + Foundation
2. Rex (with a new name)

mdiep:
✨ 

> I misunderstood that bit, I thought:
> 
> 1. Swift + Foundation
> 2. Rex (with a new name)

I'm just proposing that Rex's new name should be ReactiveCocoa and it should be merged with the things we currently have that don't fit with Swift or Obj-C.

PR #2888, opened 2016-05-09
Allow DynamicProperty to be initialized with an optional object

sharplet:
This regression was introduced in 0059830a20e29956f0d6d5dc1a7d08920e1b8b6e.


mdiep:
Thanks! 😅

Issue #2890, opened 2016-05-09
Throttle misdescribed in the README


tditchek:
The online search example in the wiki claims the `throttle` operator "prevents values from being sent less than 0.5 seconds apart, so the user must stop editing for at least that long before we’ll use their search string." However, it's not working that way for me. Instead of waiting 0.5 seconds, it is sending the values every 0.5 seconds, even as i continuously type. Here's my code:
```
textField.rac_textSignal()
.toSignalProducer()
.map { $0 as! String }
.throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
.startWithNext { text in
   print("throttled text: \(text)") }
```

Did `throttle` get changed at some point? If not, does anybody know what I'm doing wrong? Thanks!

mdiep:
Unfortunately, that's an error in the documentation in the RAC README.

RAC 2's `throttle` behaved like that. However, that operator should actually be called `debounce`. When the README was updated, it probably was mistranslated to Swift. Would you be interested in opening a PR to fix the documentation in the README?

There is an issue open to implement debounce (#2075), but it hasn't been implemented.

PR #2891, opened 2016-05-10
fixed misrepresentation of throttle in README.md (Issue #2890)

tditchek:
#2890

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,8 +183,8 @@ let searchStrings = textField.rac_textSignal()
     .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
 ```
 
-This prevents values from being sent less than 0.5 seconds apart, so the user
-must stop editing for at least that long before we'll use their string.
+This prevents values from being sent less than 0.5 seconds apart, so the user 
+must stop editing for at least that long  before we'll use their string.
 
 To do this manually would require significant state, and end up much harder to
 read! With ReactiveCocoa, we can use just one operator to incorporate _time_ into
--- a/README.md
+++ b/README.md
@@ -170,7 +170,8 @@ let searchResults = searchStrings
 
 #### Throttling requests
 
-Now, let’s say you only want to actually perform the search periodically, to minimize traffic.
+Now, let’s say you only want to actually perform the search periodically,
+to minimize traffic.
 
 ReactiveCocoa has a declarative `throttle` operator that we can apply to our
 search strings:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -170,8 +170,7 @@ let searchResults = searchStrings
 
 #### Throttling requests
 
-Now, let’s say you only want to actually perform the search when the user pauses
-typing, to minimize traffic.
+Now, let’s say you only want to actually perform the search periodically, to minimize traffic.
```
mdiep: Can you keep the line wrapping to match the rest of the file?

README.md
```diff
@@ -183,8 +182,7 @@ let searchStrings = textField.rac_textSignal()
     .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
 ```
 
-This prevents values from being sent less than 0.5 seconds apart, so the user
-must stop editing for at least that long before we’ll use their search string.
+This will result in the values being sent _every_ 0.5 seconds after the first event, so the user must stop editing for at least that long before we'll use their search string.
```
mdiep: This isn't really accurate. Values aren't sent every 0.5 seconds. That part was accurate before—it "prevents values from being sent less than 0.5 seconds apart". And the user doesn't need to stop editing before we'll use their search string—it means we'll use their search string at most every 0.5 seconds, and we'll get the final value at most 0.5 seconds after they type it.

Here's the documentation for `throttle`:

```swift
	/// Throttle values sent by the receiver, so that at least `interval`
	/// seconds pass between each, then forwards them on the given scheduler.
	///
	/// If multiple values are received before the interval has elapsed, the
	/// latest value is the one that will be passed on.
	///
	/// If the input signal terminates while a value is being throttled, that value
	/// will be discarded and the returned signal will terminate immediately.
```

mdiep:
Thanks for opening a PR! ✨ 

I left a few comments that need to be addressed. But I'm excited to get this fixed!

tditchek:
Sure thing! I'm just getting started with contributing to open source
projects, but I'll check out your comments and handle them asap. Thanks for
the feedback!

Teddy

mdiep:
@tditchek Are you still interested in taking another shot at this? 😄

tditchek:
@mdiep yeah, hope this edit's a little better

README.md
```diff
@@ -183,8 +183,8 @@ let searchStrings = textField.rac_textSignal()
     .throttle(0.5, onScheduler: QueueScheduler.mainQueueScheduler)
 ```
 
-This prevents values from being sent less than 0.5 seconds apart, so the user
-must stop editing for at least that long before we’ll use their search string.
+This prevents values from being sent less than 0.5 seconds apart, so the user 
+must stop editing for at least that long  before we'll use their string.
```
mdiep: > so the user must stop editing for at least that long  before we'll use their string.

That part is not true. The user doesn't need to stop editing. Can you take it out?

mdiep:
Thank you!!!

Issue #2879, opened 2016-05-06
Add Playground step to .travis.yml


RuiAAPeres:
The playground should be validated by travis, like [here](https://github.com/ReactiveX/RxSwift/blob/master/scripts/validate-playgrounds.sh).

Palleas:
I started working on this and will probably open a PR later today. 😁

RuiAAPeres:
@Palleas 👍

PR #2902, opened 2016-05-16
Validate playground

Palleas:
This pull request addresses #2879.

I moved the command building the framework into a bash script, called "build". Should I go for something more explicit with "lint", "check"..? Also I'm not sure I should lint the playground with all the platform, I should probably do it only for the mac.

This is a WIP until the Travis build actually passes. 🍀

### DIFF
--- a/script/build
+++ b/script/build
@@ -13,6 +13,16 @@ if [[ -z $TRAVIS_XCODE_SCHEME ]]; then
 	TRAVIS_XCODE_SCHEME="ReactiveCocoa-Mac"
 fi
 
+if [[ -z $XCODE_ACTION ]]; then
+      echo "\$XCODE_ACTION is not set. Using \`build\`"
+      XCODE_ACTION="build"
+fi
+
+if [[ -z $XCODE_SDK ]]; then
+      echo "\$XCODE_SDK is not set. Using \`macosx\`"
+      XCODE_SDK="macosx"
+fi
+
 xcodebuild $XCODE_ACTION \
       -workspace "$TRAVIS_XCODE_WORKSPACE" \
       -scheme "$TRAVIS_XCODE_SCHEME" \
--- a/script/build
+++ b/script/build
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+BUILD_DIRECTORY="build"
+CONFIGURATION=Release
+
+if [[ -z $TRAVIS_XCODE_WORKSPACE ]]; then
+	echo "\$TRAVIS_XCODE_WORKSPACE is not set. Using \`ReactiveCocoa.xcworkspace\`"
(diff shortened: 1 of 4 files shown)


script/build
```diff
@@ -0,0 +1,29 @@
...
+      -sdk "$XCODE_SDK" \
+      -derivedDataPath "${BUILD_DIRECTORY}" \
+      -configuration $CONFIGURATION \
+      ENABLE_TESTABILITY=YES \
+      GCC_GENERATE_DEBUGGING_SYMBOLS=NO \
+      RUN_CLANG_STATIC_ANALYZER=NO \
+      | xcpretty
+
+# Compile code in playgrounds 
+# TODO only run for mac ?
```
RuiAAPeres: Since the playground is targeting Mac, yes.

ikesyo:
It seems that this is failing:

```bash
script/validate-playground.sh: line 15: build/Build/Products/Release/all-playground-pages.swift: No such file or directory
Apple Swift version 2.2 (swiftlang-703.0.18.8 clang-703.0.31)
Target: x86_64-apple-macosx10.9
<unknown>:0: error: no such file or directory: 'build/Build/Products/Release/all-playground-pages.swift'
```

Palleas:
The build seems to be passing now and I've squashed my numerous commits. Please let me know if I need to change anything else in this PR. 😸

script/build
```diff
@@ -0,0 +1,43 @@
...
+
+if [[ -z $XCODE_ACTION ]]; then
+      echo "\$XCODE_ACTION is not set. Using \`build\`"
+      XCODE_ACTION="build"
+fi
+
+if [[ -z $XCODE_SDK ]]; then
+      echo "\$XCODE_SDK is not set. Using \`macosx\`"
+      XCODE_SDK="macosx"
+fi
```
mdiep: I'm not wild about duplicating these is in the script. Can we just require them to be set?

script/build
```diff
@@ -0,0 +1,43 @@
...
+      GCC_GENERATE_DEBUGGING_SYMBOLS=NO \
+      RUN_CLANG_STATIC_ANALYZER=NO \
+      | xcpretty
+
+# Compile code in playgrounds 
+if [[ $XCODE_SDK -eq "macosx" ]]; then
+      echo "SDK is $XCODE_SDK, validating playground..."
+      . script/validate-playground.sh
+else
+      echo "SDK is $XCODE_SDK, not validating playground."
```
mdiep: I don't think we need to echo in the else case.

mdiep:
Thanks for talking this! I've left a few notes. ☺️ 

FWIW, you don't need to squash your commits.

Palleas:
The script now fails if vars are not set and nothing is displayed if we don't validate the playground. Hopefully the build should still pass!

mdiep:
Looks good to me. 👍

PR #2903, opened 2016-05-17
Fix replayLazily lifetime

### LINKED ISSUE #2728: Proper use of replayLazily in ReactiveCocoa 4
After reading that `replayLazily` was [added](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2639) to ReactiveCocoa 4, I wanted to try it out, but after experiencing some issues, I'm hoping to get some advice.

I have the following:
```swift
        let fetchMenu = Menu.fetchTodaysMenuFromCacheOrRemote(self.menuAPI)
            .replayLazily(1)

        let menuProducer = merge([active, inactive])
            .filter { $0 }
            .mapError { _ in
                return Menu.FetchMenuError.UnknownError
            }
            .flatMap(.Latest) { _ in
                return fetchMenu
            }
```

The type of `menuProducer` is `SignalProducer<Menu, FetchMenuError>`. Basically the intention of the above code is: whenever the view gets active I want to fetch the menu. Because there are multiple subscribers to `menuProducer` I've added the `replayLazily` in order to avoid executing this code multiple times when the subscribers come in.

Although the above code works, the issue is that now, when the user puts the app into the background and bring it back to the foreground, a menu is not fetched. This is obviously because of `replayLazily`. So.. how do I avoid running this code multiple times when multiple subscribers subscribe to this producer, but still being able to run the code every time the view gets active?

mdiep:
Fixes #2728.

`replayLazily` was interrupting the returned producers when `self` went out of scope—even if there were active subscriptions. Adds a strong reference to the deallocation token to each producer so that they won't be interrupted until all producers have also been disposed.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1412,6 +1412,7 @@ extension SignalProducerType {
 		let token = DeallocationToken()
 
 		return SignalProducer { observer, disposable in
+			var token: DeallocationToken? = token
 			let initializedProducer: SignalProducer<Value, Error>
 			let initializedObserver: SignalProducer<Value, Error>.ProducedSignal.Observer
 			let shouldStartUnderlyingProducer: Bool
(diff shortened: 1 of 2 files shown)


ikesyo:
Makes sense. 👍

PR #2908, opened 2016-05-18
[CocoaAction] Use private setter for some properties

ikesyo:


ReactiveCocoa/Swift/CocoaAction.swift
```diff
@@ -42,15 +36,15 @@ public final class CocoaAction: NSObject {
 			.observeOn(UIScheduler())
 			.startWithNext { [weak self] value in
 				self?.willChangeValueForKey("enabled")
-				self?._enabled = value
+				self?.enabled = value
 				self?.didChangeValueForKey("enabled")
```
mdiep: Are the will/did change still needed here?

ikesyo: They aren't if we remove `public override class func automaticallyNotifiesObserversForKey` and add `dynamic` modifier to `enabled` and `executing`:

```swift
	public private(set) dynamic var enabled: Bool = false
	public private(set) dynamic var executing: Bool = false
```

Are you OK with this?

mdiep: Ah, okay. I wasn't sure. Let's keep it as-is. 👍

PR #2910, opened 2016-05-18
Break retain cycle with Action.unsafeCocoaAction

### LINKED ISSUE #2233: Using unsafeCocoaAction leads to memory leak
It's very easy to introduce retain cycle using unsafeCocoaAction, since CocoaAction and Action retains each other.
Also it has naming issue: it suggests to set your own cocoa action to this property of your input isn't Void, thus making it not unsafe anymore.

mdiep:
Fixes #2233.


ReactiveCocoa/Swift/CocoaAction.swift
```diff
@@ -77,15 +77,13 @@ public final class CocoaAction: NSObject {
...
-	/// it unsafe for use when the action is parameterized for something like `Void`
-	/// input. In those cases, explicitly assign a value to this property that transforms
-	/// the input to suit your needs.
+	/// A UI bindable `CocoaAction`. The default behavior force casts the
+	/// AnyObject? input to match the action's `Input` type. This makes it
+	/// unsafe for use when the action is parameterized for something like
+	/// `Void` input. In those cases, explicitly assign a value to this property
+	/// that transforms the input to suit your needs.
 	public var unsafeCocoaAction: CocoaAction {
-		return associatedObject(self, key: &unsafeCocoaActionKey, initial: { host in
```
ikesyo: `private var unsafeCocoaActionKey` should be removed as well.

ReactiveCocoaTests/Swift/ActionSpec.swift
```diff
@@ -176,12 +177,15 @@ class ActionSpec: QuickSpec {
 				let result = action.apply(0).first()
 				expect(result?.value) == 1
 				expect(values).toEventually(equal([ true, false, true ]))
+				
+				_ = cocoaAction
```
Palleas: Just curious, what is this assignation here for ?

ReactiveCocoa/Swift/CocoaAction.swift
mdiep: 👍

ReactiveCocoaTests/Swift/ActionSpec.swift
mdiep: That's to keep `cocoaAction` from deallocating before the end of the test—because when it deallocates, the producer will complete.

Palleas: Makes sense, thanks!

ikesyo:
👍

kaybutter:
This kind of change should not be hidden behind a simple "fixed retain cycle" in the release notes.

While this code would work fine in RAC 4.1, it compiles and fails silently in RAC 4.2:
``` swift
someButton.addTarget(someAction.unsafeCocoaAction, action: CocoaAction.selector)
```

I feel like this change kind of defeats the purpose of having `unsafeCocoaAction` in the first place. Is there a better pattern to be used in this instance?

mdiep:
Sorry for the inconvenience! Let's move this discussion to #2983.

I'll try to comment there soon, but it may to happen this week because of WWDC.

Issue #2841, opened 2016-04-29
PropertyType changes


andersio:
Just saw a few proposals for RAC 5.0, so I decided to bring this one up again. :)

If RAC 5.0 is targeting Swift 3.0 and API compatibility is not a concern, PropertyType and MutablePropertyType could use a few changes (that were rejected before due to breaking changes & impl. issue):

1. Introduce `PropertyType.withValue(_:)` and `MutablePropertyType.modify(_:)`.
2. Move the MutableProperty `swap(:_)` to `MutablePropertyType`.
3. Make `value` a default implementation of the protocols, depending on (1).

These require [SE-0049: Move @noescape and @autoclosure to be type attributes](https://github.com/apple/swift-evolution/blob/master/proposals/0049-noescape-autoclosure-type-attrs.md) to be implemented cleanly.

With these additions, new compositing/transforming operators like `map(_:)` can be introduced using `PropertyType.signal` gracefully, but need not worry about breaking conforming types with a thread safe guarantee, e.g. `MutableProperty` (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2788#issuecomment-214596381).

andersio:
Just played with the latest Swift 3 snapshot. I found that fully realising this proposal would need Swift 3 to have not only `@noescape` but `rethrows` allowed in function type annotations. Otherwise, compromises would have to be made, i.e. no `rethrows` on protocols.

PR #2911, opened 2016-05-19
Property composition and enhancements

andersio:
Here is an implementation of the #2841 proposal, seeking for your comments.
Swift 3.0 is not required.

#### Objective
This PR aims to introduce new composition operators on `PropertyType`.

On the other hand, `AnyMutableProperty` is introduced for users who would like to build higher level constructs that operate on the `MutablePropertyType` abstraction. An example would be two-way bindings between two `MutablePropertyType`s.

#### Implementation
This PR carries breaking changes to those who have custom implementations of the `PropertyType` protocols.

##### Type-erasing wrappers
1. `AnyProperty` now uses a private, type erasing box.
2. A new type erased wrapper `AnyMutableProperty`.
3. The producer and the signal of an `AnyProperty` now completes only when the source has completed.

##### Protocols
4. Introduce `withValue` and `modify` in the protocols.
5. `value` and `swap(newValue:)` are now protocol extension methods.

##### Property Composition
1. New combining operators: `combineLatest` and `zip`
2. New value operators: `combinePrevious`, `skipRepeats` and `uniqueValues`
3. New flattening operators: `flattenLatest` and `flatMapLatest`

##### Miscellaneous
1. Separated the `DynamicProperty` tests from the rest of the PropertySpec. 
2. Reorganised the `Property.swift`, with protocols & protocol extensions at the top, followed by implementations, bindings and finally the private utility classes.

##### Type erasing boxes
The current closure approach cannot be expressed non-escaping blocks and rethrowing functions in Swift 2.2 so far, and Swift 3 has tackled only one of them so far. Without these, `withValue` and `modify` cannot be brought to the protocols.

In the future (post Swift 3.0), these wrappers should be replaceable by typealiases of a generalised existential type without breaking any user code.

-
Edited on 21 May 2016

### DIFF
--- a/ReactiveCocoa/Swift/DynamicProperty.swift
+++ b/ReactiveCocoa/Swift/DynamicProperty.swift
@@ -54,7 +54,7 @@ import enum Result.NoError
 			}
 	}
 
-	/// Atomically modifies the variable.
+	/// Modifies the variable.
 	///
 	/// Returns the old value.
 	public func modify(@noescape action: (Value) throws -> Value) rethrows -> Value {
@@ -63,7 +63,7 @@ import enum Result.NoError
 		return oldValue
 	}
 
-	/// Atomically performs an arbitrary action using the current value of the
+	/// Performs an arbitrary action using the current value of the
 	/// variable.
 	///
 	/// Returns the result of the action.
--- a/ReactiveCocoa/Swift/DynamicProperty.swift
+++ b/ReactiveCocoa/Swift/DynamicProperty.swift
@@ -15,18 +15,6 @@ import enum Result.NoError
 
 	private var property: MutableProperty<AnyObject?>?
 
-	/// The current value of the property, as read and written using Key-Value
-	/// Coding.
-	public var value: AnyObject? {
-		@objc(rac_value) get {
-			return object?.valueForKeyPath(keyPath)
-		}
-
-		@objc(setRac_value:) set(newValue) {
-			object?.setValue(newValue, forKeyPath: keyPath)
-		}
-	}
-
 	/// A producer that will create a Key-Value Observer for the given object,
 	/// send its initial value then all changes over time, and then complete
 	/// when the observed object has deallocated.
@@ -64,6 +52,23 @@ import enum Result.NoError
 					self.property = nil
 				}
 			}
+	}
+
+	/// Atomically modifies the variable.
+	///
+	/// Returns the old value.
+	public func modify(@noescape action: (Value) throws -> Value) rethrows -> Value {
+		let oldValue = value
+		object?.setValue(try action(oldValue), forKey: keyPath)
+		return oldValue
+	}
+
+	/// Atomically performs an arbitrary action using the current value of the
+	/// variable.
+	///
+	/// Returns the result of the action.
+	public func withValue<Result>(@noescape action: (Value) throws -> Result) rethrows -> Result {
+		return try action(object?.valueForKeyPath(keyPath))
 	}
 }
 
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -40,32 +40,41 @@ extension PropertyType {
 	}
 }
 
+/// Workaround of the nil coalescing operator, as it does not work as
+/// intended with function types.
+///
+/// https://bugs.swift.org/browse/SR-832
+private func coalescing<T>(optional: T?, default value: T) -> T {
+	return optional ?? value
+}
+
 /// Protocol composition operators
 ///
 /// The producer and the signal of transformed properties would complete
 /// only when its source properties have deinitialized.
+///
+/// A transformed property would retain its ultimate source, but not
+/// any intermediate property during the composition.
 extension PropertyType {
 	/// Lifts an unary SignalProducer operator to operate upon PropertyType instead.
 	@warn_unused_result(message="Did you forget to use the composed property?")
 	private func lift<U>(@noescape transform: SignalProducer<Value, NoError> -> SignalProducer<U, NoError>) -> AnyProperty<U> {
+		let capturingClosure = coalescing((self as? AnyProperty<Value>)?.capturingClosure, default: { self })
+
 		return AnyProperty(propertyProducer: transform(producer),
-		                   with: ActionDisposable { self })
+		                   capturing: capturingClosure)
 	}
 
 	/// Lifts a binary SignalProducer operator to operate upon PropertyType instead.
 	@warn_unused_result(message="Did you forget to use the composed property?")
 	private func lift<P: PropertyType, U>(transform: SignalProducer<Value, NoError> -> SignalProducer<P.Value, NoError> -> SignalProducer<U, NoError>) -> P -> AnyProperty<U> {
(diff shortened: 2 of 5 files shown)


ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -65,6 +53,23 @@ import enum Result.NoError
...
 	}
+
+	/// Atomically modifies the variable.
+	///
+	/// Returns the old value.
+	public func modify(@noescape action: (Value) throws -> Value) rethrows -> Value {
+		let oldValue = value
+		object?.setValue(try action(oldValue), forKey: keyPath)
+		return oldValue
+	}
```
mdiep: This isn't atomic.

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -65,6 +53,23 @@ import enum Result.NoError
...
+
+	/// Atomically performs an arbitrary action using the current value of the
+	/// variable.
+	///
+	/// Returns the result of the action.
+	public func withValue<Result>(@noescape action: (Value) throws -> Result) rethrows -> Result {
+		return try action(object?.valueForKeyPath(keyPath))
+	}
```
mdiep: Neither is this.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -5,69 +5,182 @@ import enum Result.NoError
...
 	var producer: SignalProducer<Value, NoError> { get }
 
 	/// A signal that will send the property's changes over time.
 	var signal: Signal<Value, NoError> { get }
+
+	/// Performs an arbitrary action using the current value of the
+	/// variable. Conforming types may optionally guarantee atomicity.
+	///
+	/// Returns the result of the action.
+	func withValue<Result>(@noescape action: Value throws -> Result) rethrows -> Result
```
mdiep: I don't see the value of `withValue`. This could just as easily be a readonly property.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -5,69 +5,182 @@ import enum Result.NoError
...
+
+	/// Combines the current value and the subsequent values of two `Property`s in
+	/// the manner described by `Signal.combineLatestWith:`, and applys the supplied
+	/// transform on the combined pair.
+	///
+	/// A combined property would retain its source properties.
+	@warn_unused_result(message="Did you forget to use the composed property?")
+	public func combineLatest<P: PropertyType, U>(with other: P, combining transform: (Value, P.Value) -> U) -> AnyProperty<U> {
+		return withValue { currentValue in
+			return other.withValue { currentOtherValue in
```
mdiep: Nested locks (if these were actually atomic) are often a recipe for trouble.

mdiep:
It seems to me that the best answer to property composition is to add an internal property initializer that takes only a `SignalProducer`. Since a property's signal producer sends the initial value upon subscription, RAC could block and wait for the first value.

This isn't safe to do generally with signal producers, which is why such an initializer wouldn't be `public`, but I don't see why it couldn't be done internally. That should provide an easy way to add composition to properties.

ReactiveCocoa/Swift/Property.swift
andersio: Hmm, I now see that this is risky. If one composes `propertyA` with `propertyB`, and composes another in the reverse order at the same time, it has a chance to deadlock.

andersio:
@mdiep I agree that it would be a simpler approach for standard property compositing operators. Just that custom operators outside the RAC module could not take advantage of it.

ReactiveCocoa/Swift/Property.swift
andersio: `MutableProperty` has `withValue` (and `modify`) so that the supplied action block is executed within the critical section. But the `AnyProperty` wrapper did not expose it.

neilpa:
> Though I see no harm in opening it up as long as the caveat is explicitly documented

I disagree. For a framework of this scale there's plenty of people that don't read documentation and would misuse it. Better to keep it internal.

andersio:
I just pushed a commit with four new operators: `combinePrevious`, `uniqueValues`, `flattenLatest` and `flatMapLatest`.

As for why only exposing the switch-to-latest strategy, I cannot think of any particular use case that the merging would be needed on properties, while the concatenating doesn't make sense at all with the lifetime characteristics of properties.

andersio:
@neilpa Making a failable/throwable initializer can force error handling at call site. But the user could still do `try!` or force unwrap without reasoning about it...

andersio:
After reviewing my projects and trying to adopt the flattening property operator, it actually creates more boilerplate since the property has to be retained. I wonder if it should be taken away.

Otherwise, the rest of the PR should be fine - it is pretty useful if one's model layer could expose the attributes as a `PropertyType` conforming proxies/types.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -5,69 +5,284 @@ import enum Result.NoError
...
+	public func skipRepeats() -> AnyProperty<Value> {
+		return lift { $0.skipRepeats() }
+	}
+}
+
+extension PropertyType where Value: PropertyType {
+	/// Returns a property that forwards values from the latest property hold by
+	/// `self`, ignoring values sent on previous inner properties.
+	@warn_unused_result(message="Did you forget to use the composed property?")
+	public func flattenLatest() -> AnyProperty<Value.Value> {
```
mdiep: This should just be called `flatten`.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -5,69 +5,284 @@ import enum Result.NoError
...
+		return lift { $0.flatMap(.Latest) { $0.producer } }
+	}
+}
+
+extension PropertyType {
+	/// Maps a property to a new property, and then flattens it in the manner
+	/// described by `flattenLatest`.
+	@warn_unused_result(message="Did you forget to use the composed property?")
+	public func flatMapLatest<P: PropertyType>(transform: Value -> P) -> AnyProperty<P.Value> {
```
mdiep: And this should be `flatMap`.

ReactiveCocoa/Swift/Property.swift
mdiep: But none of the property implement locking, so this is unneeded.

mdiep:
> Note that it is not a bug: the producer of a property are said to complete when the property has reinitialised, and the <~ bindings do not retain the source property.

The lifetime of the mapped property, or at least its producer, should be tied to the root property. `property.flatMap { … }.producer` should complete when `property` completes.

andersio:
@mdiep Are we treating `AnyProperty` as a lense to the original property, or a standard `PropertyType` implementation? I made it this way to be consistent with what's written in the documentation, i.e. `producer` completes when the property has deinitialised.

andersio:
@mdiep I have made changes in response to your comments. `property.flatMap { ... }.producer` should be an equivalent of `property.producer.flatMap { ... }` now, and the deallocation of the intermediate properties should have no effect on such producer.

P.S. I rebased the PR onto the RAC5 branch, please visit #2922. 😄

PR #2912, opened 2016-05-19
[UIScheduler] Don't rely on `NSThread.isMainThread()`

### LINKED ISSUE #2635: UIScheduler not working as expected
With RAC4 rc1 `.observeOn(UIScheduler())` doesn't work as expected for me (I didn't test older versions though):

```swift
	let currentRoute = MutableProperty<MKRoute?>(nil)

    func observe() {
		self.currentRoute.producer.observeOn(UIScheduler()).startWithNext { route in
// Works:
//			dispatch_async(dispatch_get_main_queue(), {
//				if let route = route {
//					self.mapView.addOverlay(route.polyline, level: .AboveRoads)
//				}
//			})

// Crashes in -[VKRasterOverlayTileSource init] with EXC_BAD_ACCESS:
			if let route = route {
				self.mapView.addOverlay(route.polyline, level: .AboveRoads)
			}
		}
     }
```

`currentRoute` is set from the callback in `MKDirections calculateDirectionsWithCompletionHandler`. When I wrap the `addOverlay` call into a `dispatch_async(dispatch_get_main_queue()`, as shown in the commented version, it works as expected and doesn't crash.

Any hints on what I'm doing wrong?

ikesyo:
Use `dispatch_queue_set_specific` for the main queue and `dispatch_get_specific` instead.

Should fix #2635.

### DIFF
--- a/ReactiveCocoa/Swift/Scheduler.swift
+++ b/ReactiveCocoa/Swift/Scheduler.swift
@@ -52,15 +52,28 @@ public final class ImmediateScheduler: SchedulerType {
 	}
 }
 
-/// A scheduler that performs all work on the main thread, as soon as possible.
+/// A scheduler that performs all work on the main queue, as soon as possible.
 ///
-/// If the caller is already running on the main thread when an action is
+/// If the caller is already running on the main queue when an action is
 /// scheduled, it may be run synchronously. However, ordering between actions
 /// will always be preserved.
 public final class UIScheduler: SchedulerType {
+	private static var dispatchOnceToken: dispatch_once_t = 0
+	private static var dispatchSpecificKey: UInt8 = 0
+	private static var dispatchSpecificContext: UInt8 = 0
+
 	private var queueLength: Int32 = 0
 
-	public init() {}
+	public init() {
+		dispatch_once(&UIScheduler.dispatchOnceToken) {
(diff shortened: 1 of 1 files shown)


mdiep:
Very interesting! I hope this works out. ⚡ ⚡ ⚡

ikesyo:
I definitely confirmed that the fix works with the example noted in https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2635#issuecomment-167539538.

> http://www.komendera.com/bugs/rac/UISchedulerTest.zip

To correctly confirm the example, the `source` and `destination` should be visible in a map view.

```swift
// Change latitudes and longitudes if needed
		request.source = MKMapItem(placemark: MKPlacemark(coordinate: CLLocationCoordinate2D(latitude: 48.211151, longitude: 16.365379), addressDictionary: nil))
		request.destination = MKMapItem(placemark: MKPlacemark(coordinate: CLLocationCoordinate2D(latitude: 48.199167, longitude: 16.351016), addressDictionary: nil))
```

mdiep:
> I definitely confirmed that the fix works with the example noted in #2635 (comment).

Thanks for doing that. That was on my list, but I hadn't gotten around to it yet.

This fix looks great to me! 🚢

PR #2915, opened 2016-05-20
Action tweaks

ikesyo:
Additionally, I'm not sure what is the point of `executingQueue`. The targeted `_executing` is a `MutableProperty` which is thread-safe, and the access to the variable in `disposable.addDisposable { ... }` is not surrounded by `dispatch_sync(self.executingQueue) { ... }`. Can we remove it or should we add `dispatch_sync` in the disposable action?

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -69,7 +69,7 @@ public final class Action<Input, Output, Error: ErrorType> {
 		errors = events.map { $0.error }.ignoreNil()
 
 		_enabled <~ enabledIf.producer
-			.combineLatestWith(executing.producer)
+			.combineLatestWith(_executing.producer)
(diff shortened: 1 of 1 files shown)


mdiep:
> Additionally, I'm not sure what is the point of executingQueue. The targeted _executing is a MutableProperty which is thread-safe, and the access to the variable in disposable.addDisposable { ... } is not surrounded by dispatch_sync(self.executingQueue) { ... }. Can we remove it or should we add dispatch_sync in the disposable action?

That documentation on `executingQueue` is very poor. It's providing atomicity to `apply`. Without the `dispatch_sync`, there's a race condition: two threads can both check to see that the action is enabled, both succeed, and both proceed to execute—even though only one should be able to. Since `enabled` is based on `executing`, an atomic read-then-write is needed.

ikesyo:
I got it. Thank you @mdiep! 😄

PR #2917, opened 2016-05-21
Optimize Signal.zipWith

ikesyo:


ikesyo:
Really the test suites are failing: 

> Executed 1435 tests, with 4 failures (0 unexpected) in 21.741 (28.363) seconds

but the CI reported the build as successful:

> The command "script/build" exited with 0.

ikesyo:
It's ready. ⚡

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1041,22 +1041,24 @@ extension SignalType {
 			let disposable = CompositeDisposable()
 			
 			let flush = {
-				var (leftOriginal, rightOriginal) = states.modify { states in
+				var tuple: (Value, U)?
```
mdiep: If we added a `var finished: Bool = false` here and left `isFinished` like it was before, it could be set inside the `modify`. That would simplify things a bit.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1026,8 +1026,8 @@ private struct ZipState<Value> {
 	var values: [Value] = []
 	var completed = false
```
mdiep: Maybe this should have be:

```swift
private struct ZipState<Left, Right> {
    var values: (left: [Left], right: [Right) = ([], [])
    var completed: (left: Bool, right: Bool) = (false, false)

    var isFinished: Bool {
        return values.left.isEmpty && values.left.completed || values.right.isEmpty && values.right.completed
    }
}
```

(It's too bad that Swift doesn't have a builtin `Pair` type!)

ikesyo:
@mdiep Thank you for the feedback! Updated. ⚡

mdiep:
Thanks!

PR #2922, opened 2016-05-22
Property composition operators.

andersio:
A subset of #2911.

-
##### Property Composition
1. New combining operators: `combineLatest` and `zip`. Also their top-level binary to 10-ary shorthands.
2. New value operators: `combinePrevious`, `skipRepeats` and `uniqueValues`
3. New flattening operators: `flatten` and `flatMap`.

The signals and producers of `AnyProperty`s respect the lifetime of its ultimate source(s). In other words:
```
someIntProperty.map { $0 + 1 }.map { $0 + 2 }.map { $0 + 3 }.producer
```
would be alive as long as `someIntProperty` is still alive, even if all the intermediate `AnyProperty`s are deinitialised.

##### Breaking change
The producer and the signal of the resulting property of `map` has now a different lifetime semantics. That's said `map` is not yet included in a release.

##### Miscellaneous
1. Reorganised the `Property.swift`, with protocols & protocol extensions at the top, followed by implementations, bindings and finally the private utility classes.


andersio:
This branch includes a change to the lifetime of the producers and signals of transformed and non passing through properties. They now respect the lifetime of the ultimate source, instead of the internal relaying property. Moreover, it allows intermediate properties in a chained composition to be deallocated, instead of holding a series of unreachable `MutableProperty`s to propagate the changes.

It is likely a better solution to the unsafe operator chaining issue brought up by @filblue (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2788#issuecomment-208274061).

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,486 @@ public protocol PropertyType {
...
+	/// A signal that will send the property's changes over time. It
+	/// completes when the property has deinitialized, or has no further
+	/// change.
 	var signal: Signal<Value, NoError> { get }
+
+	/// Performs an arbitrary action using the current value of the
+	/// variable. Conforming types may optionally guarantee atomicity.
+	///
+	/// Returns the result of the action.
+	func withValue<Result>(@noescape action: Value throws -> Result) rethrows -> Result
```
mdiep: I don't think this is necessary. None of the property types guarantee atomicity with this PR.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,486 @@ public protocol PropertyType {
...
+/// weak references (and value types currently do not).
+public protocol MutablePropertyType: class, PropertyType {
+	/// The current value of the property.
+	var value: Value { get set }
+
+	/// Modifies the variable. Conforming types may optionally
+	/// guarantee atomicity.
+	///
+	/// Returns the old value.
+	func modify(@noescape action: Value throws -> Value) rethrows -> Value
```
mdiep: Some goes for this.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,486 @@ public protocol PropertyType {
...
-		let mappedProducer = SignalProducer<U, NoError> { observer, disposable in
-			disposable += ActionDisposable { self }
-			disposable += self.producer.map(transform).start(observer)
+	/// Initializes a property from a producer that promises to send at least one
+	/// value synchronously in its start handler before sending any subsequent event.
+	/// If the producer fails its promise, a fatal error would be raised.
+	///
+	/// The producer and the signal of the created property would complete only
+	/// when the `propertyProducer` completes.
+	private init(propertyProducer: SignalProducer<Value, NoError>, capturing closure: (() -> Void)? = nil) {
```
mdiep: Looking at this, I think it would be simpler to create 2 initializers specifically for transforming properties:

```
private init<P: PropertyType>(property: P, transform: (SignalProducer<P.Value, NoError> -> SignalProducer<Value, NoError>)

private init<P1: PropertyType, P2: PropertyType>(property1: P1, transform: (SignalProducer<P1.Value, NoError> -> SignalProducer<P2.Value, NoError> -> SignalProducer<Value, NoError>), property2: P2)
```

(I'm sure I don't have those types quite right.)

That way the implementation details about captures, etc. don't leak out into the operators.

mdiep:
> AnyProperty now uses a private, type erasing box.

What's the motivation for this?

> A new type erased wrapper AnyMutableProperty.

What's the motivation for this? Can we break it out into a separate PR?

> It now targets the RAC5 branch since it contains breaking changes

What are the breaking changes? The changes to property lifetimes?

ReactiveCocoa/Swift/Property.swift
andersio: `MutableProperty` does.

andersio:
>> AnyProperty now uses a private, type erasing box.
>
> What's the motivation for this?
To allow `AnyProperty` to be able to pass through `withValue` of its wrapping property. This is impossible in Swift 2.x due to the limitation of `@noescape`, while in Swift 3 it still lacks the ability to express a `rethrows` function type.

>> A new type erased wrapper AnyMutableProperty.
>
>What's the motivation for this? Can we break it out into a separate PR?
Sure. For the motivation, as written in the preceding PR:

AnyMutableProperty is introduced for users who would like to build higher level constructs that operate on the MutablePropertyType abstraction. An example would be two-way bindings between two MutablePropertyTypes.

>> It now targets the RAC5 branch since it contains breaking changes
> What are the breaking changes? The changes to property lifetimes?
1. New protocol requirements of `PropertyType` and `MutablePropertyType`.
2. The producer/signal lifetime of AnyProperty created using `init(initialValue:signal:)` or `init(initialValue:producer:)`.

ReactiveCocoa/Swift/Property.swift
andersio: Yeah, it would move the details from `lift` to `AnyProperty`. But the initialiser you commented on would still be there, as `init(initialValue:signal:)` and `init(initialValue:producer:)` now depends on it.

andersio:
I will try to split the property composition operators & the changes to the protocols into 2 PRs.

andersio:
@mdiep I wonder if the team might be interested in pushing this into a Swift 2.2 release? If yes, maybe I can also take the breaking change out as a separated PR

andersio:
I just realised the `AnyProperty(propertyProducer:capturing:)` depends on the order of calling observers, which is in turn the implementation detail of *unordered* `Bag`.

I have added a test case to cover this, but I just wonder if a step further should be taken to remove such dependency.

mdiep:
> I have reverted the changes in the protocols, and would perhaps push them in a future PR. I am sorry for my attempt to pack everything in the one PR.
> 
> This PR now focuses on the property composition operators.

That's great—thank you! Smaller PRs are much easier to review. I'm happy to consider those other changes in a separate PR; I was trying to figure out how they related to the property composition operators.

> I wonder if the team might be interested in pushing these addictive operators into a Swift 2.2 release? If yes, I can take the breaking change out as a separated PR.

Yes, I'd be interested in that. It seems like the breaking changes were unrelated to property composition.

andersio:
@mdiep If the PR must assume the observers of any signals are called in random order, my quick experiment results in boilerplate to enforce the ordering* on `AnyProperty`'s end.

Would it be appropriate to instead leave a note in `Bag` or `Signal(_:)` to state that the insertion order must be maintained? Or is that a test case that would fail with.


* Just for the record, I was talking about the assertion that:
```
let mappedProperty = someProperty.map { $0 + 1 }
mappedProperty.producer.startWithNext { newValue in
    assert(newValue == mappedProperty.value) // Semantically, the event is sent after the change.
}
```
which is covered by [this test case](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2922/commits/e3215a53fcfe9c1a794d987ebacde34283973929). While there is no written requirement so far, it is the behaviour of the two mutable property types.

I can verify that the assertion would fail if I change a line in `insert(value:)` of `Bag`:
```
		// elements.append(element)
		elements.insert(element, atIndex: 0)
```

andersio:
I've updated the implementation to enforce the post-change order of signal events. i.e.:
```
mappedProperty.producer.startWithNext { value in
}

**Comment by andersio** [comment]
Please note that the implementation assumes observers of a signal are called in the insertion order. This is the current behaviour of `Bag`.

A test case on this assumption (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2922/commits/e3215a53fcfe9c1a794d987ebacde34283973929) has been added, in case there is any change to `Bag` in the future.

File: ReactiveCocoa/Swift/Property.swift [hunk 4]
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
+		let initial = first.map { [$0] }
+		return GeneratorSequence(generator).reduce(initial) { property, next in
+			property.zip(with: next).map { $0.0 + [$0.1] }
+		}
+	}
+
+	return nil
+}
+
+/// A read-only, type-erased view of a property.
```
mdiep: I don't think I'd describe this as type-erased.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
 
-extension PropertyType {
-	/// Maps the current value and all subsequent values to a new value.
-	public func map<U>(transform: Value -> U) -> AnyProperty<U> {
-		let mappedProducer = SignalProducer<U, NoError> { observer, disposable in
-			disposable += ActionDisposable { self }
-			disposable += self.producer.map(transform).start(observer)
+	/// Initializes a property by applying the unary `SignalProducer` transform on
+	/// `property`. The resulting property captures `property`.
+	private init<P: PropertyType>(transforming property: P, @noescape using transform: SignalProducer<P.Value, NoError> -> SignalProducer<Value, NoError>) {
```
mdiep: This is kinda minor, but I'd name this to be similar to `public init<P: PropertyType where P.Value == Value>(_ property: P)`:

```swift
private init<P: PropertyType>(_ property: P, @noescape transform: SignalProducer<P.Value, NoError> -> SignalProducer<Value, NoError>) {
```

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
+
+/// Protocol composition operators
+///
+/// The producer and the signal of transformed properties would complete
+/// only when its source properties have deinitialized.
+///
+/// A transformed property would retain its ultimate source, but not
+/// any intermediate property during the composition.
+extension PropertyType {
+	/// Lifts an unary SignalProducer operator to operate upon PropertyType instead.
```
mdiep: This should be `a unary`. (English!)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
+	/// `property`. The resulting property captures `property`.
+	private init<P: PropertyType>(transforming property: P, @noescape using transform: SignalProducer<P.Value, NoError> -> SignalProducer<Value, NoError>) {
+		self.init(propertyProducer: transform(property.producer),
+		          capturing: AnyProperty.capture(property))
+	}
+
+	/// Initializes a property by applying the binary `SignalProducer` transform on
+	/// `property` and `anotherProperty`. The resulting property captures `property`
+	/// and `anotherProperty`.
+	private init<P1: PropertyType, P2: PropertyType>(transformingFirst property: P1, second anotherProperty: P2, @noescape using transform: SignalProducer<P1.Value, NoError> -> SignalProducer<P2.Value, NoError> -> SignalProducer<Value, NoError>) {
```
mdiep: Same here. I'd remove the names from the property arguments and remove the external name for `transform`.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
+
+public enum PropertyFlattenStrategy {
+	case Latest
+}
+
+extension PropertyType where Value: PropertyType {
+	/// Returns a property that forwards values from the latest property hold by
+	/// `self`, ignoring values sent on previous inner properties.
+	@warn_unused_result(message="Did you forget to use the composed property?")
+	public func flatten(strategy: PropertyFlattenStrategy) -> AnyProperty<Value.Value> {
```
mdiep: This doesn't need to have a strategy argument. Neither does `flatMap`. Just making it `flatten()` is enough.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
+	/// Maps the current value and all subsequent values to a new property.
+	@warn_unused_result(message="Did you forget to use the composed property?")
+	public func map<U>(transform: Value -> U) -> AnyProperty<U> {
+		return lift { $0.map(transform) }
+	}
+
+	/// Combines the current value and the subsequent values of two `Property`s in
+	/// the manner described by `Signal.combineLatestWith:`.
+	@warn_unused_result(message="Did you forget to use the composed property?")
+	public func combineLatest<P: PropertyType>(with other: P) -> AnyProperty<(Value, P.Value)> {
```
mdiep: We should call this `combineLatestWith` for now for consistency. It can be renamed in RAC5 when we Swift3-ify all our functions.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
+		return lift(SignalProducer.combineLatestWith)(other)
+	}
+
+	/// Zips the current value and the subsequent values of two `Property`s in
+	/// the manner described by `Signal.zipWith`.
+	@warn_unused_result(message="Did you forget to use the composed property?")
+	public func zip<P: PropertyType>(with other: P) -> AnyProperty<(Value, P.Value)> {
```
mdiep: Same is true here.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -8,66 +8,457 @@ public protocol PropertyType {
...
+			property.zip(with: next).map { $0.0 + [$0.1] }
+		}
+	}
+
+	return nil
+}
+
+/// A read-only, type-erased view of a property.
 public struct AnyProperty<Value>: PropertyType {
+	private let capturingClosure: (() -> Void)?
```
mdiep: It looks to me like the could be done with an array instead:

```swift
private let sources: [AnyObject]
```

mdiep:
Thank you for your continued work on this!

ReactiveCocoa/Swift/Property.swift
andersio: It erased the type of what actually `PropertyType` is.

https://github.com/apple/swift/blob/master/stdlib/public/core/ExistentialCollection.swift.gyb

ReactiveCocoa/Swift/Property.swift
andersio: *cough*

ReactiveCocoa/Swift/Property.swift
andersio: Should the merge strategy be added? Concatenation doesn't make sense, but merge should be fine. Otherwise I would remove it.

ReactiveCocoa/Swift/Property.swift
andersio: But it looks better at the call site. Instead of `AnyProperty(property, transform: transform)`, we have  `AnyProperty(property, using: transform)` that IMO makes more sense.

ReactiveCocoa/Swift/Property.swift
mdiep: That's a good question. I'm going to think about this for a bit.

ReactiveCocoa/Swift/Property.swift
mdiep: `transform:` is more inline with RAC and with the Swift sodlib.

ReactiveCocoa/Swift/Property.swift
mdiep: I think `Concat` would actually make sense—as much sense as doing any other flatten operation with `PropertyType`s. Let's either use the normal `FlattenStrategy` or remove `flatten` altogether.

andersio: @mdiep `Concat` does not make sense as a property because transformed properties would retain their source, and consequently the producer can never move past the first inner property, unless the transformed property has deinitialised.

It can still be included for extracting the producer after property composition, though.

mdiep: Let's just remove `flatten` for now. I'm not sure how useful it will be, and it will be easy to add later.

Getting the other operators out there will let people use them and make the needs for `flatten` more obvious.

andersio: TBH the most important bit of this PR for me is `flatMap(.Latest)`...

andersio: I consider `flatMap(.Latest)` pretty useful in the case of using a property to hold the view model in a reusable view. It allows you to compose at the property level, and drop down to `producer` only at the end. For sure it is just a sugar, but so as the rest of the property composition operators.

1. `viewModel.producer.flatMap(.Latest) { $0.name.producer }`
2. `viewModel.flatMap(.Latest) { $0.name }.producer`

andersio:
@mdiep Hmm... it seems I am wrong about `Concat`. It is behaving normally in my WIP test cases.

mdiep:
Looks good to me. Thanks for all your work on this!

PR #2925, opened 2016-05-22
add merge operator for SignalProducer

larryonoff:
It often usable to use merge operator for SignalProducer. `Signal` already has merge operator.


ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -552,6 +552,22 @@ extension SignalType {
...
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public static func merge<S : SequenceType where S.Generator.Element == SignalProducer<Value, Error>>(producers: S) -> SignalProducer<Value, Error> {
+		return SignalProducer(values: producers).flatten(.Merge)
+	}
+	
+	/// Merges the given producers into a single `SignalProducer` that will emit all values
+	/// from each of them, and complete when all of them have completed.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public static func merge<Value, Error: ErrorType>(producers: SignalProducer<Value, Error>...) -> SignalProducer<Value, Error> {
+		return SignalProducer(values: producers).flatten(.Merge)
```
mdiep: This could just call the above variant and pass `producers`.

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -552,6 +552,22 @@ extension SignalType {
...
+	/// from each of them, and complete when all of them have completed.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public static func merge<S : SequenceType where S.Generator.Element == SignalProducer<Value, Error>>(producers: S) -> SignalProducer<Value, Error> {
+		return SignalProducer(values: producers).flatten(.Merge)
+	}
+	
+	/// Merges the given producers into a single `SignalProducer` that will emit all values
+	/// from each of them, and complete when all of them have completed.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public static func merge<Value, Error: ErrorType>(producers: SignalProducer<Value, Error>...) -> SignalProducer<Value, Error> {
```
mdiep: Can you add a `...` variant like this to `Signal` as well if we're going to try to even out the interfaces?

mdiep:
Thanks for the PR! I agree this is a handy thing to have.

larryonoff:
Thank you for the comments! I've updated this PR according to your comments.

larryonoff:
What do you think about the case when `combineLatest` and `zip` operators are free functions, when `merge` is not.

Maybe it's better to move `merge` operator from protocols?

ikesyo:
> What do you think about the case when `combineLatest` and `zip` operators are free functions, when `merge` is not.
> 
> Maybe it's better to move `merge` operator from protocols?

@larryonoff That should be discussed in #2926.

larryonoff:
@ikesyo thanks!

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1893,6 +1893,71 @@ class SignalProducerSpec: QuickSpec {
 				expect(started) == false
 			}
 		}
+		
+		describe("SignalProducer.merge") {
```
ikesyo: I think this (and `describe("Signal.merge")` in `SignalSpec.swift`) should be defined in `FlattenSpec.swift`.

larryonoff: Will move it.

larryonoff: Done

larryonoff:
`Signal.merge()` and `SignalProducer.merge()` moved into FlattenSpec.

mdiep:
This is great! Thanks for the PR. ✨

PR #2928, opened 2016-05-23
add SignalProducer variadic init with values

larryonoff:


RuiAAPeres:
👍

ikesyo:
We have found that this is a bit problematic for type inference. See https://github.com/Carthage/Carthage/pull/1329/files#r65486309 for the details.

larryonoff:
@ikesyo I spent some time on analysis of possible solutions, but haven't find any solutions except reverting this PR and making v4.2.1 until it isn't too late.

PR #2932, opened 2016-05-24
Utilize Event.value getter

ikesyo:
Ignoring whitespace is more readable: [diff](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2932/files?w=1)


RuiAAPeres:
I like the idea of using the event's value. 👍 The only issue I have with `guard`, in this context, is that you need to add a bunch of early returns in the code. The flow becomes slightly more difficult to read, than if it was with an if/else approach.

RuiAAPeres:
Forget about my early comment, misread the diff. 👍

PR #2935, opened 2016-05-24
avoid signal retain in `bind`

tzongw:
`bind` will retain signals when subscribe, this is not necessary and is leak-prone. The code below will fail

```objc
    __weak RACSignal *s1;
    @autoreleasepool {
        RACSignal *s2 = [[RACSignal return:@123] delay:1];
        [[s2 map:^id(id value) {
            return @456;
        }] subscribeNext:^(id x) {
            NSLog(@"%@", x);
        }];
        s1 = s2;
    }
    NSAssert(s1 == nil, @"should be nil");
```

PR fix this issue and makes rac_textSignal of UITextField/UITextView simpler.

### DIFF
--- a/ReactiveCocoa/Objective-C/RACSignal.m
+++ b/ReactiveCocoa/Objective-C/RACSignal.m
@@ -21,6 +21,7 @@
 #import "RACSubject.h"
 #import "RACSubscriber+Private.h"
 #import "RACTuple.h"
+#import <libkern/OSAtomic.h>
 
 @implementation RACSignal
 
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Objective-C/RACSignal.m
```diff
@@ -107,31 +107,36 @@ - (RACSignal *)bind:(RACStreamBindBlock (^)(void))block {
 	return [[RACSignal createSignal:^(id<RACSubscriber> subscriber) {
 		RACStreamBindBlock bindingBlock = block();
 
-		NSMutableArray *signals = [NSMutableArray arrayWithObject:self];
+		__block NSUInteger signalsCount = 1;   // indicates self
```
StatusReport: Since `-bind:` is super popular, it will probably be more efficient to use atomic operations instead of a lock:
```
__block volatile int32_t signalsCount = 1;

// Then call:
OSAtomicIncrement32Barrier(&signalsCount);
OSAtomicDecrement32Barrier(&signalsCount);
```

tzongw: I think there will be a race condition between these two lines:

`OSAtomicDecrement32Barrier(&signalsCount);

if (signalsCount == 0) {`

We need an atomic 'decrement and test' operation.

StatusReport: Actually, this specific issue is solvable by checking the return value of `OSAtomicDecrement32Barrier`. However, it seems that if `signalsCount` reaches zero you must `dispose` the disposable and send completion to the subscriber, so you must put this entire logic in a critical section, as you've already did.

tzongw: Didn't notice OSAtomicDecrement32Barrier returning the new value. I think it really works! When `signalsCount ` reaches zero, only the last signal holds the subscriber and disposable, so needn't to worry about atomicity.

StatusReport: From shortly skimming the `-bind:` code, it indeed seems that it complies with its current implementation.

ReactiveCocoa/Objective-C/UITextView+RACSignalSupport.m
```diff
@@ -35,12 +35,7 @@ - (RACDelegateProxy *)rac_delegateProxy {
 }
 
 - (RACSignal *)rac_textSignal {
-	@weakify(self);
-	RACSignal *signal = [[[[[RACSignal
-		defer:^{
-			@strongify(self);
-			return [RACSignal return:RACTuplePack(self)];
-		}]
```
mdiep: I think this was done to make the returned `RACSignal` not retain `self`. I think this should be reverted.

ReactiveCocoa/Objective-C/RACSignal.m
mdiep: It looks to me like `OSAtomicDecrement32Barrier` would be safe. Once `signalsCount` is `0`, it will never be incremented again. So `sendCompleted` does not need to be called while holding the lock.

mdiep:
Thanks for the PR!

In addition to the notes above, would you be willing to add a test case that verifies this behavior?

ReactiveCocoa/Objective-C/UITextView+RACSignalSupport.m
tzongw: Since `bind`(also `map`) not retaining signals anymore, `weakify/strongify ` is not necessary in `rac_textSignal`.

mdiep: But `return:` retains its value. This changes the memory semantics of `rac_textSignal`.

tzongw:
test case added

mdiep:
Looks great! Thanks! ✨

PR #2939, opened 2016-05-25
Debounce

natestedman:
This is a potential implementation of https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2075. It's derived from a [previous implementation](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2075#issuecomment-148425140) that did not handle terminating events correctly. I use that version on a long-lived property, so terminating events hadn't come up yet.

This matches the behavior of `RACSignal` and [Rx](http://rxmarbles.com/#debounce), but it doesn't appear to match the way that the term `debounce` is commonly used in JavaScript frameworks. The goal of #2075 is to revive the `RACSignal` throttle behavior, and the term `debounce` came up later, so if it is inappropriate, it could definitely be renamed.

My use case for this operator is to throttle expensive writes of configuration changes that a user can alter rapidly, until the user stops taking actions and the configuration value reaches a resting point.

In the process, I also added [an additional check](https://github.com/ReactiveCocoa/ReactiveCocoa/commit/44ca545d30476ba1948e2cc34630531d40f31f57) to the `throttle` test, to match what I think (from the documentation of `throttle`) is the expected behavior - terminating events should be scheduled to run immediately, whereas only checking `TestScheduler`'s `run` just tests that they run eventually, without allowing any additional values to be sent. I think that this more strictly validates the behavior of `throttle` and `debounce`.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1215,6 +1215,24 @@ extension SignalType {
 			return disposable
 		}
 	}
+
+	/// Debounce values sent by the receiver, such that at least `interval`
+	/// seconds pass after the receiver has last sent a value, then
+	/// forwards the latest value on the given scheduler.
+	///
+	/// If multiple values are received before the interval has elapsed, the
+	/// latest value is the one that will be passed on.
+	///
+	/// If the input signal terminates while a value is being debounced, that value
+	/// will be discarded and the returned signal will terminate immediately.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func debounce(interval: NSTimeInterval, onScheduler scheduler: DateSchedulerType) -> Signal<Value, Error> {
+		return materialize().flatMap(.Latest, transform: { event in
+			event.isTerminating
+				? SignalProducer(value: event).observeOn(scheduler)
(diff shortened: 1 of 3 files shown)


RuiAAPeres:
@natestedman thanks for the PR. Have you looked at Rex's [`debouce`](https://github.com/neilpa/Rex/blob/master/Source/Signal.swift#L88#L94) implementation? (which we will merge eventually here #2790)

mdiep:
Thanks for the PR! ✨ 

The implementation in Rex that @RuiAAPeres linked to does seem superior. `delay` forwards failures and interrupts immediately, so you don't need to `materialize`.

If you'd like to update the PR to match that implementation, I'd be very happy to merge it! Then we'd have debounce in the upcoming `4.3` release.

mdiep:
@natestedman Are you interested in updating thiS?

mdiep:
Superseded by #2952.

natestedman:
A similar implementation was [what I started with](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2075#issuecomment-148425140) (and I've been using it for a while), but the behavior of terminating events didn't match `throttle` - I use it on a long-lived producer derived from a property, so that issue hadn't come up yet.

PR #2941, opened 2016-05-26
Add `prefix(value:)`

inamiy:
As discussed in Slack channel recently, there seem to have many reinvention of `beginWith` (old `- (instancetype)startWith:(id)value`), so I think it's good time to add one for Swift code.

Example of `beginWith` are:

- [JavaScript example](https://gist.github.com/staltz/868e7e9bc2a7b8c1f754)
- [Swift example](https://github.com/inamiy/ReactiveCocoaCatalog/blob/20a01590f4bfb056e09f8a73dffb16d1c0a5e0b3/ReactiveCocoaCatalog/Samples/WhoToFollowViewController.swift)

---

**2016/05/31 Updated:** Renamed `beginWith()` to `prefix(value:)`

### DIFF
--- a/ReactiveCocoa/Swift/Flatten.swift
+++ b/ReactiveCocoa/Swift/Flatten.swift
@@ -385,6 +385,12 @@ extension SignalProducerType {
 	public func concat(next: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
 		return SignalProducer<SignalProducer<Value, Error>, Error>(values: [ self.producer, next ]).flatten(.Concat)
 	}
+
+	/// `concat`s `self` onto initial `value`.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func beginWith(value: Value) -> SignalProducer<Value, Error> {
+		return SignalProducer(value: value).concat(self.producer)
+	}
 }
 
 private final class ConcatState<Value, Error: ErrorType> {
--- a/ReactiveCocoa/Swift/Flatten.swift
+++ b/ReactiveCocoa/Swift/Flatten.swift
@@ -388,7 +388,7 @@ extension SignalProducerType {
 
 	/// `concat`s `self` onto initial `value`.
 	@warn_unused_result(message="Did you forget to call `start` on the producer?")
-	public func beginWith(value: Value) -> SignalProducer<Value, Error> {
+	public func prefix(value value: Value) -> SignalProducer<Value, Error> {
 		return SignalProducer(value: value).concat(self.producer)
 	}
 }
--- a/ReactiveCocoaTests/Swift/FlattenSpec.swift
+++ b/ReactiveCocoaTests/Swift/FlattenSpec.swift
@@ -841,5 +841,27 @@ class FlattenSpec: QuickSpec {
 				expect(completed) == true
 			}
 		}
+
+		describe("SignalProducer.beginWith()") {
+			it("should emit initial value") {
+				let (signal, observer) = SignalProducer<Int, NoError>.pipe()
+
+				let mergedSignals = signal.beginWith(0)
+
+				var lastValue: Int?
+				mergedSignals.startWithNext { lastValue = $0 }
+
+				expect(lastValue) == 0
+
+				observer.sendNext(1)
+				expect(lastValue) == 1
+
+				observer.sendNext(2)
+				expect(lastValue) == 2
+
+				observer.sendNext(3)
+				expect(lastValue) == 3
+			}
+		}
 	}
 }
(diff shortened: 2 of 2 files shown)


mdiep:
Two bits of feedback:

1. This operator is usually called `startWith`.
2. A `Signal` variant should also be added.

I agree that it can be useful and would be 👍 to adding it once those are addressed.

inamiy:
@mdiep Thank you for feedback!

> This operator is usually called startWith.

The name `startWithXXX` is already used for activating `SignalProducer`, so I think we need to rename old `- (instancetype)startWith:(id)value)` operator to separate their meanings.

> A Signal variant should also be added.

There will be no `Signal` variant because the initial value of `beginWith` will only be emitted via cold `SignalProducer`'s `startWithSignal()` but not for hot `Signal` where initial value will be immediately emitted before `signal.observe()`.

mdiep:
> The name `startWithXXX` is already used for activating `SignalProducer`, so I think we need to rename old `- (instancetype)startWith:(id)value)` operator to separate their meanings.

I think that actually strengthens the case for naming it `startWith`—it reinforces that the value is sent when you call a `start…` method.

> There will be no `Signal` variant because the initial value of `beginWith` will only be emitted via cold `SignalProducer`'s `startWithSignal()` but not for hot `Signal` where initial value will be immediately emitted before `signal.observe()`.

Oh, right. 😆😅

inamiy:
> I think that actually strengthens the case for naming it startWith—it reinforces that the value is sent when you call a start… method.

Though there's a good naming reinforcement, I'm still a bit afraid to adopt this idea because there are already a lot of `startWithXXX()` methods and they can be easily mixed up with new `startWith()` operator. For example, we must always say **"`startWith()` does not actually _start_ `SignalProducer`"**.

As seen in slack discussion, `beginWith` sounds a good alternative, and this was indeed what I also come up with exact same name in [Swift example](https://github.com/inamiy/ReactiveCocoaCatalog/blob/20a01590f4bfb056e09f8a73dffb16d1c0a5e0b3/ReactiveCocoaCatalog/Samples/WhoToFollowViewController.swift).

I will +1 to the name `beginWith` once more. 
(but no problem to rename if many people opposed :wink:)

mdiep:
> For example, we must always say "**`startWith()` does not actually _start_ `SignalProducer`**".

The compiler will say that for us because of `@warn_unused_result`. ☺️

RuiAAPeres:
I am going with beginsWith as well. startXXX() returns a disposable, where the potential startsWith wouldnt.

mdiep:
@ReactiveCocoa/reactivecocoa This PR adds the equivalent of RAC2's `startWith` operator. There's debate over whether it should be called `startWith` or `beginsWith`. Can people please weigh in?

@inamiy and @RuiAAPeres have argued that there's potential confusion between `startWith` and `startWithNext`, etc.

@mdiep has argued that there's a helpful parallel between `startWith` and `startWithNext`, that they have different types, and that the compiler will point out if you use it wrong.

JaviSoto:
My 2 cents: in my codebase I named this in an extension `startWithValue()`. I was also concerned with the similarity with the other startWith methods, but I thought the keyword "value" did a good job at disambiguating.

However, it might be good to consider how we're gonna name these methods with Swift 3 and the new API naming guidelines. Perhaps the one that returns a disposable could be `func start(next:error:etc:)`. And that leaves this method (which btw I think is very useful to have) kind of in a tough spot. I think `func begin(value:)` would make sense.

Somewhat unrelated to this, I've also found a `startWithNil()` method that maps the signal into T? to be very useful to have!

mdiep:
Maybe it would work to go a different route and name it something like `prefix(value:)` (`prefixWith` in RAC4)?

neilpa:
Throwing out another option from left field -- `prefix`. IMO, both `startWith` and `beginWith` suggest that they subscribe the underlying producer at the call-site. Using begin**s** is @mdiep suggested is better but subtle and still potentially confusing. Something like `prefix` is more obviously declarative.

neilpa:
Lol, @mdiep beat me by seconds. Good sign that we had the same suggestion independently.

inamiy:
`prefix(value:)` sounds nicer to me!
I was worried if someone says `begin` or `beginWith` does not actually _begin_ SignalProducer :stuck_out_tongue_closed_eyes:

If it's no problem, I will rename this method and GH-issue title.

mdiep:
Go for it, @inamiy!

inamiy:
Done renaming in 608d501!

mdiep:
Thanks @inamiy! This is great. ✨ 

And I'm glad we settled on `prefix(value:)`. ☺️

inamiy:
@mdiep Thanks for merge & good naming! ✨

Issue #2875, opened 2016-05-06
Add missing Signal's operators @ Playground


RuiAAPeres:

tomj:
hey @RuiAAPeres just starting to knock this off now, do you think it best to follow the structure you've established with `SignalProducer` over [here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa.playground/Pages/SignalProducer.xcplaygroundpage/Contents.swift)?  I can drop an initial few operators on my branch and you can point out whether you think it's heading in the right direction, I'd hate to do all the operators only to have it not match the style you'd intended.

I'll start with: `collect`, `combineLatest`, `delay`, `skip` and `materialize` unless any objections 😀

RuiAAPeres:
@tomj make yourself at home. The way I went with the SignalProducer's was by order and using their own documentation.

tomj:
Rock and roll - will do.  Cheers 😄

tomj:
fyi I'm pushing my work on this up to [my fork](https://github.com/tomj/ReactiveCocoa/tree/signal_playground) just in case anyone is working on this at the same time

mdiep:
The playground has moved to ReactiveSwift, but I think this has been completed anyway.

PR #2943, opened 2016-05-27
Signal playground operators

tomj:
Following on from #2875 let me know how you guys want to do PR's like this - I could make one whopper with ~600 lines and all the operators or can do it in chunks like this and keep adding to this PR.

Hope the emoji's aren't too controversial.  I find RAC to be easier conceptualised in not just numerical terms and so in my own experience sometimes pictures can help 😄  (and they are called "play"grounds after all 😄 )

### DIFF
--- a/ReactiveCocoa.playground/Pages/Signal.xcplaygroundpage/Contents.swift
+++ b/ReactiveCocoa.playground/Pages/Signal.xcplaygroundpage/Contents.swift
@@ -19,7 +19,7 @@ import Foundation
 /*:
  ## Signal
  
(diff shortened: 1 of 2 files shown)


RuiAAPeres:
The goal of using numbers, instead of emojis, is that it's easier to understand the flow. I know that if my initial sequence is [1,2,3,4], after a 2 I will get a 3 (it's intuitive). Not so much with emojis.

tomj:
All good - just trying to mix it up a bit. 62465c7 takes a 🔫 to the emojis 😄

mdiep:
Looks good to me. 👍 Thanks!

Issue #2781, opened 2016-04-06
[RAC 2.x] incorrect output of switchToLatest when sending values on different scheduler


StatusReport:
I have a signal of signals that is being subscribed on the main scheduler. The signals that are sent on this signal are being subscribed on the default scheduler, and I'm using `switchToLatest` to switch to the values of the latest signal.

So the configuration is as follows:
```objc
RACSignal *signalA = [[RACSignal return:@"foo"]
    subscribeOn:[RACScheduler scheduler]];

RACSignal *signalB = [[RACSignal return:@"bar"]
    subscribeOn:[RACScheduler scheduler]];

__block NSMutableArray *values = [NSMutableArray array];
RACSubject *subject = [RACSubject subject];
[[subject
    switchToLatest]
    subscribeNext:^(id x) {
      [values addObject:x];
    } completed:^{
      NSAssert([values.lastObject isEqual:@"bar"]);
    }];

[subject sendNext:signalA];
[subject sendNext:signalB];
[subject sendCompleted];
``` 

Given the definition of `switchToLatest`, one would assume to have the last value of the `values` array to be equal to `@"bar"`. However, running this sample code in a loop raises an exception fairly quickly.

Is this considered a well-defined input for RAC (having signal of signals with different schedulers)? If so, I can try and work on a fix.

Note that the above question is also relevant for RAC 4, although it seems that the `switchToLatest` logic there is different and doesn't use `takeUntil:` like RAC 2.x (additional information on why this was the design decision will be appreciated, as a backport of this may be a potential fix).

mdiep:
Yes, that should work AFAICT. What exception do you get?

StatusReport:
I meant that the assert the my code fails.

StatusReport:
After digging some more, it seems that the issue stems from the fact that `switchToLatest` is using `takeUntil:` to dispose the subscription to a signal once a new signal arrives. So in the code example above, the subscription to `signalA`, which happens on the default scheduler, is disposed when a multicast connection signal that is subscribed on the main scheduler is fired (see `switchToLatest` implementation). This obviously leads to a possible race condition where the second signal is sent over `subject`, and the `takeUntil:` operator disposes the subscription just after `sendNext:` is called.

In the context of `takeUntil:` this makes complete sense, as the possible race condition is transparent to the client, but in `switchToLatest` it seems like a bug.

mdiep:
Can you explain that again? I lost you here:

[...]

PR #1610, opened 2014-12-10
Backport RACLiveSubscriber (redux)

jspahrsummers:
#1597, reopened after my premature merge.

I may or may not be the one to get to this. If someone else wants to fix up the tests, please feel free.

**To do:**

 - [ ] Get tests building

PR #2951, opened 2016-05-30
Avoid sending events after disposable have been disposed.

StatusReport:
The API of `RACPassthroughSubscriber` mentions that no more events will be sent after the disposable have been disposed. However, this is not correct if the disposable is disposed on a different thread than the thread that sends the events, as race conditions are possible inside the `send[Next|Completed|Error]` methods, after the `disposable.disposed` test.

### DIFF
--- a/ReactiveCocoa/Objective-C/RACPassthroughSubscriber.m
+++ b/ReactiveCocoa/Objective-C/RACPassthroughSubscriber.m
@@ -7,6 +7,7 @@
 //
 
 #import "RACPassthroughSubscriber.h"
+#import <ReactiveCocoa/EXTScope.h>
 #import "RACCompoundDisposable.h"
 #import "RACSignal.h"
 #import "RACSignalProvider.h"
@@ -33,7 +34,7 @@ static const char *cleanedSignalDescription(RACSignal *signal) {
 @interface RACPassthroughSubscriber ()
 
 // The subscriber to which events should be forwarded.
-@property (nonatomic, strong, readonly) id<RACSubscriber> innerSubscriber;
+@property (nonatomic, strong, readwrite) id<RACSubscriber> innerSubscriber;
 
 // The signal sending events to this subscriber.
 //
@@ -63,6 +64,14 @@ static const char *cleanedSignalDescription(RACSignal *signal) {
 	_disposable = disposable;
 
 	[self.innerSubscriber didSubscribeWithDisposable:self.disposable];
+	@weakify(self);
+	[self.disposable addDisposable:[RACDisposable disposableWithBlock:^{
+		@strongify(self);
+		@synchronized (self) {
+			self.innerSubscriber = nil;
+		}
+	}]];
+
 	return self;
 }
 
@@ -75,7 +84,9 @@ static const char *cleanedSignalDescription(RACSignal *signal) {
 		RACSIGNAL_NEXT(cleanedSignalDescription(self.signal), cleanedDTraceString(self.innerSubscriber.description), cleanedDTraceString([value description]));
 	}
 
-	[self.innerSubscriber sendNext:value];
+	@synchronized (self) {
+		[self.innerSubscriber sendNext:value];
+	}
 }
 
 - (void)sendError:(NSError *)error {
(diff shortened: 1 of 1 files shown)


StatusReport:
Notes:
- This is a _huge_ bug in RAC 2.x, IMHO. I can give many examples of bad effects, but for example think of a `UICollectionViewCell` with an image signal that is being replaced upon cell reuse.
- The solution adds a `@synchronized` block around each event that the passthrough subscriber sends. This adds about ~8%-10% performance penalty in my tests. Ideas on how to make it more efficient are welcomed.
- I tried writing test for this, but I'm not sure if it's possible to sanely test this since I need to check if a disposable has been disposed, and the `BOOL` flag in `RACCompoundDisposable` that indicates that is set *before* the inner disposables are being disposed. So without a lot of tricks that may effect the results of the tests I don't see how's that possible.

ReactiveCocoa/Objective-C/RACPassthroughSubscriber.m
```diff
@@ -75,7 +84,9 @@ - (void)sendNext:(id)value {
 		RACSIGNAL_NEXT(cleanedSignalDescription(self.signal), cleanedDTraceString(self.innerSubscriber.description), cleanedDTraceString([value description]));
 	}
```
mdiep: The dtrace macros should probably be moved inside the lock?

And I think this means that the check of `self.disposable.disposed` can be removed. It should probably be replaced with an early exited that checks `if (self.innerSubscriber == nil) return;`.

mdiep:
It doesn't seem like this should matter. When the inner subscriber is disposed, it will stop taking its actions. So even though it doesn't happen in `RACPassthroughSubscriber`, events on the inner subscriber won't be sent after it's disposed.

How did this relate to #2781?

StatusReport:
It does matter, since `RACSubscriber` allows multiple subscriptions over the same `next`, `error` and `complete` blocks. These blocks are `nil`d out when `RACSubscriber`'s disposable is disposed, but `-dispose` is only called when `error` or `complete` is sent, and inside `-dealloc`. Note that it is _not_ called when the disposable given to `-didSubscribeWithDiposable:` is disposed.

So in case we disposed a subscription disposable, the blocks will not be `nil`d out, and additional calls to `next`, `error` or `complete` will still go through to the subscriber until `RACSubscriber` is deallocated. Indeed, the retain cycle between the subscriber's disposable and the subscription disposable is not there after the subscription disposable has been disposed, but if it entered an autoreleasepool, the deallocation of `RACSubscriber` will not happen immediately (and on GCD threads without a runloop [it happens only when they're idle](http://asciiwwdc.com/2015/sessions/718)).

Regarding relation to #2781, I don't want to spoil all the future fun there, but the new implementation of `-switchToLatest` there's a need to dispose the old inner signal and make sure that no additional values from it are being sent (which is exactly the issue I was reporting).

mdiep:
> It does matter, since RACSubscriber allows multiple subscriptions over the same next, error and complete blocks. These blocks are nild out when RACSubscriber's disposable is disposed, but -dispose is only called when error or complete is sent, and inside -dealloc. Note that it is not called when the disposable given to -didSubscribeWithDiposable: is disposed.

That makes sense. Thanks. I'm not very familiar with the RAC 2 internals, so I'm trying to learn as I go here.

My comment above still needs to be addressed.

If you want to try to improve the performance, my understanding is that an `NSLock` will be faster than using `@synchronized`. You could also try using `dispatch_sync` like `RACKVOProxy`; I'm not sure how that compares in performance.

tzongw:
> However, this is not correct if the disposable is disposed on a different thread than the thread that sends the events, as race conditions are possible inside the send[Next|Completed|Error] methods, after the disposable.disposed test.

I think if `dispose` is called on a different thread, whether receiving more events is a chance, since `send` could be called before `dispose` actually taking place and race conditions are still possible.

StatusReport:
@tzongw it is fine for `send*` methods to happen before `dispose`, but after `dispose` has been returned you must not receive additional calls.

@mdiep we're trying to revive #1610, which will replace `RACPassthroughSubscriber` with `RACLiveSubscriber`. It seems that this problem exists there as well and we will fix it before sending it for review. So this PR is on hold for now.

mdiep:
The Objective-C ReactiveCocoa APIs have been moved to https://github.com/ReactiveCocoa/ReactiveObjC/. If you're interested in continuing this PR, please reopen it against that repository. Thanks!

PR #2952, opened 2016-05-31
Debounce

### LINKED ISSUE #2075: Proposal: revive debounce (the throttle of old)

mdiep:
Fixes #2075, supersedes #2939.

This adds the `precondition` that Rex had and conforms it to the RAC style.

Rex's didn't pass the (legitimate) test that was included. I think that's because termination events weren't sent on the passed-in scheduler.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1215,6 +1215,24 @@ extension SignalType {
 			return disposable
 		}
 	}
+
+	/// Debounce values sent by the receiver, such that at least `interval`
+	/// seconds pass after the receiver has last sent a value, then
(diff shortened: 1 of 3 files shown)


ikesyo:
👍

PR #2954, opened 2016-06-01
Added suffix 

RuiAAPeres:


mdiep:
I'm 👍 to add this, but I think it should be called `concat(value:)`. ☺️ Intra-operator consistency is more important than inter-operator consistency. I could also imagine a `prefix` variant that takes a `SignalProducer`.

RuiAAPeres:
@mdiep what do you think about this:

```swift
	@warn_unused_result(message="Did you forget to call `start` on the producer?")
	public func concat(next: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
		return SignalProducer<SignalProducer<Value, Error>, Error>(values: [ self.producer, next ]).flatten(.Concat)
	}
	
	/// `concat`s `value` onto `self`.
	@warn_unused_result(message="Did you forget to call `start` on the producer?")
	public func concat(value value: Value) -> SignalProducer<Value, Error> {
		return self.concat(SignalProducer(value: value))
	}
	
	/// `concat`s `self` onto initial `previous`.
	@warn_unused_result(message="Did you forget to call `start` on the producer?")
	public func prefix(previous previous: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
		return previous.concat(self.producer)
	}
	
	/// `concat`s `self` onto initial `value`.
	@warn_unused_result(message="Did you forget to call `start` on the producer?")
	public func prefix(value value: Value) -> SignalProducer<Value, Error> {
		return self.prefix(previous: SignalProducer(value: value))
	}
```

More specifically the `previous`.

mdiep:
I think the `SignalProducer` variants should use unnamed parameters. I think of those as the default parameter type. The argument name is used when the parameter is something other than a `SignalProducer`/`Signal`.

RuiAAPeres:
> The argument name is used when the parameter is something other than a SignalProducer/Signal

👍 

And as for the `previous` naming?

mdiep:
> And as for the previous naming?

That seems good. 👍

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -385,11 +385,23 @@ extension SignalProducerType {
...
+	
+	/// `concat`s `value` onto `self`.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func concat(value value: Value) -> SignalProducer<Value, Error> {
+		return self.concat(SignalProducer(value: value))
+	}
+	
+	/// `concat`s `self` onto initial `previous`.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func prefix(previous: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
```
ikesyo: I'd like to accept a `SignalProducerType` here:

```swift
public func prefix<S: SignalProducerType where P.Value == Value, P.Error == Error>(previous: P) -> SignalProducer<Value, Error> {
```

RuiAAPeres:
@ikesyo 👍

mdiep:
This is great—thanks! And just in time for the 4.2 release. ☺️

PR #2955, opened 2016-06-02
Fix a typo in the documentation playground

rpowelll:
The documentation for `startWithInterrupted` incorrectly stated that it would invoke its callback on a `failed` event


mdiep:
Thanks!

Issue #2949, opened 2016-05-29
Internal signal not being disposed by flatMap operator on outter SignalProducer


guidomb:
I'm not sure if this is a bug or expected behavior but given my interpretation of the docs, in particular to the comments of the `observe` method

> Observes the Signal by sending any future events to the given observer. If the Signal has already terminated, the observer will immediately receive an `Interrupted` event.
> 
> Returns a Disposable which can be used to disconnect the observer. Disposing of the Disposable will have no effect on the Signal itself.

which says that the disposable returned by this method is used to remove the observer and given https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2453 and https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2582. I understand that if I have a producer of signal which is flattened and then the producer is disposed, it should also dispose the internal signal inside the closure provided to flatMap.

The following is an example that reproduces the issue that I describe in the title

``` swift
let signal = Signal<Int, NoError> { observer in

    func next(count: Int = 0) {
        let delayTime = dispatch_time(DISPATCH_TIME_NOW, Int64(1 * Double(NSEC_PER_SEC)))
        dispatch_after(delayTime, dispatch_get_main_queue()) {
                observer.sendNext(count)
                next(count + 1)
        }
    }

    next()

    return nil
}

let action = Action { (_: AnyObject) -> SignalProducer<(), NSError> in
    return SignalProducer { observer, _ in
        let delayTime = dispatch_time(DISPATCH_TIME_NOW, Int64(2 * Double(NSEC_PER_SEC)))
        dispatch_after(delayTime, dispatch_get_main_queue()) {
            observer.sendNext(())
            observer.sendCompleted()
        }
    }
    .on(
        terminated: { print("action producer terminated") },
        disposed: { print("action producer disposed") },
        interrupted: { print("action producer interrupted") }
    )
}
let result = action.executing.producer
    .filter { $0 }
    .flatMap(.Latest) { executing -> Signal<[Int], NoError> in
        print("action executed")
        return signal
            .scan(Array<Int>()) { acum, current in
                print("Accumulating")
                var newAccum = acum
                newAccum.append(current)
                return newAccum
            }
            .on(
                terminated: { print("scan signal terminated") },
                disposed: { print("scan signal disposed") },
                interrupted: { print("scan signal interrupted") }
            )
    }
    .on(
        terminated: { print("result producer terminated") },
        disposed: { print("result producer disposed") },
[...]

PR #2959, opened 2016-06-03
[5.0] Changes to the lifetime semantic of `Signal`.

andersio:
A proposed solution following the discussion in #2949.
A replica of the grand old #2591.
Mutually exclusive with #3024.

Edited on 23 July to improve the writing.

### Design
https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2959#issuecomment-236131149

#### Side effect in operators without retaining the signal
```
var shinyCounter = 0
func shinyWork() {
    mySignal.map {
        shinyCounter += 1
        return $0
    }
}
shinyWork()
myObserver.sendNext(2)
/// At this point:
/// Before this PR: shinyCounter would be 1.
/// After this PR: shinyCounter would be 0.
```
would no longer be valid.

To keep the event stream alive so that the side effect can be evaluated, three alternatives can be used:

1. Retains the resulting signal
2. Attaches an empty observer.
3. Convert the side effect into an observer.

#### Intermediate signals would now be disposed of if the resulting signal is disposed of.
```
var disposable: Disposable?
func shinyWork() {
    disposable  = shinySignal
        .map { $0 + 1 } /// First mapped signal
        .map { $0 + 2 } /// Second mapped signal
        .map { $0 + 3 } /// Third mapped signal
        .observe { _ in }
}
shinyWork()
disposable?.dispose()
```
Upon exiting the scope of `shinyWork`, all the mapped signals would become **unreachable**. 

Therefore, at the time the `disposable` is disposed of, it would cause the bottommost mapped signal to be **unwatched** and consequently trigger a disposal.

Likewise, the disposal would be propagated up along the chain, until it reaches an event stream that is either (or both) reachable or watched.


guidomb:
Just to clarify what the new contract would imply. Lets say you have the following class

```swift
public final class SomeSingletonClass {
    
    public static let sharedInstance = SomeSingletonClass()
    
    public private(set) lazy var signal: Signal<Int, NoError> =
        Signal { [weak self] observer in
            
            func next(count: Int = 0) {
                let delayTime = dispatch_time(DISPATCH_TIME_NOW, Int64(1 * Double(NSEC_PER_SEC)))
                dispatch_after(delayTime, dispatch_get_main_queue()) {
                    observer.sendNext(count)
                    if self?.sendEvents ?? false {
                        next(count + 1)
                    }
                }
            }
            
            next()
            
            return ActionDisposable { [weak self] _ in
                print("Original signal disposed")
                self?.sendEvents = false
            }
        }
    
    private var sendEvents = true
    
}
```

and then is some other part of your code base

```swift
let transformedSignal = SomeSingletonClass.sharedInstance.signal
    .map { value in
        print("Multiplying \(value)")
        return value * 2
    }
    .map { (value: Int) in
        print("Adding 10")
        return value + 10
    }
    .filter { (value: Int) in
        print("Filtering")
        return value % 10 != 0
}

let disposable = transformedSignal.observeNext {
    print("")
    print($0)
}

let delayTime = dispatch_time(DISPATCH_TIME_NOW, Int64(5 * Double(NSEC_PER_SEC)))
dispatch_after(delayTime, dispatch_get_main_queue()) {
    print("disposing")
    disposable?.dispose()
}
```

Currently with the latest release of ReactiveCocoa (4.2) the console output would be

```
Multiplying 0
Adding 10
Filtering
Multiplying 1
Adding 10
Filtering

12
Multiplying 2
Adding 10
Filtering

14
Multiplying 3
Adding 10
Filtering

16
disposing
Multiplying 4
Adding 10
Filtering
Multiplying 5
Adding 10
Filtering
Multiplying 6
Adding 10
Filtering
Multiplying 7
Adding 10
Filtering
Multiplying 8
Adding 10
Filtering
```

but with the new contract it would be

```
Multiplying 0
Adding 10
Filtering
Multiplying 1
Adding 10
Filtering

12
Multiplying 2
Adding 10
Filtering

14
Multiplying 3
Adding 10
Filtering

16
disposing
```

Meaning the the intermediate signal created by `map` and `filter` will be terminated. Am I correct?

andersio:
@guidomb Yes. A timeline of what's happening:

Scope 1

1. Maps the source signal to a new signal A.
2. Maps signal A to a new signal B.
3. Filters signal B to derive a new signal C.
4. Attaches an observer to signal C.
5. Enqueue a block to GCD
6. Deinitialization of `Signal` of signal A, B and C when exiting the scope.

Scope 2 (a submitted block called by GCD)

1. Detach the observer from signal C. There are two possibilities of what would happen:
    1. If the detachment happens before `Signal` of C has deinitialised, i.e. Step 6 in Scope 1, the generator disposable of signal C would be disposed of by the deinitialiser.
    2. If the detachment happens after `Signal` of C has deinitialised, the generator disposable of signal C would be disposed of.
2. As a result of the disposal in (1), the observer of signal C observing signal B would be detached.
3. Similar to (1), the generator disposable of signal B would eventually be disposed of.
4. As a result of the disposal in (3), the observer of signal B observing signal A would be detached.
5. Similar to (1), the generator disposable of signal A would eventually be disposed of.
6. All intermediate signals have been disposed of.

mdiep:
I'm investigating this, but it will probably take me a little longer to get to. Unfortunately I'm not sure I'll get to it until after WWDC.

andersio:
(Kind of) superseded by #3024. Could be revived as needed.

andymatuschak:
Nice! This reachability semantic does seem to address the issue I encountered in #3075.

andersio:
The latest commit has taken out the implementation of the event stream from the `Signal`, which is now only a proxy to the event stream.

andymatuschak:
I haven't done a line-level review, but conceptually, this makes sense to me.

mdiep:
The idea makes sense to me. I'll try to do an in-depth review this week.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -139,22 +84,129 @@ public final class Signal<Value, Error: ErrorProtocol> {
 	///            the Signal itself.
 	@discardableResult
 	public func observe(_ observer: Observer) -> Disposable? {
+		if let disposable = stream.observe(observer) {
+			return disposable
+		} else {
+			observer.sendInterrupted()
+			return nil
+		}
```
mdiep: It seems like the `else` branch here could be handled by `stream`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -139,22 +84,129 @@ public final class Signal<Value, Error: ErrorProtocol> {
...
+	}
+}
+
+private final class EventStream<Value, Error: ErrorProtocol> {
+	/// The optional disposable returned by the signal generator.
+	/// It would be disposed of when the event stream is disposed of.
+	private var generatorDisposable: Disposable?
+
+	/// An atomic bag of observers to this event stream.
+	private let atomicObservers: Atomic<Bag<Signal<Value, Error>.Observer>?>
```
mdiep: I realize that you didn't name it, but this name is a bit silly. It should just be called `observers`. The atomicity is noted in the type.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -14,12 +14,14 @@ import Result
 /// progress,” like notifications, user input, etc. To represent streams that
 /// must first be _started_, see the SignalProducer type.
 ///
-/// Signals do not need to be retained. A Signal will be automatically kept
-/// alive until the event stream has terminated.
+/// Signals do not need to be retained to keep the event stream alive. The event
+/// stream would be automatically kept alive until either:
+/// 1. a terminating event is received; or
+/// 2. the event stream is unwatched and unreachable.
```
mdiep: I think it's simpler (and still accurate) to say:

> 2\. The signal has no observers and deallocates

mdiep:
It seems like it might be simpler to (1) have each observer retain the signal—possibly through the disposable, (2) remove any retain cycles in the `Signal`, and (3) interrupt the signal when it deallocates. What do you think?

(I wouldn't erase what you have here until/unless we know that works and is better. But it might be worth exploring in a separate PR?)

Still digging into this, but that's my initial reaction…

andersio:
Now it is like:
```
               -----------------
           ++++| inputObserver |
       weak+   ----------------- 
           +                       
  ------------------   weak   ------------------------
  |     Signal     | ++++++++ | observer disposables |
  ------------------          ------------------------
           ↑                                    
     strong|   -------------  
           ╰---| observers |  
               -------------  
```

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -14,12 +14,14 @@ import Result
...
-/// alive until the event stream has terminated.
+/// A Signal is kept alive until either of the following happens:
+///    1. its input observer receives a terminating event; or
+///    2. it has no active observers, and is not being retained.
 public final class Signal<Value, Error: ErrorProtocol> {
 	public typealias Observer = ReactiveCocoa.Observer<Value, Error>
 
-	private let atomicObservers: Atomic<Bag<Observer>?> = Atomic(Bag())
+	private var generatorDisposable: Disposable?
+	private let state: Atomic<SignalState<Value, Error>?>
```
mdiep: This could use a comment explaining that the `SignalState` will be `nil` once the `Signal` is `Interrupted`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -14,12 +14,14 @@ import Result
...
-/// Signals do not need to be retained. A Signal will be automatically kept
-/// alive until the event stream has terminated.
+/// A Signal is kept alive until either of the following happens:
+///    1. its input observer receives a terminating event; or
+///    2. it has no active observers, and is not being retained.
 public final class Signal<Value, Error: ErrorProtocol> {
 	public typealias Observer = ReactiveCocoa.Observer<Value, Error>
 
-	private let atomicObservers: Atomic<Bag<Observer>?> = Atomic(Bag())
+	private var generatorDisposable: Disposable?
```
mdiep: I think this can be `let`?

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -135,19 +146,28 @@ public final class Signal<Value, Error: ErrorProtocol> {
...
-	///            observer. Disposing of the Disposable will have no effect on
-	///            the Signal itself.
+	///            observer.
 	@discardableResult
 	public func observe(_ observer: Observer) -> Disposable? {
 		var token: RemovalToken?
-		atomicObservers.modify { observers in
-			token = observers?.insert(observer)
+		state.modify { state in
+			if state != nil {
```
mdiep: `if let state = state {` ☺️

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -135,19 +146,28 @@ public final class Signal<Value, Error: ErrorProtocol> {
...
+				if state!.observers.isEmpty {
```
mdiep: This `isEmpty` check isn't really necessary since setting `state.retainedSignal` again has no effect.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -135,19 +146,28 @@ public final class Signal<Value, Error: ErrorProtocol> {
...
+				token = state!.observers.insert(observer)
+			}
 		}
 
 		if let token = token {
 			return ActionDisposable { [weak self] in
-				_ = self?.atomicObservers.modify { observers in
-					observers?.remove(using: token)
+				_ = self?.state.modify { state in
+					if state != nil {
```
mdiep: `if let state = state {` ☺️

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -135,19 +146,28 @@ public final class Signal<Value, Error: ErrorProtocol> {
...
 		var token: RemovalToken?
-		atomicObservers.modify { observers in
-			token = observers?.insert(observer)
+		state.modify { state in
+			if state != nil {
+				if state!.observers.isEmpty {
+					state!.retainedSignal = self
+				}
+				token = state!.observers.insert(observer)
+			}
```
mdiep: Would this work?

```swift
state.modify { state in
    state?.retainedSignal = self
    token = state?.observers.insert(observer)
}
```

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -32,17 +34,28 @@ public final class Signal<Value, Error: ErrorProtocol> {
...
 
-		let generatorDisposable = SerialDisposable()
-
 		/// When set to `true`, the Signal should interrupt as soon as possible.
 		let interrupted = Atomic(false)
 
-		let observer = Observer { event in
+		let observer = Observer { [weak self] event in
+			guard let signal = self else {
+				return
```
mdiep: Can this happen?

ReactiveCocoa/Swift/Signal.swift
andersio: `self` is used in constructing the observer passed into the generator, and the setting of this comes after. So probably not until we have delayed initialisation.

ReactiveCocoa/Swift/Signal.swift
andersio: It can. For example, an implementation of an operator forgot to return the observer disposable as (part of) the generator disposable. Then this input observer would stick in the upstream forever, with its captured `self` being `nil`.

ReactiveCocoa/Swift/Signal.swift
andersio: Yeah. I just put it that way for (perhaps unnecessarily) one less `nil` check and less releases & retains.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -32,17 +38,28 @@ public final class Signal<Value, Error: ErrorProtocol> {
...
 		sendLock.name = "org.reactivecocoa.ReactiveCocoa.Signal"
 
-		let generatorDisposable = SerialDisposable()
-
 		/// When set to `true`, the Signal should interrupt as soon as possible.
 		let interrupted = Atomic(false)
 
-		let observer = Observer { event in
+		let observer = Observer { [weak self] event in
+			guard let signal = self else {
```
mdiep: It'd probably be better to use `self` here instead of `signal`.

```swift
guard let `self` = self else {
```

mdiep:
@ReactiveCocoa/reactivecocoa This looks good to me. Would anyone else like to review the semantics and/or code?

ReactiveCocoa/Swift/Signal.swift
andersio: This is an unintended compiler bug, as mentioned in the discussion thread of optional binding for `self`.

mdiep:
I don't see any objectors. Unfortunately, this now has conflicts. 😭 Once those are fixed up, we're 👍🏻to merge.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -189,26 +210,108 @@ class SignalLifetimeSpec: QuickSpec {
...
+					disposable = sourceSignal
+						.map { (value: Int) -> Int in
+							firstCounter += 1
+							return value
+						}
+						.map { (value: Int) -> Int in
+							secondCounter += 1
+							return value
+						}
+						.map { (value: Int) -> Int in
```
NachoSoto: Can you remove the `: Int` in these 3 `.map`s? They're not necessary.

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
```diff
@@ -26,37 +26,41 @@ class SignalLifetimeSpec: QuickSpec {
 				expect(signal).to(beNil())
 			}
 
-			it("should deallocate even if it has an observer") {
+			it("should deallocate if it does not have any observers") {
```
NachoSoto: I think we should write another version of this that makes the `signal` be initially retained, and verify that it is not deallocated, but then it is once the last strong reference is removed.
It should be equivalent to this one, but since the timing is slightly different, it's worth having.

NachoSoto:
:shipit: from me (though I'll review once it's rebased).
I'd still love to see that test I suggested though ⭐

ReactiveCocoaTests/Swift/SignalLifetimeSpec.swift
andersio: *should not deallocate if it has at least one observer* should be the test case you want.

NachoSoto: No, I mean when the signal never gets any observers, but it's temporarily retained.

mdiep:
🎉

PR #2961, opened 2016-06-06
Fix memory leak when binding MutablePropertyType to a producer

github-user32:
I've fixed a memory leak when binding a MutablePropertyType to a producer (`public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.Value, NoError>) -> Disposable`). The fix is similar to that already in place when binding a MutablePropertyType to a signal.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -238,14 +238,14 @@ public func <~ <P: MutablePropertyType>(property: P, signal: Signal<P.Value, NoE
 /// The binding will automatically terminate when the property is deinitialized,
 /// or when the created signal sends a `Completed` event.
 public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.Value, NoError>) -> Disposable {
-	var disposable: Disposable!
+	let disposable = CompositeDisposable()
 
 	producer.startWithSignal { signal, signalDisposable in
 		property <~ signal
-		disposable = signalDisposable
+		disposable += signalDisposable
 
-		property.producer.startWithCompleted {
-			signalDisposable.dispose()
+		disposable += property.producer.startWithCompleted {
+			disposable.dispose()
 		}
 	}
 


mdiep:
Could you add a test for the memory leak so it doesn't regress? There are lots of examples in the test suite.

github-user32:
Yep!

github-user32:
Hi Mdiep,
I've had look into updating the relevant PropertySpec.swift test (`describe("binding")/describe("from a SignalProducer")/it("should tear down the binding when disposed")`) so that it could detect the memory leak concerned, however, I can't see a clear way to do this as the memory that is leaked is all internal to the implementation of `<~` and there is no external data that can be used to detect the leak - ie. there is nothing external to `<~` that could be used to verify whether memory was leaked or not.

To do so, I'd need to be able to verify that the return value from the internal call to property.producer.startWithCompleted was disposed of, and the only way I can think to do this is to mock up an entire MutableProperty (simple) so that it uses a mocked SignalProducer (not so simple) and mocked Disposables. The mocked producer could be instrumented to keep track of the Disposables it returns and then the test could validate that those Disposables were actually disposed of.

However, given that SignalProducerType has quite a large interface and SignalProducer is a struct that can't simply be sub-classed to easily add instrumentation, I think this would all add up to sizeable amount of work, which I'm not able to do presently. 

I'd also note that the existing (accepted) implementation for binding to a Signal (`public func <~ <P: MutablePropertyType>(property: P, signal: Signal<P.Value, NoError>) -> Disposable`) is virtually the same as my proposed fix for binding to a SignalProducer and that the corresponding test case for the Signal binding operator does not verify that there are any internal implementation leaks and does not do the kind of mocking that would be necessary to verify such internal leaks either.

Does this make sense? (Let me know if I've missed an obvious way to test this. Scripting and analysing Instruments run over the tests, would be another non-trivial way to verify memory leak fixes.)

mdiep:
Makes sense! Thanks for the PR and for trying to add a test. ✨

github-user32:
Great! Thanks mdiep.

PR #2963, opened 2016-06-06
[Bugfix] Property composition operator.

andersio:
When working on #2964, I caught a bug that can crash the composition operators. It would affect `ConstantProperty`s in the current RAC5 branch.

A perfect counter example of IUO.
:-[

PR #2964, opened 2016-06-06
[5.0] Changes to producer- and signal-sourced `AnyProperty`s.

andersio:
Since the producers and signals of composed properties respect the lifetime of their ultimate source(s), IMO we should extend the behaviour to producer/signal-sourced `AnyProperty`s for consistency.

It would be a breaking change though, as currently the producers and signals of these `AnyProperty`s terminates when the property (note: not the source) are deinitialised.

Requires #2963.


andersio:
This time:
> ReactiveCocoaTests.ObjectiveCBridgingSpec
  RACScheduler__gives_current_date, expected to be close to <2016-06-06 11:46:32.6440> (within 0.0001), got <2016-06-06 11:46:32.6500>
  > /Users/travis/build/ReactiveCocoa/ReactiveCocoa/ReactiveCocoaTests/Swift/ObjectiveCBridgingSpec.swift:27

TBH, like the one failing in #2963, this often fails in Travis CI. Should these tests be tolerable, ehh?

PR #2966, opened 2016-06-07
[5.0] Changes to producer- and signal-sourced `AnyProperty`s.

andersio:
Since now we have property composition operators creating properties that respect the lifetime of their ultimate source(s), IMO producer/signal-sourced `AnyProperty`s should be treated the same way for consistency.

It would be a breaking change though, as currently the producers and signals of these `AnyProperty`s terminates when the property (note: not the source) are deinitialised.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -366,17 +366,15 @@ public struct AnyProperty<Value>: PropertyType {
 	/// Initializes a property that first takes on `initialValue`, then each value
 	/// sent on a signal created by `producer`.
 	public init(initialValue: Value, producer: SignalProducer<Value, NoError>) {
-		let mutableProperty = MutableProperty(initialValue)
-		mutableProperty <~ producer
-		self.init(mutableProperty)
+		self.init(propertyProducer: SignalProducer(value: initialValue).concat(producer),
+		          capturing: [])
 	}
 
 	/// Initializes a property that first takes on `initialValue`, then each value
 	/// sent on `signal`.
 	public init(initialValue: Value, signal: Signal<Value, NoError>) {
-		let mutableProperty = MutableProperty(initialValue)
-		mutableProperty <~ signal
-		self.init(mutableProperty)
+		self.init(propertyProducer: SignalProducer(value: initialValue).concat(SignalProducer(signal: signal)),
+		          capturing: [])
 	}
 
 	/// Initializes a property by applying the unary `SignalProducer` transform on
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -366,17 +366,15 @@ public struct AnyProperty<Value>: PropertyType {
 	/// Initializes a property that first takes on `initialValue`, then each value
 	/// sent on a signal created by `producer`.
 	public init(initialValue: Value, producer: SignalProducer<Value, NoError>) {
-		let mutableProperty = MutableProperty(initialValue)
-		mutableProperty <~ producer
-		self.init(mutableProperty)
+		self.init(propertyProducer: SignalProducer(value: initialValue).concat(producer),
```
mdiep: I just updated the `RAC5` branch from `master`. Can you merge into this branch and use the new `prefix` operator here? It will simplify it quite a bit.

mdiep:
This change makes sense to me. But I think it should use `prefix` as noted above.

andersio:
@mdiep Working on it. Though it seems there might be an issue in `zipWith` with the later commits. I am investigating.

andersio:
@mdiep Is `zipWith` supposed to never complete if both signals do not send the same amount of values? Some patches merged into RAC5 seem to have changed that.

andersio:
@mdiep I caught a bug in the new `zipWith` implementation, which causes the signal to never complete even if both sources are completed.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1026,7 +1026,7 @@ private struct ZipState<Left, Right> {
 	var completed: (left: Bool, right: Bool) = (false, false)
 
 	var isFinished: Bool {
-		return (completed.left && values.left.isEmpty) || (completed.right && values.right.isEmpty)
+		return (completed.left && values.left.isEmpty) || (completed.right && values.right.isEmpty) || (completed.left && completed.right)
```
mdiep: Could you open up a PR against `master` for this? It's unrelated to the other changes here, but more importantly it's not specific to `RAC5`

Also: If `completed.left && completed.right` and the previous conditions aren't true, that means that there are more values that should have been sent out. So the fix should be to send those values, not to prematurely end the signal.

This must be fallout from #2917. Major bonus point if you can add a test. 😄

andersio: @mdiep It doesn't prematurely end the signal. The conditions here only determines whether it should end the signal after sending a tuple (if any).

andersio: > If completed.left && completed.right and the previous conditions aren't true, that means that there are more values that should have been sent out.

Then `isFinished` would be false anyway, and the logic looking up for values works independent of this `isFinished` flag.

andersio: Done. 😺

mdiep:
It looks like this failed test requires #2972. Let's wait to merge this until that's merged through.

mdiep:
I've pulled the fix from #2972 into `RAC5` if you want to merge that into here.

andersio:
@mdiep Rebased.

mdiep:
✨

PR #2969, opened 2016-06-08
fix typo

mshibanami:


mdiep:
Thanks for the PR, but this is correct as-is. `on` is an instance method; this example example is trying to illustrate that you can use it on a `signalProducer` variable;

mshibanami:
Oh, sorry.

PR #2972, opened 2016-06-09
[Bugfix] Zipped producers do not complete when both sources are completed.

andersio:


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1023,10 +1023,10 @@ extension SignalType {
 
 private struct ZipState<Left, Right> {
 	var values: (left: [Left], right: [Right]) = ([], [])
-	var completed: (left: Bool, right: Bool) = (false, false)
+	var isCompleted: (left: Bool, right: Bool) = (false, false)
 
 	var isFinished: Bool {
-		return (completed.left && values.left.isEmpty) || (completed.right && values.right.isEmpty)
+		return (isCompleted.left && values.left.isEmpty) || (isCompleted.right && values.right.isEmpty) || (isCompleted.left && isCompleted.right && (values.left.isEmpty || values.right.isEmpty))
```
mdiep: The added case here is redundant.

Before this was `(C1 && V1) || (C2 && V2)`. You've made it `(C1 && V1) || (C2 && V2) || (C1 && C2 && (V1 || V2)). If `C1 && C2` is true, then this can be simplified to `V1 || V2 || (V1 || V2)`.

andersio: @mdiep Hmm. I was able to proof that this expression is the equivalent of the previous one. I would revert this.

mdiep:
There seem to be some legitimately failing tests.

andersio:
@mdiep Yeah. Something is wrong about `defer`.

mdiep:
Thanks!

PR #2976, opened 2016-06-10
Flatten changes

baconpat:
I ran into a couple of issues I was able to trace back to `-flatten:`. This pull request fixes them.

1. There's a race condition in `-flatten:` when it's used by `-concat`. If the current subscription completes at the same time as a new signal is added, the new signal could be subscribed to immediately in addition to the queued up signal that was supposed to be subscribed to next.

2. A memory leak where the strong reference to the `subscribeToSignal` wasn't being cleaned up.

3. A possible crash with that same `subscribeToSignal` being called after it had been set to nil.

### DIFF
--- a/ReactiveCocoa/RACSignal+Operations.m
+++ b/ReactiveCocoa/RACSignal+Operations.m
@@ -516,6 +516,12 @@ static RACDisposable *subscribeForever (RACSignal *signal, void (^next)(id), voi
 			}
 		};
 
+		[compoundDisposable addDisposable:[RACDisposable disposableWithBlock:^{
+			// A strong reference is held to `subscribeToSignal` until we're
+			// done, preventing it from deallocating early.
+			subscribeToSignal = nil;
+		}]];
+
 		// The signals waiting to be started.
 		//
 		// This array should only be used while synchronized on `subscriber`.
@@ -560,7 +566,9 @@ static RACDisposable *subscribeForever (RACSignal *signal, void (^next)(id), voi
 					pendingSubscription = YES;
 				}
 
-				subscribeToSignal(nextSignal);
+				if (subscribeToSignal) {
+					subscribeToSignal(nextSignal);
+				}
 			}];
 		};
 
@@ -579,7 +587,9 @@ static RACDisposable *subscribeForever (RACSignal *signal, void (^next)(id), voi
 				}
 			}
 
-			subscribeToSignal(signal);
+			if (subscribeToSignal) {
+				subscribeToSignal(signal);
+			}
 		} error:^(NSError *error) {
 			[subscriber sendError:error];
 		} completed:^{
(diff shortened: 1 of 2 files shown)


mdiep:
> A memory leak where the strong reference to the subscribeToSignal wasn't being cleaned up.

It looks like this is half of #2433, which I somehow missed when I was looking for patches that hadn't been back ported. Would you mind updating the changes here to match those or opening the other changes against `master`?

ReactiveCocoaTests/RACSignalSpec.m
```diff
@@ -1907,6 +1907,61 @@ + (void)configure:(Configuration *)configuration {
 			expect(@(flattenDisposable.disposed)).toEventually(beTruthy());
 		}
 	});
+
+	qck_it(@"only subscribes to one signal at a time - even when adding new signals at the same time as one completes", ^{
+		for (int i = 0; i < 100; i++) {
```
mdiep: I'm not wild about adding a long test like this. 😕 It takes 1.6s on CI.

ReactiveCocoa/RACSignal+Operations.m
```diff
@@ -560,7 +578,7 @@ - (RACSignal *)flatten:(NSUInteger)maxConcurrent {
 			NSCAssert([signal isKindOfClass:RACSignal.class], @"Expected a RACSignal, got %@", signal);
 
 			@synchronized (subscriber) {
-				if (maxConcurrent > 0 && activeDisposables.count >= maxConcurrent) {
+				if (maxConcurrent > 0 && (activeDisposables.count >= maxConcurrent || pendingSubscription)) {
```
mdiep: I don't think this works generally. It works for `-concat` because `maxConcurrent` is `1`, but `flatten:` needs to handle `n`—including `n` pending subscriptions.

Think of the case where a signal is `flatten:10`. If 3 threads race to send a signal through the `flatten:`, all 3 should be immediately subscribed to—even if one of them is `pendingSubscription`.

So I think `pendingSubscription` should be an `NSUInteger` of pending subscriptions, and this should check `maxConcurrent > 0 || (activeDisposables.count + pendingSubscriptions >= maxConcurrent)`. Or maybe it would be simpler to add a `activeCount` and use/manipulate that instead.

This is probably relatively uncommon—`flatten:` didn't seem to be widely used. But it could lead to some bad behavior.

mdiep:
@baconpat Are you still interested in pursuing this?

baconpat:
I'd still like to try to get these changes in but I haven't had a chance to address your concerns. I'll try to get to it soon.

mdiep:
The Objective-C ReactiveCocoa APIs have been moved to https://github.com/ReactiveCocoa/ReactiveObjC/. If you're interested in continuing this PR, please reopen it against that repository. Thanks!

PR #2979, opened 2016-06-12
[WIP] ReactiveSwift target

eimantas:
The "ReactiveSwift" target compiles nicely, but I'm still working on getting other targets depend on it. For now I can't get `rac_valuesForKeyPath:` working when running unit tests for "ReactiveCocoa-Mac" target.


mdiep:
Thanks for working on this!

I'd really like to do some small steps towards this in `master` first.

* Create separate directories on disk for `ReactiveObjcBridge` and `ReactiveCocoa`. (These can be done separately)

* Add framework targets for `ReactiveObjC`, `ReactiveObjCBridge`, and `ReactiveSwift`. (These can be done separately)

* Make the new `ReactiveCocoa` not depend on `ReactiveObjC`.

Once those things are done, we can think about living on the RAC5 branch.

So it'd be ✨if you could:

1. Retarget against `master`
2. Not make `ReactiveCocoa` depend on `ReactiveSwift`—just include that code in both frameworks.

PR #2984, opened 2016-06-15
Adopt `inout` parameter in `Atomic.modify`.

andersio:
~~Both~~ `modify` ~~and `withValue`~~ now expects an action closure that takes an inout parameter. ~~In other words, `withValue` can now mutate the value.~~

The motivation is to take advantage of the potential `inout` optimisation. Ideally with the optimisation in place, ~~`withValue` would enable in-place mutation with no copying at all, while~~ `modify` always makes a copy of the old value before any in-place mutation happens.

But this would likely be a trivial optimisation to the ReactiveSwift internals, since the growing data structures of most of the value type internal states are already CoW due to the use of stdlib collections.

-

Edit: Reduced the scope. Will create follow-up PRs if this is accepted.

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -20,7 +20,7 @@ public final class Atomic<Value> {
 		}
 	
 		set(newValue) {
-			modify { _ in newValue }
+			swap(newValue)
 		}
 	}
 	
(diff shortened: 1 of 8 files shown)


RuiAAPeres:
I am bit reluctant on merging this, without measuring it. @andersio could you run some benchmarks on this?

andersio:
@RuiAAPeres Here are the results using `Atomic<Int>`.

| Iterations | Pass-by-value | `inout` |
| --- | --- | --- |
| 10000 | 0.041 sec (60%) | 0.005 sec (52%) |
| 500000 | 0.540 sec (11%) | 0.207 sec (7%) |
| 5000000 | 4.251 sec (8%) | 2.101 sec (3%) |

https://gist.github.com/andersio/e586c3d979a296e312f2be3dd07ccd54

RuiAAPeres:
LGTM and thanks for the measurements. 👍

PR #2987, opened 2016-06-15
Update then() to forward interruptions on the original producer

sharplet:
The `then()` operator is documented as waiting until completion of the first producer, before subscribing to the subsequent producer. The behaviour for Interrupted events was not explicitly documented, and because of the semantics of `concat()` was being treated the same was a Completed event.

This behaviour caught me by surprise! For example, I was scratching my head for some time as to why this didn't behave the way I expected:

```swift
// Upon completion of the upload, sends the NSURL and then completes.
func uploadImage(data: NSData, to url: NSURL) -> SignalProducer<NSURL, NSError> {
  let upload = UploadRequest(data: data, type: "image/jpeg", url: NSURL)

  // Even if `api.perform(upload)` is interrupted internally, the URL is passed
  // through anyway and the producer completes!
  return api.perform(upload).then(SignalProducer(value: url))
}
```

I'm proposing this change to make the semantics of Interrupted events explicit, and hopefully it's in the spirit of the way `then()` is intended to be used. Would love to hear if you think this is a worthwhile change!

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1331,12 +1331,12 @@ extension SignalProducerType {
 	}
 
 	/// Waits for completion of `producer`, *then* forwards all events from
-	/// `replacement`. Any failure sent from `producer` is forwarded immediately, in
-	/// which case `replacement` will not be started, and none of its events will be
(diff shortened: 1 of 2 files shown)


NachoSoto:
Makes sense, this is much better :+1:

mdiep:
👏🏻

PR #2989, opened 2016-06-16
Introduce `withMutableValue`.

andersio:
The change is addictive. It simply adds `Atomic.withMutableValue(_:)` and `MutableProperty.withMutableValue(_:)`, and applies them in places where appropriate. The main difference from `modify(_:)` is that it **does not** copy and return the original value, and it was able to replace many previous uses of `modify` in the ReactiveSwift codebase that discards the returned value.

The benchmark this time uses an `Atomic` containing a struct of **eight** `Int`s to amplify the effect.

[Benchmark Source](https://gist.github.com/andersio/1c7a08b77075acc188ac3f62a13ee8fc)

##### Reading: pass-by-value (withValue) vs `inout` (withMutableValue)
| Iterations | Pass-by-value | Using `inout` |
| ---------- |  ------- | ----------- |
| 10000 | 0.004 sec<sup>(14% SD)</sup> | 0.003 sec<sup>(21% SD)</sup> |
| 500000 | 0.184 sec<sup>(6% SD)</sup> | 0.120 sec<sup>(8% SD)</sup> |
| 5000000 | 1.864 sec<sup>(2% SD)</sup> | 1.183 sec<sup>(1% SD)</sup> |

##### Mutating: `modify` vs `withMutableValue` (only for cases that the old value is discardable)
| Iterations | `modify` | `withMutableValue ` |
| ---------- |  ------- | ----------- |
| 10000 | 0.004 sec<sup>(10% SD)</sup> | 0.002 sec<sup>(21% SD)</sup> |
| 500000 | 0.194 sec<sup>(22% SD)</sup> | 0.106 sec<sup>(8% SD)</sup> |
| 5000000 | 1.752 sec<sup>(2% SD)</sup> | 1.096 sec<sup>(4% SD)</sup> |

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -20,7 +20,7 @@ public final class Atomic<Value> {
 		}
 	
 		set(newValue) {
-			swap(newValue)
+			withMutableValue { $0 = newValue }
 		}
 	}
 	
@@ -57,20 +57,31 @@ public final class Atomic<Value> {
 	///
 	/// Returns the old value.
 	public func modify(@noescape action: (inout Value) throws -> Void) rethrows -> Value {
-		return try withValue { value in
-			try action(&_value)
-			return value
+		return try withMutableValue { value in
+			let oldValue = value
+			try action(&value)
+			return oldValue
 		}
 	}
-	
+
+	/// Atomically performs an arbitrary action using an in-out reference to the current value
+	/// of the variable.
+	///
+	/// Returns the result of the action.
+	public func withMutableValue<Result>(@noescape action: (inout Value) throws -> Result) rethrows -> Result {
+		lock()
+		defer { unlock() }
+
+		return try action(&_value)
+	}
+
 	/// Atomically performs an arbitrary action using the current value of the
 	/// variable.
 	///
 	/// Returns the result of the action.
 	public func withValue<Result>(@noescape action: (Value) throws -> Result) rethrows -> Result {
-		lock()
-		defer { unlock() }
-
-		return try action(_value)
+		return try withMutableValue {
+			return try action($0)
+		}
 	}
 }
--- a/ReactiveCocoa/Swift/Disposable.swift
+++ b/ReactiveCocoa/Swift/Disposable.swift
@@ -77,7 +77,7 @@ public final class CompositeDisposable: Disposable {
 		/// that are no longer needed.
 		public func remove() {
 			if let token = bagToken.swap(nil) {
-				disposable?.disposables.modify { bag in
+				disposable?.disposables.withMutableValue { bag in
 					bag?.removeValueForToken(token)
 				}
 			}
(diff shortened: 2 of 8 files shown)


RuiAAPeres:
I had the feeling this work was already done in #2984.

RuiAAPeres:
Disregard my previous comment. Looking at the previous PR #2984:

screen shot 2016-06-15 at 23 00 06

It actually was wrong, since you are returning the new value. Or am I missing something?

andersio:
@RuiAAPeres In the previous PR, `withValue` does not use `inout` parameters. So the `value` there was a copy.

RuiAAPeres:
@andersio 👍

andersio:
Oops, I almost forgot `MutableProperty`.

andersio:
It seems due to the `copy-in-copy-out` semantic, it is impossible of `MutableProperty.withValue` to take advantage of `inout`.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -58,19 +58,20 @@ public final class Atomic<Value> {
...
 		}
 	}
-	
+
 	/// Atomically performs an arbitrary action using the current value of the
 	/// variable.
 	///
 	/// Returns the result of the action.
-	public func withValue<Result>(@noescape action: (Value) throws -> Result) rethrows -> Result {
+	public func withValue<Result>(@noescape action: (inout Value) throws -> Result) rethrows -> Result {
```
mdiep: With this change, actions sent to `withValue` can modify the value. That's a big semantic change, and it's not one that I think makes sense. The point of `withValue` is that it's readonly.

andersio:
True while being addictive at call site. Maybe it should be done as a variant of `modify` that does not copy and return the old value?

andersio:
@mdiep True. I have restored `withValue`, and make the inout variant `withMutableValue` instead.

RuiAAPeres:
@andersio looking at your testing cases, I see that when you use `withMutableValue`, you return something in the callback, for example:

```swift
+				let result: Bool = property.withMutableValue {
+					$0 = subsequentPropertyValue
+					return $0.isEmpty
+				}
```

Which is fine, so you mutate the value and you do something with it. But in some cases, where you replaced `modify` with `withMutableValue`, you don't do that, for example: 

```swift
 				innerProducer.startWithSignal { innerSignal, innerDisposable in
 -					state.modify { state in
 +					state.withMutableValue { state in
  						// When we replace the disposable below, this prevents the
  						// generated Interrupted event from doing any work.
  						state.replacingInnerSignal = true
  					}
```

Could you walk me through your reasoning?

andersio:
@RuiAAPeres These cases can be as simple as atomic in-place mutation of the state. In other words, they need neither a copy of the original state, nor propagation of results (say a disposable or a token).

But it just happens that we have a type inferrer that can infer `Void` as a return type (and it is also the cheapest type in Swift IIRC), so that `withMutableValue` can cover all cases that **do not** need the previous value copied and returned.

mdiep:
I don't think we should proceed with this change.

While this would be faster, it complicates the API. I dislike the additional semantics and the increased surface area. And while the benchmarks show that this is faster than `withValue`, I doubt that this will make a measurable impact in real-world RAC usage.

andersio:
@mdiep What about changing the semantics of `modify` instead? Then those clients explicitly needs the previous value can copy and return on their own.

PR #2991, opened 2016-06-17
Fix TestScheduler not having correct `currentDate` while advancing

inamiy:
Bug repro is well-illustrated in @bricklife's tweets (thanks!):

-  [RAC code](https://twitter.com/ooba/status/741099929427312640) (bad)
-  [RxSwift code](https://twitter.com/ooba/status/741100653351604225) (good)

We can normally work around this RAC issue by calling `testScheduler.advanceByInterval(delta)` multiple times in test codes, but many people would get stuck with this behavior (today's me :sob:),
so I would like to get this PR applied.


mdiep:
Thanks for the fix! ✨

PR #2992, opened 2016-06-17
Add ObserverType

natestedman:
This adds a protocol version of `Observer`, to be used for type-constrained extensions. Currently, Swift (2, at least) will not permit creating an `extension Observer where Value == NSError`, for example.

### DIFF
--- a/ReactiveCocoa/Swift/Observer.swift
+++ b/ReactiveCocoa/Swift/Observer.swift
@@ -6,6 +6,24 @@
 //  Copyright © 2015 GitHub. All rights reserved.
 //
 
+/// A protocol for type-constrained extensions of `Observer`.
+public protocol ObserverType {
+	associatedtype Value
+	associatedtype Error: ErrorType
+
+	/// Puts a `Next` event into the given observer.
+	func sendNext(value: Value)
+
+	/// Puts a `Failed` event into the given observer.
+	func sendFailed(error: Error)
+
+	/// Puts a `Completed` event into the given observer.
+	func sendCompleted()
+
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/Swift/Observer.swift
```diff
@@ -55,3 +55,21 @@ public struct Observer<Value, Error: ErrorType> {
 		action(.Interrupted)
 	}
 }
+
+/// A protocol for type-constrained extensions of `Observer`.
+public protocol ObserverType {
```
NachoSoto: Maybe put this before the actual type?

natestedman: Sure thing, just copied the ordering from `Signal.swift` and `SignalProducer.swift`.

RuiAAPeres:
For consistency sake, follow the same approach used in `SignalProducer` and `Signal`, make the type adopt the protocol via an extension (example [here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/Swift/SignalProducer.swift#L334) and [here](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/Swift/Signal.swift#L171)).

natestedman:
I noticed that "a"/"an" is used incorrectly in documentation (I assume `Failed` probably used to be `Error`?) and fixed that in the `ObserverType` comments, then added an additional commit to fix it in the `Observer` documentation. It's not strictly related to adding a backing protocol, so I can split that out into a different PR if desired.

RuiAAPeres:
👍

PR #3001, opened 2016-06-24
[5.0] Swift 3.0 Groundwork

andersio:
### Scope
1. General [Swift API Guidelines](https://swift.org/documentation/api-design-guidelines/) compliance.
Since the discussion of naming conventions on swift-evolution is still on-going, any renaming of operators (including `-ing` and `-ed`) and closure argument labelling would not happen in this PR, as suggested by various contributors on the channel.

2. Make it work under Xcode 8 and Swift 3.

Renaming of any operators and new features would be out of scope for this PR.

---
### List of Issues
1. `RACCommand.toAction` is  @temporarily gone, and replaced by a free function ~~`toAction`~~ `bridgedAction(from:)`. (still crashing in Xcode 8 beta 2)
2. ~~Some tests are disabled.~~
3. iOS tests cannot pass in Travis but locally. Could be due to Xcode.

---

_Last Updated: 30 June 2016_

PR #3002, opened 2016-06-24
Functional(ish) unit tests

liscio:
I've further added onto @andersio's initial swift3 work by fixing up Quick and Nimble so that they run in Xcode 8 / Swift 3 (see https://github.com/Quick/Quick/issues/551#issuecomment-228417715), and then I went on to get RAC's tests compiling and even running (and passing on the Mac target!)

There were a few crashes when running against my iOS simulator, and I nailed down an issue found by the Thread Sanitizer in the test code. Unfortunately, now Xcode 8 won't compile my iOS code at all anymore, segfaulting on everything I throw at it. 

Seems like a good place to stop for today!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "Carthage/Checkouts/Nimble"]
 	path = Carthage/Checkouts/Nimble
-	url = https://github.com/norio-nomura/Nimble.git
+	url = https://github.com/liscio/Nimble.git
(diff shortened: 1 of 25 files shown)


andersio:
Would you mind to rebase onto `RAC5-swift3`?
https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3001/commits/521a73144053ad8a3d6013008ef96398733ba604 is the squashed commit for everything that was previously in my fork.

🎉

PR #2995, opened 2016-06-18
Xcode markup documentation

tomj:
So @eimantas has cranked out an _insane_ amount of documentation over the past couple of days.  I've been proofreading and cleaning up the commit history a little but I think it looks pretty good.

I've added what has been done for `Scheduler.swift` to open up the discussion on this PR.  Given the volume of changes which have been made (and still need to be proofread/polished) we thought it would be good to hear what everyone thought before making a whopper of a PR.  Any feedback appreciated 😄 🎉

PR #3005, opened 2016-06-25
Support Swift 2.3

ikesyo:
~~This currently requires a weird workaround noted in https://github.com/Quick/Nimble/pull/302#issue-160088167.~~ The underlying issue is fixed in Xcode 8 beta 2.


.travis.yml
```diff
@@ -10,23 +10,55 @@ script:
   - script/build
 xcode_workspace: ReactiveCocoa.xcworkspace
 matrix:
+  allow_failures:
+    - xcode_scheme: ReactiveCocoa-iOS
+      osx_image: xcode8
```
mdiep: Why is iOS allowed to fail?

mdiep:
I'm very excited to see this. ⚡ ⚡ ⚡

.travis.yml
ikesyo: Because the iOS test on Xcode 8 is now randomly failing as seen in https://github.com/Quick/Nimble/pull/302#issuecomment-226390685 and https://github.com/Quick/Quick/pull/547.

andersio:
🎉 

The myth-busting PR on why I can never get green iOS checks from Travis on the Swift 3.0 branch.

.travis.yml
mdiep: It seems like it's consistently failing? 😕 Is this just a CI issue or does it fail locally too?

ikesyo: I believe this is a CI issue and it works well locally.

andersio: Same for the Swift 3 branch. iOS passes tests locally, but cannot run at all on Travis. Compilation is successful though - it just fails right after the test executable is signed.

mdiep:
I think this should specify Swift 2.3 in the Xcode file and set the Xcode upgrade check to 8.0.

.travis.yml
mdiep: The iOS tests seem to hang locally for me. 😕

andersio: You may check the scheme if the thread sanitiser has been turned on.

ikesyo:
> this should specify Swift 2.3 in the Xcode file

This was done in cfdf1e9.

> set the Xcode upgrade check to 8.0

Done in 7a88c90.

ikesyo:
It may be better to wait for the next Xcode 8 beta?

.travis.yml
ikesyo: Remove `allow_failures` in cc9a151.

lukaskubanek:
@ikesyo @mdiep Is the support for Swift 2.3 going to be released as a 4.X release before the 5.0 with Swift 3.0?

andersio:
@lukaskubanek I don't see why it wouldn't. ;)

mdiep:
Yes!

NachoSoto:
Awesome! Is this ready now then? :)

ikesyo:
The linker error does not happen on Xcode 8 beta 2, so I remove the workaround. 💥 Let's wait Travis to support beta 2.

ikesyo:
Finally Xcode 8 beta 2 has come on Travis and the CI builds are all green now. ✅ 

https://twitter.com/travisci/status/752704816116293632 🎉

andersio:
The iOS test is still shaky. #3057 initially passed, but a rerun failed.

ikesyo:
This should be good to go this time. The new `build-for-testing` and `test-without-building` build actions did the trick.

andersio:
Not sure if we should really include the Carthage test though... It would double the build time for all subsequent PRs.

mdiep:
Thanks for all the work on this, @ikesyo!

I want to see if we can get #2995 merged in the next week, but then we can release the (hopefully) last RAC 4.x release and start working on RAC 5 on `master`!

PR #3007, opened 2016-06-26
Making Travis CI work (Swift 3)

andersio:
~~Just in case, please do not merge.~~ I am looking into what's going on with the silent timeout in Travis' tests, since I didn't manage to reproduce locally...

Edit: Caught two misbehaving test cases.
:tada:


andersio:
All green checks, except for iOS that somehow failed to run any tests after code signing the test executable in Xcode 8 (as in #3005).
:tada:

PR #3011, opened 2016-06-28
Deprecated startWithNext/observeNext and exposed only for Error == NoError. Added startWithResult/observeResult

### LINKED ISSUE #3010: Expose startWithNext/observeNext only for Error == NoError
I was thinking about how easy it is to do something like:
```swift
let signal: Signal<Int, SomeError>

signal.observeNext { print($0) }
```

And totally ignore errors. `RxJava` deals with this by throwing an exception at _runtime_ if this happens, but we can do better! We could detect this at _compile time_ if `observeNext`/`startWithNext` were not exposed for any `Signal`/`SignalProducer` that has an error other than `NoError`.

Obviously we can't know what users are doing inside of `observe`/`start`, but at least we can limit the damage.

I was going to suggest having `startWithNext:failed:`/`observeNext:failed:` in the case that the error **is not** `NoError`, but unfortunately that's not doable. I filed [`rdar://23432210`](http://www.openradar.me/23432210) a while ago asking for something that would support this though.
We could always add this overload though, it's just silly for signals with `NoError`.

So to summarize, I'm asking that we _hide_ `startWithNext`/`observeNext` behind:
```swift
extension SignalType where Error == NoError
extension SignalProducerType where Error == NoError
```

NachoSoto:
Fixes #3010 

screen shot 2016-06-28 at 12 47 31

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -181,9 +181,9 @@ extension SignalType {
 		return observe(Observer(action))
 	}
 
-	@available(*, unavailable, message="This Signal may emit errors which must be handled explicitly, or observed using observeResult:")
+	@available(*, deprecated, message="This Signal may emit errors which must be handled explicitly, or observed using observeResult:")
 	public func observeNext(next: Value -> Void) -> Disposable? {
-		fatalError("Unavailable method")
+		return observe(Observer(next: next))
 	}
 
 	/// Observes the Signal by invoking the given callback when
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -181,6 +181,21 @@ extension SignalType {
 		return observe(Observer(action))
 	}
 
+	/// Observes the Signal by invoking the given callback when
+	/// `next` or `failed` event are received.
+	///
+	/// Returns a Disposable which can be used to stop the invocation of the
+	/// callback. Disposing of the Disposable will have no effect on the Signal
+	/// itself.
+	public func observeResult(result: (Result<Value, Error>) -> Void) -> Disposable? {
+		return observe(
+			Observer(
+				next: { result(.Success($0)) },
+				failed: { result(.Failure($0)) }
+			)
+		)
+	}
+
 	/// Observes the Signal by invoking the given callback when a `completed` event is
 	/// received.
 	///
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -181,6 +181,11 @@ extension SignalType {
 		return observe(Observer(action))
 	}
 
+	@available(*, unavailable, message="This Signal may emit errors which must be handled explicitly, or observed using observeResult:")
+	public func observeNext(next: Value -> Void) -> Disposable? {
+		fatalError("Unavailable method")
+	}
+
 	/// Observes the Signal by invoking the given callback when
 	/// `next` or `failed` event are received.
 	///
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -181,16 +181,6 @@ extension SignalType {
 		return observe(Observer(action))
 	}
 
-	/// Observes the Signal by invoking the given callback when `next` events are
-	/// received.
-	///
-	/// Returns a Disposable which can be used to stop the invocation of the
-	/// callbacks. Disposing of the Disposable will have no effect on the Signal
-	/// itself.
-	public func observeNext(next: Value -> Void) -> Disposable? {
-		return observe(Observer(next: next))
-	}
-
 	/// Observes the Signal by invoking the given callback when a `completed` event is
 	/// received.
 	///
@@ -221,7 +211,21 @@ extension SignalType {
 	public func observeInterrupted(interrupted: () -> Void) -> Disposable? {
 		return observe(Observer(interrupted: interrupted))
 	}
+}
 
+extension SignalType where Error == NoError {
+	/// Observes the Signal by invoking the given callback when `next` events are
+	/// received.
+	///
+	/// Returns a Disposable which can be used to stop the invocation of the
+	/// callbacks. Disposing of the Disposable will have no effect on the Signal
+	/// itself.
+	public func observeNext(next: Value -> Void) -> Disposable? {
+		return observe(Observer(next: next))
+	}
+}
+
+extension SignalType {
 	/// Maps each value in the signal to a new value.
 	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
 	public func map<U>(transform: Value -> U) -> Signal<U, Error> {
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -361,9 +361,9 @@ extension SignalProducerType {
 		return start(Observer(observerAction))
 	}
 
-	@available(*, unavailable, message="This SignalProducer may emit errors which must be handled explicitly, or observed using startWithResult:")
+	@available(*, deprecated, message="This SignalProducer may emit errors which must be handled explicitly, or observed using startWithResult:")
 	public func startWithNext(next: Value -> Void) -> Disposable {
-		fatalError("Unavailable method")
+		return start(Observer(next: next))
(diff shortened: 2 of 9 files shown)


ReactiveCocoaTests/Swift/FlattenSpec.swift
```diff
@@ -107,6 +107,7 @@ class FlattenSpec: QuickSpec {
 				var observed: Int? = nil
 				outer
 					.flatten(.Latest)
+					.assumeNoErrors()
```
JaviSoto: I love this change, because it forces the code to explicitly yell "I'm ignoring errors".

ReactiveCocoaTests/Swift/TestError.swift
```diff
@@ -14,3 +17,27 @@ enum TestError: Int {
...
+	func assumeNoErrors() -> SignalProducer<Value, NoError> {
+		return self.lift { $0.assumeNoErrors() }
+	}
+}
+
+internal extension SignalType {
+	/// Halts if an error is emitted in the receiver signal.
+	/// This is useful in tests to be able to just use `startWithNext`
+	/// in cases where we know that an error won't be emitted.
+	func assumeNoErrors() -> Signal<Value, NoError> {
```
JaviSoto: I think it would be useful to expose this function publicly?

JaviSoto:
Love it. 
tumblr_inline_mld3bl7okh1qz4rgp

ReactiveCocoaTests/Swift/TestError.swift
NachoSoto: I'd be okay with it :) I have it on every single personal project :P

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -360,14 +360,24 @@ extension SignalProducerType {
...
+
+	/// Creates a Signal from the producer, then adds an observer to
+	/// the Signal, which will invoke the given callback when `next` or `failed` 
+	/// events are received.
 	///
 	/// Returns a Disposable which can be used to interrupt the work associated
 	/// with the Signal, and prevent any future callbacks from being invoked.
-	public func startWithNext(next: Value -> Void) -> Disposable {
-		return start(Observer(next: next))
+	public func startWithResult(result: Result<Value, Error> -> Void) -> Disposable {
```
sharplet: I feel like using the singular "result" here could be a little misleading. It almost sounds too final, like there will only be one. What about `.startWithResults {}`? I do think this could use a little bikeshedding, though.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -360,14 +360,24 @@ extension SignalProducerType {
 		return start(Observer(observerAction))
 	}
 
-	/// Creates a Signal from the producer, then adds exactly one observer to
-	/// the Signal, which will invoke the given callback when `next` events are
-	/// received.
+	@available(*, deprecated=1.0, message="This SignalProducer may emit errors which must be handled explicitly, or observed using startWithResult:")
```
sharplet: If this is going to be deprecated, shouldn't it call through to a working implementation? Otherwise, I feel like `unavailable` would be more appropriate, because that would at least generate an error at compile time instead of runtime.

NachoSoto: Ooooh that's what I meant to use, thanks!

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: For sure. I matched `startWithNext`, but I agree this is misleading. I like `startWithResults` too.

Thoughts anyone else?

ReactiveCocoaTests/Swift/TestError.swift
sharplet: I have a similar operator in my project that actually accepts the failure message as a parameter:

```swift
foobar.suppressErrors(failingWith: "unexpected error in foobar")
```

The thing I like about this is that it makes the operator harder to use without thinking about the reason why you're suppressing errors. If this method became public, I think that could be a useful affordance.

ReactiveCocoa/Swift/SignalProducer.swift
NachoSoto: Done.

ReactiveCocoaTests/Swift/TestError.swift
NachoSoto: This doesn't _suppress_ them though: I feel strongly we shouldn't expose such operator. This **assumes** that none will be emitted, resulting in a runtime failure if they are.

sharplet: Yeah, I do like how this version neatly captures the assumption that there will be no errors.

I do wonder if there is a form that makes it _slightly_ less easy to introduce crashes—but on the other hand if you're going to misuse the API there's not much we can do to prevent it, other than not exposing the operator publicly.

NachoSoto: Right, that's why I just wanted to keep it private for now.

NachoSoto:
Failing test is unrelated

ReactiveCocoa/Swift/SignalProducer.swift
RuiAAPeres: It's difficult for me to have an unbiased opinion, since I am looking at the implementation, but yeah `results` might be clear.

NachoSoto: What do you think about leaving this as `startWithResult` for consistency, and leaving this discussion for #3013?

sharplet: Yeah I'm happy with that approach.

NachoSoto:
This is ready then :)

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -181,14 +181,24 @@ extension SignalType {
 		return observe(Observer(action))
 	}
 
-	/// Observes the Signal by invoking the given callback when `next` events are
-	/// received.
+	@available(*, unavailable, message="This Signal may emit errors which must be handled explicitly, or observed using observeResult:")
```
mdiep: I think this and `startWithNext` should be marked as `deprecated`, not `unavailable`, since this is targeting RAC 4.

Also, the message should probably say `observeResult()`, not `observeResult:`. ☺️

mdiep:
❤️ it. But we need to deprecate this properly.

ReactiveCocoa/Swift/Signal.swift
NachoSoto: I think unavailable is correct, this is actually sort of a bug fix. If this breaks a client... Then good! We helped someone find a potential bug :) it's not like we're not allowing them from using the method anymore, they just need to start observing errors or handling them.

Does that make sense?

mdiep: The issue is that it breaks semantic versioning. That's liable to cause issues for people.

Calling it a bug fix is a bit of a stretch. ☺️ I agree that the method is unsafe, but we've let people use it, so we can't take it away without deprecating it first.

JaviSoto: Marking it as deprecated makes more sense, but remember that Swift doesn't let you call deprecated APIs, so in practice it's the same hehe

NachoSoto: Done!

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -313,6 +313,39 @@ class SignalSpec: QuickSpec {
 				observer.sendNext(1)
 				expect(values) == [1]
 			}
+
+			//  TODO: remove when the method is marked unavailable
+			it("receives next values with erroring signal") {
```
NachoSoto: I added 2 tests (one for `SignalProducer` and one for `Signal`) to make sure that we don't break compatibility for now.

NachoSoto:
Failing tests are unrelated.

ReactiveCocoaTests/Swift/SignalSpec.swift
mdiep: 💖

ReactiveCocoa/Swift/Signal.swift
mdiep: > Swift doesn't let you call deprecated APIs

ORLY?

PR #3015, opened 2016-06-30
Deprecate SignalProducer.buffer()

mdiep:
cf. #2831

I want to get better at deprecating these things to make upgrading easier.


NachoSoto:
Awesome :shipit:

eimantas:
There should probably be "use this instead:" note.

NachoSoto:
@eimantas it says that on the message :)

eimantas:
Right. Let me get my coffee...

PR #3016, opened 2016-06-30
[5.0] Property = ConstantProperty + AnyProperty

andersio:
~~Well, a self-explanatory title.~~

Both `AnyProperty` and `ConstantProperty` are read-only entities. This PR proposes to unify them into a single type `Property`, and constant properties can be created using `Property(value:)` instead.

###### Before
```
		name = AnyProperty(ConstantProperty("undefined"))
		postingDate = AnyProperty(ConstantProperty(Date.distantFuture))
```

###### After
```
		name = Property(value: "undefined")
		postingDate = Property(value: Date.distantFuture)
```

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -32,14 +32,14 @@ public final class Action<Input, Output, Error: ErrorProtocol> {
 
 	/// Whether the action is currently executing.
 	public var isExecuting: Property<Bool> {
-		return Property(reflecting: _isExecuting)
+		return Property(_isExecuting)
 	}
 
 	private let _isExecuting: MutableProperty<Bool> = MutableProperty(false)
 
 	/// Whether the action is currently enabled.
 	public var isEnabled: Property<Bool> {
-		return Property(reflecting: _isEnabled)
+		return Property(_isEnabled)
 	}
 
 	private let _isEnabled: MutableProperty<Bool> = MutableProperty(false)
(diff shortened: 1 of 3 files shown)


NachoSoto:
Hmm can you elaborate what the goal for this is?

andersio:
@NachoSoto There was a discussion on the Slack channel concerning `AnyProperty`. It started with a suggestion to rename it to `ImmutableProperty`, then `ReadOnlyProperty` and eventually a consensus on just `Property`, considering how the standard library has named immutable variants against their mutable counterparts. Such renaming is expected to encourage the usage of `Property` for composition and encapsulation.

As for `ConstantProperty`, since `Property` is semantically read only, while having the lifetime of its producer and its signals tying to the source itself, a merge of `ConstantProperty` into `Property` was suggested. It could reduce boilerplates like `AnyProperty(ConstantProperty(defaultValue))` into `Property(defaultValue)`.

andersio:
Hmm... the constant initialiser is now made `Property(constant:)` due to the potential ambiguity when using with other properties. Moreover, type erasing wrapper is considered a full width conversion anyway, so let's stick to that convention.

dmcrodrigues:
@NachoSoto since ReactiveCocoa is going through a process of rename a few things to be more aligned with Swift conventions I started a discussion around `AnyProperty` name which I personally think can be confusing about the intent and behaviour of the property, a read-only property (independently of how has been initialized). `MutableProperty` for example is much more clearer about his intent. 

I think renaming it, and merging `ConstantProperty` in it, makes ReactiveCocoa's types more consistent and meaningful, you work with properties which are not mutable by default unless you say so, and that way they become directly related to each other which I think is not so obvious today with `AnyProperty`, `ConstantProperty` and `MutableProperty`.

What do you think?

PR #3018, opened 2016-06-30
Correctly teardown MutableProperty binding when it is disposed.

azeff:
Fixed memory leak when dispose binding of MutablePropertyType to a SignalProducer.
Added unit tests and updated existing ones that were without assertions. Unit tests required some way to know if Signal has any observers. After binding is disposed signal from MutableProperty should not have observers. That is why I've added internal property to Signal.

### DIFF
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -240,14 +240,16 @@ public func <~ <P: MutablePropertyType>(property: P, signal: Signal<P.Value, NoE
 public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.Value, NoError>) -> Disposable {
 	let disposable = CompositeDisposable()
 
-	producer.startWithSignal { signal, signalDisposable in
-		property <~ signal
-		disposable += signalDisposable
+	producer
+		.on(completed: { disposable.dispose() })
+		.startWithSignal { signal, signalDisposable in
+			disposable += property <~ signal
+			disposable += signalDisposable
 
-		disposable += property.producer.startWithCompleted {
-			disposable.dispose()
+			disposable += property.producer.startWithCompleted {
+				disposable.dispose()
+			}
 		}
-	}
 
 	return disposable
 }
--- a/ReactiveCocoaTests/Swift/PropertySpec.swift
+++ b/ReactiveCocoaTests/Swift/PropertySpec.swift
@@ -599,24 +599,26 @@ class PropertySpec: QuickSpec {
 				}
 
 				it("should tear down the binding when disposed") {
-					let signalValues = [initialPropertyValue, subsequentPropertyValue]
-					let signalProducer = SignalProducer<String, NoError>(values: signalValues)
+					let (signalProducer, observer) = SignalProducer<String, NoError>.pipe()
 
 					let mutableProperty = MutableProperty(initialPropertyValue)
 					let disposable = mutableProperty <~ signalProducer
 
 					disposable.dispose()
-					// TODO: Assert binding was torn down?
+
+					observer.sendNext(subsequentPropertyValue)
+					expect(mutableProperty.value) == initialPropertyValue
 				}
 
 				it("should tear down the binding when bound signal is completed") {
 					let (signalProducer, observer) = SignalProducer<String, NoError>.pipe()
 
 					let mutableProperty = MutableProperty(initialPropertyValue)
-					mutableProperty <~ signalProducer
+					let disposable = mutableProperty <~ signalProducer
 
 					observer.sendCompleted()
-					// TODO: Assert binding was torn down?
+
+					expect(disposable.disposed) == true
 				}
 
 				it("should tear down the binding when the property deallocates") {


ReactiveCocoa/Swift/Signal.swift
```diff
@@ -154,6 +154,15 @@ public final class Signal<Value, Error: ErrorType> {
 	}
 }
 
+extension Signal {
+	/// Number of observers observing the signal or `nil` if signal is terminated.
+	///
+	/// Property used for Unit Tests.
+	internal var observersCount: Int? {
```
NachoSoto: Noooooo. We certainly don't want this, as it would mean that we're testing implementation details 👎

ReactiveCocoa/Swift/Property.swift
```diff
@@ -241,12 +241,15 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
...
 
 	producer.startWithSignal { signal, signalDisposable in
-		property <~ signal
+		disposable += property <~ signal
 		disposable += signalDisposable
 
 		disposable += property.producer.startWithCompleted {
 			disposable.dispose()
 		}
+		disposable += signal.observeCompleted {
```
NachoSoto: Good catch

ReactiveCocoa/Swift/Property.swift
```diff
@@ -241,12 +241,15 @@ public func <~ <P: MutablePropertyType>(property: P, producer: SignalProducer<P.
 	let disposable = CompositeDisposable()
 
 	producer.startWithSignal { signal, signalDisposable in
-		property <~ signal
+		disposable += property <~ signal
```
NachoSoto: Maybe do:
```swift
producer
  .on(completed: { disposable.dispose() })
  .startWithSignal { ... }
```

That way you avoid the extra call to `observeCompleted` and needing to add that to the disposable?

NachoSoto:
Thanks for the PR! The fix looks correct, really good catch! But we need to find a different way to test this that doesn't expose implementation details.

ReactiveCocoa/Swift/Signal.swift
azeff: Maybe replace it with something like
```
internal var isObserved: Bool {
  return atomicObservers.value?.count > 0
}
```
It will be enough for the tests and exposes less details. What do you think?

NachoSoto: It's still exposing internal state.
I'd rather us test this by verifying that the subscriptions have terminated on every observer.

ReactiveCocoa/Swift/Property.swift
azeff: Sounds good.

azeff:
Test in `RACSignalSpec` has failed. Pretty sure it's not because of my changes.

azeff:
@NachoSoto after taking a second look at binding implementation I've decided not to reuse one for Signal binding. This way is simpler, easier to understand and deal with disposables.
I've also removed property from Signal and corresponding unit tests.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -240,12 +240,18 @@ public func <~ <P: MutablePropertyType>(property: P, signal: Signal<P.Value, NoE
...
-	producer.startWithSignal { signal, signalDisposable in
-		property <~ signal
-		disposable += signalDisposable
-
-		disposable += property.producer.startWithCompleted {
+	disposable += property.producer.startWithCompleted {
+		disposable.dispose()
+	}
+	
+	disposable += producer.start { [weak property] event in
```
NachoSoto: `property` should be `unowned`, to better reflect the ownership relationship, and so that we don't have to reason about what to do if it becomes `nil`.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -240,12 +240,18 @@ public func <~ <P: MutablePropertyType>(property: P, signal: Signal<P.Value, NoE
...
+		switch event {
+		case let .Next(value):
+			property?.value = value
```
NachoSoto: This is duplicated from the canonical `<~` now :(

ReactiveCocoa/Swift/Property.swift
azeff: I tend to use `weak` because it is much safer. With `unowned` we have to be absolutely sure closure is not executed after property is deallocated. I'm not ready to say may it happen or not.

ReactiveCocoa/Swift/Property.swift
azeff: Yeah :( But attempt to reuse canonical one does not look good either and there already was a PR that solved one leak. I think in this case it'd be better to have duplication of simple solution.
If you think otherwise I can revert to initial PR.

NachoSoto: Yeah I think the previous implementation was better.

azeff:
@NachoSoto update PR according to our discussion. Reverted to initial implementation and applied your comments.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -599,24 +599,26 @@ class PropertySpec: QuickSpec {
...
+					let (signalProducer, observer) = SignalProducer<String, NoError>.pipe()
 
 					let mutableProperty = MutableProperty(initialPropertyValue)
 					let disposable = mutableProperty <~ signalProducer
 
 					disposable.dispose()
-					// TODO: Assert binding was torn down?
+
+					observer.sendNext(subsequentPropertyValue)
+					expect(mutableProperty.value) == initialPropertyValue
```
NachoSoto: Awesome

NachoSoto:
Perfect! 🎉 I'll merge as soon as tests pass. Thanks a lot!

NachoSoto:
Failing test is unrelated :shipit:

azeff:
🎉 Thanks.

Issue #2926, opened 2016-05-23
Convert free functions to static functions


mdiep:
Should non-instance methods be free functions or static functions?

Because currently we have:

```swift
extension SignalType {

    /// Merges the given signals into a single `Signal` that will emit all values
    /// from each of them, and complete when all of them have completed.
    @warn_unused_result(message="Did you forget to call `observe` on the signal?")
    public static func merge<S : SequenceType where S.Generator.Element == Signal<Value, Error>>(signals: S) -> ReactiveCocoa.Signal<Self.Value, Self.Error>
}
```

But we also have:

```swift
/// Combines the values of all the given signals, in the manner described by
/// `combineLatestWith`.
@warn_unused_result(message="Did you forget to call `observe` on the signal?")
public func combineLatest<A, B, Error>(a: Signal<A, Error>, _ b: Signal<B, Error>) -> Signal.Signal<(A, B), Error>
```

We should standardize on one or the other for RAC 5.0.

I'd vote free functions, but I'm curious if someone thinks we should be using static functions. @ReactiveCocoa/reactivecocoa

JaviSoto:
Static functions are more autocomplete-friendly, so I would vote for that.
Thanks for bringing this up :) definitely agree we should unify for consistency!

mdiep:
A benefit of free functions is that you don't need to think about whether something is a `Signal` or a `SignalProducer`.

RuiAAPeres:
I agree with @JaviSoto. For people that know RAC well, free functions are not an issue, I would argue for  novice to mid level discoverability can be an issue.

andymatuschak:
The Swift API guidelines [are pretty clear on this](https://swift.org/documentation/api-design-guidelines/#general-conventions).

> Prefer methods and properties to free functions. Free functions are used only in special cases:
> 
> When there’s no obvious self:
> `min(x, y, z)`
>
> When the function is an unconstrained generic:
> `print(x)`
>
> When function syntax is part of the established domain notation:
> `sin(x)`

These functions are basically fancy constructors, so using the metatype as `self` seems plenty defensible.

mdiep:
This was done in #3001! 🎉

PR #3022, opened 2016-07-02
Swift 3.0 Groundwork

andersio:
Originally #3001. It now bases on `master` (4.x) without all the shiny features `RAC5` has.

-
A few special notes:

1. Implemented #2926.
    But `timer(_: onScheduler:withLeeway:)` is instead renamed to `SignalProducer<Date, NoError>(interval:on:leeway:)`. It is a compromise made for the type inferrer, since even if the static function does not use any associated types, the inferrer would still do its work. The original form is marked `@unavailable`.

7. `Action(enabledIf:_:)` is renamed to `Action(enabling:_:)`.
The rationale here is that `enabling` gives a sense of action (to enable something), versus the status represented by `enabled`. (Should we call it a `switch` instead? Or just revert it?)

4. Operators are still imperative verbs.
Since the discussion of naming conventions on swift-evolution is still on-going, we should perhaps delay stuffing/renaming _operators_ to a future PR, as suggested by various contributors on the channel.

4. Closure arguments aren't named yet.
Same situation as (4).


andersio:
Kind of inspired by #3024, I would suggest bringing back `sink` (or `inputObserver` if you prefer) as the parameter name for input observers, i.e. signal generator & producer start handler. This bugs me for a while when writing documentations, since `observer` in these contexts has a different role than the observers supplied to `start` or `observe`.

mdiep:
I'm not ready to accept these changes on `master`, since want to ship another 4.x release with Swift 2.2/2.3 support.

It would be better to target this against the RAC5 branch IMO since these changes are for RAC5. Otherwise we can discuss this here but wait to merge until the RAC5 branch has been merged.

andersio:
@mdiep Would you consider rebasing `RAC5` on top of this branch? Say at least we may provide a stopgap version/branch with no breaking changes, in case we do not release `RAC5` in line with Swift 3.

mdiep:
No, I plan to release RAC5 in line with Swift 3. (It necessarily needs to be a major version because it's a breaking change.)

I think reviving the old PR is a better idea.

PR #3023, opened 2016-07-02
Unifying read-only properties, and making `PropertyProtocol` class bound.

andersio:
Originally #3016. It now targets the `RAC5` branch.

-
Both `AnyProperty` and `ConstantProperty` are read-only entities. This PR proposes to unify them into a single type `Property`, so that constant properties can be created using `Property(value:)`.

###### Before
```
		name = AnyProperty(ConstantProperty("undefined"))
		postingDate = AnyProperty(ConstantProperty(Date.distantFuture))
```

###### After
```
		name = Property(value: "undefined")
		postingDate = Property(value: Date.distantFuture)
```


mdiep:
I think the intent of `ConstantProperty` was to encode the immutability in the type system—which this abandons. But I don't see why that's actually useful.

This change makes sense to me. @ReactiveCocoa/reactivecocoa: Any objections?

NachoSoto:
I'll take a look at this tomorrow.
I do think keeping `ConstantProperty` is useful though.
On Mon, Jul 4, 2016 at 6:34 PM Matt Diephouse <notifications@github.com>
wrote:

> I think the intent of ConstantProperty was to encode the immutability in
> the type system—which this abandons. But I don't see why that's actually
> useful.
>
> This change makes sense to me. @ReactiveCocoa/reactivecocoa
> <https://github.com/orgs/ReactiveCocoa/teams/reactivecocoa>: Any
> objections?
>
> —
> You are receiving this because you are on a team that was mentioned.
>
>
> Reply to this email directly, view it on GitHub
> <https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3023#issuecomment-230370858>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAp2Ke84sLWH_Kr7chCSKIRcjRD6l0fGks5qSbTCgaJpZM4JDrVz>
> .
>
-- 
Ignacio Soto

ReactiveCocoa/Swift/Property.swift
```diff
@@ -325,8 +325,27 @@ public func zip<S: SequenceType where S.Generator.Element: PropertyType>(propert
...
+/// # Composed property
+/// Created by either the compositional operators in `PropertyProtocol`, or
+/// `Property(initial:followingBy:)`, a composed property presents a
+/// composed view of its source, which can be a set of properties,
+/// a producer, or a signal.
+///
+/// A composed property respects the lifetime of its source rather than its own.
+/// In other words, its producer and signal can outlive the property itself, if
+/// its source outlives it too.
+public struct Property<Value>: PropertyType {
```
NachoSoto: I don't think this makes sense as a `struct`: given that it composes a `Signal`, it doesn't really have value semantics.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -430,33 +456,18 @@ public struct AnyProperty<Value>: PropertyType {
 		}
 	}
 
-	/// Check if `property` is an `AnyProperty` and has already captured its sources
+	/// Check if `property` is an `Property` and has already captured its sources
 	/// using a closure. Returns that closure if it does. Otherwise, returns a closure
 	/// which captures `property`.
 	private static func capture<P: PropertyType>(property: P) -> [Any] {
-		if let property = property as? AnyProperty<P.Value> {
+		if let property = property as? Property<P.Value> {
```
NachoSoto: Huh I think I missed when this happens: how come we're doing type erasure here rather than encoding this on the method signature?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -430,33 +456,18 @@ public struct AnyProperty<Value>: PropertyType {
...
 			return property.sources
 		} else {
 			return [property]
 		}
 	}
 }
 
-/// A property that never changes.
-public struct ConstantProperty<Value>: PropertyType {
```
NachoSoto: Ooooh I understand the reason to remove this now 👍

NachoSoto:
Cool this looks great, nice simplification!

A few comments:
- Can you create dummy `ConstantProperty` and `AnyProperty` with `@unavailable`, to point people to the new type?
- `AnyProperty` was previously a `struct`, but now that I think about it I don't think that's correct since it doesn't really have value semantics.
- I'm confused about the implementation of `capture` performing a runtime type check.

andersio:
> Can you create dummy ConstantProperty and AnyProperty with @unavailable, to point people to the new type?

Sure.

> AnyProperty was previously a struct, but now that I think about it I don't think that's correct since it doesn't really have value semantics.

I would argue that value types need not be of pure value semantics. Moreover, since we do not constrain `PropertyProtocol` with a class bound, it could still somehow wrap a value type. Making it `class` would probably be less confusing though.

> I'm confused about the implementation of capture performing a runtime type check.

Let's say we composed two `MutableProperty`s A and B, resulting in C. Then we compose C with another `MutableProperty` D resulting in E.

E ultimately sources from just A, B and D. So when E captures its sources, instead of capturing C, it can simply grab C's sources if C is also a composed property. This is important because if C is left unused, it allows C's `ScopedDisposable` to deinitialise, which in turn interrupts the observation on the producer. 

An extreme example would be `property.flatMap { $0 }.map { $0 }.map { $0 }.map { $0 }`. In this case, assuming that we just care about the end result, such design allows the three intermediate property to deinitialize cleanly.

NachoSoto:
> I would argue that value types need not be of pure value semantics.

Why not?

> Moreover, since we do not constrain PropertyProtocol with a class bound, it could still somehow wrap a value type.

But regardless, `Property` contains a `Signal`, which has reference semantics, making the fact that `Property` is a `struct` misleading.

andersio:
> Why not?
> But regardless, Property contains a Signal, which has reference semantics, making the fact that Property is a struct misleading.
Hmm... that makes sense, especially after we removed the only primitive value type property. I suppose we should give `PropertyType` as class bound too?

NachoSoto:
> I suppose we should give PropertyType as class bound too?

Yeah I think that'd be appropriate too.

ReactiveCocoa/Swift/Property.swift
mdiep: Oversight on my part in the review, I guess. 🙊

andersio: @mdiep See the rationale, or you may check the PR for this. This is crucial to the semantics you suggested. 🐵

andersio:
I have pushed a replacement to `Property.Type.capture` (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3023/commits/5e9aa2bae4fb6280d5889d3e357225675f05cdd5) to see if it makes sense to you guys.

andersio:
Oops, I guess with the #3042 fix it doesn't matter anymore. We can remove the capturing stuff.

andersio:
Rebased for Swift 3.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -17,17 +20,26 @@ public protocol PropertyProtocol {
...
-public protocol MutablePropertyProtocol: class, PropertyProtocol {
+public protocol MutablePropertyProtocol: PropertyProtocol {
 	/// The current value of the property.
 	var value: Value { get set }
 }
 
+extension PropertyProtocol {
+	/// The property sources of `self`, or `nil` if `self` owns the
+	/// lifetime of its signal and producer.
+	public var sources: [AnyObject]? {
```
mdiep: Why is there and why is it public?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -36,15 +48,15 @@ public protocol MutablePropertyProtocol: class, PropertyProtocol {
 /// A transformed property would retain its ultimate source, but not
 /// any intermediate property during the composition.
 extension PropertyProtocol {
-	/// Lifts a unary SignalProducer operator to operate upon PropertyProtocol instead.
-	private func lift<U>( _ transform: @noescape (SignalProducer<Value, NoError>) -> SignalProducer<U, NoError>) -> AnyProperty<U> {
-		return AnyProperty(self, transform: transform)
+	/// Lifts a unary SignalProducer operator to operate upon PropertyType instead.
```
mdiep: This should still be `PropertyProtocol`?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -350,9 +362,28 @@ extension PropertyProtocol {
...
+/// Created by either the compositional operators in `PropertyProtocol`, or
+/// `Property(initial:followingBy:)`, a composed property presents a
+/// composed view of its source, which can be a set of properties,
+/// a producer, or a signal.
+///
+/// A composed property respects the lifetime of its source rather than its own.
+/// In other words, its producer and signal can outlive the property itself, if
+/// its source outlives it too.
+public final class Property<Value>: PropertyProtocol {
+	public let sources: [AnyObject]?
```
mdiep: I don't really like this being `public`. It should be an implementation detail.

ReactiveCocoa/Swift/Property.swift
andersio: The `PropertyProtocol.sources` requirement in `PropertyProtocol` is to replace the `AnyProperty.capture` using runtime dynamic cast. Since the protocol is public, this have to be public too. Moreover, custom property types should be allowed to specify their sources.

Though whether or not a default implementation should be supplied is debatable.

mdiep: Can we go back to the runtime dynamic cast, at least for this PR? That seems orthogonal to "unifying read-only properties".

ReactiveCocoa/Swift/Property.swift
andersio: Making it non-public would require runtime casting, i.e. what is currently in `RAC5`, since we do not have private conformance in Swift.

ReactiveCocoa/Swift/Property.swift
andersio: Sure.

mdiep:
This is great. Thanks! ✨

PR #3024, opened 2016-07-02
Remove signal generator disposable.

andersio:
~~Originally #3020.~~ It now targets `RAC5`.
Related issue: #2044
Mutually exclusive with #2959 
16 July: A revised v2.

-

##### Rationale
For signals not involving multiple upstream signals or schedulers, the generator disposable is usually unused.

Moreover, the current implementation of signal operators also returns the observer disposable (obtained from observing the upstream) as the generator disposable of the composed signal. This is however unnecessary, since the terminating event from the upstream would have already done the job.

Therefore, this PR proposes to remove the generator disposable, thus transferring the responsibility of the resources disposal completely to the signal operators.

<br>

##### Removes the signal generator disposable.
A signal no longer accepts a `Disposable` from its generator, which used to be disposed of after the terminating event. 

```
/// Signal
    public init(generator: @noescape (Observer) -> Void)
```

Operators may use `ObserverProtocol.on(terminated:)` instead to inject actions upon any terminating events.
<br>

##### Some Possible Directions
1. Pass through the composite disposable during producer composition, instead of the chain of *disposable-disposing-disposable* we have now.
1. Replace the composite disposable in `SignalProducer` with a signal.

<br>

@neilpa @NachoSoto @andymatuschak @jspahrsummers


andersio:
Rebased for Swift 3, and vastly reduced the scope. :)

andymatuschak:
This is an interesting simplification—thank you! You mention in #2959 that this PR partially supersedes that one. I'm not sure I see how this resolves the tricksy signal lifetime issues mentioned in #2949 and #3075. Could you elaborate on that?

andersio:
It actually prevents #2959 from happening. That's why it is "kind of" superseded by this.

andersio:
Closed until there is a decision on the `Signal` semantics. :)

PR #3025, opened 2016-07-02
Fix the indentation of the Playground sources.

andersio:
Somehow it wasn't using tabs like the rest of the repo. 😕

A few changes have been made too, like adding punctuations, breaking long comments and breaking chained codes for better indentation. Though they are buried in the changes of indentation. Hmm.


andersio:
Gonna leave it to the future, since this patch would be disruptive to the auto merge process. :-[

PR #3030, opened 2016-07-04
Add Lifetime to enable observation of object deallocation in Swift

sharplet:
Because `rac_willDeallocSignal` doesn't have a native Swift equivalent yet, I've implemented it myself a few times in my projects. It's always felt a little complicated, and it only works for Objective-C objects.

I had an idea for a much simpler API, which works for Swift objects (and Objective-C ones), and is much simpler than `rac_willDeallocSignal`. I've seen other examples of this pattern in the wild — using `Signal.pipe()` to send a completed event on `deinit`. This PR is my take on that pattern, giving it the name `Lifetime`.

I've added a `lifetime` property to all `NSObject`s in a public extension that uses associated objects for storage.

I also added a new operator, `takeWithin`, that allows you to say the inverse of _take until this signal triggers_ — instead you can phrase it as _take within this object's lifetime_.

I'd love to know what you think!

### DIFF
--- a/ReactiveCocoa/Swift/Lifetime.swift
+++ b/ReactiveCocoa/Swift/Lifetime.swift
@@ -24,6 +24,9 @@ private var lifetimeKey: UInt8 = 0
 extension NSObject {
 	/// Returns a lifetime that ends when the receiver is deallocated.
 	@nonobjc public var rac_lifetime: Lifetime {
+		objc_sync_enter(self)
+		defer { objc_sync_exit(self) }
+
 		if let lifetime = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime? {
 			return lifetime
 		}
--- a/ReactiveCocoa/Swift/Lifetime.swift
+++ b/ReactiveCocoa/Swift/Lifetime.swift
@@ -1,0 +1,36 @@
+import enum Result.NoError
+
+/// Provides a signal that completes when the lifetime object is deinitialized.
+///
+/// When assigned to a property of another object, provides a hook to
+/// observe when that object goes out of scope.
+public final class Lifetime {
+	/// A signal that sends a Completed event when the lifetime ends.
+	public let ended: Signal<(), NoError>
+
+	private let endedObserver: Signal<(), NoError>.Observer
+
+	public init() {
+		(ended, endedObserver) = Signal.pipe()
+	}
+
+	deinit {
+		endedObserver.sendCompleted()
+	}
+}
+
+private var lifetimeKey: UInt8 = 0
+
+extension NSObject {
+	/// Returns a lifetime that ends when the receiver is deallocated.
+	@nonobjc public var rac_lifetime: Lifetime {
+		if let lifetime = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime? {
+			return lifetime
+		}
+
+		let lifetime = Lifetime()
+		objc_setAssociatedObject(self, &lifetimeKey, lifetime, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+
+		return lifetime
+	}
+}
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -958,6 +958,19 @@ extension SignalType {
 			.map { $0.0 }
 	}
 
+	/// Forwards events from `self` until `lifetime` ends, at which point the
+	/// returned signal will complete.
+	///
+	/// - parameters:
+	///   - lifetime: A lifetime whose `ended` signal will cause the returned
+	///               signal to complete.
+	///
+	/// - returns: A signal that will deliver events until `lifetime` ends.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func takeDuring(lifetime: Lifetime) -> Signal<Value, Error> {
+		return takeUntil(lifetime.ended)
+	}
+
 	/// Forward events from `self` until `trigger` sends a `Next` or
 	/// `Completed` event, at which point the returned signal will complete.
 	///
@@ -986,7 +999,7 @@ extension SignalType {
 			return disposable
 		}
 	}
-	
+
 	/// Do not forward any values from `self` until `trigger` sends a `Next` or
 	/// `Completed` event, at which point the returned signal behaves exactly
 	/// like `signal`.
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -988,6 +988,19 @@ extension SignalProducerType {
 		return lift(Signal.sampleOn)(sampler)
 	}
 
+	/// Forwards events from `self` until `lifetime` ends, at which point the
+	/// returned producer will complete.
+	///
+	/// - parameters:
+	///   - lifetime: A lifetime whose `ended` signal will cause the returned
+	///               producer to complete.
+	///
+	/// - returns: A producer that will deliver events until `lifetime` ends.
+	@warn_unused_result(message="Did you forget to call `start` on the producer?")
+	public func takeDuring(lifetime: Lifetime) -> SignalProducer<Value, Error> {
+		return takeUntil(lifetime.ended)
+	}
+
 	/// Forward events from `self` until `trigger` sends a `Next` or `Completed`
 	/// event, at which point the returned producer will complete.
 	///
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1942,10 +1942,10 @@ extension SignalProducerType {
 		// out of scope. This lets us know when we're supposed to dispose the
 		// underlying producer. This is necessary because `struct`s don't have
 		// `deinit`.
-		let token = DeallocationToken()
+		let lifetime = Lifetime()
 
 		return SignalProducer { observer, disposable in
-			var token: DeallocationToken? = token
+			var lifetime: Lifetime? = lifetime
 			let initializedProducer: SignalProducer<Value, Error>
 			let initializedObserver: SignalProducer<Value, Error>.ProducedSignal.Observer
 			let shouldStartUnderlyingProducer: Bool
@@ -1968,21 +1968,13 @@ extension SignalProducerType {
 			disposable += {
 				// Don't dispose of the original producer until all observers
 				// have terminated.
-				token = nil
+				lifetime = nil
 			}
 
 			if shouldStartUnderlyingProducer {
-				self.takeUntil(token!.deallocSignal)
+				self.takeDuring(lifetime!)
 					.start(initializedObserver)
 			}
 		}
-	}
-}
-
-private final class DeallocationToken {
-	let (deallocSignal, observer) = Signal<(), NoError>.pipe()
-
-	deinit {
-		observer.sendCompleted()
 	}
 }
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -11,6 +11,13 @@
 		02D2602B1C1D6DB8003ACC61 /* SignalLifetimeSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02D260291C1D6DAF003ACC61 /* SignalLifetimeSpec.swift */; };
 		314304171ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.h in Headers */ = {isa = PBXBuildFile; fileRef = 314304151ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		314304181ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.m in Sources */ = {isa = PBXBuildFile; fileRef = 314304161ACA8B1E00595017 /* MKAnnotationView+RACSignalSupport.m */; };
+		4A0E10FF1D2A92720065D310 /* Lifetime.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A0E10FE1D2A92720065D310 /* Lifetime.swift */; };
+		4A0E11001D2A92720065D310 /* Lifetime.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A0E10FE1D2A92720065D310 /* Lifetime.swift */; };
+		4A0E11011D2A92720065D310 /* Lifetime.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A0E10FE1D2A92720065D310 /* Lifetime.swift */; };
+		4A0E11021D2A92720065D310 /* Lifetime.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A0E10FE1D2A92720065D310 /* Lifetime.swift */; };
+		4A0E11041D2A95200065D310 /* LifetimeSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A0E11031D2A95200065D310 /* LifetimeSpec.swift */; };
+		4A0E11051D2A95200065D310 /* LifetimeSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A0E11031D2A95200065D310 /* LifetimeSpec.swift */; };
+		4A0E11061D2A95200065D310 /* LifetimeSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A0E11031D2A95200065D310 /* LifetimeSpec.swift */; };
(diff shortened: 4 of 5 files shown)


andersio:
Just want to note that this is not the equivalent of `rac_willDeallocSignal`, since the token's deinitializer is called after its owner. If the observer uses `[unowned self]`, this might lead to surprising behaviour.

andersio:
Here is an example:

Say if I am observing an `NSManagedObjectContext`'s deallocation so as to remove an observer from the `NotificationCenter`, using this lifetime token would not do the job actually, since the weakly reference (or unowned) `self` needed by `removeObserver` by that time would be `nil`.

sharplet:
Yeah, UI bindings are what I had in mind.

Your example's a good one, I guess this implementation isn't a replacement for `rac_willDealloc`, as it's more of a _did_ dealloc notification.

mdiep:
I don't think the addition of the `Lifetime` type here is an improvement over the existing pattern. It's neither more nor less code, and it's less generic/flexible.

The lifetime semantics that @andersio mention are also a big concern.

How does this improve things for Swift objects?

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,36 @@
...
+
+	private let endedObserver: Observer<(), NoError>
+
+	public init() {
+		(ended, endedObserver) = Signal.pipe()
+	}
+
+	deinit {
+		endedObserver.sendCompleted()
+	}
```
NachoSoto: This is redundant, but I guess it's good to be explicit here.

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,36 @@
...
+}
+
+private var lifetimeKey: UInt8 = 0
+
+extension NSObject {
+	/// Returns a lifetime that ends when the receiver is deallocated.
+	@nonobjc public var lifetime: Lifetime {
+		if let lifetime = objc_getAssociatedObject(self, &lifetimeKey) as? Lifetime {
```
NachoSoto: This should be `as! Lifetime?`.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -821,6 +821,13 @@ extension SignalType {
 			.map { $0.0 }
 	}
 
+	/// Forwards events from `self` until `lifetime` ends, at which point the
+	/// returned signal will complete.
+	@warn_unused_result(message="Did you forget to call `observe` on the signal?")
+	public func takeWithin(lifetime: Lifetime) -> Signal<Value, Error> {
```
NachoSoto: Why not make this take an `NSObject` instead, and make `Lifetime` an implementation detail?

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,36 @@
...
+	deinit {
+		endedObserver.sendCompleted()
+	}
+}
+
+private var lifetimeKey: UInt8 = 0
+
+extension NSObject {
+	/// Returns a lifetime that ends when the receiver is deallocated.
+	@nonobjc public var lifetime: Lifetime {
```
NachoSoto: TIL `@nonobjc`.

NachoSoto:
I like the idea of finally exposing this 👍 but I agree with @mdiep, I don't see why we need to expose `Lifetime`, it's an implementation detail.

I want to note that this is already implemented for [`SignalProducer.replayLazily`](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/Swift/SignalProducer.swift#L1503-L1509), so I'd implement this in terms of that.

andersio:
@NachoSoto AFAIK we have neither means to extend `AnyObject`, nor storage in protocols with de deinitialisers. So the only way to get a pre- or a post-dealloc signal in pure Swift classes is defining your own.

andersio:
My second thought is that this has a vastly limited scope for only pure Swift controllers and non-Cocoa users (which we probably have none yet).

Under Cocoa, users can always grab a `willDeallocSignal`, and I assume we gonna make a Swift re-implementation of it before we are ready to split ObjC and Swift. Moreover, we usually recommend the `MutableProperty` + flattening approach for reuseable cells, whereas `MutableProperty` already has the semantics of `Lifetime`.

andersio:
That's said I think it might worth creating a new protocol to cover the token suggested by @sharplet, `NSObject`s and also voluntarily conforming Swift classes. The target at call site is unambiguous across the cases: `.take(withinLifetime: object)` and `.take(withinLifetime: object.token)`.

sharplet:
@andersio:

> My second thought is that this has a vastly limited scope for only pure Swift controllers and non-Cocoa users (which we probably don't have one yet).

@mdiep:

> How does this improve things for Swift objects?

The intent was that if you needed this for a Swift value, it would be as easy as adding a property:

```swift
class MyClass {
  private let lifetime = Lifetime()
  
  func start() {
    observeStuff.takeWithin(lifetime).start()
  }
}
```

@NachoSoto:

> Why not make this take an NSObject instead, and make Lifetime an implementation detail?

At best, I think this idea might be slightly more _learnable_ by being modelled as a real type. That is, if you were to read `takeWithin`'s signature or see the `Lifetime` class, you'd hopefully grasp its purpose easily.

But you're right, it also may not be beneficial enough to justify its existence. It's not super discoverable in its current form.

---

I've worked on a few teams with less experienced RAC users and it seems pretty common to do things like starting an unbounded timer or `NSNotificationCenter` observation.

Originally I had some ideas around actually incorporating the idea of a "lifetime" into the type of a producer, and then preventing it from being started until the lifetime has been explicitly specified (e.g., something like `takeWithin(lifetime)`, or `MutableProperty` which has the same semantics implicitly).

This PR is actually the result of some ideas I've been having around modelling signal lifetimes, but stripped back to be the simplest useful change I could think of.

If we agree that it's not enough of an improvement I'm happy to close this for now. Thanks for your input!

andersio:
@sharplet I'd like to redact my comment. After performing a few tests, `rac_willDeallocSignal` would not provide the semantics I said if `weak` or `unowned` references are used. Only `Unmanaged` (`unsafe unretained`) references would work.

ReactiveCocoa/Swift/Lifetime.swift
sharplet: Good catch 👍

andersio:
I have a snap idea though.

Say we have a protocol hierarchy of `protocol LifetimeProviding: LifetimeObservable`.

`LifetimeProviding` requires a lifetime object, while providing default implementations for `LifetimeObservable`. `LifetimeObservable` requires a EOL signal, and was implemented for `NSObject` and various `RAC` constructs.

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,36 @@
...
+
+/// Provides a signal that completes when the lifetime object is deinitialized.
+///
+/// When assigned to a property of another object, provides a hook to
+/// observe when that object goes out of scope.
+public final class Lifetime {
+	/// A signal that sends a Completed event when the lifetime ends.
+	public let ended: Signal<(), NoError>
+
+	private let endedObserver: Observer<(), NoError>
```
NachoSoto: Can you use `Signal<(), NoError>.Observer`? That way it's clearer to see the parallel between the signal and its observer :)

NachoSoto:
I want to note that `SignalProducer.replayLazily` would have to be changed to be implemented using this new operator.

ReactiveCocoa/Swift/Lifetime.swift
mdiep: How is it redundant? I don't see how it would complete otherwise.

mdiep:
> The intent was that if you needed this for a Swift value, it would be as easy as adding a property:

That makes sense now. 👍🏻

ReactiveCocoa/Swift/Signal.swift
mdiep: > Why not make this take an NSObject instead, and make Lifetime an implementation detail?

I think the idea is to make this usable with non-`NSObject` classes.

How about naming this `take(during lifetime: Lifetime)`. That seems like it would read a little more naturally.

```swift
.take(during: myObject.lifetime)
```

mdiep:
I'm into this. 👍🏻

There are some conflicts and a few unaddressed items above. Would you be up for updating this?

_(Sorry for the delay on this!)_

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,36 @@
...
+		endedObserver.sendCompleted()
+	}
+}
+
+private var lifetimeKey: UInt8 = 0
+
+extension NSObject {
+	/// Returns a lifetime that ends when the receiver is deallocated.
+	@nonobjc public var lifetime: Lifetime {
+		if let lifetime = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime? {
```
andersio: Apparently we could have a race condition here. This would need to be guarded by the `objc_sync_*` pair or some kind of locks.

andersio:
`NSObject.lifetime` would need thread safety. Otherwise LGTM now.  🎉

-
Second thought on protocols:
> A protocol like `LifetimeProviding` might be less useful, if Lifetime is part of the public API. Say if we write a function `func bind(to: Something, during lifetime: Lifetime`, there is already an implicit constraint of objects that can provide a lifetime.

andersio:
I can now see `take(within:)` (or `take(during:)`?) that takes a `Lifetime` is optimal. :tada:

Constraining it with a protocol doesn't seem necessary to me, since we have cases that would use a `Lifetime` object directly. e.g. `replayLazily` as mentioned by @NachoSoto.

ReactiveCocoa/Swift/Lifetime.swift
sharplet: Done.

sharplet:
@andersio I like `take(during:)`! I've pushed an amended version with the operator renamed.

ReactiveCocoa/Swift/Lifetime.swift
sharplet: Yep, that's a good call. I've pushed a change which synchronises on `self` here.

> P.S. Why not simply write as? Lifetime?

I originally used that, see [this discussion](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3030#discussion_r69509093). There is no valid reason for the type of the associated object to be something other than `Lifetime`. If it is something else, then it should be caught as early as possible in development.

sharplet:
Ok, I think I've addressed all the feedback. Thanks so much!

I think what this could really use now is some proper API documentation, so this addition is more discoverable than our release notes or autocompletion. Should I take a stab at that as part of this PR, or can that be deferred until closer to the next release?

sharplet:
Oh, and one other thing: do you think it would be appropriate to add a Swift-specific deprecation warning to either/both of `rac_willDeallocSignal` and `rac_deallocDisposable`, encouraging Swift users to adopt `Lifetime` instead?

sharplet:
FWIW I think a Swift-specific deprecation should be achievable with a combination of `NS_REFINED_FOR_SWIFT` and a new Swift API using an appropriate availability attribute and calling through to the original.

andersio:
The Objective-C APIs would be moved to a separate module (ReactiveObjC?), and IIRC the ultimate goal is to completely drop the dependency of the Swift land on ReactiveObjC. So I don't see the need to mark it `@unavailable`.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1942,10 +1942,10 @@ extension SignalProducerType {
 		// out of scope. This lets us know when we're supposed to dispose the
 		// underlying producer. This is necessary because `struct`s don't have
 		// `deinit`.
-		let token = DeallocationToken()
+		let token = NSObject()
```
andersio: I think we should use a `Lifetime` here directly. `NSObject.lifetime` is available only on Darwin platforms.

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,36 @@
+import enum Result.NoError
+
+/// Provides a signal that completes when the lifetime object is deinitialized.
+///
+/// When assigned to a property of another object, provides a hook to
+/// observe when that object goes out of scope.
+public final class Lifetime {
+	/// A signal that sends a Completed event when the lifetime ends.
+	public let ended: Signal<(), NoError>
```
andersio: [Bikeshedding] We should use a noun here, shouldn't we?

mdiep:
Should we use a prefix for `NSObject.lifetime`? Since it's a property, the type can't be used to disambiguate and it may conflict with existing methods/properties. cc #3049

andersio:
True... It should be prefixed before we ever figure out a new approach.

ReactiveCocoa/Swift/SignalProducer.swift
sharplet: Using `NSObject` here didn't bother me because of the use of `NSLock` above (although I suppose that will have a counterpart in swift-corelibs-foundation).

I agree though, using `Lifetime` directly makes sense!

sharplet: Done.

andersio:
An alternative would be built around a protocol and avoid using a property.

sharplet:
> The Objective-C APIs would be moved to a separate module (ReactiveObjC?). So I don't see the need to mark it unavailable.

The repo split won't happen until RAC 5, which means (unless I'm mistaken) that this is likely to be part of a 4.x release. In my own projects, I've wrapped `rac_signalForSelector`/`rac_deallocDisposable` in a number of places in order to expose the functionality to Swift in a type-safe way. So I think have a deprecation warning (as opposed to marking the API unavailable) could prove useful for any other users who have the same kind of wrappers and would benefit from being aware of the new feature. Do you agree?

andersio:
True for the split planned for RAC 5. But there _could_ be people using the ObjC API in Swift though, without caring about our shiny Swift offerings.

sharplet:
Personally, prefixing the name bothers me (but only a little), and I feel it hampers readability. Having said that, I totally understand the concern and agree that libraries should be very careful forcing extensions of types we don't on users, without any workaround (other than _rename your API_).

I actually feel fairly strongly that the naming is very much _the whole point_ of this change, which is what's giving me uncomfortable feels about it. 😕

I have to wonder whether there is actually any instance of `NSObject.lifetime` out there amongst RAC's users? Problem is the only way to find out would seem to be shipping a potentially breaking change and waiting to _maybe_ find out.

andersio:
I don't like it either. But in 4.x everything (aka the two Foundation extensions and the Rex bindings...) is prefixed. So we should be consistent, shouldn't we?

That's said there is sadly no way (AFAIK) we can completely safe from collision, except prefixing `rac` that is safe... ehm... conventionally.

mdiep:
Unfortunately, I think we need the prefix. 😞 Once that's done, I'm 👍🏻 to merge.

(Also, FWIW I don't think there will be any more 4.x releases. We'll merge `RAC5` to `master` once Travis updates to ß3.)

andersio:
Last Minute Question:

Having written:
```
object.lifetime.ended.observeCompleted(observer.sendCompleted)
```
makes me wonder if we should provide a convenience `observeCompleted` on `Lifetime`. Seems a bit awkward at the call site.

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,39 @@
...
+	@nonobjc public var lifetime: Lifetime {
+		objc_sync_enter(self)
+		defer { objc_sync_exit(self) }
+
+		if let lifetime = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime? {
+			return lifetime
+		}
+
+		let lifetime = Lifetime()
+		objc_setAssociatedObject(self, &lifetimeKey, lifetime, .OBJC_ASSOCIATION_RETAIN)
```
andersio: I suggest using `RETAIN_NONATOMIC` instead. Using `RETAIN` (atomic) would somehow hinder the ability to use a `unowned(unsafe)` reference to the object.

mdiep:
@sharplet Are you interested in finishing this up? Not trying to pressure you; I just don't want this to fall between the cracks. 😄

sharplet:
Ok I think that's everything, except for this comment:

> [Bikeshedding] We should use a noun here, shouldn't we?

FWIW I really like the way `takeUntil(lifetime.ended)` reads. Although I'm certainly open to suggestions.

In any case, thanks for the nudge @mdiep and let me know if there's anything else! ✌️

mdiep:
This is great! Thanks @sharplet! 🎉

sharplet:
Thanks for merging @mdiep! And thanks everyone for the great feedback. I'd like to contribute some more documentation around this in the form of a usage guide, if you think that's appropriate. Perhaps an addition to the [Framework Overview](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/Documentation/FrameworkOverview.md)?

mdiep:
Documentation is always appreciated. 😄

PR #3031, opened 2016-07-04
Sandbox

RuiAAPeres:
I sometimes want to experiment something quick, and I find myself picking a random page, removing everything except the imports and then start doing my experiments. The sandbox page, will have this clean page, where people can simply start working.


iv-mexx:
Great idea,

> I find myself picking a random page, removing everything except the imports and then start doing my experiments.

That's something I frequently do/did as well

Palleas:
Love the idea 👍

mdiep:
👍

PR #3033, opened 2016-07-05
Atomic value for PropertyType

filblue:
1. Separates `Mutex` implementation and `Atomic` value storage
2. Reduces code duplication and maintainability in `MutableProperty` by reusing higher level `Atomic` API instead of ad-hoc reimplementation.

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -35,22 +35,22 @@ public final class PosixThreadMutex: MutexType {
 	
 	public init() {
 		let result = pthread_mutex_init(&mutex, nil)
-		assert(result == 0, "Failed to initialize mutex with error \(result).")
+		precondition(result == 0, "Failed to initialize mutex with error \(result).")
 	}
 	
 	deinit {
 		let result = pthread_mutex_destroy(&mutex)
-		assert(result == 0, "Failed to destroy mutex with error \(result).")
+		precondition(result == 0, "Failed to destroy mutex with error \(result).")
 	}
 	
 	public func lock() {
 		let result = pthread_mutex_lock(&mutex)
-		assert(result == 0, "Failed to lock \(self) with error \(result).")
+		precondition(result == 0, "Failed to lock \(self) with error \(result).")
 	}
 	
 	public func unlock() {
 		let result = pthread_mutex_unlock(&mutex)
-		assert(result == 0, "Failed to unlock \(self) with error \(result).")
+		precondition(result == 0, "Failed to unlock \(self) with error \(result).")
 	}
 }
 
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -8,9 +8,56 @@
 
 import Foundation
 
+public protocol MutexType {
+	func lock()
+	func unlock()
+}
+
+public final class RecursiveLock: MutexType {
+	private var _lock: NSRecursiveLock
+
+	public init(_ name: String) {
+		_lock = NSRecursiveLock()
+		_lock.name = name
+	}
+
+	public func lock() {
+		_lock.lock()
+	}
+
+	public func unlock() {
+		_lock.unlock()
+	}
+}
+
+public final class PosixThreadMutex: MutexType {
+	private var mutex = pthread_mutex_t()
+	
+	public init() {
+		let result = pthread_mutex_init(&mutex, nil)
+		assert(result == 0, "Failed to initialize mutex with error \(result).")
+	}
+	
+	deinit {
+		let result = pthread_mutex_destroy(&mutex)
+		assert(result == 0, "Failed to destroy mutex with error \(result).")
+	}
+	
+	public func lock() {
+		let result = pthread_mutex_lock(&mutex)
+		assert(result == 0, "Failed to lock \(self) with error \(result).")
+	}
+	
+	public func unlock() {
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
...
+		_lock.unlock()
+	}
+}
+
+public final class PosixThreadMutex: MutexType {
+	private var mutex = pthread_mutex_t()
+	
+	public init() {
+		let result = pthread_mutex_init(&mutex, nil)
+		assert(result == 0, "Failed to initialize mutex with error \(result).")
```
NachoSoto: These should be `precondition`s, I don't think we want to ignore a potential error in production.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -25,25 +72,17 @@ public final class Atomic<Value> {
 	}
 	
 	/// Initializes the variable with the given initial value.
-	public init(_ value: Value) {
+	public init(_ value: Value, mutex: MutexType = PosixThreadMutex()) {
```
NachoSoto: I don't think I see why we'd want this to be a public API: why would anyway want to pick a particular mutex implementation?

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -74,3 +121,47 @@ public final class Atomic<Value> {
 		return try action(_value)
 	}
 }
+
+public struct AnyAtomic<Value> {
```
NachoSoto: What does this represent?

filblue:
@andersio:
> Is there a benefit to reuse the implementation of Atomic? 

Removing duplication of low-level code and improving maintainability by separating concerns seem beneficial to me. `Properties` do not have to reimplement `lock/unlock/withValue` as it is already implemented in `Atomic`.

> Moreover, I don't see the point of abolishing PropertyType.value and requires Atomic on PropertyType. DynamicProperty cannot be atomic, for example. The only clear intention I can see is exposing withValue, but that can be done on PropertyType.

`DynamicProperty` is not a really native `Swift/RAC` citizen, and IMO its inner issues should not leak to `PropertyType`. If we think about _sane_ `PropertyType` implementations, seems totally reasonable to require atomicity of its inner value storage.

As the `DynamicProperty` docs say:
```
/// Wraps a `dynamic` property, or one defined in Objective-C, using Key-Value
/// Coding and Key-Value Observing.
///
/// Use this class only as a last resort! `MutableProperty` is generally better
/// unless KVC/KVO is required by the API you're using (for example,
/// `NSOperation`).
```

Swift team will not consider implementing `KVO` in the foreseeable future, and `PropertyType.withValue` has the same problem with `DynamicProperty`.

ReactiveCocoa/Swift/Atomic.swift
filblue: Picking a particular mutex implementation is exactly what happens in `MutableProperty`

NachoSoto: That's an implementation detail though.

ReactiveCocoa/Swift/Atomic.swift
filblue: I'm not happy with this name )). This thing is basically a wrapper around `AnyAtomicState`, hiding its details, i.e. separate `withValue` logic for constant and changeable things.

filblue: Ideas on a better name would be appreciated

ReactiveCocoa/Swift/Atomic.swift
filblue: `withValue` logic built on `mutex` implementation details are not exactly `MutableProperty`'s concern, that same thing was implemented in `Atomic`.

NachoSoto: What do you mean?

ReactiveCocoa/Swift/Atomic.swift
filblue: Fixed, thanks

ReactiveCocoa/Swift/Atomic.swift
filblue: `lock`, `unlock`, `modify` and `withValue` routines where implemented twice, in `MutableProperty` and `Atomic`. `MutableProperty` knows that she wants `atomicity` built on `recursive` locking strategy, but implementation of these requirements can be delegated.

mdiep:
I think this conflates a few different changes, some of which are controversial.

1. Extract the locking behavior from `MutableProperty` into a separate class
2. Combining that class with `Atomic`
3. Exposing `AnyAtomic`
4. Adding `atomic` properties to properties

I think proposing these in smaller steps would lead to more productive discussion and easier reviewing. In particular, I think (1) seems like a good idea by itself.

andersio:
@filblue The problem is that `PropertyType` does not mandate thread safety at this moment, nor would it be necessary for producers/signals to be composed safely. For example, I have a `ManagedObjectProperty` which makes no sense to use `Atomic`.

andersio:
(P.S. there are legitimate cases* of read-read race conditions. But then `withValue` or just `producer.take(1)` on `PropertyType` can cover those cases in case the impl. is atomic.)

-
_* But are IMO rare due to the shiny composition operators. An example would be taking `value` and observing `signal`._

filblue:
Hey everyone, thanks for you feedback!

By @mdiep suggestion  I have extracted the least controversial change into #3046

Issue #3040, opened 2016-07-07
[Pitch] Replace `PropertyType.signal`.


andersio:
While it makes absolute sense to expose `signal` on `MutableProperty`, the situation has slightly changed as now we have composed properties.

Due to the semantics of `Signal` which is uninterruptible from the downstream observers, any `signal`s created from a composed property but then left unused would actually **leak** until the upstream is terminated.

```
var disposable: Disposable?
func scope() {
    disposable = property.map { $0 + 1 }.map { $0 + 1 }.signal.observeNext { _ in }
}

scope()
disposable?.dispose()

// The signal is still alive at this point, and will terminate only when `property` is deinitialised.
```

Together with the semantics of composed properties (_respects the lifetime of its ultimate sources_), there is hardly a solution to this, unless we change the semantics of `Signal` to allow signals without observers to be automatically disposed of.

Therefore, I would like to propose that `PropertyType.signal` should be replaced by a producer ~~`PropertyType.changes`, and consequently `PropertyType.producer` should be renamed to `PropertyType.values` to make its role clear.~~

(The renaming can be done regardless of the change in type.)

-
Edited: Expand the scope for a bit.

RuiAAPeres:
> Therefore, I would like to propose that PropertyType.signal should be replaced by a producer PropertyType.changes, and consequently PropertyType.producer should be renamed to PropertyType.values to make its role clear.

What would be the difference between these two producers then? I like the hot semantics that the exposed signal provides. Specially when we are dealing with things like `rex_text` (versus the producer's cold nature).

andersio:
Still the same difference: emitting the current value, or not. Moreover, you still have access to `signal` on the concrete types in the case of `MutableProperty`.

It is worth noting that the `AnyProperty` in `RAC5` has a different lifetime w.r.t. `master`'s. The one in `RAC5` is a lens to the original property, so that:
```
/// property: MutableProperty<MutableProperty<Value>>
property.producer.flatMap { $0.producer }
property.flatMap { $0 }.producer
```
are identical. The one in `master` is a standalone property that determines its own lifetime and just happens to have a binding.

andersio:
Hmm, a more illustrative example I can think of:
```
let tweetListViewModel = appViewModel.flatMap { $0.tweetList }
```
[...]

PR #3041, opened 2016-07-07
Makes composed properties safe.

andersio:
(Splitting PR, Rebasing)

##### Completed
As mentioned in #3040, `PropertyType.signal` is not safe on composed properties due to the potential silent leakage caused by the semantics of `Signal`. Currently, it was just denoted in the comments that the use of `signal` on composed properties is discouraged.

This PR proposes further to replace it with an interruptible `SignalProducer` - `PropertyType.changes`. Meanwhile, it proposes also to rename `PropertyType.producer` to `PropertyType.values`, which describes now its role rather than its type.

> https://swift.org/documentation/api-design-guidelines/#promote-clear-usage
>
> Name variables, parameters, and associated types according to their roles, rather than their type constraints.

P.S. The difference between `changes` and `values` is _still_ whether they emit the current value when started.

### DIFF
--- a/ReactiveCocoa/Swift/Action.swift
+++ b/ReactiveCocoa/Swift/Action.swift
@@ -68,8 +68,8 @@ public final class Action<Input, Output, Error: ErrorType> {
 		values = events.map { $0.value }.ignoreNil()
 		errors = events.map { $0.error }.ignoreNil()
 
-		_enabled <~ enabledIf.producer
-			.combineLatestWith(_executing.producer)
+		_enabled <~ enabledIf.values
+			.combineLatestWith(_executing.values)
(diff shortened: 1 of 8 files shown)


andersio:
The issue comes at two different levels:

1. The latest value from `PropertyType.values` may not be `PropertyType.value`.
2. The latest value from a composed property may not be logically consistent with its source(s).

While (1) can be solved in a similar way as `sources`, (2) requires a relay as a point of consistency to cache & forward the value. This means we probably need to revert the awesome but inconsistent lazy behaviour.

andersio:
I will reopen this PR when this is discussed & the dependencies are dropped and merged, perhaps.

PR #3042, opened 2016-07-07
Fix the inconsistency of composed properties.

andersio:
##### The inconsistency
Unfortunately, we overlooked the semantic for a couple of operators that are **not memoryless*** in #2646.

While the `SignalProducer` used in the composition producing different instances of signals is fine for memoryless systems, it leads to multiple instances of truth for non-memoryless operators.

Consequently, the following may happen to these operators at some point:

1. The latest value from `producer` and `signal` of a composed property is not always consistent with `value`.
[(Failing Test case on `RAC5`)](https://github.com/ReactiveCocoa/ReactiveCocoa/commit/05f1570abfa786deffacc4d97164d3d175dc7e16)
_Note: It is `secondResult` which failed the expectation._

2. Similar to (1), a composed property is not always consistent with its sources of any kind.
[(Failing Test case `RAC5`)](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3042/commits/309f3880f57c2a38d42c88084b3078b350570ce9)

##### Proposed solution
Revert to use a relay object (`MutableProperty`) to stage the values, so that `value`, `producer` and `signal` can share a single source of truth to deliver values consistently.

The relay object is retained by not only the `AnyProperty` which created it, but also all produced signals **and** the lazily initialised signal of the `AnyProperty`.

In other words, when the `AnyProperty` is released, the relay object would:

1. Deinitialize, iff `AnyProperty.signal` is not ever used, and no signals was produced from `AnyProperty.producer`.

1. Live until the upstream sends a terminating event, iff `AnyProperty.signal` has ever been requested.

2. Live until all produced signals are interrupted, or until the upstream sends a terminating event, whichever happens earlier.

Note that from the user's point of view, the semantic is still untouched: the intermediate `AnyProperty` would still deinitialize if they go out of scope.

##### Notes
\* An operator is said to be **memoryless** if its output at time t depends on inputs at **only** time t. For example. `map` is simply a system of `y[t] ← x[t]`. Likewise, `combineLatest(with:)`, `flatten(.latest)`, `flatMap(.latest)` and `skipRepeats` can also be qualified as memoryless.

-
_P.S. Really thank you @NachoSoto for reviewing again the old RAC 5 patches, or otherwise I won't catch this shaky spot._


andersio:
I have cherry-picked the test cases to #3023 to show how it is gonna fail.

[A direct link to the build log.](https://travis-ci.org/ReactiveCocoa/ReactiveCocoa/jobs/143349682)

andersio:
Rebased for Swift 3.

andersio:
It is possible to keep the previous behavior specifically for the memoryless operator though.

But it would require 2 new variants of `lift`, 2 convenience initializer and 1 designated initializer for lifting them.

mdiep:
This has conflicts and could use an update.

PR #3046, opened 2016-07-08
Reuse Atomic in MutableProperty

filblue:
1. Allows `Atomic` to work with different `Mutex` implementations
2. Uses `Atomic`'s `modify` and `withValue` in `MutableProperty`

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -8,9 +8,55 @@
 
 import Foundation
 
+internal protocol MutexType: class {
+	func lock()
+	func unlock()
+}
+
+final class RecursiveLock: MutexType {
+	private var _lock: NSRecursiveLock
+
+	init(_ name: String) {
+		_lock = NSRecursiveLock()
+		_lock.name = name
+	}
+
+	func lock() {
+		_lock.lock()
+	}
+
+	func unlock() {
+		_lock.unlock()
+	}
+}
+
+final class PosixThreadMutex: MutexType {
+	private var _mutex = pthread_mutex_t()
+
+	init() {
+		let result = pthread_mutex_init(&_mutex, nil)
+		precondition(result == 0, "Failed to initialize mutex with error \(result).")
+	}
+
+	deinit {
+		let result = pthread_mutex_destroy(&_mutex)
+		precondition(result == 0, "Failed to destroy mutex with error \(result).")
+	}
+
+	func lock() {
+		let result = pthread_mutex_lock(&_mutex)
+		precondition(result == 0, "Failed to lock \(self) with error \(result).")
+	}
+
+	func unlock() {
+		let result = pthread_mutex_unlock(&_mutex)
+		precondition(result == 0, "Failed to unlock \(self) with error \(result).")
+	}
+}
+
 /// An atomic variable.
 public final class Atomic<Value> {
-	private var mutex = pthread_mutex_t()
+	private var _mutex: MutexType
 	private var _value: Value
 	
 	/// Atomically get or set the value of the variable.
@@ -28,25 +74,22 @@ public final class Atomic<Value> {
 	/// 
 	/// - parameters:
 	///   - value: Initial value for `self`.
-	public init(_ value: Value) {
-		_value = value
-		let result = pthread_mutex_init(&mutex, nil)
-		assert(result == 0, "Failed to initialize mutex with error \(result).")
+	public convenience init(_ value: Value) {
+		self.init(value, mutex: PosixThreadMutex())
 	}
 
-	deinit {
-		let result = pthread_mutex_destroy(&mutex)
-		assert(result == 0, "Failed to destroy mutex with error \(result).")
+	/// Initializes the variable with the given initial value.
+	internal init(_ value: Value, mutex: MutexType) {
+		_value = value
+		_mutex = mutex
 	}
 
 	private func lock() {
-		let result = pthread_mutex_lock(&mutex)
-		assert(result == 0, "Failed to lock \(self) with error \(result).")
+		_mutex.lock()
 	}
 	
 	private func unlock() {
-		let result = pthread_mutex_unlock(&mutex)
-		assert(result == 0, "Failed to unlock \(self) with error \(result).")
+		_mutex.unlock()
 	}
 	
 	/// Atomically replace the contents of the variable.
@@ -66,8 +109,16 @@ public final class Atomic<Value> {
 	///
 	/// - returns: The old value.
 	public func modify(@noescape action: (Value) throws -> Value) rethrows -> Value {
+		return try modify(action, completion: { _ in })
+	}
+
+	/// Atomically modifies the variable.
+	///
+	/// Returns the old value.
+	public func modify(@noescape action: (Value) throws -> Value, @noescape completion: (Value) -> ()) rethrows -> Value {
 		return try withValue { value in
 			_value = try action(value)
+			completion(_value)
 			return value
 		}
 	}
--- a/ReactiveCocoa/Swift/Property.swift
+++ b/ReactiveCocoa/Swift/Property.swift
@@ -123,17 +123,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 
 	private let observer: Signal<Value, NoError>.Observer
 
-	/// Need a recursive lock around `value` to allow recursive access to
-	/// `value`. Note that recursive sets will still deadlock because the
-	/// underlying producer prevents sending recursive events.
-	private let lock: NSRecursiveLock
-
-	/// The getter of the underlying storage, which may outlive the property
-	/// if a returned producer is being retained.
-	private let getter: () -> Value
-
-	/// The setter of the underlying storage.
-	private let setter: Value -> Void
+	private let _atomic: Atomic<Value>
 
 	/// The current value of the property.
 	///
@@ -141,7 +131,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// created from the `values` producer.
 	public var value: Value {
 		get {
-			return withValue { $0 }
+			return _atomic.withValue { $0 }
 		}
 
 		set {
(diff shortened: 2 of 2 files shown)


ReactiveCocoa/Swift/Property.swift
```diff
@@ -135,7 +128,7 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// followed by all changes over time, then complete when the property has
 	/// deinitialized.
 	public var producer: SignalProducer<Value, NoError> {
-		return SignalProducer { [getter, weak self] producerObserver, producerDisposable in
+		return SignalProducer { [_atomic, weak self] producerObserver, producerDisposable in
 			if let strongSelf = self {
```
andersio: This could be simplified a bit if we grab the value via `withValue` first, and evaluate whether the property is deinitialized within the action. The locking is mandatory now anyway, so the comments should be removed too.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -57,8 +96,16 @@ public final class Atomic<Value> {
 	///
 	/// Returns the old value.
 	public func modify(@noescape action: (Value) throws -> Value) rethrows -> Value {
+		return try modify(action, completion: { _ in })
+	}
+
+	/// Atomically modifies the variable.
+	///
+	/// Returns the old value.
+	public func modify(@noescape action: (Value) throws -> Value, @noescape completion: (Value) -> ()) rethrows -> Value {
```
andersio: Is there a reason why `completion` cannot be an optional?

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
 
 import Foundation
 
+public protocol MutexType {
```
andersio: This should be a class bound protocol anyway.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
...
+	func lock()
+	func unlock()
+}
+
+public final class RecursiveLock: MutexType {
```
andersio: Should we use a pthread recursive mutex here, if the Objective-C runtime dispatch cost is/was a concern for `Atomic`?

Otherwise, instead of reinventing the wheel, `Locking`(`NSLocking`) available in both Objective-C and Swift Foundation sounds a better option.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -25,25 +72,17 @@ public final class Atomic<Value> {
 	}
 	
 	/// Initializes the variable with the given initial value.
-	public init(_ value: Value) {
+	public init(_ value: Value, mutex: MutexType = PosixThreadMutex()) {
```
andersio: I think it makes more sense to have `Atomic` parameterised with a `MutexType`. It also allows the compiler to specialise the calls to the `MutexType` implementation.

ReactiveCocoa/Swift/Atomic.swift
dmcrodrigues: I think we should, at least to keep code consistent. And doing that we can have a single type, instead this two, by making it configurable with a particular strategy, `Recursive` or `NonRecursive`, during initialization.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
...
+		_lock.lock()
+	}
+
+	public func unlock() {
+		_lock.unlock()
+	}
+}
+
+public final class PosixThreadMutex: MutexType {
+	private var mutex = pthread_mutex_t()
```
dmcrodrigues: To keep consistency this should be prefixed with a `_`.

ReactiveCocoa/Swift/Property.swift
filblue: agree, will fix

ReactiveCocoa/Swift/Atomic.swift
filblue: good idea, thanks!

ReactiveCocoa/Swift/Atomic.swift
filblue: I see your point guys, but I would prefer to refrain from changing particular recursive mutex implementation in this particular PR.

andersio:
Just wanna note that the use of `MutexType` existential is a degradation on paper. An existential has to be dynamically dispatched, which is absolutely unnecessary for the lock in `Atomic` or `MutableProperty`.

So if you meant to merge against `master`, I am not quite fond of refactoring a working implementation with a negative impact. Fixing it however would likely require breaking changes, which would mean `RAC5`... Hmm...

andersio:
Just want to note that using a `MutexType` existential is a slight degradation over the inlined locks we currently have, due to dynamic dispatch. Bringing it up par would need to scrap the existential and very likely introduce breaking changes.

ReactiveCocoa/Swift/Atomic.swift
andersio: Then you might as well conform `NSRecursiveLock` to `MutexType`, instead of wrapping it.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
 
 import Foundation
 
+public protocol MutexType {
+	func lock()
```
andersio:
```
func lock<Result>(@noescape action: () throws -> Result) rethrows -> Result
```
sounds a nice Swifty contender.

ReactiveCocoa/Swift/Atomic.swift
filblue: hm, cannot keep `@noescape` when trying to make `completion` optional

ReactiveCocoa/Swift/Atomic.swift
NachoSoto: I'll repeat myself: I think this is a great refactor to take this out of `MutableProperty`, but I don't think this constructor should be `public` .

ReactiveCocoa/Swift/Atomic.swift
andersio: It can be expressed in Swift 3.0 though...

ReactiveCocoa/Swift/Atomic.swift
mdiep: I agree with @NachoSoto. I don't think this API should be public as-is. Let's keep the public API for `Atomic` the same.

ReactiveCocoa/Swift/Atomic.swift
filblue: I'l be happy to port this to RAC5 via another PR!

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
 
 import Foundation
 
+public protocol MutexType: class {
```
mdiep: Can you make this protocol and the other added types also `private` or `internal`? There's no reason to expose them since they can't be used with `Atomic` externally.

mdiep:
This needs a merge from master. (Sorry!)

NachoSoto:
I'll review this tomorrow too :) (or when it gets rebased)

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
 
 import Foundation
 
+protocol MutexType: class {
```
NachoSoto: Maybe make this explicitly `internal`, to remind us that it's not public on purpose?

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -8,9 +8,56 @@
...
+	func unlock() {
+		let result = pthread_mutex_unlock(&_mutex)
+		precondition(result == 0, "Failed to unlock \(self) with error \(result).")
+	}
+}
+
 /// An atomic variable.
 public final class Atomic<Value> {
-	private var mutex = pthread_mutex_t()
+
```
NachoSoto: Extra line.

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -25,25 +72,22 @@ public final class Atomic<Value> {
...
-		assert(result == 0, "Failed to initialize mutex with error \(result).")
+	public convenience init(_ value: Value) {
+		self.init(value, mutex: PosixThreadMutex())
 	}
 
-	deinit {
-		let result = pthread_mutex_destroy(&mutex)
-		assert(result == 0, "Failed to destroy mutex with error \(result).")
+	/// Initializes the variable with the given initial value.
+	init(_ value: Value, mutex: MutexType) {
```
NachoSoto: Ditto: explicitly `internal`?

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -25,25 +72,22 @@ public final class Atomic<Value> {
 	}
 	
 	/// Initializes the variable with the given initial value.
-	public init(_ value: Value) {
-		_value = value
-		let result = pthread_mutex_init(&mutex, nil)
-		assert(result == 0, "Failed to initialize mutex with error \(result).")
+	public convenience init(_ value: Value) {
+		self.init(value, mutex: PosixThreadMutex())
```
NachoSoto: Awesome

ReactiveCocoa/Swift/Property.swift
```diff
@@ -104,22 +104,15 @@ public final class MutableProperty<Value>: MutablePropertyType {
 	/// Need a recursive lock around `value` to allow recursive access to
```
NachoSoto: This comment is no longer relevant here. Can you move it to where you're instantiating the particular lock in the constructor?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -104,22 +104,15 @@ public final class MutableProperty<Value>: MutablePropertyType {
...
 	/// underlying producer prevents sending recursive events.
-	private let lock: NSRecursiveLock
-
-	/// The getter of the underlying storage, which may outlive the property
-	/// if a returned producer is being retained.
-	private let getter: () -> Value
-
-	/// The setter of the underlying storage.
-	private let setter: Value -> Void
+	private var _atomic: Atomic<Value>
```
NachoSoto: This doesn't need to be mutable, it should be `let`!

filblue:
More comments addressed (thanks @NachoSoto) and rebased on master.

NachoSoto:
Perfect 🎉

PR #3047, opened 2016-07-08
Generic `skip(until:)` and `take(until:)`.

andersio:
This PR proposes to alter the signature of `skip(until:)` and `take(until:)` for both `Signal` and `SignalProducer` to:

```
func take<U>(until: Signal<U, NoError>) -> Signal<Value, Error>
func skip<U>(until: Signal<U, NoError>) -> Signal<Value, Error>
```

So that a trigger can be any arbitrary signals that never fails. These operators do not care about the actual value anyway, but only two specific types of events.

The changes are purely addictive.

An obvious use case would be accepting a zipped signal of `willDeallocSignal`, which is of type `((), ())` and is not accepted by the original signature:
```
NotificationCenter.default.rac_notifications(.aNotificationThatIAmInterestedIn, object: object)
    .take(until: self.willDeallocSignal.zip(with: another.willDeallocSignal))
    .startWithNext(process(_:))
```

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -793,7 +793,7 @@ extension SignalProtocol {
 
 	/// Forwards events from `self` until `trigger` sends a Next or Completed
 	/// event, at which point the returned signal will complete.
-	public func take(until trigger: Signal<(), NoError>) -> Signal<Value, Error> {
+	public func take<U>(until trigger: Signal<U, NoError>) -> Signal<Value, Error> {
 		return Signal { observer in
 			let disposable = CompositeDisposable()
 			disposable += self.observe(observer)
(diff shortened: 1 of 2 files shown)


andersio:
The test would fail anyway, since Travis does not support beta 2 yet.

NachoSoto:
Thanks for the idea! Unfortunately, this was done on purpose: we want users to be 100% aware of the fact that the `trigger` values are completely ignored, by making them have to explicitly ignore them.

andersio:
@NachoSoto Would it be appropriate to provide overloads for tuples of empty tuples (say up to arity 3) then?

NachoSoto:
Like, `Signal<((), ()), NoError>`?

andersio:
Yes.

PR #3052, opened 2016-07-12
Lifetime

andersio:
Built on top of @sharplet's #3030.

Introduced a `LifetimeProviding` protocol which allows objects to declare their ability to notify others of its deinitialization. So far only `NSObject` conforms to this protocol.

`Property` is an ideal participant, but its lifetime semantic & fixes have to be settled first IMO. May as well do that separately.


andersio:
If we provide only a hot lifetime signal, composed properties are forced to create a hot lifetime signal that *could* be leaked.

So introducing `lifetimeProducer` at least provides a chance for the composed properties to workaround this if they are used with cold signals, assuming we just implement the fix in #3042 without changing the semantic.

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -0,0 +1,35 @@
+import enum Result.NoError
+
+/// Describes the ability of an object to notify others of its deinitialization.
+/// Proxy objects may pass through the lifetime of its underlying object.
+public protocol LifetimeProviding: class {
+	/// A signal representing the lifetime of `self`.
+	///
+	/// The signal emits `completed` when the object completes, or
+	/// `interrupted` after the object is completed.
```
JaviSoto: This sounds a bit confusing

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -791,6 +791,12 @@ extension SignalProtocol {
 			.map { $0.0 }
 	}
 
+	/// Forwards events from `self` until `lifetime` ends, at which point the
+	/// returned signal will complete.
+	public func take<U: LifetimeProviding>(withinLifetime object: U) -> Signal<Value, Error> {
```
JaviSoto: Should the parameter be `withinLifetimeOf:`?

JaviSoto:
I like this!

I think we should be able to do it for any `AnyObject`: by holding a `weak` reference to it. Something like:

```swift
final class DeallocationObserver: LifetimeProviding {
    weak var object: AnyObject? {
        didSet {
            if object == nil {
                self.lifetimeObserver.sendCompleted()
            }
        }
    }

    let (lifetime, lifetimeObserver) = Signal<(), NoError>.pipe()

    init(object: AnyObject) {
        self.object = object
    }
}
```

We would just need to make sure the `DeallocationObserver` instance stays alive.

andersio:
Property observer would not be called on weak references turning `nil`.

sharplet:
Yeah, the only way to find out if a weak reference is `nil` is to try to unwrap it.

sharplet:
This is why #3030 is implemented the way it is: so that you can do something like

```swift
class Foo {
  private let lifetime = Lifetime()
}
```

and have something that will be implicitly deallocated at the same time as `Foo`.

JaviSoto:
> Property observer would not be called on weak references turning nil. :-[

Nooooooooooooooooo :(

ReactiveCocoa/Swift/Signal.swift
andersio: `take(ifRetained:)`? #NamingHardestProblem

sharplet: My vote is for `take(withinLifetimeOf:)`.

andersio:
@sharplet We may add a `LifetimeTokenProviding` protocol that accepts a `DeallocationToken` instead of a signal, and let it provide default impl. for `LifetimeProviding`. But it is purely sugar in the sense that it reduces 6 lines of code into one.

mdiep:
#3030 seems much simpler than this and thus preferable since: (a) Swift adoption becomes a single line of code and (b) the `NSObject` implementation is also simpler.

Adding a protocol does seem like it might be worthwhile though.

andersio:
The NSObject implementation is different in only returning the signal or the `Lifetime` object (which is made `internal` in this patch). The objc_sync calls have to be added anyway, since there is no thread safe guarantees on the ObjC association APIs.

andersio:
A protocol like `LifetimeProviding` might be less useful, if `Lifetime` is part of the public API. Say if we write a function `func bind(to: Something, during lifetime: Lifetime`, we are already implicitly constraining to objects that can provide a lifetime.

Having said that, my second thought says `.take(during: myObject.lifetime)` does look better than the super long label suggested here. So I gonna close this.

PR #3053, opened 2016-07-12
Drop `rac_values` in `DynamicProperty`.

andersio:
Depends on #3052.
An all-Swift implementation of `DynamicProperty`.


andersio:
Wrong base branch. Oops

PR #3062, opened 2016-07-15
Introduce the `BindingTarget` protocol.

andersio:
_Quick v3._

-

This PR introduces a new `BindingTraget` protocol, so that we can precisely model non-observable properties  that makes sense as targets of bindings.

Note that this PR is just the setting the stage. To migrate binding operators to use `BindingTarget`, we would need to have `Property` exposing `Lifetime` tokens.


mdiep:
I'm not sure I follow.

`BindableProperty` conforms to `ObservableProperty` and `SourceProperty` conforms to `BindableProperty` with no additions. So I'm not sure what this gives us. It's no easier to conform to `BindableProperty` than it is to conform to `SourceProperty`.

I'm also not wild about the names, but it seems like that's not the main point of the PR.

andersio:
Oops, it seems I handled the merge conflict incorrectly. `BindableProperty` should not conform to `ObservableProperty`.

andersio:
Oops, I handled the merge conflict incorrectly. `BindableProperty` should not conform to `ObservableProperty`:

```
`ObservableProperty` ---> `SourceProperty` <--- `BindableProperty`
```

The distinction between `BindableProperty` and `SourceProperty` is the presence of `signal` and `producer` requirement, inherited from `ObservableProperty`.

P.S. `BindableProperty` may be further restricted to require just a setter, e.g.:
```
protocol BindableProperty: class {
    associatedtype Value
    func set(_ value: Value)
}
```

andersio:
Oops, I must have handled the merge conflict incorrectly. `BindableProperty` should not conform to `ObservableProperty`:

-

But anyway, I thought of a simpler approach, so here comes the v3.

andersio:
This PR alone doesn't make much sense... I would push this again after sharplet's `Lifetime` PR is merged.

PR #3074, opened 2016-07-22
[MutexType] Just extend NSRecursiveLock and remove RecursiveLock class

ikesyo:


NachoSoto:
Ooooh of course, great idea 👍

filblue:
Is `NSRecursiveLock` going to exist on Linux? 

`RecursiveLock` frees `Property.swift` from being dependent on OS X-specific mutex implementation. When we start porting `ReactiveSwift` to other platforms, we'd better reap the fruits of this abstraction layer instead of polluting both `MutableProperty.swift` and `Atomic.swift` with platform checks.

ikesyo:
That exists in swift-corelibs-foundation: https://github.com/apple/swift-corelibs-foundation/blob/swift-3.0-PREVIEW-2/Foundation/NSLock.swift#L138

so this should not be an issue.

andersio:
If we do not care about the cost of dispatching, why don't we use `Locking` directly?

Issue #3076, opened 2016-07-23
[5.0] The current Atomic implementation and the cost of dynamic dispatch.


andersio:
I have put together a quick benchmark to justify the claimed regression introduced by #3046.

-
#### The test
Xcode 8 b2, Swift 3.0
Source Code: https://gist.github.com/andersio/b2aefdf7aa41641fce9119b9e1658b20
The test is compiled with -O.

1 million Iterations

##### Dynamic Dispatch (via Protocol Type or Existential Container)
1. `Foundation.Locking` with `Foundation.Lock`: **0.056 sec** (1.760% sd)
1. `Foundation.Locking` with `PthreadMutex`: 0.080 sec (8.625% sd)
1. `MutexType` with `Foundation.Lock`: 0.079 sec (4.961% sd)
1. `MutexType` with `PthreadMutex`: 0.078 sec (3.659% sd)

##### Static Dispatch (`final` subclasses + compiler optimisation)
1. `Foundation.Locking` with `Foundation.Lock`: 0.060 sec (4.370% sd)
1. `Foundation.Locking` with `PthreadMutex`: **0.041 sec** (3.211% sd)
1. `MutexType` with `Foundation.Lock`: 0.058 sec (3.524% sd)
1. `MutexType` with `PthreadMutex`: **0.039 sec** (4.343% sd)

##### Ideal performance (inlined lock)
**0.040 sec** (3.188%)

-
The results show that:

1. Swift's dynamic dispatch for existentials is slower than that of Objective-C, even if the protocol has a class bound...
2. `Foundation.Locking` drops the `@objc` requirement on its methods. So calls do not go through the Objective-C runtime if the type is a native Swift class.
3. We can achieve the ideal performance even with an abstraction over the lock.
<br />
<br />

andersio:
All solved. :tada:

PR #3077, opened 2016-07-23
[5.0] Merge `Atomic` changes from `master`, and adopt Foundation's `Locking`.

andersio:
Justification: #3076 

This PR merely replaces `MutexType` with `Foundation.Locking`. `Locking` has dropped the the `@objc` annotation for its method requirements in Swift 3.0, so it seems we needn't reinvent the protocol anymore.

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -8,14 +8,7 @@
 
 import Foundation
 
-internal protocol MutexType: class {
-	func lock()
-	func unlock()
-}
-
-extension RecursiveLock: MutexType {}
-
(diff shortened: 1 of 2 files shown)


mdiep:
This looks good to me. 👍 It makes sense to reuse the Foundation protocol.

I'll merge when CI passes.

NachoSoto:
It appears that `Foundation.Locking` is gone from Swift 3 beta 4 😨 still investigating.

NachoSoto:
Oh no, of course, this seems to have been fixed by removing derived data...

NachoSoto:
No, never mind, it is indeed hone?

ikesyo:
See https://github.com/apple/swift-evolution/pull/421.

PR #3078, opened 2016-07-23
Fix the inconsistency of composed properties

mdiep:
This is an alternative to #3042 that I think is quite a bit simpler. cc @andersio 

The problem, restated:

```swift
let A = MutableProperty(1)
let B = MutableProperty(100)
let AB = A.zip(with: B)
A.value = 2

// This mistakenly prints 102 because even though `AB.value` is 101,
// the producer is passed through. So when `AB.producer` is used,
// the initial producer is started again, yielding a potentially different
// set of zipped values.
AB.producer.startWithNext { print($0) }
```

This PR uses a replayed producer to provide a consistent value.


ReactiveCocoa/Swift/Property.swift
```diff
@@ -481,88 +481,38 @@ public final class Property<Value>: PropertyProtocol {
...
-			var signal: Signal<Value, NoError>!
-			atomicSignal.modify { innerSignal in
-				if signal == nil {
-					innerSignal = prepareRelaySignal(relay.signal)
-				}
-				signal = innerSignal
-			}
-			return signal
+			var extractedSignal: Signal<Value, NoError>!
+			producer.skip(first: 1).startWithSignal { signal, _ in extractedSignal = signal }
```
andersio: `skip(first: 1)` is not necessary. The current value would be sent right after the `setUp` closure is invoked.

mdiep: Ah, right.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -481,88 +481,38 @@ public final class Property<Value>: PropertyProtocol {
...
-
-		_value = { relay.value! }
-		_producer = { prepareRelayProducer(relay.producer) }
-
-		// Lazily initializes the signal of this composed property.
-		//
-		// The created signal would retain `scopedDisposable` to keep the relay
-		// alive, thus inevitably binding the relay to the lifetime of the upstream
-		// for good.
-		let atomicSignal = Atomic<Signal<Value, NoError>?>(nil)
```
andersio: This is still a problem unless we change the semantics of `Signal`. Uniquing the signal at least limited the scope of *potential* leaking.

andersio: Consider this snippet:
```
let property = source.map { $0 + 1 }
for i in 0 ..< 100000 { _ = property.signal }
```

100000 signals would be inevitably kept alive until `source` deinitializes...

mdiep: Want to open an issue to consider that separately? I think that ties into our discussion about the lifetime semantics of `Signal`.

andersio:
LGTM with a nitpick on the change to `signal`. It is pretty awesome to see `replayLazily` fit in. Haven't thought of it before. :tada:

ReactiveCocoa/Swift/Property.swift
andersio: Sure. But I just think we should stick with the status quo of the branch, shouldn't we? We might end up not changing the semantics.

mdiep: That's why I'd open a separate issue: if we don't change the semantics, we'll have an issue that says we should fix this.

andersio: Okay.

mdiep:
I think this is ready to be merged then?

PR #3079, opened 2016-07-23
[5.0] Static Dispatch for the lock of `Atomic`.

andersio:
Justification: #3076

This branch uses type erasure, and adds an internal `RecursiveAtomic` for `MutableProperty`. The downside is that `_AtomicBase` would be leaked to the public API.

An alternative would be a parameterised `Atomic` with the lock type. e.g. `Atomic<Value, PosixThreadMutex>`.

### DIFF
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -34,7 +34,7 @@ final class PosixThreadMutex: Locking {
 
 /// An atomic variable.
 public final class Atomic<Value>: AtomicProtocol {
-	private var lock: PosixThreadMutex
+	private let lock: PosixThreadMutex
 	private var _value: Value
 
 	/// Initialize the variable with the given initial value.
@@ -78,7 +78,7 @@ public final class Atomic<Value>: AtomicProtocol {
 
 /// An atomic variable which uses a recursive lock.
 internal final class RecursiveAtomic<Value>: AtomicProtocol {
-	private var lock: RecursiveLock
+	private let lock: RecursiveLock
 	private var _value: Value
 	private let didSetObserver: ((Value) -> Void)?
 
--- a/ReactiveCocoa/Swift/Atomic.swift
+++ b/ReactiveCocoa/Swift/Atomic.swift
@@ -33,8 +33,33 @@ final class PosixThreadMutex: Locking {
 }
 
 /// An atomic variable.
-public final class Atomic<Value> {
-	private var _mutex: Locking
+public final class Atomic<Value>: _AtomicBase<Value, PosixThreadMutex> {
+	/// Initialize the variable with the given initial value.
+	/// 
+	/// - parameters:
+	///   - value: Initial value for `self`.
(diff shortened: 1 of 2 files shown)


mdiep:
I'm pretty 👎🏻 on leaking details like that. 😕 

If it came down to it, I'd rather duplicate the code in `Atomic` or parameterize `Atomic` like you suggested. It's too bad that Foundation doesn't expose something like `PosixThreadMutex`.

andersio:
I have opted for duplication in take 2, since parametrising `Atomic` would lose the convenient shorthand e.g. `let interrupted = Atomic(false)`. We got some shared impl. in AtomicProtocol though.

But due to overload ambiguity, the `completion` parameter of `modify` is removed, and `RecursiveAtomic` accepts a didSet observer in its initializer instead.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -580,14 +580,14 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
...
 		/// `value`. Note that recursive sets will still deadlock because the
 		/// underlying producer prevents sending recursive events.
-		let lock = RecursiveLock()
-		lock.name = "org.reactivecocoa.ReactiveCocoa.MutableProperty"
-		_atomic = Atomic(initialValue, mutex: lock)
-
-		(signal, observer) = Signal.pipe()
+		_atomic = RecursiveAtomic(initialValue,
+															name: "org.reactivecocoa.ReactiveCocoa.MutableProperty",
+															didSet: observer.sendNext)
```
NachoSoto: [Indentation here is wrong](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/dc104c86b59533a5de2c75813d65bf96b24c516d/ReactiveCocoa/Swift/Property.swift#L589).

ReactiveCocoa/Swift/Atomic.swift
```diff
@@ -33,77 +33,79 @@ final class PosixThreadMutex: Locking {
...
+	/// Initialize the variable with the given initial value.
+	/// 
+	/// - parameters:
+	///   - value: Initial value for `self`.
+	///   - name: An optional name used to create the recursive lock.
+	internal init(_ value: Value, name: StaticString? = nil, didSet action: ((Value) -> Void)? = nil) {
+		_value = value
+
+		lock = RecursiveLock()
+		lock.name = name.map { String($0) }
```
NachoSoto: This can be `name.map(String.init)`, no?

andersio: It cannot. Overloading doesn't seem working.

andersio: We need to specify which variant though. So it would be `String.init(_:)`.

andersio:
The changes in the property names are bikeshedding. But there seems having no point to underscore the property except for a name clash with public API. (Note: AFAIK we previously use them only for `_value`.) The stdlib also uses this as a convention for the leaked implementation details, e.g. `Array._buffer`, which aren't the case of these changes.

PR #3082, opened 2016-07-24
Melt `buffer` down. 

andersio:
~~All hail the `ReplayState`. Just simplified how `replayLazily` initialises the replaying producer.~~
`replayLazily` now subsumes the implementation of `buffer`.


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1695,26 +1682,27 @@ private final class DeallocationToken {
 	}
 }
 
-extension SignalProducer {
-	private static func bufferingProducer(upTo capacity: Int) -> (SignalProducer, Signal<Value, Error>.Observer) {
-		precondition(capacity >= 0, "Invalid capacity: \(capacity)")
+private struct ReplayState<Value, Error: ErrorProtocol> {
+	let producer: SignalProducer<Value, Error>
+	let observer: SignalProducer<Value, Error>.ProducedSignal.Observer
```
mdiep: This seems a bit odd to me for a few reasons:

1. `ReplayState` is used with `Atomic` by `replayLazily` but itself uses an `Atomic<BufferState<Value, Error>>`.
2. `BufferState` is only used inside `ReplayState`.
3. `ReplayState` has a producer and observer, which means that it's more than just state information

How would you feel about attempting to unify this code a bit more?

andersio: Sure. 👍

andersio:
(Not ready yet)

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1745,75 +1730,42 @@ extension SignalProducer {
...
-			return
-		}
-
-		values.append(value)
-
-		let overflow = values.count - capacity
-		if overflow > 0 {
-			values.removeSubrange(0..<overflow)
-		}
-	}
```
mdiep: This seems like it's still useful to have as a `mutating func`. Keeping the state manipulation here makes it a little easier to follow the higher-level threading logic in the producer.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1640,81 +1638,68 @@ extension SignalProducerProtocol {
...
-
-		let lock = Lock()
-		lock.name = "org.reactivecocoa.ReactiveCocoa.SignalProducer.replayLazily"
-
 		// This will go "out of scope" when the returned `SignalProducer` goes
 		// out of scope. This lets us know when we're supposed to dispose the
 		// underlying producer. This is necessary because `struct`s don't have
 		// `deinit`.
-		let token = DeallocationToken()
+		let lifetime = DeallocationToken()
```
mdiep: This should be able to use a `Lifetime` now?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,175 @@ extension SignalProducerProtocol {
...
+			// attachments.
+			terminationEvent = event
+			observers = nil
 		}
 	}
+
+	/// Broadcast the event to all observers.
+	///
+	/// - parameter:
+	///   - event: The event to be broadcasted.
```
andersio: Oops. Spelling mistake. Will fix ASAP.

andersio:
An experiment of encapsulating all state manipulations.

ReactiveCocoa/Swift/Observer.swift
```diff
@@ -26,7 +26,7 @@ public protocol ObserverProtocol {
 
 /// An Observer is a simple wrapper around a function which can receive Events
 /// (typically from a Signal).
-public struct Observer<Value, Error: ErrorProtocol> {
+public class Observer<Value, Error: ErrorProtocol> {
```
NachoSoto: +1 I think this is more correct.
It should be `final` though!

ReactiveCocoa/Swift/Observer.swift
```diff
@@ -70,6 +70,16 @@ public struct Observer<Value, Error: ErrorProtocol> {
 	}
 }
 
+extension Observer: Hashable {
```
NachoSoto: Hmm I don't think it makes sense to give identity to `Observer` though :/

ReactiveCocoa/Swift/Observer.swift
NachoSoto: Bump

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,187 @@ extension SignalProducerProtocol {
...
-				if let value = event.value {
-					state.add(value, upTo: capacity)
-				} else {
-					// Disconnect all observers and prevent future
-					// attachments.
-					state.terminationEvent = event
-					state.observers = nil
+					// Terminate the replay loop.
+					return
+				} catch ReplayError<Value>.pending(let values) {
```
andersio: The compiler somehow does not permit `catch let ReplayError<Value>.pending(values)`.

andersio:
Please note that the semantic has slightly changed: the underlying producer is captured by the bootstrap disposable, so it would be released after it is being started. In the original implementation, it was captured by the replayed producer.

mdiep:
I merged the `RAC5` branch, so this will need to be reopened against `master`. Sorry!

PR #3084, opened 2016-07-25
Handle the case when throttle: previous date is newer than current da…

Patrickuang:
…te. For example, previous date is 5 mins ahead of the standard time, adjust the device time 5 mins back to the standard time and trigger the signal again, the scheduleDate is (5 mins + interval) later.

### DIFF
--- a/ReactiveCocoa/Swift/Scheduler.swift
+++ b/ReactiveCocoa/Swift/Scheduler.swift
@@ -452,12 +452,12 @@ public final class TestScheduler: DateSchedulerType {
 		advanceToDate(NSDate.distantFuture())
 	}
 	
-	/// Retreats the virtualized clock by the given interval.
+	/// Rewinds the virtualized clock by the given interval.
 	/// This simulates that user changes device date.
 	///
 	/// - parameters:
 	///   - interval: Interval by which the current date will be retreated.
-	public func retreatCurrentDateByInterval(interval: NSTimeInterval) {
+	public func rewindByInterval(interval: NSTimeInterval) {
 		lock.lock()
 		
 		let newDate = currentDate.dateByAddingTimeInterval(-interval)
--- a/ReactiveCocoa/Swift/Scheduler.swift
+++ b/ReactiveCocoa/Swift/Scheduler.swift
@@ -451,4 +451,20 @@ public final class TestScheduler: DateSchedulerType {
 	public func run() {
 		advanceToDate(NSDate.distantFuture())
 	}
+	
+	/// Retreats the virtualized clock by the given interval.
+	/// This simulates that user changes device date.
+	///
+	/// - parameters:
+	///   - interval: Interval by which the current date will be retreated.
+	public func retreatCurrentDateByInterval(interval: NSTimeInterval) {
+		lock.lock()
+		
+		let newDate = currentDate.dateByAddingTimeInterval(-interval)
+		assert(currentDate.compare(newDate) != .OrderedAscending)
+		_currentDate = newDate
+		
+		lock.unlock()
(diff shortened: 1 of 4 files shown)


liscio:
This is a great bug fix, but could you also add a test that exercises (and hence describes) the bug that you resolved and the behavior you're asserting? Thanks! 👍

Patrickuang:
Yeah, working on UT. :smile:

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1452,7 +1452,13 @@ extension SignalType {
 					var state = state
 					state.pendingValue = value
 
-					let proposedScheduleDate = state.previousDate?.dateByAddingTimeInterval(interval) ?? scheduler.currentDate
+					let proposedScheduleDate: NSDate
+					if let previousDate = state.previousDate where previousDate.compare(scheduler.currentDate) == .OrderedAscending {
```
mdiep: I think this deserves a comment explaining why `previousDate` could be at or after `scheduler.currentDate`.

Patrickuang: Sorry for the late update on the UT. I was stuck in some personal affairs and still working on that.
Let me give more detail on the situation you mentioned. `previousDate` could be after `scheduler.currentDate` if the device time is changed back during `throttle` . For example, let 's set `throttle` duration to **60 seconds**.  

| Standard Date  | Device Date | Event | 
|---|---|---|
|**00:00:00 AM**|**00:05:00 AM**|Value 0 is sent, and the `proposedScheduleDate` is **00:06:00 AM**, `previousDate` is **00:05:00 AM** |
|**00:00:30 AM**|**00:05:30 AM**|user finds the device date is 5 minutes ahead of standard date, so change device date back to **00:00:30 AM**, `proposedScheduleDate` is still **00:06:00 AM**|
|**00:00:45 AM**|**00:00:45 AM**|Value 1 is sent, and `previousDate` is **00:05:00 AM**, `scheduler.currentDate` is **00:00:45 AM**. So based on old code, `proposedScheduleDate` is **00:06:00 AM**. Thus, there will be a **6 minutes** delay (not **1 minute**) to send Next event.   | 

And the root cause of this problem is the change of device date.

mdiep: Oh, I meant that a comment should be added to the code. ☺️

Patrickuang:
Does anyone know why the Carthage job was failed?

RuiAAPeres:
@Patrickuang sometimes it randomly fails. I restarted it.

ReactiveCocoa/Swift/Scheduler.swift
```diff
@@ -451,4 +451,20 @@ public final class TestScheduler: DateSchedulerType {
 	public func run() {
 		advanceToDate(NSDate.distantFuture())
 	}
+	
+	/// Retreats the virtualized clock by the given interval.
+	/// This simulates that user changes device date.
+	///
+	/// - parameters:
+	///   - interval: Interval by which the current date will be retreated.
+	public func retreatCurrentDateByInterval(interval: NSTimeInterval) {
```
mdiep: I think the naming of this method is a bit off. It doesn't match `advanceByInterval` because it includes `CurrentDate`, but I also don't love the verb _retreat_ in this context.

What would you think of renaming this `rewindByInterval`?

PatKuang: Thanks for pointing out this. Renaming has been committed.

mdiep:
Thanks for the PR! This is great. ✨

PR #3085, opened 2016-07-25
Signal and SignalProducer: added timeoutWithError overloads for NoError

### LINKED ISSUE #3083: timeoutWithError for NoError should not require promoteErrors
If signal/producer does not initiate any error (NoError), currently we have to adopt it to any error type with promoteErrors if we want to use timeoutWithError on it.

Can we add timeoutWithError for NoError type also, in similar way it was added for flatMap functions  https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2542

NachoSoto:
Fixes #3083.

This is similar to the overloads introduced in #2542.

### DIFF
--- a/ReactiveCocoa/Swift/Signal.swift
+++ b/ReactiveCocoa/Swift/Signal.swift
@@ -1821,4 +1821,31 @@ extension SignalType where Error == NoError {
 			}
 		}
 	}
+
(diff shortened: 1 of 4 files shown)


mdiep:
Nice. 👍 

I'm not sure what's going on with CI though. :confused:

mdiep:
Looks like Quick needs to be updated.

PR #3086, opened 2016-07-26
Update Quick back to Quick/Quick

mdiep:
The commit we were pinned to seems to have been deleted.


sharplet:
Wow that explains what I was seeing earlier today, I couldn't figure out why `carthage update` was failing! :sweat_smile:

Issue #1122, opened 2014-02-19
Crash [RACKVOTrampoline retain] message sent to deallocated instance


rspeyer:
I have been observing an issue in our iOS application where users are crashing in KVO notification flows for subscriptions managed by ```RACObserve```. Based on what I was seeing, it appeared that the problem has to do with the object which owns the subscription being managed on a different queue than where the property change occurs.

I was able to write test code to reproduce this crash consistently. which I have pasted below. Please let me know if I can provide any other information

```objective-c
@interface AppDelegate()
@property int testProperty;
@end

@implementation AppDelegate
- (BOOL)          application:(UIApplication *)application
didFinishLaunchingWithOptions:(NSDictionary *)launchOptions  {
    int numIterations = 10000;
    __block RACDisposable *dispose;
    for (int i=0; i< numIterations; ++i) {
        dispatch_async(dispatch_get_main_queue(), ^{
            [dispose dispose];
            dispose = nil;
            
            dispose = [RACObserve(self, testProperty) subscribeNext:^(id x) {
                NSLog(@"XXXXXXXX %@", x);
            }];
            
            dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
                self.testProperty++;
            });
        });
    }
}
@end
```

joshaber:
Hopefully fixed by #1627. :pray:

PR #3087, opened 2016-07-26
Remove ObjC API dependency in `DynamicProperty`.

andersio:
Depends on #3030.
(Changed also the association policy of `NSObject.lifetime`)


ikesyo:
Maybe we would need a singleton observer instance that was originally implemented in #1276 and #1627, discussed in #1122.

mdiep:
> Maybe we would need a singleton observer instance

That's a good idea. 👍🏻 Thanks for bringing that up!

mdiep:
I also this we should expose a method like `rac_valuesForKeyPath`—that's still useful in many circumstances.

andersio:
I was thinking about moving `DynamicProperty` initialisers to be private, and in favour of something like `NSObject.property(forKeyPath:[bridging|casting]:)`.

Since DynamicProperty can be made a stateless proxy anyway, its producer would just be the same as `rac_values`.

andersio:
The previous version passed all the tests I ported from `RACKVOProxySpec`. But I have implemented the singleton proxy anyway.

andersio:
The approach of this PR passed the tests ported from `RACKVOProxySpec` (with automatic KVO notification).

AFAICU, the singleton proxy is a workaround to prevent observed objects from sending messages to any deinitialising observing objects.

The `KeyValueObserver` in this PR has itself retained its `disposable`, which is used to abolish the observation. The end result is effectively the same, but without serialising all KVO accesses with one shared lock.

Note that this approach does not guard against the race condition of notifications vs. removal of observers. Nor do I think this should be addressed in the `KeyValueObserver`, since setting a property on multiple threads without proper synchronisation should result in undefined behavior in the first place.

That's said we can always go for the battle-tested singleton proxy, if you guys prefer it.

ReactiveCocoa/Swift/NSObject+KeyValueObserving.swift
```diff
@@ -0,0 +1,66 @@
...
+	///   A disposable that would tear down the observation upon disposal.
+	static func observe(object: NSObject, keyPath: String, options: NSKeyValueObservingOptions, action: ([String: AnyObject?]) -> Void) -> Disposable {
+		let observer = KeyValueObserver(action: action)
+
+		object.addObserver(observer,
+		                   forKeyPath: keyPath,
+		                   options: [.Initial, .New],
+		                   context: keyValueObserverKey)
+
+		unowned(unsafe) let unsafeObject = object
```
mdiep: This seems like it should be `weak`?

ReactiveCocoa/Swift/NSObject+KeyValueObserving.swift
```diff
@@ -0,0 +1,66 @@
...
+
+	let action: ([String: AnyObject?]) -> Void
+
+	private init(action: ([String: AnyObject?]) -> Void) {
+		self.action = action
+	}
+
+	override func observeValueForKeyPath(keyPath: String?, ofObject object: AnyObject?, change: [String : AnyObject]?, context: UnsafeMutablePointer<Void>) {
+		if context == keyValueObserverKey {
+			action(change!)
```
mdiep: How about if we send the new value here instead of passing the change dictionary?

That would keep the knowledge about KVO centralized in this class.

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -52,24 +54,12 @@ import enum Result.NoError
 	public init(object: NSObject?, keyPath: String) {
```
mdiep: Outside the scope of this PR, but why is this `object` an optional? 😕

mdiep:
This is looking pretty good. Just a few notes.

But we'll need to wait on #3030.

ReactiveCocoa/Swift/NSObject+KeyValueObserving.swift
andersio: It is expected to be callable during deinitialization, at which point the `weak` references would be `nil` by that time.

ReactiveCocoa/Swift/DynamicProperty.swift
andersio: I don't know either. But how do you think about hiding the initialisers and instead exposing `NSObject.property(forKeyPath:...)`?

ReactiveCocoa/Swift/NSObject+KeyValueObserving.swift
andersio: New+Old and Prior+New might worth exposing. So I changed to send the dictionary instead. Moreover, the key of the dictionary would be typed as a string enum in Swift 3.0. So wrapping it doesn't worth much.

mdiep: I think we should default to sending the simplest thing that we need now. Let's not try to anticipate future needs.

ReactiveCocoa/Swift/DynamicProperty.swift
mdiep: I think I'd rather keep the initializers here.

We should consider trying to remove the `Optional` here in a separate PR though.

mdiep:
#3030 has been merged, so this needs an update and then I think it will be ready to go.

andersio:
The Xcode 7.3 Carthage job keeps failing like #3098. 😕

mdiep:
Could you add some tests for key paths that have multiple keys? (`foo.bar.baz`) I want to make sure we handle that correctly. The RAC2 code goes through quite a bit of work for those, but I'm not entirely sure why.

andersio:
It seems functional without all the workaround in the `RACKVOWrapper` category.

mdiep:
It looks like there's a legitimate failure. 😢 

```
ReactiveCocoaTests.PropertySpec
  DynamicProperty__should_observe_changes_in_a_nested_weak_key_path, failed: caught "NSInternalInconsistencyException", "An instance 0x7fc42b747fc0 of class _TtC18ReactiveCocoaTestsP33_14EF8FEBE6D14DD4F5893E451ED0400216ObservableObject was deallocated while key value observers were still registered with it. Current observation info: <NSKeyValueObservationInfo 0x7fc42b743390> (
```

We probably need to add dealloc observables like `RACKVOWrapper` does.

andersio:
Hmm, interesting. I didn't catch it locally.

andersio:
(Not yet ready, messed up a few bits when cleaning things up)

andersio:
`RACPropertySubscribing` has some kind of locking which IMO isn't needed here, since `SignalProducer` is synchronous by default, and we have the control of the lifetime of the observer.

Having said that, do we need the producer to weakly reference the object?

andersio:
P.S. some comments are just copied from `RACKVOWrapper`. But there is one I am not quite fond of, given how `rac_valuesForKeyPath` is working fine with Core Data at least for my projects.

>		// Attempting to observe non-weak properties using dynamic getters will
>		// result in broken behavior, so don't even try.

mdiep:
Thanks! I wish we could have stuck with your initial implementation. It was so much simpler! 😅

PR #3097, opened 2016-07-29
Pre- and post-start hooks in `SignalProducer.on`.

### LINKED ISSUE #3095: Executing order of SignalProducer.on(start:) is reversed compared to SignalProducer.on(next:)
I am using `on(start: next:)` to delivery states from view model and refresh on view controller, as follows:

```
// in view model
loadDataProducer.on(start: { // closure 1
    loading = true
})
// in view controller
.on(start: { // closure 2
    cover.hidden = viewModel.loading
})
```

It is a trivial example above, but the point is that it cannot make state dependent chaining closure of `on(start:)`, since `closure 2` is executed before `closure 1`, which is against order of signal flow like `next` and `complete`. 

I wonder if this is intentional? If so, how should I use `on(start:)` to react on each layer?

andersio:
Fix #3095.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1481,8 +1481,12 @@ public func timer(interval: NSTimeInterval, onScheduler scheduler: DateScheduler
 extension SignalProducerType {
 	/// Injects side effects to be performed upon the specified producer events.
 	///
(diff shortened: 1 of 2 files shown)


andersio:
Xcode 7.3 Carthage job keeps failing. Hmm...

mdiep:
👍🏻

PR #3096, opened 2016-07-29
Start SignalProducers in order as defined for combineLatest() and zip()

chrrasmussen:
Fixes #3094

PR #3099, opened 2016-07-29
SignalProducer.combineLatest: removed no longer necessary workaround

NachoSoto:
This was originally introduced in #2751, but it's not necessary anymore. Tests pass with `-O` now!

This will have to be changed to `liftLeft` once rebasing #3096.


mdiep:
👍🏻

PR #3106, opened 2016-08-05
Introduce non-lazy replayed producers.

andersio:
Depend on #3082.

1. Renamed `replayLazily(upTo:)` to `replay(upTo:)`. (`SignalProducer` is generally lazy anyway...)
2. Introduced `replay(upTo:startsLazily:)` to allow starting the underlying producer in place.

Motivation: Needed in a WIP `DynamicProperty` patch to cache the latest value without attaching a dummy observer.


NachoSoto:
I'm very very 👎. **Operators should not have side effects**.

andersio:
Hmm after the second thought, it is not actually necessary with the cons I haven't considered before.

Issue #3111, opened 2016-08-05
Reliability of `Lifetime`.


andersio:
Consider this snippet:
```
var object: NSObject? = NSObject()
var lifetime: Lifetime = object.rac_lifetime

object!.values(forKeyPath: "...").start()

var object = nil /// KVO Exception: There are still registered observers.
var lifetime = nil
```
The external retainment of the lifetime token causes the havoc. Specifically, the completion is being deferred from the actual deinitialization due to the external retainment.

To uphold the reliability of the deallocation signal, the `Lifetime` API has to be decoupled with the deallocation detection, at least for `NSObject.rac_lifetime`. Moreover, adopting classes have to kept the lifetime token private to themselves by convention.

#### Possible solutions

1. Revive `DeallocationToken`, and make `Lifetime` a wrapper of the `DeallocationToken`'s `ended` signal.
 ```
	private let token = DeallocationToken()
	public var rac_lifetime: Lifetime { return Lifetime(token) }
 ```

2. Establish a convention that requires an lifetime ended **signal** instead, so that the token can be declared `private`. Use `take(until:)` directly.
 ```
	private let lifetime = Lifetime()
	public var rac_lifetime: Signal<(), NoError> { return lifetime.ended }
 ```

Either solutions cannot be implemented as a one liner (for native Swift citizens) as what `Lifetime` is at the moment, AFAICT.

PR #3112, opened 2016-08-05
Decoupled the deinitialization detection from `Lifetime`.

andersio:
Solution 1 in issue #3111 (Reliability of `Lifetime`).

Recap: The split prevents the completion of `Lifetime`'s signal from being deferred from the actual deinitialization due to external retainments.

### DIFF
--- a/ReactiveCocoa/Swift/Lifetime.swift
+++ b/ReactiveCocoa/Swift/Lifetime.swift
@@ -1,21 +1,41 @@
 import enum Result.NoError
 
-/// Provides a signal that completes when the lifetime object is deinitialized.
-///
-/// When assigned to a property of another object, provides a hook to
-/// observe when that object goes out of scope.
+/// Represent the lifetime of an object, and provide a hook to observe when the
+/// object deinitializes.
 public final class Lifetime {
 	/// A signal that sends a Completed event when the lifetime ends.
 	public let ended: Signal<(), NoError>
 
-	private let endedObserver: Signal<(), NoError>.Observer
-
-	public init() {
-		(ended, endedObserver) = Signal.pipe()
+	public init(_ token: Token) {
+		ended = token.ended
 	}
 
-	deinit {
-		endedObserver.sendCompleted()
+	/// A token object which completes its signal when it deinitializes.
+	///
+	/// It is generally used in conjuncion with `Lifetime` as a private
+	/// deinitialization trigger.
+	///
+	/// ```
+	/// class MyController {
+	///		private let token = Lifetime.Token()
+	///		public var lifetime: Lifetime {
+	///			return Lifetime(token)
+	///		}
+	/// }
+	/// ```
+	public final class Token {
+		/// A signal that sends a Completed event when the lifetime ends.
+		private let ended: Signal<(), NoError>
+
+		private let endedObserver: Signal<(), NoError>.Observer
+
+		public init() {
+			(ended, endedObserver) = Signal.pipe()
+		}
+
+		deinit {
+			endedObserver.sendCompleted()
+		}
 	}
 }
 
@@ -27,13 +47,13 @@ extension NSObject {
 		objc_sync_enter(self)
 		defer { objc_sync_exit(self) }
 
-		if let lifetime = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime? {
-			return lifetime
+		if let token = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime.Token? {
+			return Lifetime(token)
 		}
 
-		let lifetime = Lifetime()
-		objc_setAssociatedObject(self, &lifetimeKey, lifetime, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+		let token = Lifetime.Token()
+		objc_setAssociatedObject(self, &lifetimeKey, token, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
 
-		return lifetime
+		return Lifetime(token)
 	}
 }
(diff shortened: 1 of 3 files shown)


sharplet:
What do you think about using a nested type `Lifetime.Token`?

Also, in your Swift usage sample, prefixing the property as `rac_lifetime` isn't strictly necessary because it's not an extension of a type in another module. What do you think about just naming the property `lifetime` in the example?

andersio:
Updated with your suggestions. :tada:

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -1,21 +1,41 @@
 import enum Result.NoError
 
-/// Provides a signal that completes when the lifetime object is deinitialized.
-///
-/// When assigned to a property of another object, provides a hook to
-/// observe when that object goes out of scope.
+/// Represent the lifetime of an object, and provide a hook to observe when the
```
NachoSoto: This should be third person: `Represents`.

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -27,13 +47,13 @@ extension NSObject {
 		objc_sync_enter(self)
 		defer { objc_sync_exit(self) }
 
-		if let lifetime = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime? {
-			return lifetime
+		if let token = objc_getAssociatedObject(self, &lifetimeKey) as! Lifetime.Token? {
+			return Lifetime(token)
```
NachoSoto: Any reason not to store the `Lifetime`, to avoid allocating a different one each time?

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -1,21 +1,41 @@
...
+/// object deinitializes.
 public final class Lifetime {
 	/// A signal that sends a Completed event when the lifetime ends.
 	public let ended: Signal<(), NoError>
 
-	private let endedObserver: Signal<(), NoError>.Observer
-
-	public init() {
-		(ended, endedObserver) = Signal.pipe()
+	public init(_ token: Token) {
```
NachoSoto: Can you document this? In particular I would mention that `Lifetime` does not hold a strong reference to `Token`, as that would defeat the purpose.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1710,11 +1715,11 @@ extension SignalProducerProtocol {
 			disposable += {
 				// Don't dispose of the original producer until all observers
 				// have terminated.
-				lifetime = nil
+				token = nil
 			}
 
 			if shouldStartUnderlyingProducer {
-				self.take(during: lifetime!)
+				self.take(during: lifetime)
```
NachoSoto: Can you put `.take` in its own line while you're here?

ReactiveCocoaTests/Swift/LifetimeSpec.swift
```diff
@@ -42,6 +42,34 @@ final class LifetimeSpec: QuickSpec {
...
+				}
+
+				let lifetime = object.value!.lifetime
+
+				// Suppress the "never read" warning.
+				_ = lifetime
+
+				object.value = nil
+
+				expect(events) == ["completed"]
```
NachoSoto: Instead of having magic strings I'd recommend storing the events themselves. Otherwise this test is pretty fragile.

ReactiveCocoaTests/Swift/LifetimeSpec.swift
```diff
@@ -42,6 +42,34 @@ final class LifetimeSpec: QuickSpec {
...
+					case .completed:
+						events.append("completed")
+					case .failed:
+						events.append("failed")
+					case .interrupted:
+						events.append("interrupted")
+					}
+				}
+
+				let lifetime = object.value!.lifetime
```
NachoSoto: I would define this before the observation to avoid duplicating `object.value!.lifetime`.

ReactiveCocoaTests/Swift/LifetimeSpec.swift
```diff
@@ -42,6 +42,34 @@ final class LifetimeSpec: QuickSpec {
...
+
+				// Suppress the "never read" warning.
+				_ = lifetime
```
NachoSoto: Then you wouldn't need this.

NachoSoto:
> Also, in your Swift usage sample, prefixing the property as rac_lifetime isn't strictly necessary because it's not an extension of a type in another module. What do you think about just naming the property lifetime in the example?

:+1:

ReactiveCocoa/Swift/Lifetime.swift
andersio: The compiler has stack promotions for objects.

NachoSoto: Huh? I mean for every caller to this property.

mdiep:
I merged the `RAC5` branch, so this will need to be reopened against `master`. Sorry!

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1684,10 +1684,15 @@ extension SignalProducerProtocol {
...
-		let lifetime = Lifetime()
+		let token = Lifetime.Token()
+		let lifetime = Lifetime(token)
 
 		return SignalProducer { observer, disposable in
-			var lifetime: Lifetime? = lifetime
+			var token: Lifetime.Token? = token
+
+			// Suppress the "never read" warning.
+			_ = token
```
mdiep: I think you could create the `lifetime` here instead?

mdiep:
This bums me out, but it makes sense.

NachoSoto:
> I merged the RAC5 branch, so this will need to be reopened against master. Sorry!

And please remember to address the comments in here!

PR #3119, opened 2016-08-07
Decouple the deinitialization detection from `Lifetime`.

andersio:
Originally #3112.
Solves #3111.

-

> Instead of having magic strings I'd recommend storing the events themselves. Otherwise this test is pretty fragile.

Since `Event` is not equatable in this case, this is probably why @sharplet had written the test in such way. Anyway, I have made them to now check only for completion, which is what the lifetime contract is all about.

### DIFF
--- a/ReactiveCocoa/Swift/Lifetime.swift
+++ b/ReactiveCocoa/Swift/Lifetime.swift
@@ -1,25 +1,55 @@
 import enum Result.NoError
 
-/// Provides a signal that completes when the lifetime object is deinitialized.
-///
-/// When assigned to a property of another object, provides a hook to
-/// observe when that object goes out of scope.
+/// Represents the lifetime of an object, and provides a hook to observe when
+/// the object deinitializes.
 public final class Lifetime {
 	/// A signal that sends a Completed event when the lifetime ends.
 	public let ended: Signal<(), NoError>
 
-	private let endedObserver: Signal<(), NoError>.Observer
-
-	public init() {
-		(ended, endedObserver) = Signal.pipe()
+	/// Initialize a `Lifetime` from a lifetime token, which is expected to be
+	/// associated with an object.
+	///
+	/// - important: The resulting lifetime object does not retain the lifetime
+	///              token.
+	///
+	/// - parameters:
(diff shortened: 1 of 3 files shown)


ReactiveCocoaTests/Swift/LifetimeSpec.swift
```diff
@@ -42,27 +42,25 @@ final class LifetimeSpec: QuickSpec {
...
-					case .failed:
-						events.append("failed")
-					case .interrupted:
-						events.append("interrupted")
-					}
-				}
+				var isCompleted = false
+				lifetime.ended.observeCompleted { isCompleted = true }
 
 				object.value = nil
```
NachoSoto: I would check that  `isCompleted` is `false` here.

NachoSoto:
:shipit:

ReactiveCocoaTests/Swift/LifetimeSpec.swift
andersio: observation should have no side effect, shouldn't it?

ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -1,25 +1,55 @@
...
-		(ended, endedObserver) = Signal.pipe()
+	/// Initialize a `Lifetime` from a lifetime token, which is expected to be
+	/// associated with an object.
+	///
+	/// -important: The resulting lifetime object does not retain the lifetime
+	///             token.
+	///
+	/// -parameters:
+	///  -token: A lifetime token for detecting the deinitialization of the
+	///          associated object.
```
mdiep: There should be a ` ` after each `-`s here.

ReactiveCocoaTests/Swift/LifetimeSpec.swift
mdiep: I think it mostly makes it a bit clearer that you're testing that `object.value = nil` is expected to trigger `isCompleted = true`. But I can envision a bug where the completed block could complete immediately.

mdiep:
Needs a merge, then it's ready to 🚢.

PR #3120, opened 2016-08-07
Streamline the `replayLazily` implementation.

andersio:
Originally #3082 

> Please note that the semantic has changed slightly: the underlying producer is captured by the bootstrap disposable only. So the producer would be released after it is being started for the first time.

> In the original implementation, it was captured by the start handler of the replayed producer, and was retained as long as there is still a copy of the replayed producer.

### DIFF
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1783,29 +1783,36 @@ private struct ReplayState<Value, Error: Swift.Error> {
 		// the `Observer` to track them directly.
 		let id = ObjectIdentifier(observer)
 
-		func replayCompletion() -> Result<RemovalToken?, ReplayError<Value>> {
-			if let event = terminationEvent {
-				observer.action(event)
-			}
-			return .success(observers?.insert(observer))
+		switch replayBuffers[id] {
+		case .none where !values.isEmpty:
+			// No in-flight replay buffers was found, but the `ReplayState` has one or
+			// more cached values in the `ReplayState`. The observer should replay
+			// them before attempting to observe again.
+			replayBuffers[id] = []
+			return .failure(ReplayError(values: values))
+
+		case let .some(buffer) where !buffer.isEmpty:
+			// An in-flight replay buffer was found with one or more buffered values.
+			// The observer should replay them before attempting to observe again.
+			defer { replayBuffers[id] = [] }
+			return .failure(ReplayError(values: buffer))
+
+		case let .some(buffer) where buffer.isEmpty:
+			// Since an in-flight but empty replay buffer was found, the observer is
+			// ready to be attached to the `ReplayState`.
+			replayBuffers.removeValue(forKey: id)
+
+		default:
+			// No values has to be replayed. The observer is ready to be attached to
+			// the `ReplayState`.
+			break
 		}
 
-		if let buffer = replayBuffers[id] {
-			if buffer.isEmpty {
-				replayBuffers.removeValue(forKey: id)
-				return replayCompletion()
-			} else {
-				defer { replayBuffers[id] = [] }
-				return .failure(ReplayError(values: buffer))
-			}
-		} else {
-			if values.isEmpty {
-				return replayCompletion()
-			} else {
-				replayBuffers[id] = []
-				return .failure(ReplayError(values: values))
-			}
+		if let event = terminationEvent {
+			observer.action(event)
 		}
+
+		return .success(observers?.insert(observer))
 	}
 
 	/// Enqueue the supplied event to the replay state.
@@ -1819,30 +1826,25 @@ private struct ReplayState<Value, Error: Swift.Error> {
 				replayBuffers[key]!.append(value)
 			}
 
-			if capacity == 0 {
-				// `state.values` cannot be non-empty with a capacity of zero.
-				return
-			}
+			switch capacity {
+			case 0:
+				// With a capacity of zero, `state.values` can never be filled.
+				break
 
-			if capacity == 1 {
+			case 1:
 				values = [value]
-				return
+
+			default:
+				values.append(value)
+
+				let overflow = values.count - capacity
+				if overflow > 0 {
+					values.removeFirst(overflow)
+				}
 			}
 
-			values.append(value)
-
-			let overflow = values.count - capacity
-			if overflow > 0 {
-				values.removeSubrange(0 ..< overflow)
-			}
-
-		case .failed:
-			fallthrough
-		case .completed:
-			fallthrough
-		case .interrupted:
-			// Disconnect all observers and prevent future
-			// attachments.
+		case .completed, .failed, .interrupted:
+			// Disconnect all observers and prevent future attachments.
 			terminationEvent = event
 			observers = nil
 		}
--- a/ReactiveCocoa/Swift/SignalProducer.swift
+++ b/ReactiveCocoa/Swift/SignalProducer.swift
@@ -1662,8 +1662,6 @@ extension SignalProducerProtocol {
 	/// This operator is only recommended when you absolutely need to introduce
 	/// a layer of caching in front of another `SignalProducer`.
 	///
-	/// - note: This operator has the same semantics as `SignalProducer.buffer`.
-	///
 	/// - precondtion: `capacity` must be non-negative integer.
 	///
 	/// - parameters:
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
-			let replayBuffer = ReplayBuffer<Value>()
-			var replayValues: [Value] = []
-			var replayToken: RemovalToken?
-			var next = state.modify { state in
-				replayValues = state.values
-				if replayValues.isEmpty {
-					token = state.observers?.insert(observer)
-				} else {
-					replayToken = state.replayBuffers.insert(replayBuffer)
+		let bootstrap = ActionDisposable { [ended = lifetime.ended] in
```
mdiep: Using a disposable, while convenient, is somewhat misleading. I think it'd be preferable to use a `Atomic<(() -> Void)?>` directly.

I'm also not wild about the name `bootstrap` here. I think `start`, or something like it, would probably be more accurate.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
+					return
+				} catch ReplayError<Value>.pending(let values) {
+					values.forEach(observer.sendNext)
+				} catch let error {
+					preconditionFailure("Unexpected error: \(error).")
 				}
-			}
-
-			originalState.observers?.forEach { $0.action(event) }
+			} while true
```
mdiep: This could just be a `while` loop since the condition is `true`.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
+	///   pending for being replayed by the observer.
+	///
+	/// - returns:
+	///   The removal token of the observer, or `nil` if the replay state has
+	///   already terminated.
+	mutating func observe(_ observer: Signal<Value, Error>.Observer) throws -> RemovalToken? {
+		// The only use case is in `replayLazily` which has always a unique observer
+		// created for each produced signal. So we can use the ObjectIdentifier to
+		// track them directly.
+		let id = ObjectIdentifier(observer)
```
mdiep: TIL!

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
+				replayBuffers.removeValue(forKey: id)
+				_ = terminationEvent.map(observer.action)
+				return observers?.insert(observer)
+			} else {
+				defer { buffer.values.removeAll() }
+				throw ReplayError<Value>.pending(values: buffer.values)
+			}
+		} else {
+			if values.isEmpty {
+				_ = terminationEvent.map(observer.action)
```
mdiep: I don't love the side-effect use of `map` here and above.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
+/// A uniquely identifying token for `Observer`s that are replaying values in
+/// `ReplayState`.
 private final class ReplayBuffer<Value> {
 	private var values: [Value] = []
 }
 
-private struct BufferState<Value, Error: Swift.Error> {
-	/// All values in the buffer.
+private enum ReplayError<Value>: Error {
+	case pending(values: [Value])
```
mdiep: This could just be a `struct` since it only has one case?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
-					// Disconnect all observers and prevent future
-					// attachments.
-					state.terminationEvent = event
-					state.observers = nil
+					// Terminate the replay loop.
+					return
+				} catch ReplayError<Value>.pending(let values) {
+					values.forEach(observer.sendNext)
+				} catch let error {
+					preconditionFailure("Unexpected error: \(error).")
```
mdiep: I think using `Result` instead of `throws` would make for clearer code and let us avoid this unreachable case.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
-				next = state.modify { state in
-					replayValues = replayBuffer.values
-					replayBuffer.values = []
-					if replayValues.isEmpty {
-						if let replayToken = replayToken {
-							state.replayBuffers.remove(using: replayToken)
+					if let token = token {
+						disposable += {
+							state.modify { state in
+								state.removeObserver(using: token)
```
mdiep: I think using `$0` might actually be clearer here. What do you think?

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
-/// BufferState.
+/// A uniquely identifying token for `Observer`s that are replaying values in
+/// `ReplayState`.
 private final class ReplayBuffer<Value> {
 	private var values: [Value] = []
 }
 
-private struct BufferState<Value, Error: Swift.Error> {
-	/// All values in the buffer.
+private enum ReplayError<Value>: Error {
```
mdiep: `ReplayBuffer` and `ReplayError` could be inner class/structs on `ReplayState`.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
-	/// The observers currently attached to the buffered producer, or nil if the
-	/// producer was terminated.
+	/// The observers currently attached to the caching producer, or `nil` if the
+	/// caching producer was terminated.
 	var observers: Bag<Signal<Value, Error>.Observer>? = Bag()
 
-	/// The set of unused replay token identifiers.
-	var replayBuffers: Bag<ReplayBuffer<Value>> = Bag()
+	/// The set of in-flight replay tokens.
+	var replayBuffers: [ObjectIdentifier: ReplayBuffer<Value>] = [:]
```
mdiep: This makes `ReplayBuffer` unnecessary; a `[Value]` could be used instead.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
-		if capacity == 0 {
-			values = []
-			return
-		}
+	/// Enqueue the supplied event to the replay state.
+	///
+	/// - parameter:
+	///   - event: The event to be cached.
+	mutating func enqueue(_ event: Event<Value, Error>) {
+		if let value = event.value {
```
mdiep: I think it'd be preferable to use a `switch` here and explicitly list the cases in the `else` condition.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,188 @@ extension SignalProducerProtocol {
...
+			terminationEvent = event
+			observers = nil
 		}
 	}
+
+	/// Broadcast the event to all observers.
+	///
+	/// - parameter:
+	///   - event: The event to be broadcast.
+	func broadcast(_ event: Event<Value, Error>) {
```
mdiep: I don't love having a side-effecting method on a value struct like this. I think it'd be better to move the code inline.

mdiep:
I left some more suggestions for how I think this can be improved. Sorry for accidentally closing!

ReactiveCocoa/Swift/SignalProducer.swift
andersio: We preferred named parameters for `Atomic.modify` in the rest of the codebase though. Should we change those at some point too?

ReactiveCocoa/Swift/SignalProducer.swift
andersio: We can't nest types inside generic types. :-[

ReactiveCocoa/Swift/SignalProducer.swift
andersio: The intention here is to keep all properties encapsulated via `private`, and exposes only the behaviors via `fileprivate` once implemented.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,189 @@ extension SignalProducerProtocol {
...
-					replayToken = state.replayBuffers.insert(replayBuffer)
+		return SignalProducer { observer, disposable in
+			// Don't dispose of the original producer until all observers
+			// have terminated.
+			disposable += { _ = lifetime }
+
+			while true {
+				var result: Result<RemovalToken?, ReplayError<Value>>!
+				state.modify {
+					result = $0.observe(observer)
```
mdiep: Maybe we should add a version of `modify` that could help here?

```swift
func modify<Result>(_ action: @noescape (inout Value) throws -> Result) rethrows -> Result
```

🤔

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,189 @@ extension SignalProducerProtocol {
...
+			}
 
-		let overflow = values.count - capacity
-		if overflow > 0 {
-			values.removeSubrange(0..<overflow)
+		case .failed:
+			fallthrough
+		case .completed:
+			fallthrough
+		case .interrupted:
```
mdiep: Can you list these like `case .failed, .completed, .interrupted:` instead?

(Sorry that wasn't clear before.)

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1674,172 +1672,189 @@ extension SignalProducerProtocol {
...
+
+			if capacity == 0 {
+				// `state.values` cannot be non-empty with a capacity of zero.
+				return
+			}
+
+			if capacity == 1 {
+				values = [value]
+				return
+			}
```
mdiep: What do you think of `switch`ing this instead? 🤔

```swift
			switch capacity {
			case 0:
				// `state.values` cannot be non-empty with a capacity of zero.
				break
				
			case 1:
				values = [value]

			default:
				values.append(value)

				let overflow = values.count - capacity
				if overflow > 0 {
					values.removeSubrange(0 ..< overflow)
				}
			}
```

mdiep:
[One note](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3120#discussion_r74175557) that needs to be addressed, but the other items are just suggestions.

ReactiveCocoa/Swift/SignalProducer.swift
mdiep: Ha! I just realized you opened #3123 for this.

ReactiveCocoa/Swift/SignalProducer.swift
andersio: Oops. I forgot we can match multiple cases at that time, well, in the midnight. 😂

mdiep:
✨

PR #3121, opened 2016-08-07
Changes to `DynamicProperty`.

andersio:
Depend on #3120, especially the said change in the semantic.

https://github.com/ReactiveCocoa/ReactiveCocoa/commit/cc066ab6f8b4045c4f2b8f641739307b14d60246
`DynamicProperty` accepts `NSObject` instead of `NSObject?`.

https://github.com/ReactiveCocoa/ReactiveCocoa/commit/88a64e72e15ff7f611c4c1ddf02d977d7cb544f1
`DynamicProperty` adopts the lifetime semantics of composed property. It now retains its source object (i.e. the `NSObject`), while its producer and signal respect the source's lifetime.


andersio:
Hmm, I'd like to work on the binding operators & property lifetime first, since the semantic change would break stuff without them ready.

PR #3122, opened 2016-08-08
Rename ignoreNil() to skipNil()

### LINKED ISSUE #2990: Rename ignoreNil to skipNil?
I was thinking about some RAC operators yesterday and realized that we use both `ignore` and `skip` to skip events. These do the same-ish thing, so they should probably use the same name. `ignoreNil` does change the type of `Value`, but I'd prefer to think of that as type-refinement.

Likewise, if we add `ignoreErrors`, it should probably be called `skipErrors`.

mdiep:
Resolves #2990.

May require some consensus.

### DIFF
--- a/ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
+++ b/ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
@@ -74,7 +74,7 @@ class SignalProducerLiftingSpec: QuickSpec {
 			}
 		}
 
-		describe("ignoreNil") {
+		describe("skipNil") {
 			it("should forward only non-nil values") {
 				let (producer, observer) = SignalProducer<Int?, NoError>.pipe()
 				let mappedProducer = producer.skipNil()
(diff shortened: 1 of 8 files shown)


NachoSoto:
Yup, agreed :+1:

neilpa:
I'll throw out one other alternative of `dropNil()`

mdiep:
I'm not a fan of the `drop` prefix, even though it would match the Swift stdlib. (Note that all the `skip*` methods would need to be renamed.)

This is ready to go now as I've fixed the failing tests.

ReactiveCocoaTests/Swift/SignalSpec.swift
```diff
@@ -400,7 +400,7 @@ class SignalSpec: QuickSpec {
 		describe("ignoreNil") {
```
NachoSoto: Needs to change this too :)

ReactiveCocoaTests/Swift/SignalProducerLiftingSpec.swift
```diff
@@ -77,7 +77,7 @@ class SignalProducerLiftingSpec: QuickSpec {
 		describe("ignoreNil") {
```
NachoSoto: And this

NachoSoto:
:shipit: after those last 2 things. I think this name is much better, consistent with the other skip operators.

mdiep:
🚢

PR #3123, opened 2016-08-08
Alter `Atomic.modify` to return the result of the supplied action.

andersio:
After revisiting all the use of `Atomic.modify`, it is observed that for those who need the old state, most of them needs only one or two specific bits. However, the current `Atomic.modify` would relentlessly copy everything (or make CoW types non-uniquely referenced).

Therefore, this PR altered `modify` to be more primitive and efficiently covering all the cases. The only downside is the lost of sugar for those who need a full copy of the old value. 🌴 

Having said that, a caveat is that [the Swift compiler can infer the return type of a closure, if the closure has a single expression only](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20151214/002583.html) at this moment. In other words, the return type of multi-line closures have to be stated explicitly, if it cannot be inferred from the LHS of an assignment statement (or `Void`).

```
// OK, since return type is obviously `Void`.
value.modify { $0 += 1 }
value.modify {
    $0 += 1
    $0 += 2
}
```

```
// This needs explicit type annotation, or the compiler
// will complain about not being able to infer the return
// type.
let previousValue = value.modify {
    let old = $0
    $0 += 1
    return old
}

// This works.
let previousValue = value.modify { value -> Int in
    let old = value
    value += 1
    return old
}

// Or this
let previousValue: Int = value.modify {
    let old = $0
    $0 += 1
    return old
}
```

### DIFF
--- a/ReactiveCocoa/Swift/Flatten.swift
+++ b/ReactiveCocoa/Swift/Flatten.swift
@@ -502,8 +502,11 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 	private func observeMerge(_ observer: Observer<Value.Value, Error>, _ disposable: CompositeDisposable) -> Disposable? {
 		let inFlight = Atomic(1)
 		let decrementInFlight = {
-			let orig = inFlight.modify { $0 -= 1 }
-			if orig == 1 {
+			let new: Int = inFlight.modify {
+				$0 -= 1
+				return $0
+			}
+			if new == 0 {
 				observer.sendCompleted()
 			}
 		}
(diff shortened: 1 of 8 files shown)


ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -429,13 +429,11 @@ private final class ConcatState<Value, Error: Swift.Error> {
...
 
-		var shouldStart = true
-
-		queuedSignalProducers.modify { queue in
+		let shouldStart: Bool = queuedSignalProducers.modify { queue in
 			// An empty queue means the concat is idle, ready & waiting to start
 			// the next producer.
-			shouldStart = queue.isEmpty
-			queue.append(producer)
+			defer { queue.append(producer) }
```
andersio: Shorthand arguments aren't used in any closures with `defer`, which somehow prohibits the use.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -1286,14 +1296,14 @@ extension SignalProtocol {
...
+				state.modify {
+					guard !$0.values.left.isEmpty && !$0.values.right.isEmpty else {
+						isFinished = $0.isFinished
 						return
 					}
 
-					tuple = (state.values.left.removeFirst(), state.values.right.removeFirst())
-					isFinished = state.isFinished
+					tuple = ($0.values.left.removeFirst(), $0.values.right.removeFirst())
+					isFinished = $0.isFinished
```
mdiep: In multi-line cases like this, I think using a named variable _is_ helpful.

mdiep:
This needs a merge as well.

andersio:
The line is set to be linear (usually 1-2 lines anyway) vs divergent control flow.

mdiep:
👍🏻

PR #3129, opened 2016-08-11
Streamlining a bit the `Property.value` impl. of composed properties.

andersio:
We need not start the replayed producer for the latest value every single time.

Since the only consistency guarantee of all `Property`s is that all `next` events are "_didSet_" events, we can exploit the observer call-out order (albeit not guaranteed by contract) and just manually cache the value .

Note that there is a test case "*should have the latest value available before sending any value*" guarding for any ordering changes of `Bag`.


ReactiveCocoa/Swift/Property.swift
```diff
@@ -501,13 +504,15 @@ public final class Property<Value>: PropertyProtocol {
 				fatalError("Receive unexpected error from a producer of `NoError` type: \(error)")
 			}
 		}
+
+		// Verify that an initial is sent. This is friendlier than deadlocking
+		// in the event that one isn't.
 		guard value != nil else {
```
NachoSoto: Since you're accessing this later you should do `guard let value = value` here to make line 515 safe at compile time.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -501,13 +504,15 @@ public final class Property<Value>: PropertyProtocol {
...
 			fatalError("A producer promised to send at least one value. Received none.")
 		}
-		disposable.dispose()
 
 		self.sources = sources
-		_value = { producer.take(first: 1).single()!.value! }
+		_value = { value! }
```
NachoSoto: This is better :+1:

ReactiveCocoa/Swift/Property.swift
```diff
@@ -374,6 +374,7 @@ extension PropertyProtocol {
 /// its source outlives it too.
 public final class Property<Value>: PropertyProtocol {
 	private let sources: [AnyObject]
+	private let disposable: Disposable?
```
NachoSoto: But I don't understand why we'd need to store the disposable now.

ReactiveCocoa/Swift/Property.swift
andersio: Then you would have created a new constant `value` to shadow the original `value`.

ReactiveCocoa/Swift/Property.swift
andersio: The property relies on a captured variable to present the latest value, and there is an observer to the replayed producer to feed this variable. (The variable is also used to check for the existence of an initial value.)

If this `Property` deinitializes, this captured variable is no longer accessible (in terms of reads), so the observer can be disposed of.

andersio:
Note that `Property.value` would no longer interlock the underlying `ReplayState`. This isn't really matter for such read-only construct IMO, because `Property.value` _alone_ is not enough to semantically guarantee it being latest, even if we lock against the `ReplayState`.

Meanwhile, we have `producer` and `signal`. ;)

andersio:
Updated with a new approach: an internal version of `replayLazily` which returns also a lense to the latest values within the `ReplayState`.

andersio:
A new approach: composed properties use a lense to peek the value cached by the `ReplayState`. It needs a new variant of `replayLazily` though, which is made `internal`.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1742,6 +1740,18 @@ extension SignalProducerProtocol {
 	}
 }
 
+internal final class ReplayValuesLense<Value, Error: Swift.Error> {
```
andersio: Related benchmark: https://gist.github.com/andersio/f6af7b65a99cccbe77acb5fdbaf67303
Swift 3 beta 4. Compiled with `-O`. Results show that using closure is 20x slower. Since only `ReplayState.values` has to be lensed so far, a hardcoded lense is used.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -484,30 +484,19 @@ public final class Property<Value>: PropertyProtocol {
...
+		// Verify that an initial is sent. This is friendlier than deadlocking in
+		// the event that one isn't.
+		guard !lense().isEmpty else {
 			fatalError("A producer promised to send at least one value. Received none.")
 		}
-		disposable.dispose()
 
 		self.sources = sources
-		_value = { producer.take(first: 1).single()!.value! }
+		_value = { lense()[0] }
```
NachoSoto: I would use `first!` here.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1670,6 +1670,10 @@ extension SignalProducerProtocol {
 	/// - returns: A caching producer that will hold up to last `capacity`
 	///            values.
 	public func replayLazily(upTo capacity: Int) -> SignalProducer<Value, Error> {
+		return replay(upTo: capacity).producer
+	}
+
+	internal func replay(upTo capacity: Int) -> (producer: SignalProducer<Value, Error>, lense: () -> [Value]) {
```
NachoSoto: How about making lense just give the last value?

ReactiveCocoa/Swift/Property.swift
NachoSoto: If the lense is just the last value then you can do `_value = lense`

NachoSoto:
I'm not against this, but what's the benefit here exactly?

andersio:
*... need not start the replayed producer for the latest value every single time...*. Starting a producer involves steps and resources (locks), while the value can be read out in a more direct way.

ReactiveCocoa/Swift/Property.swift
andersio: Either way it would trap if something unexpectedly happens, and it is asserted that the array should have always one element.

NachoSoto: Yes, but there's a reason for each suggestion:
- `first` is more explicit in 2 ways:
  - it avoids the magic number `0`.
  - it makes it explicit that the collection might be empty.
- `_value = lense`: this avoids an extra level of indirection.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1670,6 +1670,10 @@ extension SignalProducerProtocol {
 	/// - returns: A caching producer that will hold up to last `capacity`
 	///            values.
 	public func replayLazily(upTo capacity: Int) -> SignalProducer<Value, Error> {
+		return _replayLazily(upTo: capacity).producer
+	}
+
+	internal func _replayLazily(upTo capacity: Int) -> (producer: SignalProducer<Value, Error>, state: Atomic<ReplayState<Value, Error>>) {
```
NachoSoto: This is now exposing an implementation detail :-1:

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1737,30 +1743,30 @@ private struct ReplayError<Value>: Error {
 	let values: [Value]
 }
 
-private struct ReplayState<Value, Error: Swift.Error> {
-	let capacity: Int
+internal struct ReplayState<Value, Error: Swift.Error> {
```
NachoSoto: Evidenced by this change, too.

ReactiveCocoa/Swift/SignalProducer.swift
```diff
@@ -1670,6 +1670,20 @@ extension SignalProducerProtocol {
...
+	///
+	/// - precondtion: `capacity` must be non-negative integer.
+	///
+	/// - parameters:
+	///   - capcity: Number of values to hold.
+	///
+	/// - returns: `producer`: a caching producer that will hold up to last
+	///            `capacity` values; `first`: a lense for composed properties to
+	///            read the head of the values cache.
+	internal func _replayLazily(upTo capacity: Int) -> (producer: SignalProducer<Value, Error>, first: () -> Value) {
```
mdiep: I'm pretty 👎🏻 on this.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -484,30 +484,24 @@ public final class Property<Value>: PropertyProtocol {
...
+
+		// Verify that an initial is sent. This is friendlier than deadlocking in
+		// the event that one isn't.
+		guard hasInitial else {
 			fatalError("A producer promised to send at least one value. Received none.")
 		}
-		disposable.dispose()
 
 		self.sources = sources
-		_value = { producer.take(first: 1).single()!.value! }
```
mdiep: I agree that starting a producer is heavy-weight for accessing the value.

Maybe instead we could use a local or instance `Atomic` variable to set the value in the above `start` block and return it here?

andersio:
Comments were addressed. :)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -490,29 +493,18 @@ public final class Property<Value>: PropertyProtocol {
 		// they see a consistent view of the `self.value`.
 		// https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3042
 		let producer = unsafeProducer.replayLazily(upTo: 1)
-		
+
+		let atomic = Atomic<Value?>(nil)
+		disposable = producer.startWithNext { atomic.value = $0 }
```
mdiep: Would it make sense to use a `Lifetime` here instead of a `Disposable`?

mdiep:
Looks good to me. Just one question.

NachoSoto:
:shipit:

ReactiveCocoa/Swift/Property.swift
andersio: Using `Disposable` should mean less hops, and we have just this one client listening for the deinitialization anyway.

PR #3131, opened 2016-08-11
Introduce the `BindingTarget` protocol.

andersio:
1. `BindingTarget` models the minimum viable requirements of a consumer in a unidirectional binding.
1. All `<~` operators are now settled in `BindingTarget `.
2. All `MutablePropertyProtocol` conforming type must be a `BindingTarget`, and thus must provide a `lifetime` for the binding operators.


ReactiveCocoa/Swift/Lifetime.swift
```diff
@@ -3,9 +3,54 @@ import enum Result.NoError
...
+	private init(ended signal: Signal<(), NoError>) {
+		ended = signal
+	}
+
+	/// Initialize a `Lifetime` object which ends when the supplied producer
+	/// terminates.
+	///
+	/// - parameters:
+	///   - producer: The producer to be observed for termination.
+	public convenience init<U, E: Error>(_ producer: SignalProducer<U, E>) {
```
NachoSoto: :-1:

This initialization has side effects. Additionally, I don't see how this would be useful given that each subscriber sees potentially a different lifetime...

ReactiveCocoa/Swift/Property.swift
```diff
@@ -11,6 +11,9 @@ public protocol PropertyProtocol: class {
 	/// The current value of the property.
 	var value: Value { get }
 
+	/// The lifetime of the property.
```
NachoSoto: I don't understand what this would represent :/

andersio:
I have eliminated the unnecessary bits after giving a deep thought to @NachoSoto's comments. I have also brought in the changes to the binding operators, since they are highly related.

A code snippet to demonstrate the final goal of this PR:
```
extension UILabel {
    // Got rid of the `MutableProperty`. :-]
    public var rac_text: AnyBindingConsumer<String> {
        return AnyBindingConsumer(sink: { [weak self] in self?.text = $0 }, lifetime: rac_lifetime)
    }
}
```

RuiAAPeres:
> // Binding still works.
myLabel.rac_text <~ myStringProperty

👍

> // But all of these are now prohibited, since they were just implementation
// artefacts.
myMutableStringProperty <~ myLabel.rac_text

Given the MVVM pattern, how would you update a ViewModel's exposed property, every time a `UITextField`'s text changes? My expectation would be something like:

```swift
myMutableStringProperty <~ myTextField.rac_text
```

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,315 @@
...
+	/// ````
+	///
+	/// - parameters:
+	///   - consumer: A consumer to bind to.
+	///   - signal: A signal to bind.
+	///
+	/// - returns: A disposable that can be used to terminate binding before the
+	///            deinitialization of property or signal's `completed` event.
+	@discardableResult
+	static func <~ <Source: SignalProtocol where Source.Value == ValueType, Source.Error == NoError>(consumer: Self, signal: Source) -> Disposable
```
mdiep: Why is only this 1 variant of `<~` included here? It seems like it shouldn't be needed since it's implemented below? I wouldn't expect anyone to override this.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,315 @@
+import enum Result.NoError
+
+infix operator <~ {
+	associativity right
+
+	// Binds tighter than assignment but looser than everything else
+	precedence 93
+}
+
+public protocol BindingConsumer: class {
```
mdiep: `Bindable` might make more sense as a name?

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,315 @@
...
+	///            deinitialization of destination property or source property
+	///            producer's `completed` event.
+	@discardableResult
+	public static func <~ <Source: PropertyProtocol where Source.Value == ValueType.Wrapped>(consumer: Self, property: Source) -> Disposable {
+		return consumer <~ property.producer
+	}
+}
+
+/// A type-erased view of a binding consumer.
+public class AnyBindingConsumer<Value>: BindingConsumer {
```
mdiep: This isn't used, so I don't think it's worth including.

ReactiveCocoa/Swift/Property.swift
```diff
@@ -25,11 +25,19 @@ public protocol PropertyProtocol: class {
 }
 
 /// Represents an observable property that can be mutated directly.
-public protocol MutablePropertyProtocol: PropertyProtocol {
+public protocol MutablePropertyProtocol: BindingConsumer, PropertyProtocol {
 	/// The current value of the property.
 	var value: Value { get set }
 }
 
+extension MutablePropertyProtocol {
```
mdiep: I would declare the conformance to `BindingConsumer` here since this is where it's implemented.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
liscio: @mdiep The trouble with a generic `Bindable` name is that it's easily confused with two-way "bindings". I don't love the `BindingConsumer` name despite using it myself, and perhaps `Consumer` would be more apt.

`ValueConsumer` could work as well, however the `<~` operator really sends events.

mdiep: Maybe `BindingDestination` would be a better name then, since the documentation already refers to _sources_ and _destinations_?

ReactiveCocoa/Swift/UnidirectionalBinding.swift
andersio: `DynamicProperty` would override it, if we would match its semantics to the one of `Property`s.

Specifically, if we reverse the reference so that `DynamicProperty` now holds a strong reference to the `NSObject`, the `DynamicProperty` must be able to supply its own implementation of bindings that interacts directly with the underlying `NSObject`.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
andersio: It is made for all non-KVO compliant properties/keypaths, which are all currently in Rex.

ReactiveCocoa/Swift/Property.swift
andersio: Protocols cannot conform to another protocol in an extension.

andersio:
@RuiAAPeres For `UITextView` or `UITextField`, I would expect something similar to `MutableProperty` but customised, so that you can do both: `textField.rac_text <~ viewModel.text` and `viewModel.text <~ textField.rac_text`.

That's said we might also have to consider bidirectional bindings, shall we included it as part of our toolbox in the future.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
neilpa: Or maybe `BindingTarget` (I find _sources_ and _targets_ a more natural pair).

mdiep:
I'll give this another look after we get #3136 merged.

andersio:
Updated for Preview 6. :)

ReactiveCocoa/Swift/Property.swift
```diff
@@ -25,11 +25,20 @@ public protocol PropertyProtocol: class {
...
 
 /// Represents an observable property that can be mutated directly.
-public protocol MutablePropertyProtocol: PropertyProtocol {
+public protocol MutablePropertyProtocol: BindingTarget, PropertyProtocol {
 	/// The current value of the property.
 	var value: Value { get set }
 }
 
+/// Default implementation of `MutablePropertyProtocol` for `BindingTarget`.
+extension MutablePropertyProtocol {
```
NachoSoto: The conformance to the protocol should be here.

ReactiveCocoa/Swift/Signal.swift
```diff
@@ -177,6 +177,13 @@ public final class Signal<Value, Error: Swift.Error> {
 	}
 }
 
+// Testability
+extension Signal {
+	internal var test_observerCount: Int? {
```
NachoSoto: 👎👎👎👎 a thousand times no, this is exposing a private implantation detail.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,313 @@
...
+precedencegroup Binding {
+	associativity: right
+
+	// Binds tighter than assignment but looser than everything else
+	higherThan: AssignmentPrecedence
+}
+
+infix operator <~ : Binding
+
+/// Describes a target to which can be bond.
```
NachoSoto: Bond?

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,313 @@
...
+	/// - returns: A disposable that can be used to terminate binding before the
+	///            deinitialization of the target or the source property.
+	@discardableResult
+	public static func <~ <Source: PropertyProtocol>(target: Self, property: Source) -> Disposable where Source.Value == ValueType.Wrapped {
+		return target <~ property.producer
+	}
+}
+
+/// A type-erased view of a binding consumer.
+public class AnyBindingTarget<Value>: BindingTarget {
```
NachoSoto: `final`

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,313 @@
...
+	///
+	/// - parameter:
+	///   - consumer: The binding consumer to be wrapped.
+	public init<U: BindingTarget>(_ consumer: U) where U.ValueType == Value {
+		self.sink = consumer.consume
+		self.lifetime = consumer.lifetime
+	}
+
+	/// Create a binding consumer.
+	///
```
NachoSoto: `private`

ReactiveCocoaTests/Swift/LifetimeSpec.swift
```diff
@@ -5,43 +5,7 @@ import Result
 
 final class LifetimeSpec: QuickSpec {
 	override func spec() {
-		describe("SignalProducerProtocol.takeDuring") {
-			it("completes a signal when the lifetime ends") {
```
NachoSoto: Why are these tests gone?

ReactiveCocoa/Swift/Property.swift
andersio: Extensions to protocols cannot have any inheritance clauses.

ReactiveCocoaTests/Swift/LifetimeSpec.swift
andersio: Moved to SignalProducerSpec.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
andersio: This is expected to be public.

ReactiveCocoa/Swift/Signal.swift
andersio: This is a consequence of the simplified binding implementation. As the operators now return directly the disposables from `Signal.observe` or `SignalProducer.startWithSignal`, inspecting through `isDisposed` of those now just means whether it is manually disposed of.

I consider this a middle ground between making everything `internal` for tests, and having no hooks to test for.

mdiep: I'm with @NachoSoto on this one.

👎🏻

andersio:
`Signal.test_observerCount` has been replaced in one test case "_should tear down the binding when the property deallocates_".

However, the "_should tear down the binding when the bound signal/producer terminates_" test cases are removed, since there are no obvious means to test them.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,313 @@
...
+	public static func <~ <Source: PropertyProtocol>(target: Self, property: Source) -> Disposable where Source.Value == ValueType.Wrapped {
+		return target <~ property.producer
+	}
+}
+
+/// A type-erased view of a binding consumer.
+public final class AnyBindingTarget<Value>: BindingTarget {
+	public typealias ValueType = Value
+
+	let sink: @escaping (Value) -> ()
```
NachoSoto: Github moved my comment. I meant that this should be private.

ReactiveCocoaTests/Swift/PropertySpec.swift
```diff
@@ -379,44 +398,40 @@ class PropertySpec: QuickSpec {
...
-							var producerCompleted = 0
+					it("should return a producer and a signal which respect the lifetime of the source property instead of the read-only view itself") {
+						var signalCompleted = 0
+						var producerCompleted = 0
 
-							var property = Optional(MutableProperty(1))
-							var firstMappedProperty = Optional(property!.map { $0 + 1 })
-							var secondMappedProperty = Optional(firstMappedProperty!.map { $0 + 2 })
-							var thirdMappedProperty = Optional(secondMappedProperty!.map { $0 + 2 })
+						var property = Optional(MutableProperty(1))
```
NachoSoto: I would make these implicitly unwrapped.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,313 @@
+import enum Result.NoError
+
+precedencegroup Binding {
```
ikesyo: The name is not obvious. I think we should consistently add `Precedence` suffix.

ReactiveCocoaTests/Swift/PropertySpec.swift
andersio: We are not promising these properties would _eventually_ be here when needed, but asserting that they would be gone eventually. Optional seems fine & on point.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,313 @@
...
+
+	// Binds tighter than assignment but looser than everything else
+	higherThan: AssignmentPrecedence
+}
+
+infix operator <~ : BindingPrecedence
+
+/// Describes a target to which can be bound.
+public protocol BindingTarget: class {
+	associatedtype ValueType
```
mdiep: I think this could probably be called `Value`?

ReactiveCocoa/Swift/Property.swift
```diff
@@ -581,6 +593,8 @@ public final class MutableProperty<Value>: MutablePropertyProtocol {
 	///   - initialValue: Starting value for the mutable property.
 	public init(_ initialValue: Value) {
 		(signal, observer) = Signal.pipe()
+		token = Lifetime.Token()
+		lifetime = Lifetime(token)
```
mdiep: These could probably be declared inline above?

ReactiveCocoa/Swift/UnidirectionalBinding.swift
```diff
@@ -0,0 +1,313 @@
...
+	/// - returns: A disposable that can be used to terminate binding before the
+	///            deinitialization of the target or the source property.
+	@discardableResult
+	public static func <~ <Source: PropertyProtocol>(target: Self, property: Source) -> Disposable where Source.Value == ValueType.Wrapped {
+		return target <~ property.producer
+	}
+}
+
+/// A type-erased view of a binding consumer.
+public final class AnyBindingTarget<Value>: BindingTarget {
```
mdiep: You mentioned that this was destined to work with the `Rex` bindings. I think we should leave this out until Rex is integrated with RAC. I dislike having untested code and it's not really essential to this PR.

ReactiveCocoa/Swift/UnidirectionalBinding.swift
mdiep: I think that's what it should be called. That would match our other protocols.

mdiep:
👍🏻

Just a few items to fix up. Sorry for the delayed review on this!

ReactiveCocoa/Swift/Property.swift
andersio: Every stored property is initialised in the initializer, so I am just following the pattern.

Besides, `lifetime` needs the `token`, which I don't think it can be expressed inline as `self` cannot be used, unless we make it `lazy` or something like `(..., ...) = Lifetime.create()` (but then we cannot declare different access levels...).

ReactiveCocoa/Swift/UnidirectionalBinding.swift
andersio: Anyway, it was made a workaround as sometimes somehow the type parameters cannot be matched for associated type requirements. It seems it was solved by reordering the protocol inheritance clause...

mdiep:
Looks good to me. 👍🏻

PR #3136, opened 2016-08-15
Changes for Swift 3 beta 6

NachoSoto:
# Changes

- `dynamictype` -> `type(of:)` (SE-0096)
- Added explicit casting
- Argument labels have been removed from function types (SE-0111)
- Moved location of where statements in generic declarations
- `private` -> `fileprivate` (SE-0025) 
- `@noescape` is now default, `@escaping` is not (SE-0103)
- `id` is now imported as `Any` (SE-0116)
- Updated usage of `String` constructor.

# TODO:

- [x] Update `Result`: https://github.com/antitypical/Result/pull/181.
- [x] Fix `UnsafeMutablePointer` changes.
- [x] Other compile errors.

PR #3139, opened 2016-08-17
Fix the remaining build errors in ObjectiveCBridging.swift for Xcode 8 beta 6

ikesyo:
Against to #3136.

Looks like SE-0112 and SE-0116 are related to this.


NachoSoto:
:shipit:

PR #3141, opened 2016-08-19
Looping instead of recursive for flatten with .Concat

### LINKED ISSUE #3140: Flatten with .Concat stacks overflow bug when tons of completed producer need to be processed
We ran into a bug, when we have tons of completed producer in the SignalProducer pipeline inside a flatten with .Concat, as it seems .Concat was implemented with recursive calls for the next value, the stack will overflow then

Here is the simple code to reproduce it

```Swift
  func testConcatOverflow() {
    var values: [Int] = []
    for n in 0..<10000 {
      values.append(n)
    }

    SignalProducer<Int, NSError>(values: values)
      .flatMap(.Concat) { n -> SignalProducer<Int, NSError> in
        if n == 0 {
          return SignalProducer<Int, NSError>(value: n)
            .delay(0.1, onScheduler: QueueScheduler.mainQueueScheduler)
        }
        return SignalProducer<Int, NSError>(value: n)
      }
      .startWithNext { n in
        print("Num \(n)")
      }
    let exp = expectationWithDescription("")
    waitForExpectationsWithTimeout(100, handler: nil)
  }
```

Surely the bug can be workarounded via adding a delay to any queue for the SignalProducer, but I wonder why not implement .Concat with loop instead of recursion?

It's kind of surprising behavior to me, also very hard to debug, took me whole day.

I will try to fix the problem with a loop implementation of .Concat and see what you guys think about it.

victorlin:
Fix #3140

### DIFF
--- a/ReactiveCocoa/Swift/Flatten.swift
+++ b/ReactiveCocoa/Swift/Flatten.swift
@@ -392,19 +392,19 @@ extension SignalProducerProtocol {
 	public func concat(_ next: SignalProducer<Value, Error>) -> SignalProducer<Value, Error> {
 		return SignalProducer<SignalProducer<Value, Error>, Error>(values: [ self.producer, next ]).flatten(.concat)
 	}
-	
+
 	/// `concat`s `value` onto `self`.
 	public func concat(value: Value) -> SignalProducer<Value, Error> {
 		return self.concat(SignalProducer(value: value))
 	}
-	
+
 	/// `concat`s `self` onto initial `previous`.
 	public func prefix<P: SignalProducerProtocol>(_ previous: P) -> SignalProducer<Value, Error>
 		where P.Value == Value, P.Error == Error
 	{
 		return previous.concat(self.producer)
 	}
-	
+
 	/// `concat`s `self` onto initial `value`.
 	public func prefix(value: Value) -> SignalProducer<Value, Error> {
 		return self.prefix(SignalProducer(value: value))
@@ -493,7 +493,7 @@ private final class ConcatState<Value, Error: Swift.Error> {
 
 						if let nextSignalProducer = self.dequeueSignalProducer() {
 							if currentProducer.modify({ value -> Bool in
-    							// We are still in the loop, use looping
+								// We are still in the loop, use looping
 								if isLooping {
 									value = nextSignalProducer
 									// return false to not to call startNextSignalProducer
@@ -506,7 +506,7 @@ private final class ConcatState<Value, Error: Swift.Error> {
 								self.startNextSignalProducer(nextSignalProducer)
 							}
 						}
-    				case .next, .failed:
+					case .next, .failed:
 						self.observer.action(event)
 					}
 				}
@@ -607,7 +607,7 @@ extension SignalProtocol {
 
 		return result
 	}
-	
+
 	/// Merges the given signals into a single `Signal` that will emit all
 	/// values from each of them, and complete when all of them have completed.
 	public static func merge<S: SignalProtocol>(_ signals: S...) -> Signal<Value, Error>
@@ -626,7 +626,7 @@ extension SignalProducerProtocol {
 	{
 		return SignalProducer(values: producers).flatten(.merge)
 	}
-	
+
 	/// Merges the given producers into a single `SignalProducer` that will emit
 	/// all values from each of them, and complete when all of them have
 	/// completed.
@@ -756,7 +756,7 @@ extension SignalProducerProtocol where Value: SignalProducerProtocol, Error == V
 private struct LatestState<Value, Error: Swift.Error> {
 	var outerSignalComplete: Bool = false
 	var innerSignalComplete: Bool = true
-	
+
 	var replacingInnerSignal: Bool = false
 }
 
@@ -771,7 +771,7 @@ extension SignalProtocol {
 	public func flatMap<U>(_ strategy: FlattenStrategy, transform: @escaping (Value) -> SignalProducer<U, Error>) -> Signal<U, Error> {
 		return map(transform).flatten(strategy)
 	}
-	
+
 	/// Maps each event from `signal` to a new signal, then flattens the
 	/// resulting producers (into a signal of values), according to the
 	/// semantics of the given strategy.
@@ -813,14 +813,14 @@ extension SignalProtocol where Error == NoError {
 	public func flatMap<U, E>(_ strategy: FlattenStrategy, transform: @escaping (Value) -> SignalProducer<U, E>) -> Signal<U, E> {
 		return map(transform).flatten(strategy)
 	}
-	
+
 	/// Maps each event from `signal` to a new signal, then flattens the
 	/// resulting signals (into a signal of values), according to the
 	/// semantics of the given strategy.
 	public func flatMap<U>(_ strategy: FlattenStrategy, transform: @escaping (Value) -> SignalProducer<U, NoError>) -> Signal<U, NoError> {
 		return map(transform).flatten(strategy)
 	}
-	
+
 	/// Maps each event from `signal` to a new signal, then flattens the
 	/// resulting signals (into a signal of values), according to the
 	/// semantics of the given strategy.
@@ -830,7 +830,7 @@ extension SignalProtocol where Error == NoError {
 	public func flatMap<U, E>(_ strategy: FlattenStrategy, transform: @escaping (Value) -> Signal<U, E>) -> Signal<U, E> {
 		return map(transform).flatten(strategy)
 	}
-	
+
 	/// Maps each event from `signal` to a new signal, then flattens the
 	/// resulting signals (into a signal of values), according to the
 	/// semantics of the given strategy.
@@ -849,7 +849,7 @@ extension SignalProducerProtocol {
 	public func flatMap<U>(_ strategy: FlattenStrategy, transform: @escaping (Value) -> SignalProducer<U, Error>) -> SignalProducer<U, Error> {
 		return map(transform).flatten(strategy)
 	}
-	
+
 	/// Maps each event from `self` to a new producer, then flattens the
 	/// resulting producers (into a producer of values), according to the
 	/// semantics of the given strategy.
@@ -891,7 +891,7 @@ extension SignalProducerProtocol where Error == NoError {
 	public func flatMap<U, E>(_ strategy: FlattenStrategy, transform: @escaping (Value) -> SignalProducer<U, E>) -> SignalProducer<U, E> {
 		return map(transform).flatten(strategy)
 	}
-	
+
 	/// Maps each event from `self` to a new producer, then flattens the
 	/// resulting producers (into a producer of values), according to the
 	/// semantics of the given strategy.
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -241,23 +241,48 @@ private final class ConcatState<Value, Error: ErrorType> {
...
-			let handle = self.disposable?.addDisposable(disposable) ?? nil
-
-			signal.observe { event in
-				switch event {
-				case .Completed, .Interrupted:
-					handle?.remove()
+		// Current signal producer
+		var signalProducer: SignalProducer<Value, Error>? = signalProducer
+		// Are we still inside the loop
+		var inLoop = true
```
andersio: These states may need mutual exclusion, as producers may start asynchronously.

andersio:
`master` now targets Swift 3.0. Would you like to rebase it?

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -241,23 +241,48 @@ private final class ConcatState<Value, Error: ErrorType> {
...
 	func startNextSignalProducer(signalProducer: SignalProducer<Value, Error>) {
-		signalProducer.startWithSignal { signal, disposable in
-			let handle = self.disposable?.addDisposable(disposable) ?? nil
-
-			signal.observe { event in
-				switch event {
-				case .Completed, .Interrupted:
-					handle?.remove()
+		// Current signal producer
+		var signalProducer: SignalProducer<Value, Error>? = signalProducer
```
andersio: [Nit] `producer` and `isLooping`?

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -241,23 +241,48 @@ private final class ConcatState<Value, Error: ErrorType> {
...
+		//     startNextSignalProducer(signalProducer0) --Completed-->
+		//         startNextSignalProducer(signalProducer1) --Completed-->
+		//             startNextSignalProducer(signalProducer2) --Completed-->
+		//                 startNextSignalProducer(signalProducer3) --Completed-->
+		//                     ....
+		//
+		while let producer = signalProducer {
+			signalProducer = nil
+			producer.startWithSignal { signal, disposable in
+				let handle = self.disposable?.addDisposable(disposable) ?? nil
```
andersio: The optional chaining operator should have given you an optional already.

ReactiveCocoa/Swift/Flatten.swift
victorlin: @andersio thanks for catching the problem, yeah, also the loop flag may have some corner case race condition

like, say `signalProducer = nextSignalProducer` kicks in right after `startNextSignalProducer` leaves while loop but before setting inLoop to false

```
Thread 1                                Thread 2
left while loop
                                        signalProducer = nextSignalProducer
inLoop = false  
startNextSignalProducer returns
```

then the next signal producer will never be started

I am thinking how to avoid that problem. Had an idea already

victorlin:
@andersio I rebased on the master, and guard `isLooping` with critical section now. Ideally this should get rid of the race condition issue. Sorry I don't have Swift 3 installed on my machine and I don't have spare time to do it this week, not sure will it work or not. If you don't mind, could you take over this PR and make it works with Swift 3? Or I can come back later on this PR.

victorlin:
@andersio and by the way, do you have plan for back porting bugfixes back to RAC4? Which branch it is?

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1105,6 +1105,36 @@ class SignalProducerSpec: QuickSpec {
 						expect(completed) == true
 					}
 				}
+
+				describe("non-recursive") {
```
NachoSoto: This isn't very... Descriptive :P

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1105,6 +1105,36 @@ class SignalProducerSpec: QuickSpec {
...
+							.startWithNext { _ in
+								called += 1
+							}
+					}
+
+					afterEach {
+						disposable?.dispose()
+					}
+
+					it("send next 10001 times") {
```
NachoSoto: This doesn't explain what the expectation is.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1105,6 +1105,36 @@ class SignalProducerSpec: QuickSpec {
...
+				describe("non-recursive") {
+					var disposable: Disposable?
+					var called = 0
+
+					beforeEach {
+						var producers: [SignalProducer<(), NoError>] = [
+							SignalProducer<(), NoError>(value: ())
+								.delay(0.1, onScheduler: QueueScheduler())
+						]
+						for _ in 0..<10000 {
```
NachoSoto: This number is repeated here and in the expectation.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
```diff
@@ -1105,6 +1105,36 @@ class SignalProducerSpec: QuickSpec {
 						expect(completed) == true
 					}
 				}
+
+				describe("non-recursive") {
+					var disposable: Disposable?
```
NachoSoto: You want an implicitly unwrapped optional here.

andersio:
I don't think we would have any further 4.x releases. @mdiep might have a few words though.

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
victorlin: @NachoSoto will fix it

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
victorlin: @NachoSoto will fix it

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
victorlin: @NachoSoto will fix it

ReactiveCocoaTests/Swift/SignalProducerSpec.swift
victorlin: copy pasta from somewhere else :P, will fix it

victorlin:
@andersio @NachoSoto I've already addressed all the issues, can you guys review again?

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -459,20 +459,56 @@ private final class ConcatState<Value, Error: Swift.Error> {
 
 	/// Subscribes to the given signal producer.
 	func startNextSignalProducer(_ signalProducer: SignalProducer<Value, Error>) {
```
mdiep: It seems like this should be modified to not take any arguments since it's going to start _n_ producers. It can check whether the `queue` is empty. This should also simplify the code in `enqueueSignalProducer` above.

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -459,20 +459,56 @@ private final class ConcatState<Value, Error: Swift.Error> {
...
-			signal.observe { event in
-				switch event {
-				case .completed, .interrupted:
-					handle?.remove()
-
-					if let nextSignalProducer = self.dequeueSignalProducer() {
-						self.startNextSignalProducer(nextSignalProducer)
+		// Are we still inside the loop
+		var isLooping = true
+		// Current signal producer
```
mdiep: These comments aren't really helpful and should be removed. They're just restating what the code already tells us, rather than providing explanations/background.

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -459,20 +459,56 @@ private final class ConcatState<Value, Error: Swift.Error> {
...
+		// We need to do this instead of recursive, otherwise when we encounter tons of completed
+		// SignalProducer, it's very liekly we will have stack overflow issue like this
+		//
+		//     startNextSignalProducer(signalProducer0) --Completed-->
+		//         startNextSignalProducer(signalProducer1) --Completed-->
+		//             startNextSignalProducer(signalProducer2) --Completed-->
+		//                 startNextSignalProducer(signalProducer3) --Completed-->
+		//                     ....
+		//
+		while let producer = currentProducer.modify({ (value: inout SignalProducer<Value, Error>?) -> SignalProducer<Value, Error>? in
```
mdiep: Using a structure like this would probably be more readable:

```swift
while true {
   guard let … = ….modify {
       …
   } else {
      break
   }
}
```

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -459,20 +459,56 @@ private final class ConcatState<Value, Error: Swift.Error> {
...
-
-			signal.observe { event in
-				switch event {
-				case .completed, .interrupted:
-					handle?.remove()
-
-					if let nextSignalProducer = self.dequeueSignalProducer() {
-						self.startNextSignalProducer(nextSignalProducer)
+		// Are we still inside the loop
+		var isLooping = true
```
mdiep: I think this needs to be an `Atomic`

mdiep:
I'm not wild about this approach. 😕 I may play around with some alternatives here.

ReactiveCocoa/Swift/Flatten.swift
victorlin: @mdiep make sense, will do

ReactiveCocoa/Swift/Flatten.swift
victorlin: will do

ReactiveCocoa/Swift/Flatten.swift
victorlin: will do

ReactiveCocoa/Swift/Flatten.swift
victorlin: @mdiep hmmm, but it's already protected by the critical section of `currentProducer`

andersio: I would suggest you using a tuple here, and fit `isLooping` into the `Atomic`. It would communicate the intention better.

ReactiveCocoa/Swift/Flatten.swift
andersio: You may consider specifying the return type on the LHS instead.

```
let bacon: Bacon = fridge.modify {
   return $0.takeBacon()
}
```

ReactiveCocoa/Swift/Flatten.swift
victorlin: @andersio I tried that before, somehow ran into strange Swift compiler error, not sure is that a Swift 3.0 compiler bug or what, but will try to do that way again see if it works.

victorlin: @andersio I tried

```Swift
let loopingState = Atomic<(SignalProducer<Value, Error>?, Bool)>((.optional(signalProducer), true))
```

and it says

```
/Users/victorlin/workspace/open_source/ReactiveCocoa/ReactiveCocoa/Swift/Flatten.swift:462:22: Expression type 'Atomic<(SignalProducer<Value, Error>?, Bool)>' (aka 'Atomic<(Optional<SignalProducer<Value, Error>>, Bool)>') is ambiguous without more context
```

any idea?

mdiep:
Superseded by #3161.

PR #3145, opened 2016-08-22
Update tests for Xcode 8 b6

andersio:
Depends on @ikesyo's forks of Nimble and Quick.


andersio:
Nested arrays were bridged back as `NSArray` for equivalent testing, since Nimble currently does not have matchers for nested arrays.

mdiep:
🎉

PR #3149, opened 2016-08-24
then should be available for NoError

Igor-Palaguta:


mdiep:
Could you also handle the case where the new signal has `NoError`?

That requires adding 2 more variants, just like `flatMap`, because you need to add one where both signals are `NoError`—otherwise it's ambiguous for those signals.

NachoSoto:
:ship it: after @mdiep's comment. That also shows that this need an extra test to cover the ambiguity.

Igor-Palaguta:
Please check

mdiep:
Looks good (although it's curious that you didn't need a 3rd overload; maybe this is changed in Swift 3).

Just waiting on CI. Travis is a bit backed up due to maintenance, so it may be a while longer.

mdiep:
Thanks for the pull request! 🎉

PR #3161, opened 2016-08-31
Rewrite concat to start producers iteratively

### LINKED ISSUE #3140: Flatten with .Concat stacks overflow bug when tons of completed producer need to be processed

mdiep:
Recursively starting the inner producers can blow the stack if they complete immediately.

Fixes #3140.

Supersedes #3141.

### DIFF
--- a/ReactiveCocoa/Swift/Flatten.swift
+++ b/ReactiveCocoa/Swift/Flatten.swift
@@ -355,9 +355,12 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 						case .completed, .interrupted:
 							handle?.remove()
 							
-							state.modify { $0.active = nil }
+							let shouldStart: Bool = state.modify {
+								$0.active = nil
+								return !$0.isStarting
+							}
 
-							if !state.value.isStarting {
+							if shouldStart {
 								startNextIfNeeded()
 							}
 
--- a/ReactiveCocoa/Swift/Flatten.swift
+++ b/ReactiveCocoa/Swift/Flatten.swift
@@ -346,7 +346,7 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 		let state = Atomic(ConcatState<Value.Value, Error>())
 		
 		func startNextIfNeeded() {
-			while let producer = state.modify({ $0.willStart() }) {
+			while let producer = state.modify({ $0.dequeue() }) {
 				producer.startWithSignal { signal, inner in
 					let handle = disposable?.add(inner) ?? nil
 
@@ -354,8 +354,10 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
 						switch event {
 						case .completed, .interrupted:
 							handle?.remove()
+							
+							state.modify { $0.active = nil }
 
-							if state.modify({ $0.didTerminate() }) {
+							if !state.value.isStarting {
 								startNextIfNeeded()
 							}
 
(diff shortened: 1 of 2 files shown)


andersio:
It feels like the `ConcatState` would get meaningful method names if we take out the recursive/loop tracking aspect.

mdiep:
That seems to me like an integral part of the improvement here. 😕

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -341,25 +341,50 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
...
+				producer.startWithSignal { signal, inner in
+					let handle = disposable?.add(inner) ?? nil
+
+					signal.observe { event in
+						switch event {
+						case .completed, .interrupted:
+							handle?.remove()
 
-		return self.observe { event in
+							if state.modify({ $0.didTerminate() }) {
```
andersio: We might need some clarity at the call site.

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -341,25 +341,50 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
...
 		}
 	}
-
+	
 	fileprivate func observeConcat(_ observer: Observer<Value.Value, Error>, _ disposable: CompositeDisposable? = nil) -> Disposable? {
-		let state = ConcatState(observer: observer, disposable: disposable)
+		let state = Atomic(ConcatState<Value.Value, Error>())
+		
+		func startNextIfNeeded() {
+			while let producer = state.modify({ $0.willStart() }) {
```
andersio: How about `dequeue` and `acknowledgeStarted`?

mdiep:
I went with `dequeue`, but got rid of the other methods and just set the properties directly. The names were one thing I didn't like about this, and I think this is an improvement.

ReactiveCocoa/Swift/Flatten.swift
```diff
@@ -341,25 +341,52 @@ extension SignalProtocol where Value: SignalProducerProtocol, Error == Value.Err
...
+
+					signal.observe { event in
+						switch event {
+						case .completed, .interrupted:
+							handle?.remove()
+							
+							state.modify { $0.active = nil }
 
-		return self.observe { event in
+							if !state.value.isStarting {
```
andersio: You can return `isStarting` from the action above to avoid acquiring the lock twice.

mdiep:
Closed in favor of https://github.com/ReactiveCocoa/ReactiveSwift/issues/5.

PR #3162, opened 2016-08-31
Changes to the `DynamicProperty` semantics.

andersio:
Make it more `Property` alike - a lens to a key path of an `NSObject` without its own lifetime.

1. ~~A `DynamicProperty` retains its lensing object.~~
2. The producer and signal respects the lifetime of the object being lensed.
3. Bindings targeting `DynamicProperty` ties directly to the object being lensed.
4. `DynamicProperty` is now leak safe - reusing instances is no longer necessary.

 For example:
 ```
 DynamicProperty<Int>(object: controller,
                       keyPath: #keyPath(Controller.value)) <~ aProducer
 ```
 would still be valid even if the `DynamicProperty` has gone out of scope. It would last until `controller` deinitializes, or `aProducer` terminates.


NachoSoto:
> A `DynamicProperty` retains its lensing object.

:-1: I don't think this is the way to go. A `DynamicProperty` represents the values over time for a given object's property, **which does not mean that it //owns// the object**. In fact, it means the complete opposite. This is equivalent to how `NSNotificationCenter` does not retain the object (#2859, for example).

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -5,71 +5,70 @@ import enum Result.NoError
 /// types, including generic types when boxed via `AnyObject`).
 private protocol ObjectiveCRepresentable {
 	associatedtype Value
-	static func extract(from representation: Any) -> Value
-	static func represent(_ value: Value) -> Any
+	static func extract(from representation: Any?) -> Value
```
NachoSoto: Why `Optional`? `representation` can always be an optional (since it's `Any`).

ReactiveCocoa/Swift/DynamicProperty.swift
```diff
@@ -5,71 +5,70 @@ import enum Result.NoError
...
+///               `DynamicProperty` would not be solely teared down by its
+///              deinitialization.
+///
+/// - warning: Use this class only as a last resort. For just observations to
+///            KVO-compliant key paths, use `NSObject.values(forKeyPath:)`.
+///            `MutableProperty` is generally better unless KVC/KVO is
+///            required by the API you're using (for example, `NSOperation`).
 public final class DynamicProperty<Value>: MutablePropertyProtocol {
-	private weak var object: NSObject?
+	private let object: NSObject
```
NachoSoto: 👎

andersio:
It creates a reduced view to the NSObject, which concerns only a certain key path. You are still accessing the object directly, and letting go the "lens" means you are telling the system "I am no longer interested in accessing this directly".

This is actually similar to Groff's idea of "property lenses", where a full property interface capturing the object being lensed can be obtained, and passed around.

ReactiveCocoa/Swift/DynamicProperty.swift
andersio: If you pass an optional to `Any`, it would be recognised as an opaque box (`SwiftValue`).

mdiep:
I agree with @NachoSoto on this.

I think the `weak` semantics are more common here; although you'll certainly want `strong` on occasion. But it's also easier to make the semantics stronger than to make them weaker when you want the other behavior.

I definitely get what you're saying about the lens behavior. But if we want to support those semantics, they need to be in addition to the current semantics—not as a replacement.

mdiep:
I don't think we should give `DynamicProperty` lens semantics. I don't think that's the role it plays. It's meant to make values reactive/observable. It's similar to a lens, but I think going down that route is a mistake that distracts from the core value of RAC.

Going back to your original example:

```swift
DynamicProperty<Int>(object: controller,
                     keyPath: #keyPath(Controller.value)) <~ aProducer
```

I don't think that provides much value. If the `DynamicProperty` isn't kept around to reuse or observe, then there's no point in using a property. You can just as easily use `startWithValues`.

andersio:
Yeah, and that part has been removed. What's left here is the removal of the strong reference, so that `DynamicProperty` needn't be deliberately uniqued.

mdiep:
Yeah, and I just don't think those are the right semantics for `DynamicProperty`. 😞

andersio:
I don't think the semantics have changed per the state of PR, other than lifting the need of uniquing. `DynamicProperty` was bound to the underlying object before so that the bindings can propagate values as usual. This hasn't changed at all.

If uniquing is a deliberate feature, I would argue that the initializer should hide behind something that does the uniquing, instead of the slippery ground we have now.

PR #3166, opened 2016-09-03
[Travis] Use Debug configuration for `carthage build` to reduce the build time

ikesyo:
I'm not sure if this should be merged or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ matrix:
     - script:
         - brew update
         - brew outdated carthage || brew upgrade carthage
-        - carthage build --no-skip-current
+        - carthage build --no-skip-current --configuration Debug
       env:
         - JOB=CARTHAGE
     - os: osx


mdiep:
It looks like it saves some time. So I guess so? I don't anticipate that the release build will ever fail if the debug build works. 😕

NachoSoto:
Nope! I don't think we should. I think we're better off paying the time cost, for the benefit of knowing that RAC works in a production environment. There are many things that are different:
- Swift optimizations.
- Code that may only be compiled under debug.
- Whole module optimization, which has been known to make the Swift compiler crash altogether.

So I'm :-1: on this one.

ikesyo:
> Nope! I don't think we should. I think we're better off paying the time cost, for the benefit of knowing that RAC works in a production environment. There are many things that are different:

For the purposes, instead of building with Release configuration through Carthage, we should build/run the tests with `-configuration Release ENABLE_TESTABILITY=YES` I think.

ikesyo:
> we should build/run the tests with `xcodebuild ... -configuration Release ENABLE_TESTABILITY=YES` I think.

Oh, that is already done:

- https://github.com/ReactiveCocoa/ReactiveCocoa/blob/v4.2.2/script/build#L4
- https://github.com/ReactiveCocoa/ReactiveCocoa/blob/v4.2.2/script/build#L38-L39

NachoSoto:
Example of something we wouldn't have found with this: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/2751

ikesyo:
Anyway, if we attach a prebuilt binary of `carthage build --no-skip-current` to a GitHub release within Travis in future, we cannot use `--configuration Debug` for the artifact. So I'm closing this.

ikesyo:
@NachoSoto FYI at the time of the issue, we didn't use Release configuration for CI builds:

- https://github.com/ReactiveCocoa/ReactiveCocoa/blob/v4.0.1/.travis.yml#L7
- https://github.com/ReactiveCocoa/ReactiveCocoa/blob/v4.0.1/script/cibuild#L109-L121

PR #3167, opened 2016-09-04
Fix casing for enum values in markup

eimantas:
The documentation still contained old naming convention for enum cases (first letter was uppercase). Fixed this.

Also renamed `NotEnabled` case of `ActionError` to `disabled`.


eimantas:
Not sure why checks have failed. The changes are only in the documentation .(

NachoSoto:
Thanks!

PR #3181, opened 2016-09-11
Depend on ReactiveSwift through Carthage

mdiep:


NachoSoto:
Looks like only a flaky test failed. Ship it1

PR #3193, opened 2016-09-17
Adapting to latest changes in ReactiveSwift.

andersio:
`next` -> `value` renaming, specifically.


mdiep:
It looks like a few tests were missed.

Thanks for tackling this!

PR #3195, opened 2016-09-18
Remove Instruments templates

ikesyo:
Those should be only in ReactiveObjC.


mdiep:
Thanks! I wasn't sure. ☺️

PR #3199, opened 2016-09-19
Make `CocoaAction.init` generic over its `Action`

sharplet:
This allows `CocoaAction` to be used to safely type-erase any kind of action.

For a specific example of why this might be useful: https://github.com/ReactiveCocoa/ReactiveSwift/pull/22


mdiep:
This makes sense if we want `ActionProtocol` to be used to create other action types. I think it currently exists only to create constrained extensions on `Action`. Let's resolve https://github.com/ReactiveCocoa/ReactiveSwift/pull/22 first and then discuss this.

sharplet:
On https://github.com/ReactiveCocoa/ReactiveSwift/pull/22 we ended up working out a way to remove the need for a different concrete `ActionProtocol` implementation. Closing this for now.

PR #3200, opened 2016-09-19
Add convenience `CocoaAction` initialiser for `Void` actions

sharplet:
`unsafeCocoaAction` is just that: unsafe. It's not initially obvious how to bind `Void` actions to buttons, and this is a step towards a convenient and safe way of achieving that.

### DIFF
--- a/ReactiveCocoa/CocoaAction.swift
+++ b/ReactiveCocoa/CocoaAction.swift
@@ -70,7 +70,14 @@ public final class CocoaAction: NSObject {
 	public convenience init<Action: ActionProtocol>(_ action: Action, input: Action.Input) {
 		self.init(action, { _ in input })
 	}
-	
+
+	/// Initializes a Cocoa action that will ignore its input.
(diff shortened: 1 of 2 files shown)


sharplet:
If #3199 is merged this should be re-targeted to master.

mdiep:
This will have to be revisited after #3210 is merged.

liscio:
@sharplet Does this need to be updated / revisited? I was about to re-target but realized it also pulled in the `as-generic-cocoa-action` branch which had its related PR closed.

sharplet:
With https://github.com/ReactiveCocoa/ReactiveSwift/pull/22 still up in the air it's unclear what the direction will be here for generic actions. I'd like to try and revisit it soon.

andersio:
Just found that It is already present.
https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/CocoaAction.swift#L50

PR #3202, opened 2016-09-20
Specified commit of ReactiveSwift in Cartfile

pawel-sekara:
Just to prevent conflicts with Cartfile.resolved
Updated renamed Observer methods in tests.


mdiep:
Thanks! ✨

Issue #3049, opened 2016-07-10
How should we deal with the `rac_` prefixes in Swift 3?


andersio:
The Swift API guidelines set a convention of lower camel cases.

Currently, we have two exceptions in the Swift codebase: `NotificationCenter.rac_notifications(forName:object:)` and `URLSession.rac_data(with:)`. Moreover, since IIRC we are going to merge Rex in as part of the Cocoa Bindings repo at some point, how the `rex_` prefixes are handled should be considered too.

Should we follow, or leave them as is?

-
~~For the two extensions in this repo, the most uncreative name is probably `makeProducer`. But it might not suit everyone's tastes, hmm.~~

> Begin names of factory methods with “make”, e.g. x.makeIterator().

Edit: After a second thought, for producers, they do not have any side effect applied until they are started. So a non-mutating name describing what it would deliver would be appropriate IMO, e.g. just `notifications(forName:object:)`.

mdiep:
We could drop the prefixes since the types make the names unambiguous—at least in most cases. Or we could keep them. I'm on the fence.

If we do keep them, we should use `rac_` here (including the code from Rex) and find a new prefix for ReactiveSwift.

andersio:
Another approach can be introducing a "magic" variable that returns a wrapper of binding points, so that we have things like `view.rac.text <~ property`. It could make the bindings easier to discover too.

mdiep:
ReactiveCocoa 5+ doesn't need to be compatible with Rex. RAC will subsume Rex.

andersio:
Sure, I just meant those bindings brought in by Rex. Sorry if it reads like a mess. :smh

andersio:
`outlets` would be a contender too.

bobgodwinx:
"reactive" is suites better with swift language. "rac_" and "rex_" are very bad namings. I will go for "reactive", "active" or "signal"  so we could have view.reactiveText or view.activeText or view.signalText

NachoSoto:
With Swift and modules there's no longer a risk for runtime collisions, and therefore no need for prefixes. So let's just drop 'em! :D

andersio:
It is fine for the methods. But we can't have overloads for the properties.

andersio:
I think it is safe to draw the conclusion that we cannot avoid/disambiguate compile-time collisions in any way. Suggestions like using operators or magic property just limit the reach of collision to one entity.

[...]

PR #3210, opened 2016-09-22
Cocoa extensions from Rex.

andersio:
#### General
- [x] Updated for Swift 3.0 GM.
- [x] A standalone `rac_signalForSelector` port.
- [x] ~~Hide the `ReactiveCocoaPrivate` module.~~
- [x] Adopted `BindingTarget` for all non-observable key paths.
- [x] Update the documentation.

#### The `Reactive` proxy
- [x] Moved all extension method and properties to `Reactive`.
- [x] Point to a new ReactiveSwift (pre-)release with `Reactive`.

#### CocoaAction
- [x] Consolidated the API of `CocoaAction`.
- [x] Parameterized `CocoaAction` with an assumed sender type.

#### Control bindings and values
- [x] Extensions for `NSControl`.
- [x] Text controls now expose an end-of-editing signal by default, with continuous counterparts added.
- [x] Control values are split into a pair of binding target and signal.
- [x] `UIControl.reactive.trigger(for:)` for all `UIControlEvents`.

## Blocked by:
- [x] `Reactive`. https://github.com/ReactiveCocoa/ReactiveSwift/pull/51
- [x] Scheduler in `BindingTarget`. https://github.com/ReactiveCocoa/ReactiveSwift/pull/49
- [x] API Design: Exposing the control values. https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3229
- [x] API Design: Control Binding API for `Action`s. https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3233

## Review 

- [x] @mdiep 
- [x] @NachoSoto
- Others


ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,74 @@
...
+			property <~
+				host.trigger(for: [.valueChanged, .editingChanged])
+					.map { [unowned host] in getter(host) }
+		}
+	}
+	#endif
+
+	/// Wraps a control's `enabled` state in a bindable property.
+	public var rac_enabled: MutableProperty<Bool> {
+		return associatedProperty(self, key: &enabledKey, initial: { $0.isEnabled }, setter: { $0.isEnabled = $1 })
```
raylillywhite: `rac_enabled`, `rac_selected`, and any other properties like this seem like they should be either:
1) Made to use KVO for observing changes the happen via the original UIKit setter.
2) Left out of ReactiveCocoa until they do provide that two-way binding.
or 3) All be documented with a warning indicating that they're effectively write-only bindings.

andersio:
@raylillywhite It would be addressed by something like `AnyBindingTarget`.

andersio:
Requires https://github.com/ReactiveCocoa/ReactiveSwift/pull/30 to fix a deadlock in `rac_dismissAnimated`, which should be write only anyway.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -1,9 +1,40 @@
 import Foundation
 import ReactiveSwift
 import enum Result.NoError
+import ReactiveCocoaPrivate
 
 extension NSObject {
 	public func signal(for selector: Selector) -> Signal<(), NoError> {
```
sharplet: Do we expect to support accessing the arguments?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -1,9 +1,40 @@
...
-		return .empty
+		objc_sync_enter(self)
+		defer { objc_sync_exit(self) }
+
+		let map: NSMutableDictionary = {
```
sharplet: Any reason not to make this `[Selector: Signal<(), NoError>]`?

ReactiveCocoa/NSObject+Intercepting.swift
andersio: I have made progress on this. Do you think the arguments should be bridged as `[Any]`?

andersio:
The `signal(for:)` with argument would not be in this PR.

Some notes about my attempt:
 1. TypeEncoding: https://gist.github.com/andersio/844e7a1d953b0acbd5e0b5a70b2c0501
 2. I wrapped `NSInvocation` in a dummy ObjC class, and pass it via `rac_receiver_t`. Hence, there is a private `signal(for:action:)` which allows `trigger(for:)` to ignore the arguments at no cost.

andersio:
BindingTarget in action:
https://github.com/ReactiveCocoa/ReactiveCocoa/compare/rex...rex-landing?expand=1

andersio:
A little note on the API naming in `NSObject+Intercepting`:

1. `trigger(for:)` returns a "trigger" in ReactiveSwift's convention.
2. `signal(for:)` is _reserved_ for the variant which pushes arguments.

ReactiveCocoa/UIKit/UIViewController.swift
```diff
@@ -48,22 +48,9 @@ extension UIViewController {
...
-
-		let property = associatedProperty(self, key: &dismissModally, initial: initial, setter: setter) { property in
-			property <~ self.trigger(for: #selector(UIViewController.dismiss))
-				.take { _ in property.value != nil }
-				.map { _ in return nil }
-		}
-
-		return property
+	public var rac_dismissAnimated: BindingTarget<DismissingInformation> {
+		/// TODO: Convert into `Action`?
```
andersio: Related: https://github.com/ReactiveCocoa/ReactiveSwift/pull/38

ReactiveCocoa.xcodeproj/project.pbxproj
```diff
@@ -949,6 +1260,8 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = A97451351B3A935E00F48E55 /* watchOS-Framework.xcconfig */;
 			buildSettings = {
+				CLANG_ENABLE_MODULES = YES;
+				DEFINES_MODULE = YES;
```
mdiep: It seems like most of these built setting changes should be reverted.

ReactiveCocoa/AppKit/NSTextField.swift
```diff
@@ -0,0 +1,22 @@
...
+//  Copyright (c) 2015 Neil Pankey. All rights reserved.
+//
+
+import ReactiveSwift
+import AppKit
+import enum Result.NoError
+
+extension NSTextField {
+	/// Sends the field's string value whenever it changes.
+	public var rac_textSignal: SignalProducer<String, NoError> {
```
mdiep: This should probably be a property?

ReactiveCocoa/CocoaAction.swift
```diff
@@ -19,7 +27,19 @@ public final class CocoaAction: NSObject {
...
+	public var isEnabledProducer: SignalProducer<Bool, NoError> {
+		return values(forKeyPath: #keyPath(CocoaAction.isEnabled))
+			.map { $0 as! Bool }
+	}
+
+	/// Creates a producer for the `executing` state of a CocoaAction.
+	public var isExecutingProducer: SignalProducer<Bool, NoError> {
+		return values(forKeyPath: #keyPath(CocoaAction.isExecuting))
+			.map { $0 as! Bool }
+	}
```
mdiep: Should we just have `public var isEnabled: Property<Bool>` here like `Action` does?

I'm not sure why `CocoaAction` is different.

ReactiveCocoa/Association.swift
```diff
@@ -0,0 +1,76 @@
...
+		}, setter: setter)
+}
+
+/// Attaches a `MutableProperty` value to the `host` object under `key`. The property is
+/// initialized with the result of `initial`. Changes on the property's producer are
+/// monitored and written to `setter`.
+///
+/// This can be used as an alternative to `DynamicProperty` for creating strongly typed
+/// bindings on Cocoa objects.
+public func associatedProperty<Host: AnyObject, T>(_ host: Host, key: UnsafeRawPointer, initial: (Host) -> T, setter: @escaping (Host, T) -> (), setUp: (MutableProperty<T>) -> () = { _ in }) -> MutableProperty<T> {
```
mdiep: If this used `key: StaticString = #function`, I think it could simplify many of the callers.

ReactiveCocoa/NSObject+BindingTarget.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+import ReactiveSwift
+
+internal protocol NSObjectBindingTargetProtocol: class {
```
mdiep: Why not extend `NSObject` directly?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+import ReactiveSwift
+import enum Result.NoError
+import ReactiveCocoaPrivate
+
+extension NSObject {
+	public func trigger(for selector: Selector) -> Signal<(), NoError> {
```
mdiep: Needs documentation

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -0,0 +1,44 @@
...
+		return signal(for: selector) { observer in
+			return { _ in observer.send(value: ()) }
+		}
+	}
+
+	private func signal<U>(for selector: Selector, action: (Observer<U, NoError>) -> rac_receiver_t) -> Signal<U, NoError> {
```
mdiep: Needs documentation

ReactiveCocoa/ReactiveCocoaPrivate/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module ReactiveCocoaPrivate {
+	header "../RACObjCRuntimeUtilities.h"
+	export *
+}
```
mdiep: Is this needed? 😕

ReactiveCocoa/Reusable.swift
```diff
@@ -0,0 +1,45 @@
...
+	///         // do other things
+	///      }
+	///
+	/// label.rac_text <~
+	///     titleProperty
+	///         .producer
+	///         .takeUntil(self.rac_prepareForReuse)
+	/// ```
+	///
+	public var rac_prepareForReuse: Signal<Void, NoError> {
```
mdiep: This should use `()` instead of `Void` for consistency with the rest of the codebase.

ReactiveCocoa/UIKit/UIActivityIndicatorView.swift
```diff
@@ -0,0 +1,26 @@
...
+
+import ReactiveSwift
+import UIKit
+
+extension UIActivityIndicatorView {
+
+	/// Wraps an indicator's `isAnimating()` state in a bindable property.
+	/// Setting a new value to the property would call `startAnimating()` or
+	/// `stopAnimating()` depending on the value.
+	public var rac_animating: BindingTarget<Bool> {
```
mdiep: I think we should add `Reactive` to make this `rac.animating` here before people start using it. (#3049)

ReactiveCocoa/UIKit/UIActivityIndicatorView.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+import ReactiveSwift
+import UIKit
+
+extension UIActivityIndicatorView {
+
+	/// Wraps an indicator's `isAnimating()` state in a bindable property.
+	/// Setting a new value to the property would call `startAnimating()` or
+	/// `stopAnimating()` depending on the value.
```
mdiep: This documentation needs updating now that it's not a `MutableProperty`.

ReactiveCocoa/Association.swift
```diff
@@ -0,0 +1,76 @@
...
+import ReactiveSwift
+
+/// Attaches a `MutableProperty` value to the `host` object using KVC to get the initial
+/// value and write subsequent updates from the property's producer. Note that `keyPath`
+/// is a `StaticString` because it's pointer value is used as key value when associating
+/// the property.
+///
+/// This can be used as an alternative to `DynamicProperty` for creating strongly typed
+/// bindings on Cocoa objects.
+public func associatedProperty(_ host: AnyObject, keyPath: StaticString) -> MutableProperty<String> {
```
mdiep: Maybe we should make these `internal` for now?

I'd rather be conservative in terms of API and encourage people to contribution missing items.

ReactiveCocoa/UIKit/UIBarItem.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by Bjarke Hesthaven Søndergaard on 24/07/15.
+//  Copyright (c) 2015 Neil Pankey. All rights reserved.
+//
+
+import ReactiveSwift
+import UIKit
+
+extension UIBarItem {
+	/// Wraps a UIBarItem's `enabled` state in a bindable property.
```
mdiep: Documentation should be updated.

ReactiveCocoa/NSObject+BindingTarget.swift
andersio: `Self` is available only in protocol extensions.

ReactiveCocoa/ReactiveCocoaPrivate/module.modulemap
andersio: Yes, it is necessary to hide the private API in the ObjC headers.

ReactiveCocoa.xcodeproj/project.pbxproj
```diff
@@ -1294,7 +1579,6 @@
 					"$(inherited)",
 				);
 				INFOPLIST_FILE = ReactiveCocoaTests/Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
```
andersio: I am not so sure about `LD_RUNPATH_SEARCH_PATHS`. But in many cases the additions in the target are the same as the inherited paths.

mdiep: In Xcode 8, you can view the _Customized_ build settings with _Levels_. I'd toggle that and delete anything that's redundant.

andersio:
Test failure in `UIButtonTests` would be fixed by https://github.com/ReactiveCocoa/ReactiveSwift/pull/49.

ReactiveCocoa.xcodeproj/project.pbxproj
andersio: Yeah, they are all removed. ReactiveSwift seems to have the same redundancy.

ReactiveCocoa/CocoaAction.swift
andersio: Done in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/6952c5310449f27a8ddd3ac199bbf0c2872dcc41.

ReactiveCocoa/Association.swift
andersio: Done in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/2bacf27709efef538fc896b3100e098807d7e33a.

ReactiveCocoa.xcodeproj/project.pbxproj
andersio: Done in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/23ae7c002fa2b3c05fb04797ed0457c0c8e05ee4.

ReactiveCocoa/UIKit/UIActivityIndicatorView.swift
andersio: I think `rac` is debatable. We _might_ want binding target extensions of Foundation in ReactiveSwift too, which makes `rac` a bit less ideal. But `rac` is probably also the only choice anyway.

Or we may encourage operators instead, say `%view.isAnimating` as an equivalent of `view.reactive.isAnimating`

mdiep: I think `rac` makes sense for now. It's easily pronounceable and fits with the history of the project.

We can consider an operator in the future, but that could be an additive change.

andersio:
Should we move also our public extension methods to `.rac`? Say `NSObject.values`.

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,69 @@
...
+		}
+	}
+
+	#if os(iOS)
+	/// Creates a bindable property to wrap a control's value.
+	///
+	/// This property uses `UIControlEvents.ValueChanged` and `UIControlEvents.EditingChanged`
+	/// events to detect changes and keep the value up-to-date.
+	//
+	internal func value<T>(getter: @escaping (Reactant) -> T, setter: @escaping (Reactant, T) -> ()) -> MutableProperty<T> {
```
andersio: Was there a specific reason why tvOS didn't use this for `UITextField`?

/cc @neilpa @RuiAAPeres @dmcrodrigues

neilpa: IIRC `UIControlEvents` weren't available in the tvOS SDK when I added build support for it.

https://github.com/RACCommunity/Rex/commit/e4c4aa4594c732dc5c7d9535d26a88218e734eaf

Didn't comment it but this commit seems to suggest it.

andersio:
A new type `MutablePropertyFacade` is introduced for UIControl/NSControl for two main reasons:

1. The current implementation would abuse the fact that manually setting the state would not send an action message.

    In other words, for each user-triggered action message, the property would manually set the state _again_ with the fresh value. While it seems to do no harm in practice, shall we all agree this is not ideal.

2. Vulnerable to accidental retainment.

andersio:
While `dismissAnimated` the fringe case in the PR feels appropriate as an `Action`, it could open a confusing gap.

```
vc.rac.dismissAnimated.events.filter { $0 == .completed }
vc.rac.trigger(for: #selector(dismiss(animated:)))
```

They look the same, and the users may expect them to be equivalent. But the first one only works on calls done through the `Action`. 

Therefore, it was axed in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/1797e79ae8c4e401ba52a40a13e2a412d0497b76, and replaced with just a `dismissedPresented` trigger.

ReactiveCocoa/UIKit/UIViewController.swift
```diff
@@ -35,23 +35,9 @@ extension Reactivity where Reactant: UIViewController {
...
-	/// It mimics the same input as `dismissViewControllerAnimated`: a `Bool` flag for the animation
-	/// and a `(Void -> Void)?` closure for `completion`.
-	/// E.g:
-	/// ```
-	/// //Dismissed with animation (`true`) and `nil` completion
-	/// viewController.dismissAnimated <~ aProducer.map { _ in (true, nil) }
-	/// ```
-	/// The dismissal observation can be made either with binding (example above)
-	/// or `viewController.dismissViewControllerAnimated(true, completion: nil)`
-	public var dismissAnimated: BindingTarget<DismissingInformation> {
```
andersio: While `dismissAnimated` the fringe case feels appropriate as an `Action`, it could open a confusing gap.

```
vc.rac.dismissAnimated.events.filter { $0 == .completed }
vc.rac.trigger(for: #selector(dismiss(animated:)))
```

They look the same, and the users would likely expect them to be equivalent. But the first one only works on calls done through the `Action`, while the later is triggered upon all calls.

Therefore, it was axed in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/1797e79ae8c4e401ba52a40a13e2a412d0497b76, and replaced with just a `dismissedPresented` trigger.

ReactiveCocoa/ControlSubject.swift
```diff
@@ -0,0 +1,44 @@
+import ReactiveSwift
+import enum Result.NoError
+
+/// Represents a subject of an interactive control.
+///
+/// Generally speaking, `ControlSubject` emits changes only for the user
+/// interactions in which the subject specializes.
+public final class ControlSubject<Value> {
```
liscio: @mdiep see my concerns w/ properties in UI controls here: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3229#issuecomment-251395000

andersio: @mdiep The problem here is that the assumed behavior of `Property` does not match how `Control` behaves.

1. For every value you set, a `next` event is emitted by the property. For `UIControl`, this never happens for any values being set manually.
2. Because of (1), the latest value emitted by `producer` is NOT guaranteed to be `value` in the context of `UIControl`.

liscio: @sharplet The trouble boils down to multiple "sources of truth" in one's code. Ultimately we want a UI control to reflect a value that we've set somewhere in the (View)Model, but we rarely/never want to initialize our model using the current value of a control.

Yes, the *implementation* of a control certainly requires the storage of its own value so that it can be redrawn, replaced dynamically, etc. But we _should not_ expect that the control contains _any_ truth at all and hence should not be asked to store values on our behalf. It is a simple contract to fulfill: just display the value I ask you to, and notify me when the user makes changes to that value.

In contrast, to expose a "proper property" then we are asking the control to let us know if anything programmatically changes that value as well as the current/initial value of the control. In order to do that, it must also behave like a "proper property" in that our setting of the property's value directly should propagate back to us.

The alternative to the above was to make a "kinda-sorta property" that didn't fulfill the expectations of all the other Properties in the framework, which was a huge red flag in my eyes.

(Also, what @andersio said above.)

mdiep: That all makes sense.

But the pragmatism of conforming to `PropertyProtocol` might be preferable to the ideal of not asking controls for a value.

Or maybe this just needs a better name to really clarify the concept.

andersio: After reviewing my words, item (2) in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210#discussion_r81995270 was actually false. Sorry about that.

-

Anyway, I think we might be able to reduce the problem into whether the one-value-for-one-event relationship is significant enough in practical.

liscio: @mdiep I agree regarding the name, but disagree about "softening" the protocol behavior so we don't have to come up with a new thing. :)

Consider for example the `Sequence` vs `Collection` in the standard library. Imagine if `Sequence`s never existed, and "some `Collection`s just return -1 for `count` when they are infinite in size.

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -10,50 +10,61 @@ import ReactiveSwift
...
-	@objc fileprivate func sendNext(_ receiver: Any?) {
-		observer.send(value: receiver as! Control)
+	public var controlEventsForAction: UIControlEvents? {
+		return associatedAction.value?.controlEvents
 	}
-}
 
-extension Reactivity where Reactant: UIControl {
-	public func trigger(for events: UIControlEvents) -> Signal<Reactant, NoError> {
+	public func setAction(_ action: CocoaAction<Reactant>?, for controlEvents: UIControlEvents) {
```
sharplet: Did we agree to go with `ReactiveExtensionProvider`/`Base` instead of `Reactivity`/`Reactant`?

sharplet: I'm aware I could be out of context and missing something

ReactiveCocoa/Association.swift
andersio: Implemented in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/7e369e5d19bff97080a08bdc4dbfbc1cc628b456#diff-448eaf1c1acb3e1c0522257283b94f72R22.

ReactiveCocoa/NSObject+BindingTarget.swift
```diff
@@ -0,0 +1,25 @@
...
+	///
+	/// - important: The binding target is bound to the main queue.
+	///
+	/// - parameters:
+	///   - action: The action to consume values from the bindings.
+	///
+	/// - returns:
+	///   A binding target that holds no strong references to the object.
+	internal func makeBindingTarget<U>(action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
+		let scheduler = associatedValue { _ in UIScheduler() }
```
mdiep: Why is the scheduler created with an associated value?

ReactiveCocoa/NSObject+BindingTarget.swift
```diff
@@ -0,0 +1,25 @@
...
+	/// the object has not deinitialized.
+	///
+	/// - important: The binding target is bound to the main queue.
+	///
+	/// - parameters:
+	///   - action: The action to consume values from the bindings.
+	///
+	/// - returns:
+	///   A binding target that holds no strong references to the object.
+	internal func makeBindingTarget<U>(action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
```
mdiep: The _make_ prefix here seems weird. Maybe just `bindingTarget`?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -0,0 +1,54 @@
...
+		assert(isSuccessful)
+
+		lifetime.ended.observeCompleted(observer.sendCompleted)
+		map.setObject(signal, forKey: selectorName)
+
+		return signal
+	}
+}
+
+private var mapKey = 0
```
mdiep: This seems to be unused?

ReactiveCocoa/NSObject+BindingTarget.swift
andersio: Just following the `make` rule for factory methods here. e.g. `makeIterator`.

ReactiveCocoa/NSObject+BindingTarget.swift
andersio: It is intended to be one UIScheduler per binding target (across instances).

ReactiveCocoa/NSObject+Intercepting.swift
andersio: Removed in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/a7431e4f8f314e3dbc5a58efaa4f51b9cb9248d6.

ReactiveCocoa/RACObjCRuntimeUtilities.h
```diff
@@ -0,0 +1,10 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+typedef void (^rac_receiver_t)(void);
```
mdiep: I think this block makes things more confusing—at least on the Swift side. If we don't use a `typedef`, does it come through as `() -> Void`?

ReactiveCocoa/UIKit/UIButton.swift
```diff
@@ -0,0 +1,32 @@
...
+		return makeBindingTarget { $0.setTitle($1, for: .normal) }
+	}
+
+	/// Sets the title of the button for the specified state.
+	public func title(for state: UIControlState) -> BindingTarget<String> {
+		return makeBindingTarget { $0.setTitle($1, for: state) }
+	}
+}
+
+private var pressedKey: UInt8 = 0
```
mdiep: This is unused.

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,93 @@
+import ReactiveSwift
+import UIKit
+import enum Result.NoError
+
+private class UnsafeControlReceiver: NSObject {
```
mdiep: Seems like this and the AppKit version could be combined?

ReactiveCocoa/UIKit/UIProgressView.swift
```diff
@@ -0,0 +1,9 @@
+import ReactiveSwift
+import UIKit
+
+extension Reactive where Base: UIProgressView {
+	/// Sets the reative progress to be reflected by the progress view.
```
mdiep: _reactive_ is misspelled here, but can probably be removed completely

ReactiveCocoa/RACObjCRuntimeUtilities.h
andersio: This is deliberately made so that it can be extended as something like `(RACUnsafeArgumentTuple) -> Void`.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -0,0 +1,52 @@
...
+	/// of `selector` on the object.
+	///
+	/// - parameters:
+	///   - selector: The selector to observe.
+	///
+	/// - returns:
+	///   A trigger signal.
+	public func trigger(for selector: Selector) -> Signal<(), NoError> {
+		return signal(for: selector) { observer in
+			return { _ in observer.send(value: ()) }
```
andersio: This was deliberately made for the scrapped variant with bridged arguments. But yeah, I would revert this and leave it to @sharplet.

ReactiveCocoa/UIKit/UIProgressView.swift
andersio: Fixed

ReactiveCocoa/UIKit/UIControl.swift
andersio: Succeeded by `CocoaTrigger`.

ReactiveCocoa/UIKit/UIButton.swift
andersio: Removed.

ReactiveCocoa/AppKit/NSControl.swift
```diff
@@ -0,0 +1,124 @@
...
+			base.action = #selector(ActionMessageReceiver.receive)
+
+			return receiver
+		}
+
+		return receiver.trigger
+	}
+}
+
+private class ActionMessageReceiver: NSObject {
```
andersio: Succeeded by `CocoaTrigger`.

ReactiveCocoa/CocoaTrigger.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import ReactiveSwift
+import enum Result.NoError
+
+internal class CocoaTrigger: NSObject {
```
mdiep: I'd call this `CocoaTarget`. It provides a trigger, but it is a target.

mdiep:
@NachoSoto This is looking good to me. Do you want to take a look?

There's some relevant discussion in #3229 and #3233 about the approach taken here.

mdiep:
I'll give @NachoSoto a couple more days to look at or respond to this before moving forward. ☺️

NachoSoto:
I'll look at this first time Monday!

AlbertoGuerreroMartin:
@andersio what about UITableViewCell's rex_prepareForReuse property? I'm having problems using UIButton's `reactive.trigger(for: UIControlEvents)` inside UITableViewCell, due to cell reuse. Is there any way to handle this?
Thanks in advance 😊

andersio:
@AlbertoGuerreroMartin You can use `trigger(for: #selector(self.prepareForReuse))`.

AlbertoGuerreroMartin:
@andersio thanks a lot!

ReactiveCocoa/Reusable.swift
```diff
@@ -0,0 +1,10 @@
+import ReactiveSwift
+import enum Result.NoError
+
+public protocol Reusable: class {}
```
andersio: `prepareForReuse` wasn't listed as a requirement due to `NSCollectionViewItem`. The compiler complains about the availability of `prepareForReuse` only since `OS X application extensions 10.9` even with the presence of `@available` annotation.

ReactiveCocoa/Reusable.swift
```diff
@@ -0,0 +1,10 @@
...
+
+extension Reactive where Base: NSObject, Base: Reusable {
+	public var prepareForReuse: Signal<(), NoError> {
+		return trigger(for: Selector(("prepareForReuse")))
```
andersio: Quoted to suppress compiler warning.

sharplet:
Along the same lines as `prepareForReuse`, I think it would be great to expose convenience methods for `viewWillAppear`, `viewWillDisappear`, _et al._

liscio:
I think that for the sake of keeping ourselves from going crazy, we should stop adding too many more of these convenience methods and make sure we've got the baseline tools sorted out first. :)

I'd much rather that @andersio rolls back https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/commits/e48c43694c212939997f9c382c1886cd99d846e6 and we add a bunch of these kinds of convenience additions at once in a separate PR. It's already taking quite long to get this PR reviewed & merged as it stands without it gaining new features so quickly. :)

sharplet:
Agreed.

ReactiveCocoa/AppKit/NSTextField.swift
```diff
@@ -0,0 +1,34 @@
+import ReactiveSwift
+import AppKit
+import enum Result.NoError
+
+extension Reactive where Base: NSTextField {
+	/// A signal of values in `String` from the text field upon any changes.
+	public var continuousStringValues: Signal<String, NoError> {
+		var signal: Signal<String, NoError>!
```
NachoSoto: This pattern is used a lot. I created https://github.com/ReactiveCocoa/ReactiveSwift/pull/66 in `ReactiveSwift` so that at least we don't repeat it there, we might want to do something similar in `ReactiveCocoa`, but this is fine for now.

ReactiveCocoa/AppKit/NSTextField.swift
```diff
@@ -0,0 +1,34 @@
...
+	/// A signal of values in `NSAttributedString` from the text field upon any
+	/// changes.
+	public var continuousAttributedStringValues: Signal<NSAttributedString, NoError> {
+		var signal: Signal<NSAttributedString, NoError>!
+
+		NotificationCenter.default
+			.reactive
+			.notifications(forName: .NSControlTextDidChange, object: base)
+			.take(during: lifetime)
+			.map { ($0.object as! NSTextField).attributedStringValue }
```
NachoSoto: I would separate this:
```swift
.map { $0.object as! NSTextField }
.map { $0.attributedStringValue }
```

ReactiveCocoa/CocoaTarget.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+import ReactiveSwift
+import enum Result.NoError
+
+/// A target that accepts action messages.
+internal class CocoaTarget: NSObject {
```
NachoSoto: This should be `final`.

ReactiveCocoa/NSObject+Association.swift
```diff
@@ -0,0 +1,22 @@
...
+	/// associated subsequently.
+	///
+	/// - parameters:
+	///   - key: An optional key to differentiate different values.
+	///   - initial: The action that supples an initial value.
+	///
+	/// - returns:
+	///   The associated value for the specified key.
+	internal func associatedValue<T>(forKey key: StaticString = #function, initial: (Base) -> T) -> T {
+		var value = objc_getAssociatedObject(base, key.utf8Start) as? T
```
NachoSoto: `key.utf8Start` is duplicated in this method. Can you extract it?

NachoSoto: What you want here is `as! T?`, not `as? T`. We don't want to ignore a potentially existing associated object of the wrong type.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -0,0 +1,37 @@
...
+	/// - returns:
+	///   A trigger signal.
+	public func trigger(for selector: Selector) -> Signal<(), NoError> {
+		objc_sync_enter(self)
+		defer { objc_sync_exit(self) }
+
+		let map = associatedValue { _ in NSMutableDictionary() }
+
+		let selectorName = String(describing: selector) as NSString
+		if let signal = map.object(forKey: selectorName) as? Signal<(), NoError> {
```
NachoSoto: This should be `as! Signal<(), NoError>?`

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -0,0 +1,37 @@
...
+			return signal
+		}
+
+		let (signal, observer) = Signal<(), NoError>.pipe()
+		let isSuccessful = base._rac_setupInvocationObservation(for: selector,
+		                                                       protocol: nil,
+		                                                       receiver: observer.send(value:))
+		assert(isSuccessful)
```
NachoSoto: This should be `precondition`, we don't want to ignore this.

ReactiveCocoa/Reusable.swift
NachoSoto: This is extremely fragile though because `extension Reactive where Base: NSObject, Base: Reusable` does not guarantee that `Base` has a `prepareForReuse` method 👎

Where is that error exactly?

ReactiveCocoa/Reusable.swift
NachoSoto: This should be `#selector(self.prepareForReuse)`.

ReactiveCocoa/UIKit/UIBarButtonItem.swift
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	/// The action to be triggered when the button is pressed. It also controls
+	/// the enabled state of the button.
+	public var pressed: CocoaAction<Base>? {
+		get {
+			return associatedAction.value?.action
+		}
+
+		nonmutating set {
```
NachoSoto: TIL

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,81 @@
...
+	/// control events specified by `controlEvents`.
+	///
+	/// - parameters:
+	///   - action: The action to be associated.
+	///   - controlEvents: The control event mask.
+	internal func setAction(_ action: CocoaAction<Base>?, for controlEvents: UIControlEvents) {
+		associatedAction.modify { associatedAction in
+			if let old = associatedAction {
+				old.disposable.dispose()
+			}
```
NachoSoto: This is simpler as `associatedAction?.disposable.dispose()`

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,81 @@
...
+
+			if let action = action {
+				base.addTarget(action, action: CocoaAction<Base>.selector, for: controlEvents)
+
+				let disposable = CompositeDisposable()
+				disposable += isEnabled <~ action.isEnabled
+				disposable += { [weak base = self.base] in
```
NachoSoto: This can be `unowned`, right? The lifetime guarantees that it won't outlive `self.base`, so no reason to use `weak`.

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,81 @@
...
+	/// - parameters:
+	///   - controlEvents: The control event mask.
+	///
+	/// - returns:
+	///   A trigger signal.
+	public func trigger(for controlEvents: UIControlEvents) -> Signal<(), NoError> {
+		return Signal { observer in
+			let receiver = CocoaTarget(observer)
+			base.addTarget(receiver,
+			                   action: #selector(CocoaTarget.sendNext),
```
NachoSoto: This should be `#selector(receiver.sendNext)` to ensure that the receiver actually has that selector.

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,81 @@
...
+			                   for: controlEvents)
+
+			let disposable = lifetime.ended.observeCompleted(observer.sendCompleted)
+
+			return ActionDisposable { [weak base = self.base] in
+				disposable?.dispose()
+
+				base?.removeTarget(receiver,
+				                   action: #selector(CocoaTarget.sendNext),
```
NachoSoto: Ditto

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -0,0 +1,81 @@
...
+	public func trigger(for controlEvents: UIControlEvents) -> Signal<(), NoError> {
+		return Signal { observer in
+			let receiver = CocoaTarget(observer)
+			base.addTarget(receiver,
+			                   action: #selector(CocoaTarget.sendNext),
+			                   for: controlEvents)
+
+			let disposable = lifetime.ended.observeCompleted(observer.sendCompleted)
+
+			return ActionDisposable { [weak base = self.base] in
```
NachoSoto: Ditto `unowned`.

ReactiveCocoa/UIKit/UITextView.swift
```diff
@@ -0,0 +1,38 @@
...
+
+	/// A signal of text values emitted by the text view upon end of editing.
+	public var textValues: Signal<String, NoError> {
+		var signal: Signal<String, NoError>!
+
+		NotificationCenter.default
+			.reactive
+			.notifications(forName: .UITextViewTextDidEndEditing, object: base)
+			.take(during: lifetime)
+			.map { ($0.object as! UITextView).text! }
```
NachoSoto: Ditto separate this into 2 `map`s.

ReactiveCocoa/UIKit/UITextView.swift
```diff
@@ -0,0 +1,38 @@
...
+
+extension Reactive where Base: UITextView {
+	/// Sets the text of the text view.
+	public var text: BindingTarget<String> {
+		return makeBindingTarget { $0.text = $1 }
+	}
+
+	/// A signal of text values emitted by the text view upon end of editing.
+	public var textValues: Signal<String, NoError> {
+		var signal: Signal<String, NoError>!
```
NachoSoto: Actually it looks like this pattern is repeated way too much.
Can you add an `internal` method to `RAC` like https://github.com/ReactiveCocoa/ReactiveSwift/pull/66 to simplify all these?

ReactiveCocoaTests/KeyValueObservingSpec.swift
```diff
@@ -156,7 +158,9 @@ class KeyValueObservingSpec: QuickSpec {
 				it("should handle changes being made on another queue") {
 					var observedValue = 0
 
-					testObject.values(forKeyPath: "rac_value")
+					testObject
+						.reactive
+						.values(forKeyPath: "rac_value")
```
NachoSoto: Can you use `#keypath` for all these?

ReactiveCocoaTests/UIKit/ReusableComponentsSpec.swift
```diff
@@ -0,0 +1,52 @@
...
+
+		describe("UICollectionReusableView") {
+			let cell = UICollectionReusableView()
+
+			var isTriggered = false
+			cell.reactive.prepareForReuse.observeValues {
+				isTriggered = true
+			}
+
+			expect(isTriggered) == false
```
NachoSoto: These are not inside of a test (`it`) so it's not running at all!

ReactiveCocoaTests/UIKit/UIActivityIndicatorViewTests.swift
```diff
@@ -0,0 +1,26 @@
+import XCTest
+import ReactiveSwift
+import ReactiveCocoa
+import Result
+
+class UIActivityIndicatorTests: XCTestCase {
```
NachoSoto: Why not use `Quick` for this?

ReactiveCocoaTests/UIKit/UIBarButtonItemTests.swift
```diff
@@ -0,0 +1,38 @@
+import ReactiveSwift
+import ReactiveCocoa
+import UIKit
+import XCTest
+import enum Result.NoError
+
+class UIBarButtonItemTests: XCTestCase {
```
NachoSoto: Ditto. Please use `Quick`.

ReactiveCocoaTests/UIKit/UIButtonTests.swift
```diff
@@ -0,0 +1,88 @@
...
+import enum Result.NoError
+
+extension UIButton {
+	static func button() -> UIButton {
+		let button = UIButton(type: UIButtonType.custom)
+		return button
+	}
+}
+
+class UIButtonTests: XCTestCase {
```
NachoSoto: Ditto

ReactiveCocoaTests/UIKit/UIButtonTests.swift
```diff
@@ -0,0 +1,88 @@
+import ReactiveSwift
+import ReactiveCocoa
+import UIKit
+import XCTest
+import enum Result.NoError
+
+extension UIButton {
```
NachoSoto: This should be `fileprivate`, we don't want to leak it to other files

ReactiveCocoaTests/UIKit/UIButtonTests.swift
```diff
@@ -0,0 +1,88 @@
...
+	static func button() -> UIButton {
```
NachoSoto: Maybe `create()` instead?

ReactiveCocoaTests/UIKit/UIButtonTests.swift
```diff
@@ -0,0 +1,88 @@
...
+		let button = UIButton(type: UIButtonType.custom)
```
NachoSoto: Can just be `.custom`

ReactiveCocoaTests/UIKit/UIButtonTests.swift
```diff
@@ -0,0 +1,88 @@
...
+		return button
```
NachoSoto: No need to define the output and return it, just inline it please.

ReactiveCocoaTests/UIKit/UIButtonTests.swift
```diff
@@ -0,0 +1,88 @@
...
+class UIButtonTests: XCTestCase {
+	weak var _button: UIButton?
+
+	override func tearDown() {
+		XCTAssert(_button == nil, "Retain cycle detected in UIButton properties")
+		super.tearDown()
+	}
+
+	func testEnabledPropertyDoesntCreateRetainCycle() {
+		let button = UIButton(frame: CGRect.zero)
```
NachoSoto: This can be just `.zero`

ReactiveCocoaTests/UIKit/UIControl+EnableSendActionsForControlEvents.swift
```diff
@@ -0,0 +1,46 @@
...
+}()
+
+/// Unfortunately, there's an apparent limitation in using `sendActionsForControlEvents`
+/// on unit-tests for any control besides `UIButton` which is very unfortunate since we
+/// want test our bindings for `UIDatePicker`, `UISwitch`, `UITextField` and others
+/// in the future. To be able to test them, we're now using swizzling to manually invoke
+/// the pair target+action.
+extension UIControl {
+	override open class func initialize() {
+		if self !== UIControl.self {
```
NachoSoto: This would be better as `guard` so that it ensures it returns.

ReactiveCocoaTests/UIKit/UIControlTests.swift
```diff
@@ -0,0 +1,99 @@
+import ReactiveSwift
+import ReactiveCocoa
+import UIKit
+import XCTest
+import enum Result.NoError
+
+class UIControlTests: XCTestCase {
```
NachoSoto: Ditto.

ReactiveCocoaTests/UIKit/UIDatePickerTests.swift
```diff
@@ -0,0 +1,38 @@
+import ReactiveSwift
+import ReactiveCocoa
+import UIKit
+import XCTest
+
+class UIDatePickerTests: XCTestCase {
```
NachoSoto: Ditto.

ReactiveCocoaTests/UIKit/UIImageViewTests.swift
```diff
@@ -0,0 +1,64 @@
+import ReactiveSwift
+import ReactiveCocoa
+import UIKit
+import XCTest
+import enum Result.NoError
+
+class UIImageViewTests: XCTestCase {
```
NachoSoto: Ditto

ReactiveCocoaTests/UIKit/UILabelTests.swift
```diff
@@ -0,0 +1,80 @@
+import ReactiveSwift
+import ReactiveCocoa
+import UIKit
+import XCTest
+import enum Result.NoError
+
+class UILabelTests: XCTestCase {
```
NachoSoto: Ditto

NachoSoto:
Awesome job dude!

ReactiveCocoa/Reusable.swift
andersio: As said the compiler complains about the conformance of `NSCollectionViewItem`'s `prepareForReuse` to `Resuable`.

screen shot 2016-10-17 at 11 59 19 pm

ReactiveCocoa/Reusable.swift
andersio: Only if `Reusable` has the requirement.

ReactiveCocoa/Reusable.swift
andersio: I will leave the `Reusable` thingy in a separate PR.

ReactiveCocoa/UIKit/UIControl.swift
andersio: The `ScopeDisposable` deinitializes in the dealloc routine of `base`, so `base` would be nil in this case. `unowned` would cause an issue.

ReactiveCocoa/UIKit/UITextView.swift
andersio: This would no longer be necessary after https://github.com/ReactiveCocoa/ReactiveSwift/pull/64 is landed.

ReactiveCocoa/NSObject+Association.swift
andersio: Is it necessary to extract it, when `key` is an immutable owner of the pointer and `utf8Start` does not have a non-trivial cost behind its getter?

andersio:
@NachoSoto I have gone through your comments, and addressed some of those.

Would be great if someone can take over and convert the tests, perhaps in a separate PR?

ReactiveCocoa/UIKit/UIControl.swift
andersio: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210/files/e48c43694c212939997f9c382c1886cd99d846e6#r83744869

andersio:
Test fails because `UISwitch`, `UITextField` and `UITextView` themselves leaks in Xcode 8.0 GM. Should we simply disable them?

ReactiveCocoa/AppKit/NSTextField.swift
NachoSoto: What do you mean by indirection?

ReactiveCocoa/UIKit/UITextView.swift
NachoSoto: No, that won't be `public`.

NachoSoto: Did you mean ReactiveCocoa/ReactiveSwift#66? That won't be `public` for very important reasons.

andersio: With https://github.com/ReactiveCocoa/ReactiveSwift/pull/64, this would source directly from a signal.

ReactiveCocoa/AppKit/NSTextField.swift
andersio: Mapping twice means creating an extra signal that incurs extra locking and dispatch costs.

ReactiveCocoaTests/KeyValueObservingSpec.swift
NachoSoto: Bump.

ReactiveCocoaTests/UIKit/UIButtonSpec.swift
```diff
@@ -0,0 +1,61 @@
+import Quick
+import Nimble
+import ReactiveSwift
+import ReactiveCocoa
+import UIKit
+import enum Result.NoError
+
+class UIButtonSpec: QuickSpec {
```
NachoSoto: Thanks for converting these! so much better :)

ReactiveCocoa/UIKit/UITextView.swift
NachoSoto: Ooooh I see :) makes sense.

andersio:
@NachoSoto Here is a reduced test case:
https://github.com/andersio/TestUIKitWeakRef/blob/master/TestUIKitWeakRefTests/TestUIKitWeakRefTests.swift

It passes on `UIDatePicker`, but fails on `UISwitch`.

liscio:
@andersio Not sure if you came across this already, but here's some insight into what's happening: https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/iOS/UISwitch%2BRx.swift#L22

liscio:
@andersio Also, the related/source SO question — https://github.com/ReactiveX/RxSwift/issues/842

andersio:
Eww. Should we set the lifetime as its superview (or when it is removed from the superview, whichever the earlier) then?

liscio:
@andersio I think that might be equally—if not, more—gross. Possibly even dangerous in cases where folks might be intentionally reusing views.

_For the meantime_ what are your thoughts on changing `isOnValues` to `isOnValues(until: Lifetime? = nil)`, however _for now_ we can make the `Lifetime` non-Optional and not specify a default value.

This would require the caller to specify a lifetime that works for them while this bug (/feature?) exists in the framework, and later on we can deprecate the function in favor of a property?

Alternatively, we leave it as it is, and put a similar "DANGER, DANGER!" comment in the interface in hopes that nobody hits this until it's resolved. (The fact that iOS 8.1 still has it given that the original Radar filing in August, means that it could be a long while still…)

**This should not hold up this PR from being merged, however. I'm going to give the tests a run-through shortly.**

andersio:
@liscio I think it is a reasonable, temporary workaround. At least it would break your bindings instead of holding them indefinitely, when you expect the control to deinitialize with its containing view. It might be gross, but at least it is safer than letting it be.

liscio:
@andersio but what happens if someone programmatically creates their `UISwitch` and adds it to a view after setting up its bindings? Or, worse, bindings are set up in `viewDidLoad`, and with the trait collections stuff, it's feasible that you'd have views dynamically being added/removed from their parents quite regularly.

liscio:
@andersio on an unrelated note, have you run the unit tests with the Thread Sanitizer enabled? There are a few issues reported in some of the tests. They might be harmless (test-only objects that lack mutexes to protect their values), but it's worth a double-check.

andersio:
@liscio Address Sanitizer and Thread Sanitiser are all turned off.

andersio:
I would prefer to just document the bug then, instead of bending the API to workaround a bug in the SDK.

andersio:
Opened issue #3251 to track this. The dealloc checks are disabled in 5e7534c.

mdiep:
Temporarily disabling the tests (with a comment linking to the radar) seems like the right way to go.

bitomule:
Any way I can help?

I've started moving from Swift 2 to 3 and ReactiveCocoa changes are huuuuge. I need "rac_signalForSelector" and this seems the place where it's implemented (trigger)

andersio:
`trigger(for:)` has been implemented, but the variant with function arguments is not. @sharplet said he is interested in doing it.

bitomule:
I'm trying to use rex branch (created a fork to add cocoapods spec) but can't make it work. In this branch I see a trigger(for: selector. Should that work?

I'm going to migrate to carthage (bye cocoapods) and give it a try because cocoapods doesn't pull some modules and branch doesn't work.

Isn't there any migration guide to Swift 3? I would love helping but don't have enough experience (I don't know what was removed, why...)

liscio:
@andersio While I agree that they should be kept off for CI checks, I think that it's wise to run the diagnostics & find/fix any errors that might pop up along the way. I'll try and gather up the specific issues I found into issues / PRs as I'm able.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -0,0 +1,37 @@
...
+	/// - returns:
+	///   A trigger signal.
+	public func trigger(for selector: Selector) -> Signal<(), NoError> {
+		objc_sync_enter(self)
+		defer { objc_sync_exit(self) }
+
+		let map = associatedValue { _ in NSMutableDictionary() }
+
+		let selectorName = String(describing: selector) as NSString
+		if let signal = map.object(forKey: selectorName) as! Signal<(), NoError>? {
```
andersio: It is for certain an optional of `Signal<(), NoError>`, and we do not expect it to be something else. Using `as?` would mean allowing types other than it to slip through.

https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210#discussion_r83721833

andersio:
@bitomule Can you be more specific of what didn't work? It should be accessible as `NSObject.reactive.trigger(for:)`

bitomule:
@andersio This branch doesn't have a podspec, added one but don't know how to include the "ReactiveCocoaUtils.Private" module. I don't know how to add it. Should it work with a UIViewController?

`myController.reactive.trigger`

for example.

andersio:
You need to tell CocoaPods the existence of the modulemap at `ReactiveCocoa/ReactiveCocoaPrivate/`.

andersio:
I managed to make the spec of `UITextField` and `UITextView` works. But `UISwitch` probably gonna stay as is.

andersio:
@bitomule Try to use the latest commit, and specify the module map as `ReactiveCocoa/module.modulemap` in your Podfile.

bitomule:
Thanks @andersio , finally made it work merging that commit, adding the module map to the spec and also adding h and m files as source. I don't like using so many alpha frameworks but there doesn't seem to be other option for swift 3.

If anyone needs what I've done is all in my fork rex branch: https://github.com/bitomule/ReactiveCocoa/tree/rex

Is there any way I can help moving this to an stable version? I can't understand most of the code but I would be happy to help.

bitomule:
Also, there's an issue with NSTextField and NSControl as both import AppKit and crash on iOS. I've fixed it just excluding those files (my app only works on iPhone) but there should be a better way to skip those files on iOS builds.

andersio:
@bitomule Again the PodSpec has to be adjusted to accomodate new changes. CocoaPods use their own build system via PodSpec and ignores the project.

bitomule:
@andersio yeah. I'm moving to carthage to avoid this kind of issues. Cocoapods is easier but too much custom things over projects. Thanks and great work ;)

bitomule:
Is there any extension like the good old "rac_gestureSignal"?

andersio:
Not in this PR, since it is just porting the important bits of the Rex codebase to RAS 1.0 and Swift 3. But you can always file feature requests or even pull requests.

If you wish you could gather a list of what's being missed from ReactiveObjC. :)

mdiep:
🎉

liscio:
Yay! 🎆 🎉

PR #3235, opened 2016-10-05
Allow any type to be used with DynamicProperty

sharplet:
Since [SE-0116](https://github.com/apple/swift-evolution/blob/master/proposals/0116-id-as-any.md), any Swift type can be bridged to Objective-C, even if it can't be represented in Objective-C. This change removes all explicit use of `_ObjectiveCBridgeable` in favour of the bridging semantics implemented in Swift 3.

### DIFF
--- a/ReactiveCocoa/DynamicProperty.swift
+++ b/ReactiveCocoa/DynamicProperty.swift
@@ -2,14 +2,6 @@ import Foundation
 import ReactiveSwift
 import enum Result.NoError
 
-/// Models types that can be represented in Objective-C (i.e., reference
-/// types, including generic types when boxed via `AnyObject`).
-private protocol ObjectiveCRepresentable {
-	associatedtype Value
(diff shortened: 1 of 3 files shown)


NachoSoto:
This is a big win! You just killed a lot of complexity 👍

ReactiveCocoa/DynamicProperty.swift
```diff
@@ -60,85 +49,20 @@ public final class DynamicProperty<Value>: MutablePropertyProtocol {
 	}()
 
 	/// Initializes a property that will observe and set the given key path of
-	/// the given object, using the supplied representation.
+	/// the given object, where `Value` is a reference type that can be
+	/// represented directly in Objective-C via `AnyObject`.
```
sharplet: Oops, these docs are a bit out of date now

sharplet: #3236

PR #3249, opened 2016-10-17
DynamicProperty: don't expose signal setter

NachoSoto:


andersio:
The `lazy` could be stripped anyway. We won't leak by creating signals through it anymore.

NachoSoto:
@andersio it has to be `lazy` because its initialization relies on `self`. I did remove the `unowned` though which is useless.

andersio:
@NachoSoto What I mean is it can be simply a `producer.startAndRetrieveSignal()` now.

NachoSoto:
> @NachoSoto It can be simply a producer.startAndRetrieveSignal() now.

That's in `ReactiveSwift` and it's not exposed **[for very important reasons](https://github.com/ReactiveCocoa/ReactiveSwift/blob/67bfbd4c186f54477ea8dc228c5f5c7993ac7e9d/Sources/SignalProducer.swift#L336-L340)**.

PR #3219, opened 2016-09-26
[WIP] Add official podspec

ikesyo:
See also https://github.com/ReactiveCocoa/ReactiveSwift/pull/14.

PR #3252, opened 2016-10-19
Use custom module map for the framework + Add podspec (#3219)

andersio:
Merged in also @ikesyo's #3219 (official podspec).

-
This is a change for #3219, since CocoaPods doesn't seem to support `SWIFT_INCLUDE_DIR`. This approach also seems to hide the private headers without problem, too.


andersio:
Seems running fine. ;)
https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210#issuecomment-255038660

dylanmaryk:
Exactly what I'm looking for! Once someone merges this, can "master" please be merged into "add-podspec" so those of us using CocoaPods can utilise this change?

ikesyo:
Thanks @andersio, I've slightly updated the podspec in this branch.

ReactiveCocoa/module.modulemap
```diff
@@ -0,0 +1,7 @@
+framework module ReactiveCocoa {
```
NachoSoto: :+1:

Issue #3264, opened 2016-10-25
Subtle bug with NSObject.reactive.trigger(for:) in relation to URLSession


chrrasmussen:
# Problem description

I recently stumbled upon a subtle bug with `NSObject.reactive.trigger(for:)` (at least that is what it seems). I have an application that downloads some data from a JSON API using `URLSession.reactive.data(with:)`. It will download new data when navigating around in the application. After a while I noticed that the application stopped downloading more data, i.e. the signal producer from `URLSession.reactive.data(with:)` did not send "next" events anymore.

When the problem occurs, URLSession stops responding (does not call the callback function) to calls like `URLSession.dataTask()`, `URLSession.getAllTasks()`, etc. This also explains why `URLSession.reactive.data(with:)` does not send the "next" event.

After some digging I found out that the code started working again if I removed a call to `NSObject.reactive.trigger(for:)` (which was called in the same runloop iteration as `URLSession.reactive.data(with:)`). This function call seems irrelevant at first but somehow it affects the URLSession.

Other information that may be relevant:

- The problem occurs on both iOS 9 and iOS 10
- The problem occurs on both the simulator and the device
- The problem happens in normal usage, not only in tight loops (as seen in the example)
- I tried running all requests on the main thread but still got the problem

I havn't looked much into what can cause this behavior, as I opted for rewriting the code in question.

# Sample code

The minimal code needed to reproduce the problem is shown below. I have created an Xcode 8.1 project including ReactiveCococa 5.0.0-alpha.2 (using CococaPods, but the files are included in the repository for convenience): https://github.com/chrrasmussen/ReactiveCocoa-Trigger

```swift
import UIKit
import ReactiveSwift
import ReactiveCocoa

class ViewController: UIViewController {

    override func viewDidLoad() {
        super.viewDidLoad()

        // Do any additional setup after loading the view.
        let config = URLSessionConfiguration.default
        let urlSession = URLSession(configuration: config)

        let request = URLRequest(url: Bundle.main.url(forResource: "Test", withExtension: "txt")!)

        for i in 1...200 {
            print("Requesting: #\(i)")
            let task = urlSession.reactive.data(with: request)

            task.startWithCompleted {
                print("Completed: #\(i)")
            }

            // This is the problematic code. Remove comment from line 26 to introduce bug.
            let viewController = UIViewController()
//            _ = viewController.reactive.trigger(for: #selector(UIViewController.viewDidAppear(_:)))
[...]

PR #3265, opened 2016-10-25
Fixed a bug which caused a deadlock in `trigger(for:)`.

andersio:
... a programmer mistake. Kinda side effect of `id` as `Any`.

### DIFF
--- a/ReactiveCocoa/NSObject+Synchronize.swift
+++ b/ReactiveCocoa/NSObject+Synchronize.swift
@@ -1,0 +1,7 @@
+extension NSObject {
+	internal func synchronize<Result>(execute: () throws -> Result) rethrows -> Result {
+		objc_sync_enter(self)
+		defer { objc_sync_exit(self) }
+		return try execute()
+	}
+}
--- a/ReactiveCocoa/NSObject+Synchronize.swift
+++ b/ReactiveCocoa/NSObject+Synchronize.swift
@@ -1,5 +1,5 @@
 extension NSObject {
-	internal func synchronize<Result>(execute: () throws -> Result) rethrows -> Result {
+	internal func synchronized<Result>(execute: () throws -> Result) rethrows -> Result {
 		objc_sync_enter(self)
 		defer { objc_sync_exit(self) }
 		return try execute()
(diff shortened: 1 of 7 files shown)


sharplet:
Nice.

Do you think it would be worth adding a private/internal `synchronized` helper for `Reactive`, to help avoid this happening with future extensions? Something like

```swift
extension Reactive {
  func synchronized(body: () throws -> Void) rethrows {
    objc_sync_enter(base)
    defer { objc_sync_exit(base) }
    try body()
  }
}
```

Then use `synchronized { }` wherever we currently use `objc_sync_*` directly.

sharplet:
This helper would probably need to be able to return a value, too.

andersio:
BTW, I attempted to write test cases to reproduce #3264 but failed. 🙈

andersio:
@sharplet I constrained it against `NSObject` directly, since it very likely won't be our public API.

sharplet:
I think it would be better to use `synchronized` instead of `synchronize`, in keeping with the `@synchronized` keyword in Onjectove-C.

ReactiveCocoa/NSObject+Synchronize.swift
```diff
@@ -0,0 +1,7 @@
+extension NSObject {
+	internal func synchronized<Result>(execute: () throws -> Result) rethrows -> Result {
```
NachoSoto: This can be `final`.

ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -71,6 +71,24 @@ class InterceptingSpec: QuickSpec {
 				expect(counter1) == 2
 				expect(counter2) == 2
 			}
+
+			it("should not deadlock") {
```
NachoSoto: Awesome

ReactiveCocoaTests/LifetimeSpec.swift
```diff
@@ -0,0 +1,43 @@
...
+			}
+
+			afterEach {
+				object = nil
+				expect(_object).to(beNil())
+			}
+
+			it("should not deadlock") {
+				for _ in 1 ... 10 {
+					var isDeadlock = true
```
NachoSoto: I'd say `isDeadlocked`.

PR #3267, opened 2016-10-29
Add cross reference notes to UIKit text observation methods

sharplet:
Hopefully helps users to be more aware of the difference between the different ways of observing text changes.

### DIFF
--- a/ReactiveCocoa/UIKit/UITextField.swift
+++ b/ReactiveCocoa/UIKit/UITextField.swift
@@ -9,12 +9,17 @@ extension Reactive where Base: UITextField {
 	}
 
 	/// A signal of text values emitted by the text field upon end of editing.
+	///
+	/// - note: To observe text values that change on all editing events,
+	///   see `continuousTextValues`.
 	public var textValues: Signal<String?, NoError> {
 		return trigger(for: .editingDidEnd)
 			.map { [unowned base = self.base] in base.text }
 	}
 
 	/// A signal of text values emitted by the text field upon any changes.
+	///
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/UIKit/UITextField.swift
```diff
@@ -9,12 +9,17 @@ extension Reactive where Base: UITextField {
 	}
 
 	/// A signal of text values emitted by the text field upon end of editing.
+	///
+	/// - note: To observe text values that change on all editing events,
+	///   see `continuousTextValues`.
```
andersio: Align the start of the lines please. 😸

ReactiveCocoa/UIKit/UITextView.swift
```diff
@@ -9,6 +9,9 @@ extension Reactive where Base: UITextView {
 	}
 
 	/// A signal of text values emitted by the text view upon end of editing.
+	///
+	/// - note: To observe text values that change on all editing events,
+	///   see `continuousTextValues`.
```
andersio: Ditto.

ReactiveCocoa/UIKit/UITextField.swift
```diff
@@ -9,12 +9,17 @@ extension Reactive where Base: UITextField {
 	}
 
 	/// A signal of text values emitted by the text field upon end of editing.
+	///
+	/// - note: To observe text values that change on all editing events,
```
andersio: Perhaps just _text values being emitted upon any changes_? IIRC it does not emit values upon end of editing, for example.

Issue #3269, opened 2016-10-30
RACSignalForSelectorAliasPrefix should be renamed


inamiy:
There are two exact `static NSString * const RACSignalForSelectorAliasPrefix = @"rac_alias_"` in both `ReactiveCocoa` and `ReactiveObjC`.

When **`self.reactive.trigger(for:)` and `self.rac_signal(for: selector)` are both called**, swizzling of `forwardInvocation:` takes place and **`ReactiveObjC` tries to extract `RACSubject`** from associated object (and then calls `sendNext:`), but **`ReactiveCocoa` sets a closure** (in `_rac_setupInvocationObservationForSelector:`) instead, and this causes the following error:

```
-[__NSMallocBlock__ sendNext:]: unrecognized selector sent to instance
```

I think renaming one prefix in either framework will solve this issue.

PR #3270, opened 2016-10-31
Changed the method prefix and class suffix used for swizzling.

andersio:
Fixes https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3269.


sharplet:
Do you think it would be worth having an integration test for this that pulls in ReactiveObjC?

andersio:
Perhaps ReactiveObjCBridge would be a better place, hmm?

sharplet:
Yeah, I suppose? It is about integration by definition. We'd have to change it to pull in ReactiveCocoa instead of just ReactiveSwift.

mdiep:
I don't think an integration test is necessary here. It's a pretty easy, non-breaking change if this ever breaks again.

PR #3276, opened 2016-11-02
Make reactive `backgroundColor` for `UIView`

eimantas:


mdiep:
This is great, thanks!

PR #3280, opened 2016-11-04
Adds the compatibility with libraries such as Aspects, jsPatch or wax…

doggy:
…Patch whom hook functions with forwardInvocation:.

### DIFF
--- a/ReactiveCocoa/NSObject+RACSelectorSignal.m
+++ b/ReactiveCocoa/NSObject+RACSelectorSignal.m
@@ -58,13 +58,6 @@ static BOOL RACForwardInvocation(id self, NSInvocation *invocation) {
 
 static void RACSwizzleForwardInvocation(Class class) {
 	SEL forwardInvocationSEL = @selector(forwardInvocation:);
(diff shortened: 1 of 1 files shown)


doggy:
In common case, it is efficiently if we preserve the implementation of forwardInvocation: at calling rac_signalForSelector: and use it when necessary.

But it should be updated if it has been replaced after calling rac_signalForSelector:.
Holding and invoke the old implementation will lead to an unexpected situation. (infinite loops, or crashes)

doggy:
Library such as Aspects, jsPatch or waxPatch will swizzle forwardInvocation: at any time during the app running, probably.

P.S. This is an issue reported at [jsPatch's issue list](https://github.com/bang590/JSPatch/issues/292)
And I decided make this PR to RAC after reviewed all related codes between RAC and jsPatch.

andersio:
We have moved the ObjC framework to [ReactiveObjC](https://github.com/ReactiveCocoa/ReactiveObjC). Would you mind to reopen it there?

doggy:
acknowledge

PR #3284, opened 2016-11-05
Add binding targets for reloadData

sharplet:
Works for `UITableView`, `UICollectionView`, `NSTableView` and `NSCollectionView`.


andersio:
Is there a strong reason to wrap it? It seems to be in the same situation as `rex_dismissAnimated`.

sharplet:
Remind of the context of `rex_dismissAnimated`?

The main reason for me is memory management. With regards to differentiating it from something like dismissal, reloading/refreshing data happens frequently and feels like a much more naturally "reactive" action. Feels kind of like an indirect form of data binding, where the table/collection pulls the new data, instead of you pushing it into the view.

andersio:
https://github.com/ReactiveCocoa/ReactiveSwift/pull/30#issuecomment-249360530

mdiep:
Looks like the AppKit files are mistakenly part of the tvOS target though.

andersio:
@mdiep I thought you'd think it should be an `Action`.

mdiep:
I do still wonder if it should be an `Action`.

sharplet:
What advantages would we get making it an action over the simpler `BindingTarget`?

mdiep:
None, necessarily.

Before `BindingTarget` was a thing, I'd argued that we could probably use `Action`. It's a little weird to me that we have two executable concepts. But I think `BindingTarget` is actually fine here, which is why I approved above.

PR #3285, opened 2016-11-06
Trigger signal for optional, unbacked protocol requirements.

andersio:
Extend `trigger(for:)` to provide an equivalent of `rac_signalForSelector:fromProtocol:`.

### DIFF
--- a/ReactiveCocoaTests/InterceptingSpec.swift
+++ b/ReactiveCocoaTests/InterceptingSpec.swift
@@ -39,6 +39,24 @@ class InterceptingSpec: QuickSpec {
 				expect(object.counter) == 2
 			}
 
+			it("should send a value when the selector is invoked without implementation") {
+				let selector = #selector(TestProtocol.optionalMethod)
+
+				let signal = object.reactive.trigger(for: selector,
+				                                     from: TestProtocol.self)
+
+				var counter = 0
+				signal.observeValues { counter += 1 }
+
+				expect(counter) == 0
+
+				(object as TestProtocol).optionalMethod!()
+				expect(counter) == 1
+
+				(object as TestProtocol).optionalMethod!()
+				expect(counter) == 2
+			}
+
 			it("should complete when the object deinitializes") {
 				let signal = object.reactive.trigger(for: #selector(object.increment))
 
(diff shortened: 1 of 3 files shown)


ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -93,8 +113,13 @@ class InterceptingSpec: QuickSpec {
...
 }
 
-class InterceptedObject: NSObject {
+@objc protocol TestProtocol {
+	@objc optional func optionalMethod()
+}
+
+class InterceptedObject: NSObject, TestProtocol {
 	var counter = 0
+	var testProtocolCounter = 0
```
NachoSoto: This isn't used?

ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -39,6 +39,26 @@ class InterceptingSpec: QuickSpec {
...
+
+				let signal = object.reactive.trigger(for: selector,
+				                                     from: TestProtocol.self)
+				expect(object.responds(to: selector)) == true
+
+				var counter = 0
+				signal.observeValues { counter += 1 }
+
+				expect(counter) == 0
+
```
NachoSoto: I'm so confused, shouldn't this crash because the type of `object` doesn't implement the method? 😳

andersio: Not if you specify a protocol. Then in `NSObject._rac_setupInvocationObservationForSelector`, it would look if the selector exists in the protocol and create an implementation for the selector (`_objc_msgForward `) correspondingly.

andersio: It does not crash because RAC adds an implementation for it at the time you swizzle it via ` trigger(for:from:)` with the protocol specified.

If you do not swizzle it, it crashes.

mdiep:
I'm not sure this is worth doing.

If you have this it won't work:

```swift
let selector = #selector(TestProtocol.optionalMethod)
let signal = object.reactive.trigger(for: selector)
```

You need to remember to add `from: TestProtocol.self`. But that doesn't seem any easier than adding `func optionalMethod() { }` to `object`'s class. 😕 

RAC 2.x had this before, so we don't have to implement it. But we could remove it and simplify this code.

Is there a benefit I'm not seeing?

andersio:
I guess the point is exactly not needing to implement an empty method. Though we can also achieve the same goal by providing proper delegate interception.

PR #3286, opened 2016-11-07
Selector interception with arguments.

andersio:
~~Includes #3285.~~

~~Test coverage is limited.~~

This PR adds `NSObject.reactive.signal(for:from:)`, which returns a signal that emits a collection of bridged arguments `[Any?]` after every invocation of the selector.

It relies on the [the `NSNumber` and `NSValue` bridging](https://github.com/apple/swift-evolution/blob/master/proposals/0139-bridge-nsnumber-and-nsvalue.md) introduced in Swift 3.0.1 to bridge values.

- [x] ~~NSInvocation & NSMethodSignature wrapper~~
- [x] A working prototype
- [x] Bridge numbers as `NSNumber` instead of Swift built-in types.
- [x] Bridge structs and other unsupported arguments as `NSValue`.
- [x] Full test coverage on the numeric types and reference types.
- [x] Ported suitable test cases from ReactiveObjC.


andersio:
Oops, Travis isn't on Swift 3.0.1 yet.

andersio:
Would it be better to expose an API that emits tuple instead?

e.g.
```
func signal<A, B, C>(
    for selector: Selector, 
    from protocol: Protocol? = nil,
    type: (A, B, C).Type? = nil
) -> Signal<(A, B, C), NoError>
```

It is still unsafe since we cannot constraint the selector, but the consumption would be safer than `[Any?]`. The only downside is like `combineLatest`, we would have a bunch of overloads.

andersio:
Had thought about emitting tuple. But IMO it doesn't add much in terms of type safety, since we cannot constraint against `#selector`.

andersio:
The disabled test cases would be reenabled when https://github.com/ReactiveCocoa/ReactiveObjC/pull/33 is ported back.

PR #3288, opened 2016-11-10
Make some MKMapView's properties reactive

eimantas:
Extension on tvOS available only from version 9.2.

Following properties are reactive based on availability:
- `mapType` - on all platforms;
- `isZoomEnabled` - on all platforms;
- `isScrollEnabled` - on all platforms;
- `isPitchEnabled` - disabled on tvOS;
- `isRotateEnabled` - disabled on tvOS;

### DIFF
--- a/ReactiveCocoa/Shared/MKMapView.swift
+++ b/ReactiveCocoa/Shared/MKMapView.swift
@@ -1,0 +1,12 @@
+import ReactiveSwift
+import MapKit
+
+@available(tvOS 9.2, *)
+extension Reactive where Base: MKMapView {
+
+	/// Sets the map type.
+	public var mapType: BindingTarget<MKMapType> {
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/Shared/MKMapView.swift
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	/// Sets if scroll is enabled for map.
+	public var isScrollEnabled: BindingTarget<Bool> {
+		return makeBindingTarget { $0.isScrollEnabled = $1 }
+	}
+
+	#if !os(tvOS)
+	/// Sets if pitch is enabled for map.
+	@available(tvOS, unavailable)
```
eimantas: Yep, they didn't even worked. Removed!

Issue #3263, opened 2016-10-24
Controls as binding targets and signals


andersio:
The majority of controls consumes only one single type of values. This makes me wonder if we might conform these controls to `BindingTargetProtocol` or even `SignalProtocol` to reduce boilerplate for common usage.

Not saying that we should ditch binding targets in `.reactive`, but we might allow specific subclasses to expose a "default" binding target and signal.

```
titleLabel.reactive.text <~ viewModel.title
viewModel.title <~ titleLabel.reactive.textValues
```

versus

```
titleLabel <~ viewModel.title
viewModel.title <~ titleLabel
```

liscio:
I'd say this is probably not necessary, however if it doesn't interfere with the ability to expose >1 value (i.e. I can do `someTextControl <~ someSignal` in addition to `someTextControl.attributedText <~ someSignal`) then I'm indifferent.

andersio:
Yeah, options are still provided via `.reactive`, just that the controls may declare one as the default.

mdiep:
How about adding overloads of `<~` for this?

```swift
// simplified because I'm writing in the browser 🙃
public func <~ (lhs: UILabel, rhs: Signal<String>) {
    lhs.reactive.text <~ rhs
}
public func <~ (lhs: UILabel, rhs: Signal<NSAttributedString>) {
    lhs.reactive.attributedText <~ rhs
}
```

That would let us set `text` and `attributedText` without extending `UILabel`.

It's also kinda need that you could update your view model type and it would just do the right thing.

sharplet:
I think RAC is the first release to really make `<~` shine, so it's reasonable to expect that its usage will go up. So I'm a little concerned about getting too eager with the operator overloads, because it could degrade compile time performance pretty quickly.

Some anecdotal experience to back that: on my current project we're using an Auto Layout wrapper that overloads a bunch of standard operators for specifying constraints in code. So I've felt what it's like to have methods that take 5 or 10 seconds to compile (writing all constraints in code tends to result in long methods) — and I'm a little shy about it! 😬 

Granted, I haven't done the maths on this, so we might be nowhere near that critical mass yet.

mdiep:
Measuring the compile-time impact sounds like a great idea. 👍

andersio:
We can define them by extending the type, instead of introducing global overloads. I am not sure if the type checker improvements are already in place, but that's the direction.

[...]

PR #3289, opened 2016-11-11
Default binding source and target providers.

andersio:
#3263 

Allow types to be used directly in a unidirectional binding, by declaring a default binding target and a default signal.

```
myLabel <~ viewModel.title
viewModel.searchString <~ myTextField
```

--

It seems bad in terms of discoverability & local reasoning though. Makes me wonder if an operator would be a better alternative, e.g. `myLabel%.text`

### DIFF
--- a/ReactiveCocoa/AppKit/NSControl.swift
+++ b/ReactiveCocoa/AppKit/NSControl.swift
@@ -109,3 +109,77 @@ extension Reactive where Base: NSControl {
 		}
 	}
 }
+
+extension NSControl {
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+	) -> Disposable? where Source.Value == String, Source.Error == NoError {
+		return target.reactive.stringValue <~ source
+	}
+
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+	) -> Disposable? where Source.Value == NSAttributedString, Source.Error == NoError {
+		return target.reactive.attributedStringValue <~ source
+	}
+
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+	) -> Disposable? where Source.Value == Bool, Source.Error == NoError {
+		return target.reactive.boolValue <~ source
+	}
+
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+	) -> Disposable? where Source.Value == Int32, Source.Error == NoError {
+		return target.reactive.intValue <~ source
+	}
+
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+	) -> Disposable? where Source.Value == Int, Source.Error == NoError {
+		return target.reactive.integerValue <~ source
+	}
+
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+	) -> Disposable? where Source.Value == Double, Source.Error == NoError {
+		return target.reactive.doubleValue <~ source
+	}
+
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+	) -> Disposable? where Source.Value == Float, Source.Error == NoError {
+		return target.reactive.floatValue <~ source
+	}
+
+	@discardableResult
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSControl,
+		source: Source
+		) -> Disposable? where Source.Value == Any, Source.Error == NoError {
+		return target.reactive.objectValue <~ source
+	}
+
+	@discardableResult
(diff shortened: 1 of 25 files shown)


mdiep:
Do we know for sure that adding a operator per class would be worse compile time performance-wise? That seems like it'd have some benefits (such as the ability to add operators for `UILabel` for both `NSString` and `NSAttributedString`).

andersio:
@mdiep I don't think it would worsen the performance much, unless we define them as top-level operator.

mdiep:
What if we did it without a protocol?

```swift
extension UILabel {
    @discardableResult
    public static func <~ <Source: SignalProtocol>(label: UILabel, signal: Source) -> Disposable? where Source.Value == String, Source.Error == NoError {
        return label.reactive.text <~ signal
    }

    @discardableResult
    public static func <~ <Source: SignalProtocol>(label: UILabel, signal: Source) -> Disposable? where Source.Value == NSAttributedString, Source.Error == NoError {
        return label.reactive.attributedText <~ signal
    }
}
```

andersio:
Yeah, but then we will need to define also the producer and the property variants, and the optional ones.

andersio:
`gyb` isn't part of the build phase, since it has to be regenerated only when the config is changed.

mdiep:
This will need an update now that changes have been made to ReactiveSwift. But I don't think this is essential for 5.0 since it's an additive API.

andersio:
Yeah, I am waiting on the ReactiveSwift alpha/RC.

ReactiveCocoa/UIKit/iOS/BindingTargetSubstituting.generated.swift
```diff
@@ -0,0 +1,194 @@
...
+extension NSLayoutConstraint {
+
+	// constant: CGFloat
+
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: NSLayoutConstraint,
+		source: Source
+	) -> Disposable? where Source.Value == CGFloat, Source.Error == NoError {
+		return target.reactive.constant <~ source
+	}
```
mdiep: This one feels weird to me. I don't think it's obvious that it'd be targeting the content. With something that's essentially a value, defaults make sense. But I don't think this fits in that category.

ReactiveCocoa/UIKit/iOS/BindingTargetSubstituting.generated.swift
```diff
@@ -0,0 +1,194 @@
...
+
+	// text: String
+
+	public static func <~ <Source: BindingSourceProtocol>(
+		target: UITextField,
+		source: Source
+	) -> Disposable? where Source.Value == String, Source.Error == NoError {
+		return target.reactive.text <~ source
+	}
+
```
mdiep: This could use `NSAttributedString` variants.

ReactiveCocoa/UIKit/tvOS/BindingTargetSubstituting.generated.swift
```diff
@@ -0,0 +1,160 @@
+// This is auto-generated by gyb.
+
+import Foundation
+import ReactiveSwift
+import enum Result.NoError
+import UIKit
+
+extension UIActivityIndicatorView {
```
mdiep: I don't think we should duplicate shared operators. They should be in a shared source file.

script/gyb/gyb
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env python2.7
+import gyb
+gyb.main()
```
mdiep: I still don't like including `gyb`. It's no more work to copy-and-paste additions than it is to add them to the gyb files—and it's simpler and requires less knowledge.

mdiep:
This looks good to me, but I'd like to get some more opinions on whether this is a good idea.

@ReactiveCocoa/reactivecocoa What do you think of adding these shortcuts?

sharplet:
It's borderline, but I think I'm slightly 👎 on this.

The main reason I think is that it slightly increases the convenience of writing this bindings, but detracts from their readability and understandability (as mentioned in previous comments). More broadly though, I think RAC 5 already marks a big leap forward in terms of convenience when writing UI code. There are already a bunch of things that will feel new to users, and the small gains of this PR don't seem worth the potential disadvantages to me, _at this point in time_. I think this kind of change would make more sense after some time, when the community has a more solid grasp of the new stuff.

liscio:
I am indifferent, but I'll keep tooting my "teach the users" horn 🎺 and say that it's probably not going to help with self-documenting code. When bindings are completely missing from controls (which is going to continue being a problem as new API is added) then it will not be immediately obvious to seek out the `.reactive` member to see what bindings are available.

Also? It will be very common to mix the default binding with another binding in very close proximity in code. For instance:

``` swift
someTextField <~ someStringSignal
someTextField.reactive.enabled <~ someBooleanSignal
```

So it's not even buying much convenience for a great deal of users.

Now, with that said, I agree with @sharplet that it might be good in the future. For example, such a "default binding" system would work well with my (long-neglected, but still somewhat interesting) stuff over at https://github.com/liscio/ReactiveCocoaSample where a special `BindingValue` type can be used to wrap up the control's state (e.g. `.multipleValues`, `.emptySelection`) along with the value itself.

mdiep:
Let's sit on this for now and revisit for RAC 6 or 7.

PR #3291, opened 2016-11-12
Make `UISlider` great again

eimantas:
- Make `value` reactive;
- Make `minimumValue` reactive;
- Make `maximumValue` reactive;

### DIFF
--- a/ReactiveCocoaTests/UIKit/UISliderSpec.swift
+++ b/ReactiveCocoaTests/UIKit/UISliderSpec.swift
@@ -52,5 +52,17 @@ class UISliderSpec: QuickSpec {
 			observer.send(value: 0.7)
 			expect(slider.maximumValue) ≈ 0.7
 		}
+
+		it("should emit user's changes for its value") {
+			slider.value = 0.25
+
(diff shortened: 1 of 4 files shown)


andersio:
What about the signal?

eimantas:
Should the signal be called `values` or `continuousValues`?

andersio:
I'd name it `values`, which fits well in both cases, if we aren't interested in providing both variants.

ReactiveCocoaTests/UIKit/UISliderSpec.swift
```diff
@@ -0,0 +1,68 @@
...
+
+		it("should emit user's changes for its value") {
+			slider.value = 0.25
+
+			var updatedValue: Float?
+			slider.reactive.values.observeValues { value in
+				updatedValue = value
+			}
+
+			slider.sendActions(for: .valueChanged)
```
andersio: May you assert `updatedValue` to be `nil` before triggering the action?

altyus:
😢  but only because of the PR name

andersio:
🙁

PR #3299, opened 2016-11-16
Cleanup specs

andersio:


ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -76,13 +76,25 @@ class InterceptingSpec: QuickSpec {
...
+					} else {
 						DispatchQueue.global(priority: .high).async {
 							_ = object.reactive.trigger(for: #selector(object.increment))
 
-							isDeadlocked = false
+							DispatchQueue.global(priority: .high).async {
+								_ = object.reactive.trigger(for: #selector(object.increment))
+
+								isDeadlocked = false
+							}
```
mdiep: These two things seem mostly the same. It looks like only the queue is different. So we should extract that and reuse everything else.

```swift
let queue: DispatchQueue
if #available(*, macOS 10.10 ) {
    queue = ...
} else {
    queue = ...
}
queue.async { ... }
```

ReactiveCocoaTests/LifetimeSpec.swift
```diff
@@ -25,13 +25,25 @@ class LifetimeSpec: QuickSpec {
...
+					} else {
 						DispatchQueue.global(priority: .high).async {
 							_ = object.reactive.lifetime
 
-							isDeadlocked = false
+							DispatchQueue.global(priority: .high).async {
+								_ = object.reactive.lifetime
+
+								isDeadlocked = false
+							}
```
mdiep: Same comment here about the dispatch queue.

ReactiveCocoaTests/KeyValueObservingSpec.swift
```diff
@@ -299,37 +299,55 @@ class KeyValueObservingSpec: QuickSpec {
...
+						}
 					}
 
-					let event = replayProducer.last()
-					expect(event).toNot(beNil())
+					testObject = nil
+					replayProducer = nil
+					iterationQueue.resume()
+
+					iterationQueue.sync(flags: .barrier, execute: {})
```
mdiep: I'm confused about this change. What's the intent?

It seems to be more code and be more complex.

andersio: The original test case wasn't "async disposal" after all, since `dispatch_apply` is synchronous, causing the disposal happening after the setting of the value.

Issue #3300, opened 2016-11-16
`trigger(for:)` KVO interoperability bug


andersio:
ReactiveCocoa and ReactiveObjC do not correctly swizzle an instance if it had already been swizzled by KVO. Any invocation would raise an unrecognised selector exception.

```
// Swizzled by KVO.
object.reactive.values(forKeyPath: #keyPath(object.value)).start()

// Swizzled by RAC.
_ = object.reactive.trigger(for: #selector(setter: object.value))

// Unrecognized selector exception is raised.
object.value = 1
```

[Failing test case](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3286/commits/fe2f27c7207f398b474e6f4cefe2dea2060db1fe#diff-f1f2377061825177a6887d9ace987d1cR150
)

Expected to be fixed when ~~https://github.com/ReactiveCocoa/ReactiveObjC/pull/33 is back-ported~~ #3302 is merged.

mdiep:
Fixed by #3302.

PR #3302, opened 2016-11-17
ObjC object interception enhancements.

andersio:
Includes #3286.

The ObjC object interception logic behind `trigger(for:)` and the newly introduced `signal(for:)` was rewritten to be a Swift citizen (`NSObject+Intercepting.swift`).

The implementation was based on https://github.com/ReactiveCocoa/ReactiveObjC/pull/33, but the ability to dynamically implement methods (rejected in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3285) was stripped off.

`signal(for:)` relies on the [the `NSNumber` and `NSValue` bridging](https://github.com/apple/swift-evolution/blob/master/proposals/0139-bridge-nsnumber-and-nsvalue.md) introduced in Swift 3.0.1 to bridge values. In Swift 3.0, it still works provided that the arguments are casted as Cocoa container types.

- [x] Rewrite the ObjC object interception logic in Swift.
- [x] Add `NSObject.reactive.signal(for:)` as part of the public API.
- [x] Bridge numbers as `NSNumber` instead of Swift built-in types.
- [x] Bridge structs and other unsupported arguments as `NSValue`.
- [x] Full test coverage on the numeric types and reference types.
- [x] Ported suitable test cases from ReactiveObjC.


andersio:
Implementation note:

| Optimization | Time for 50000 messages |
| ------ | ------- |
| No Interception | Baseline |
| [Unoptimized](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/commits/a39a59809464ec658016946d019351653a1ee7e6) | >200x slower |
| [Dumping selector aliases](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/commits/0b88ed17d5605bc08d795eeb6e1bac6e5fdbf28e) | ~125x slower |
| [Method signature caching + Value type boxing](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/commits/43e732913eafb3bb21e33b187941a285d4070136) | ~70x slower |

andersio:
It is ready for review. 😸 

andersio:
The predefined template branch with gyb in action:
https://github.com/ReactiveCocoa/ReactiveCocoa/tree/predefined-impls

I have tried to build a custom forwarder for amd64, and the 1.4x speedup isn't as great as I expected. Predefined templates give a 1.91x speedup, albeit it not being (quite) a generic solution.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,436 @@ extension Reactive where Base: NSObject {
...
+// Runtime subclasses:
+// - `&AssociationKey.intercepted` (ObjC+RuntimeSubclassing.swift)
+//   Nullable. Exists only in runtime subclasses generated by external parties.
+//   `true` indicates the runtime subclass has already been swizzled.
+//
+// - `&AssociationKey.intercepted` (This file)
+//   Nullable. `true` indicates the runtime subclass has already been prepared for method
+//   interception.
+//
+// - `&AssociationKey.signatureCache`
```
mdiep: These should go on the properties themselves. (Then they'll also be viewable in Xcode directly at the usage site.)

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,436 @@ extension Reactive where Base: NSObject {
...
+/// Setup the method interception.
+///
+/// - parameters:
+///   - object: The object to be intercepted.
+///   - selector: The selector of the method to be intercepted.
+///
+/// - returns:
+///   A signal that sends the corresponding `NSInvocation` after every
+///   invocation of the method.
+private func setupInterception(_ object: NSObject, for selector: Selector) -> Signal<AnyObject, NoError> {
```
mdiep: How about making this a method on `NSObject`?

```swift
extension NSObject {
    private func intercept(_ selector: Selector) -> Signal<AnyObject, NoError>
}
```

ReactiveCocoa/NSObject+Association.swift
```diff
@@ -12,11 +23,35 @@ extension Reactive where Base: NSObject {
...
+
+extension NSObject {
+	/// Retrieve the associated value for the specified key.
+	///
+	/// - parameters:
+	///   - key: The key.
+	///
+	/// - returns:
+	///   The associated value, or `nil` if no value is associated with the key.
+	@nonobjc internal func value(forAssociationKey key: UnsafeRawPointer) -> Any? {
```
mdiep: I think this would be better named `associatedObject(forKey:)`.

andersio:
@mdiep 

1. The ability to implement methods dynamically is removed.

2. `newForwardInvocation` now invokes the superclass' `forwardInvocation` dynamically, instead of caching them upon creation. (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/files#diff-34b169256c54cb8dd90c09513687db9aR190

3. `newForwardInvocation` would now update the alias to reflect the latest impl. of the superclass. (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/files#diff-34b169256c54cb8dd90c09513687db9aR171)

4. For interoperability, if we swizzle an external runtime subclass (e.g. KVO), we would cache any immediate impl. of the selector we detected upon creation. (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/files#diff-34b169256c54cb8dd90c09513687db9aR100). `newForwardInvocation` would invoke it instead of the superclass impl. (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/files#diff-34b169256c54cb8dd90c09513687db9aR153) This fixes #3300.

5. A method signature cache is added to speed up the invocation forwarding chain. (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/files#diff-34b169256c54cb8dd90c09513687db9aR215)

6. A selector cache is added to speed up the retrieval of selector aliases. (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302/files#diff-34b169256c54cb8dd90c09513687db9aR141)

ReactiveCocoa/NSObject+Association.swift
```diff
@@ -1,5 +1,29 @@
 import ReactiveSwift
 
+internal struct AssociationKey {
```
mdiep: These seem mostly unrelated, so I think it actually makes sense to separate these into different `fileprivate` keys.

ReactiveCocoa/NSObject+Association.swift
```diff
@@ -12,11 +36,35 @@ extension Reactive where Base: NSObject {
...
+
+extension NSObject {
+	/// Retrieve the associated value for the specified key.
+	///
+	/// - parameters:
+	///   - key: The key.
+	///
+	/// - returns:
+	///   The associated value, or `nil` if no value is associated with the key.
+	@nonobjc internal func associatedValue(forKey key: UnsafeRawPointer) -> Any? {
```
mdiep: Would it make sense for this to be generic over the value type?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+	/// Setup the method interception.
+	///
+	/// - parameters:
+	///   - object: The object to be intercepted.
+	///   - selector: The selector of the method to be intercepted.
+	///
+	/// - returns:
+	///   A signal that sends the corresponding `NSInvocation` after every
+	///   invocation of the method.
+	@nonobjc fileprivate func setupInterception(for selector: Selector) -> Signal<AnyObject, NoError> {
```
mdiep: What do you think of the name `intercept(_ selector: Selector)`?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
-			let selectorName = String(describing: selector) as NSString
-			if let signal = map.object(forKey: selectorName) as! Signal<(), NoError>? {
-				return signal
+				if !isSwizzled {
+					objc_setAssociatedObject(subclass, AssociationKey.signatureCache, signatureCache, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+					objc_setAssociatedObject(subclass, AssociationKey.selectorCache, selectorCache, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+					objc_setAssociatedObject(subclass, AssociationKey.intercepted, true, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+
+					enableMessageForwarding(subclass, selectorCache)
+					setupMethodSignatureCaching(subclass, signatureCache)
```
mdiep: Would it make sense to move these lines into the `else` above?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+				// preserve the implementation.
+				//
+				// Example: KVO setters if the instance is swizzled by KVO before RAC
+				//          does.
+				if !class_respondsToSelector(subclass, interopAlias) {
+					let immediateMethod = class_getImmediateMethod(subclass, selector)
+
+					let immediateImpl: IMP? = immediateMethod.flatMap {
+						let immediateImpl = method_getImplementation($0)
+						return immediateImpl.flatMap { $0 != _rac_objc_msgForward ? $0 : nil }
```
mdiep: I think you could move this inner `flatMap` out a level?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+			// forwarder afterwards.
+			//
+			// However, the IMP cache would be thrashed due to the swapping.
 
-			lifetime.ended.observeCompleted(observer.sendCompleted)
-			map.setObject(signal, forKey: selectorName)
+			let interopImpl = class_getMethodImplementation(realClass, interopAlias)
+			let previousImpl = class_replaceMethod(realClass, selector, interopImpl, typeEncoding)
+			invocation.invoke()
+			_ = class_replaceMethod(realClass, selector, previousImpl, typeEncoding)
```
mdiep: I wonder if it'd make sense to wrap some of the objc runtime APIs in a more Swift-friendly package internally to make the actual logic here more readable. 🤔

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+	private var map: [Selector: (main: Selector, interop: Selector)] = [:]
+
+	init() {}
+
+	/// Cache the aliases of the specified selector in the cache.
+	///
+	/// - warning: Any invocation of this method must be synchronized against the
+	///            runtime subclass.
+	@discardableResult
+	func allocate(_ selector: Selector) -> (main: Selector, interop: Selector) {
```
mdiep: `allocate` seems like a really strange choice here. Maybe `cache` would work better?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+		if let pair = map[selector] {
+			return pair
+		}
+
+		var copy = map
+		let aliases = (selector.alias, selector.interopAlias)
+		copy[selector] = aliases
+		map = copy
```
mdiep: Why is this copying instead of `map[selector] = aliases`?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+
+		return aliases
+	}
+
+	/// Get the alias of the specified selector.
+	///
+	/// - parameters:
+	///   - selector: The selector alias.
+	subscript(main selector: Selector) -> Selector {
```
mdiep: I don't love the use of labeled subscripts for these. I'd personally rather use a method or have a single subscript that returns a `(main: Selector, alias: Selector)`.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+	///   - selector: The method signature.
+	subscript(selector: Selector) -> AnyObject? {
+		get {
+			return map[selector]
+		}
+		set {
+			if map[selector] == nil {
+				var newMap = map
+				newMap[selector] = newValue
+				map = newMap
```
mdiep: Same here: why explicitly copy?

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -6,30 +6,419 @@ extension Reactive where Base: NSObject {
...
+		case .double:
+			bridged.append(NSNumber(value: extract(CDouble.self)))
+		case .bool:
+			bridged.append(NSNumber(value: extract(CBool.self)))
+		case .object:
+			bridged.append(extract((AnyObject?).self))
+		case .type:
+			bridged.append(extract((AnyClass?).self))
+		case .selector:
+			bridged.append(extract((Selector?).self))
```
mdiep: I'd move the `bridge.append` out of the `switch` so that the logic isn't repeated for every case. Then the compiler can also enforce that we set append one value in every case.

```swift
let value: Any?
switch ... {
    case .selector:
        value = extract((Selector?).self)
    ...
}
bridged.append(value)
```

ReactiveCocoa/NSObject+Lifetime.swift
```diff
@@ -20,66 +17,57 @@ extension Reactive where Base: NSObject {
...
-					// Call the existing implementation if one has been caught. Otherwise,
-					// call the one first available in the superclass hierarchy.
-					if let existingImpl = existingImpl {
-						impl = existingImpl
-					} else {
-						let superclass: AnyClass = class_getSuperclass(objcClass)
-						impl = class_getMethodImplementation(superclass, deallocSelector)
+			synchronized(objcClass) {
+				// Swizzle the class only if it has not been swizzled before.
+				if objc_getAssociatedObject(objcClass, AssociationKey.lifetime) == nil {
```
mdiep: Seems like we could use custom methods on `AnyClass` too to have nicer APIs for this.

ReactiveCocoa/NSObject+Intercepting.swift
mdiep: Or is there an existing OSS framework that we could use? (Maybe that's a question for a separate PR—either before or after this one.)

ReactiveCocoa/NSObject+Association.swift
andersio: A central location is easier to manage with all those offsets.

ReactiveCocoa/NSObject+Intercepting.swift
andersio: When I started the PR, I did wrap them in `ObjCClass`, `ObjCMethod` and `CFunction`. But then generally it is one-to-one mapping, and doesn't seem making a huge difference to me.

😐

ReactiveCocoa/NSObject+Intercepting.swift
andersio: Hmm, yeah. The copying is unnecessary in this case.

ReactiveCocoa/NSObject+Lifetime.swift
andersio: `AnyClass` and `AnyObject` cannot be extended.

ReactiveCocoa/NSObject+Association.swift
andersio: It makes no difference in the call site. The type parameter still needs to be specified.

ReactiveCocoa/NSObject+Association.swift
mdiep: Why do they need to be a contiguous block?

ReactiveCocoa/NSObject+Association.swift
mdiep: My thought was that it would isolate the force casting to this method. The call site would still need to declare the type, but `let foo: String = bar.associatedValue(forKey: blah)` might be nicer.

I don't feel strongly about this.

andersio: It has to be `Value?` though. `as!` fails in `any as! U` when `U` is an optional and `any` is nil of type `Any?`.

ReactiveCocoa/NSObject+Association.swift
andersio: Just being a nicer citizen to the memory allocator, instead of having six scattered bytes that fragments the address space.

ReactiveCocoa/NSObject+Association.swift
mdiep: Oh, right. I should have made it optional. But that still seems like it might be nicer?

ReactiveCocoa/NSObject+Intercepting.swift
andersio: Done. (This didn't get identified as outdated diff.)

ReactiveCocoa/NSObject+Association.swift
mdiep: I don't think that optimization is worth it in that case. Unless there's big benefit to performance, we should optimize of clarity.

ReactiveCocoa/NSObject+Lifetime.swift
mdiep: TIL!

ReactiveCocoa/NSObject+Association.swift
```diff
@@ -1,5 +1,28 @@
...
+	// Byte usage.
+	//
+	// +0: intercepted
+	// +1: signatureCache
+	// +2: selectorCache
+	// +3: runtimeSubclassed
+	// +4: lifetime
+	// +5: lifetimeToken
+	// +6: knownRuntimeSubclass
+	internal static let contiguous = UnsafeMutablePointer<UInt8>.allocate(capacity: 7)
```
mdiep: We should get rid of this and just let the `AssociationKey` allocate a single `UInt8`—without worrying about them being contiguous. I think this method is error prone and an unnecessary optimization. I was concerned about this before, but didn't mention it cause I thought we should split them up, but now it's especially true.

ReactiveCocoa/NSObject+Association.swift
```diff
@@ -12,11 +35,122 @@ extension Reactive where Base: NSObject {
...
+	}
+}
+
+/// Retrieve the associated value for the specified key.
+///
+/// - parameters:
+///   - key: The key.
+///
+/// - returns:
+///   The associated value, or `nil` if no value is associated with the key.
```
mdiep: This returns the default, not `nil`.

ReactiveCocoa/NSObject+Association.swift
```diff
@@ -12,11 +35,122 @@ extension Reactive where Base: NSObject {
...
+internal func value<Value>(forKey key: AssociationKey<Value>, in object: AnyObject) -> Value {
```
mdiep: Do we need to have both this and the method? That seems confusing.

ReactiveCocoa/NSObject+Lifetime.swift
```diff
@@ -1,14 +1,22 @@
...
-private var lifetimeTokenKey: UInt8 = 0
+extension Association {
+	/// Holds the `Lifetime` of the object.
+	fileprivate static let isSwizzled = AssociationKey<Bool>(contiguous + 4, default: false)
+
+	/// Holds the `Lifetime` of the object.
+	fileprivate static let lifetime = AssociationKey<Lifetime?>(contiguous + 4, default: nil)
+
+	/// Holds the `Lifetime.Token` of the object.
+	fileprivate static let lifetimeToken = AssociationKey<Lifetime.Token?>(contiguous + 5, default: nil)
```
mdiep: I'd put these on `AssociationKey` instead. Then we should be able to use a shortcut too: `base.value(forKey: .lifetime)`

andersio: We can't have stored static property in generic types. So it wouldn't be a huge difference. It still needs to be backed by a top level constant.

PR #3305, opened 2016-11-20
Sandboxes for macOS and iOS.

andersio:


mdiep:
Can the iOS playground be added to CI as well?

andersio:
@mdiep Tried my best, but got stuck at https://travis-ci.org/ReactiveCocoa/ReactiveCocoa/jobs/178011306#L632.

mdiep:
In that case, we can live without the CI. 👍

PR #3309, opened 2016-11-22
`values(forKeyPath:)` overload for NSManagedObject.

andersio:
```
 +	/// Create a producer which sends the current value and all the subsequent
 +	/// changes of the key path, but ignore `nil`s emitted due to the managed
 +	/// object being turned into a fault, or being deleted.
 +	///
 +	/// The producer completes when the object deinitializes.
 +	///
 +	/// - note: The resulting producer can be safety transformed as a strong-typed
 +	///         non-optional producer. To opt out of the filtering, coerce the
 +	///         object as `NSObject`.
```
This is [what I have used for quite a while](https://github.com/andersio/MantleData/blob/master/MantleData/ManagedObject%2BExtension.swift#L39) to make KVO with `NSManagedObject` behave more like normal objects.

It filters the `nil`s emitted due to the state transition of `NSManagedObject`. A prominent use is rebinding `Value` of the producer to a non-optional type safety, e.g. `SignalProducer<Int, NoError>` for a non-optional Int64 attribute.

I cannot assure that it got all the corner cases covered, but so far it has been fine for me.

Note that `ReactiveProtocol` here is a (temporary?) workaround, since somehow overloads with tighter constraints in concrete type extension do not override the looser ones like their friends in protocol extensions.

Bug Report: https://bugs.swift.org/browse/SR-3257

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -148,11 +148,19 @@
 		9A6AAA2B1DB8F85C0013AAEA /* ReusableComponentsSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6AAA271DB8F7EB0013AAEA /* ReusableComponentsSpec.swift */; };
 		9A6AAA2E1DB903A20013AAEA /* ReusableComponentsSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6AAA291DB8F7F10013AAEA /* ReusableComponentsSpec.swift */; };
 		9A6AAA2F1DB903A40013AAEA /* ReusableComponentsSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6AAA291DB8F7F10013AAEA /* ReusableComponentsSpec.swift */; };
+		9A92AC201E064F2500D16628 /* CoreData.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9AA140951DE9F2E000C5E04F /* CoreData.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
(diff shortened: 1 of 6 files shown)


sharplet:
Can we weakly link CoreData.framework, to make the dependency optional for consumers?

mdiep:
I'm not wild about including this in ReactiveCocoa. It seems like it might be a better fit for a separate framework—especially since it's natural to want reactive ways to fetch, etc. as well.

Mostly I'm concerned about taking on something here that we can't support and don't want future additions to. I don't use (and don't like) Core Data, so it would be very hard for me to review changes. But I don't know how everyone else feels.

andersio:
IMO Core Data is part of Cocoa, after all, and not quite an unpopular framework.

One can use Core Data directly, or use RAC infused high-level wrapper that evolves outside RAC. But this is a rather essential one (and perhaps the only one needed in RAC) to support all kinda of Core Data use to RAC/RAS primitives without hassle.

liscio:
Part of me wants to suggest we break out certain framework wrappers (GameKit, SpriteKit, CoreData, AVFoundation, etc), and part of me wants to just keep "the apple stuff" to a single repo. But once you get into the weak linking/etc, things get really crappy really quickly (like everyone getting looped into the Photos privacy issue w/ iOS app review, for instance.)

What about making separate, optional .framework targets? I suspect that makes things difficult for Carthage, since it expects a 1:1 repo:deliverable ratio. But it would certainly be a good compromise for letting folks pick & choose the bits they require in a larger workspace setup.

If we *did* make separate repos for the separate pieces, we'd have to open a lot of RAC's `internal` methods up to make extensions easier. Also there's the annoying management of releases/etc.

Kind of a crappy situation to deal with, no matter what we choose. *Not* offering support at all for the other frameworks is definitely not the right answer, either…

mdiep:
> What about making separate, optional .framework targets? I suspect that makes things difficult for Carthage, since it expects a 1:1 repo:deliverable ratio.

That'd totally be fine, as far as Carthage is concerned. The only caveat is that you can't choose to only build some of the frameworks.

I don't have much experience with weak linking. Are there other drawbacks, besides the privacy string issue?

PR #3313, opened 2016-11-25
Fix possible DynamicProperty crash when incorrect value is passed

olejnjak:
I've been wondering why `DynamicProperty<Value>` uses producer and signal of type `Value?` and `signal`. I've been digging into its implementation and I assume the force cast to `Value` useless as it should return an optional value.

I thought whether its appropriate to return optional value and IMO it is correct as KVO compliant values will always be of type `Any`. But in case that incorrect value is passed I think that as a developer I want to know about it, but in production code I'd prefer the app not to crash. That's the reason for the assertion.

### DIFF
--- a/ReactiveCocoa/DynamicProperty.swift
+++ b/ReactiveCocoa/DynamicProperty.swift
@@ -37,7 +37,13 @@ public final class DynamicProperty<Value>: MutablePropertyProtocol {
 	///              Most UI controls are not!
 	public var producer: SignalProducer<Value?, NoError> {
 		return (object.map { $0.reactive.values(forKeyPath: keyPath) } ?? .empty)
-			.map { $0 as! Value }
+			.map {
+				guard let value = $0 as? Value else {
+					assertionFailure("Expected value to be \(Value.self), but \(type(of: $0)) found.")
+					return nil
+				}
+				return value
+		}
 	}
 
 	public private(set) lazy var signal: Signal<Value?, NoError> = {


andersio:
The crash Is likely because somehow Swift 3.0.1 does not allow casting a `nil` in `Any?` to a generic `U` type when `U` is an optional.

```
func test<U>(_ any: Any?) -> U {
	return any as! U
}

let a: Int = test(1)
let b: Int? = test(nil) // Crash
```

As for why `DynamicProperty` uses `Value?`, it is because the object is referenced weakly.

ReactiveCocoa/DynamicProperty.swift
```diff
@@ -37,7 +37,13 @@ public final class DynamicProperty<Value>: MutablePropertyProtocol {
 	///              Most UI controls are not!
 	public var producer: SignalProducer<Value?, NoError> {
 		return (object.map { $0.reactive.values(forKeyPath: keyPath) } ?? .empty)
-			.map { $0 as! Value }
```
sharplet: The subtle thing here is there's a difference between `PropertyProtocol.Value` (associated type) and `DynamicProperty<Value>` (generic type parameter). It's really unclear, but `Value` in the context of this dynamic cast is actually `Value?`, whereas the same name in the context of the type declaration is `DynamicProperty.Value` (non-optional).

If it was possible to namespace the type names you'd have something like `PropertyProtocol.Value == DynamicProperty.Value?`.

I believe the only way to cause this to actually crash would be to pass an object of an unexpected type into `setValue(_:forKey:)`, or if an object of unexpected type was assigned from Objective-C.

The type parameter naming could certainly be improved.

sharplet:
If you're getting an unexpected crash in your code, please feel free to open an issue (or PR with a failing test case). Closing this for now.

ReactiveCocoa/DynamicProperty.swift
andersio: @sharplet There is an existing case though. https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3313#issuecomment-262998365

andersio:
By the way, while it is true that KVO compliant values in Swift are of type `Any` (with the possibility of being `NSNull`). Our KVO implementation does not use the KVO change dictionary.

PR #3319, opened 2016-11-29
Keyboard Frame Notification

andersio:
Not quite sure about the name and where to place it though. `UIView.Type` seems a better place than `NotificationCenter`, since this is tied to the default notification center.

```swift
public struct KeyboardChangeContext {
	public var beginFrame: CGRect { get }
	public var endFrame: CGRect { get }
	public var animationCurve: UIViewAnimationCurve { get }
	public var animationDuration: Double { get }
	@available(iOS 9.0, *)
	public var isLocal: Bool { get }
}

extension Reactive where Base: UIView {
	public static var keyboardChange: Signal<KeyboardChangeContext, NoError>
}
```


ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
```diff
@@ -0,0 +1,40 @@
+import UIKit
+import ReactiveSwift
+import enum Result.NoError
+
+public struct KeyboardChangeContext {
```
NachoSoto: Awesome. Can you add a quick docstring?

ReactiveCocoaTests/UIKit/UIKeyboardSpec.swift
```diff
@@ -0,0 +1,37 @@
...
+import ReactiveCocoa
+import Quick
+import Nimble
+
+class UIKeyboardSpec: QuickSpec {
+	override func spec() {
+		describe("UIView.Type.reactive.keyboardChange") {
+			it("should emit a `value` event when the notification is posted") {
+				var context: KeyboardChangeContext?
+
```
NachoSoto: No need to give the parameter a name if it's going to be abbreviated :P just use `$0`?

ReactiveCocoaTests/UIKit/UIKeyboardSpec.swift
```diff
@@ -0,0 +1,37 @@
...
+				}
+
+				let dummyInfo: [AnyHashable: Any] = [
+					UIKeyboardFrameBeginUserInfoKey: CGRect(x: 10, y: 10, width: 10, height: 10),
+					UIKeyboardFrameEndUserInfoKey: CGRect(x: 20, y: 20, width: 20, height: 20),
+					UIKeyboardAnimationDurationUserInfoKey: 1.0,
+					UIKeyboardAnimationCurveUserInfoKey: UIViewAnimationCurve.easeInOut
+				]
+
+				NotificationCenter.default.post(name: .UIKeyboardWillChangeFrame,
```
NachoSoto: Using `NotificationCenter.default` could potentially affect other tests. Can you make the signal take a `NotificationCenter` (defaulting to this) so the test can use its own one?

calebd:
Personally I don't think `UIView` is the appropriate place for this. There are plenty of notifications that are only posted in the default notification center but that doesn't mean they should be chained off of random classes in UIKit or Foundation.

andersio:
Hmm, what about the `NotificationCenter` type?

ReactiveCocoaTests/UIKit/UIKeyboardSpec.swift
andersio: It would then have to be exposed as an extension to `NotificationCenter` instances, I'd think.

PR #3320, opened 2016-11-29
Add ReactiveCocoa specific documentation

iv-mexx:
This document explains the additions that ReactiveCocoa adds to ReactiveSwift.

Right now, it points to the documentation in ReactiveSwift, so I think some duplicate documents (Framework Overview, Basic Operators) could be deleted here.

On the other hand, it might make sense to keep them to have a complete Documentation in ReactiveCocoa as well...


mdiep:
This is great stuff. 👏 

> I think some duplicate documents (Framework Overview, Basic Operators) could be deleted here.

I think that makes sense. 👍

Issue #3315, opened 2016-11-26
CocoaAction may not observe changes to its underlying action on the main thread


sharplet:
CocoaAction documents that its properties will only change [on the main thread](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/CocoaAction.swift#L15-L16). However it ultimately just [makes a reference](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/CocoaAction.swift#L40-L41) to the underlying action's properties:

```swift
isEnabled = action.isEnabled
isExecuting = action.isExecuting
```

Looks like we should be observing these properties on `UIScheduler`.

PR #3321, opened 2016-11-29
Add reactive refresh control.

calebd:
- [x] Add `isRefreshing` binding
- [x] Add `refresh` action
- [x] Add `attributedTitle` binding
- [x] Bind action `isExecuting` to `isRefreshing`

### DIFF
--- a/ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
+++ b/ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
@@ -25,8 +25,8 @@ extension Reactive where Base: UIRefreshControl {
 		}
 
 		nonmutating set {
-			let disposable = action.map { isRefreshing <~ $0.isExecuting }
-			setAction(newValue, for: .touchUpInside, disposable: disposable)
+			let disposable = newValue.map { isRefreshing <~ $0.isExecuting }
+			setAction(newValue, for: .valueChanged, disposable: disposable)
 		}
 	}
 }
--- a/ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
+++ b/ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
@@ -14,7 +14,7 @@ extension Reactive where Base: UIRefreshControl {
 	}
 
 	/// The action to be triggered when the refresh control is refreshed. It
-	/// also controls the enabled state of the refresh control.
+	/// also controls the enabled and refreshing states of the refresh control.
 	public var refreshed: CocoaAction<Base>? {
 		get {
 			return associatedAction.withValue { info in
@@ -25,7 +25,26 @@ extension Reactive where Base: UIRefreshControl {
 		}
 
 		nonmutating set {
-			setAction(newValue, for: .valueChanged)
+			associatedAction.modify { associatedAction in
+				associatedAction?.disposable.dispose()
+
+				let controlEvents = UIControlEvents.valueChanged
+
+				if let action = newValue {
+					base.addTarget(action, action: CocoaAction<Base>.selector, for: controlEvents)
+
+					let disposable = CompositeDisposable()
+					disposable += isEnabled <~ action.isEnabled
+					disposable += isRefreshing <~ action.isExecuting
+					disposable += { [weak base = self.base] in
+						base?.removeTarget(action, action: CocoaAction<Base>.selector, for: controlEvents)
+					}
+
+					associatedAction = (action, controlEvents, ScopedDisposable(disposable))
+				} else {
+					associatedAction = nil
+				}
+			}
 		}
 	}
 }
--- a/ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
+++ b/ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
@@ -25,26 +25,8 @@ extension Reactive where Base: UIRefreshControl {
 		}
 
 		nonmutating set {
-			associatedAction.modify { associatedAction in
-				associatedAction?.disposable.dispose()
-
-				let controlEvents = UIControlEvents.valueChanged
-
-				if let action = newValue {
-					base.addTarget(action, action: CocoaAction<Base>.selector, for: controlEvents)
-
-					let disposable = CompositeDisposable()
-					disposable += isEnabled <~ action.isEnabled
-					disposable += isRefreshing <~ action.isExecuting
-					disposable += { [weak base = self.base] in
-						base?.removeTarget(action, action: CocoaAction<Base>.selector, for: controlEvents)
-					}
-
-					associatedAction = (action, controlEvents, ScopedDisposable(disposable))
-				} else {
-					associatedAction = nil
-				}
-			}
+			let disposable = action.map { isRefreshing <~ $0.isExecuting }
+			setAction(newValue, for: .touchUpInside, disposable: disposable)
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
```diff
@@ -0,0 +1,20 @@
...
+		return makeBindingTarget { $1 ? $0.beginRefreshing() : $0.endRefreshing() }
+	}
+
+	/// Sets the attributed title of the refresh control.
+	public var attributedTitle: BindingTarget<NSAttributedString?> {
+		return makeBindingTarget { $0.attributedTitle = $1 }
+	}
+
+	/// A trigger that sends `next` when the user performs a refresh.
+	public var refresh: Signal<Void, NoError> {
```
andersio: That's said we didn't provide any `trigger(for:)` shorthands in the initial Rex port. I will let others decide whether we should keep that stance.

calebd: `refreshing` is too close to `isRefreshing`. Perhaps a `CocoaAction` would provide a more appropriate API?

calebd:
The test failure above is from `MKMapViewSpec.swift`. Unrelated to this pull request.

ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
andersio: Yeah. That would be a nice addition. By the way, does `valueChanged` cover both the starting and the ending? If yes, it might better be `isRefreshingValues` instead.

calebd: Nope it only sends `.valueChanged` when the user pulls down on the containing scroll view. The user wants to perform an action. I think I'll make it a `CocoaAction` instead.

andersio: Maybe we can consider adding `trigger` to buttons and this altogether after we are done with this.

ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
```diff
@@ -13,8 +13,19 @@ extension Reactive where Base: UIRefreshControl {
 		return makeBindingTarget { $0.attributedTitle = $1 }
 	}
 
-	/// A trigger that sends `next` when the user performs a refresh.
-	public var refresh: Signal<Void, NoError> {
-		return trigger(for: .valueChanged)
+	/// The action to be triggered when the refresh control is refreshed. It
+	/// also controls the enabled state of the refresh control.
+	public var refreshed: CocoaAction<Base>? {
```
andersio: Maybe `pulled`? It sounds better than `refreshing` and `refreshed`.

calebd: I think refreshed is an appropriate name given the name of the class. Yes the user is physically pulling but that's not semantically what this action represents.

calebd: I don't feel strongly about it. Happy with whatever name makes sense as long as the functionality is there.

andersio: It seems to clash with `isRefreshing` though. That's why I brought it up. Consider this: when your `refreshed` action is invoked, you observe a positive `isRefreshing`.

We are communicating the intent here. A little bike-shedding won't hurt.

calebd: The documentation calls these events "refresh operations" and nowhere is the notion of "pulling" mentioned. I would personally prefer to remain close to the language defined by the type, its functions, and its documentation.

andersio: That's a valid point. But for the moment, we also have stuff that are not named accordingly, e.g. `UIButton.pressed`, inherited from Rex. It seems we do need to figure out which way to go, hmm.

mdiep: I think we should call this `refresh`.

@andersio is right that the past tense isn't really accurate.

ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
```diff
@@ -0,0 +1,31 @@
...
+		get {
+			return associatedAction.withValue { info in
+				return info.flatMap { info in
+					return info.controlEvents == .valueChanged ? info.action : nil
+				}
+			}
+		}
+
+		nonmutating set {
+			setAction(newValue, for: .valueChanged)
```
mdiep: Should we bind `isRefreshing` to the `Action`'s `isExecuting`?

ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
calebd: It was originally called refresh. I think that is the best name for it but a change was requested.

I used past a past tense command name to match the past tense command names found on other types in ReactiveCocoa.

ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
calebd: Yep that's a good idea!

ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
mdiep: We're still working through how we should name all these. But I think your original name was good!

ReactiveCocoa/UIKit/iOS/UIRefreshControl.swift
calebd: @mdiep Is there a way to implement that without copying all of `Reactive<UIControl>.setAction(for:)`? None of the other `Reactive<UIControl>` extensions attempt to add extra bindings to their actions.

calebd:
I would love to find a better way to implement `isRefreshing <~ action.isExecuting` but I feel like augmenting the `Reactive<UIControl>` API surface is outside the scope of this pull request. Open to feedback!

mdiep:
This PR has conflicts that need to be resolved.



> I would love to find a better way to implement isRefreshing <~ action.isExecuting but I feel like augmenting the Reactive<UIControl> API surface is outside the scope of this pull request. Open to feedback!

It seems like the way to do that would be to add an optional disposable parameter to `setAction(_:for:)`.

calebd:
Resolving them now. I am ok with that API change. Any preferences on the name of the parameter?

mdiep:
> Any preferences on the name of the parameter?

`disposable:`?

calebd:
`disposable` gets shadowed by the local `let` 😕

mdiep:
I'd rename the `let` variable to `compositeDisposable`.

calebd:
> I'd rename the let variable to compositeDisposable.

...doing that now

ReactiveCocoaTests/UIKit/UIRefreshControlSpec.swift
```diff
@@ -0,0 +1,84 @@
...
+			expect(refreshControl.attributedTitle).to(beNil())
+
+			observer.send(value: string)
+			expect(refreshControl.attributedTitle) == string
+
+			observer.send(value: nil)
+			expect(refreshControl.attributedTitle).to(beNil())
+		}
+
+		it("should execute the `refreshed` action upon receiving a `valueChanged` action message.") {
```
andersio: `refreshed` -> `refresh`.

calebd:
Should these binding targets explicitly jump to the main thread? It doesn't seem like any of them are, and at least in this case, `isExecuting` could send events on a background thread.

calebd:
> Should these binding targets explicitly jump to the main thread?

Oh nvm I see it now.

andersio:
We'd solve it separately. https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3315

calebd:
In `Reactive<NSObject>`, the `BindingTarget` created in `makeBindingTarget` is created with a `UIScheduler` so the targets themselves are main-thread-safe even if the events delivered to it are on other threads.

PR #3324, opened 2016-12-01
Send the UIControl sender for each control event

sharplet:
Changes `Reactive<UIControl>.trigger(for:)` to emit the control itself for each value, and as it is no longer just a trigger signal, renames it `controlEvents(_:)`.

Streamlines the definition of the various value-changed convenience signals, and removes the need to capture an extra unowned reference.

### DIFF
--- a/ReactiveCocoa/UIKit/UIControl.swift
+++ b/ReactiveCocoa/UIKit/UIControl.swift
@@ -35,20 +35,20 @@ extension Reactive where Base: UIControl {
 		}
 	}
 
-	/// Create a signal which sends a `next` event for each of the specified
+	/// Create a signal which sends a `value` event for each of the specified
 	/// control events.
 	///
 	/// - parameters:
 	///   - controlEvents: The control event mask.
 	///
 	/// - returns:
-	///   A trigger signal.
-	public func trigger(for controlEvents: UIControlEvents) -> Signal<(), NoError> {
+	///   A signal that sends the control each time the control event occurs.
+	public func controlEvents(_ controlEvents: UIControlEvents) -> Signal<Base, NoError> {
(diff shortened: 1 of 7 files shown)


sharplet:
We don't necessarily need the rename fixit for `trigger(for:)`, but I thought it would be a nice convenience for users of the current alpha, and could be removed in the subsequent release (or 5.0).

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -62,6 +62,11 @@ extension Reactive where Base: UIControl {
 		}
 	}
 
+	@available(*, unavailable, renamed: "controlEvents(_:)")
```
andersio: `Deprecations+Removal.swift` please? 😸

sharplet: I went back and forth on that, and thought I'd just put it here because the file doesn't exist for RAC and this is temporary (to be removed by 5.0). But on the other hand I can see how creating that file makes it more visible than "burying" it here alongside the renamed method.

andersio:
I wonder what's your opinion on providing pretty shorthands though.

sharplet:
What do you mean? Like keeping `trigger(for:)` around but discarding the sender?

sharplet:
So for now I've created https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3325 so that we remember to delete the fixit before 5.0. But if anyone has any objections to that approach please let me know.

andersio:
Hmm, never mind. It is just something missed from the button extensions.

sharplet:
I think it would be worth making an equivalent change for `NSControl`, too.

PR #3335, opened 2016-12-09
Add reactive UISearchBar

dcaunt:
`UISearchBar` supports a number of properties (and events via its delegate) – these are probably the minimum required:

- [x] Add `text` binding
- [x] Add `text` values signal
- [x] Add continuous `text` values signal

This requires becoming the delegate of the search bar, so this should probably be documented somewhere.

### DIFF
--- a/ReactiveCocoa/UIKit/iOS/UISearchBar.swift
+++ b/ReactiveCocoa/UIKit/iOS/UISearchBar.swift
@@ -32,9 +32,9 @@ extension Reactive where Base: UISearchBar {
 
 	/// A signal of text values emitted by the search bar upon end of editing.
 	///
-	/// - important: Creating this `Signal` requires the reactive extension
-	///	  provider to become the `delegate` of the search bar. Setting your own 
-	///   `delegate` is not supported and will result in a runtime error.
+	/// - important: Creating this Signal will make the reactive extension
+	///   provider the delegate of the search bar. Setting your own delegate is
+	///   not supported and will result in a runtime error.
 	///
 	/// - note: To observe text values that change on all editing events,
 	///   see `continuousTextValues`.
@@ -45,9 +45,9 @@ extension Reactive where Base: UISearchBar {
 
 	/// A signal of text values emitted by the search bar upon any changes.
 	///
-	/// - important: Creating this `Signal` requires the reactive extension
-	///	  provider to become the `delegate` of the search bar. Setting your own
-	///   `delegate` is not supported and will result in a runtime error.
+	/// - important: Creating this Signal will make the reactive extension 
+	///   provider the delegate of the search bar. Setting your own delegate is 
+	///   not supported and will result in a runtime error.
 	///
 	/// - note: To observe text values only when editing ends, see `textValues`.
 	public var continuousTextValues: Signal<String?, NoError> {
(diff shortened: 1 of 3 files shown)


mdiep:
> This requires becoming the delegate of the search bar, so this should probably be documented somewhere.

Hmm… that's tough. We definitely need to decide on a consistent approach for this. At a minimum, it should probably be document in an `- important: ` block on each property that sets the delegate.

I'd love to do this in a way that preserves the ability to set another delegate. But it looks like that's not possible, so I guess documentation is the way to go.

ReactiveCocoa/UIKit/iOS/UISearchBar.swift
```diff
@@ -0,0 +1,48 @@
...
+
+	private var delegate: ReactiveUISearchBarDelegate {
+		if let delegate = objc_getAssociatedObject(base, &delegateKey) as? ReactiveUISearchBarDelegate {
+			return delegate
+		}
+
+		let delegate = ReactiveUISearchBarDelegate()
+		base.delegate = delegate
+		objc_setAssociatedObject(base, &delegateKey, delegate, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+		return delegate
```
mdiep: I don't think we need an associated object here. Instead I'd do this:

```swift
if let delegate = base.delegate as? ReactiveUISearchBarDelegate {
    return delegate
} else if let delegate = base.delegate {
    fatalError("Cannot use reactive values on UISearchBar with a custom delegate!)
} else {
    let delegate = ReactiveUISearchBarDelegate()
    base.delegate = delegate
    return delegate
}
```

andersio:
@mdiep 

> I'd love to do this in a way that preserves the ability to set another delegate. 

It is possible, just that to do it with lowest cost and less boilerplate it involves interacting with the ObjC runtime, and the precursor is #3302 which is jampacked (yet IMO indivisible).

It is fine to have an interim solution though.

ReactiveCocoa/UIKit/iOS/UISearchBar.swift
dcaunt: Makes sense, I've pushed another commit which adds the `fatalError`. But I think the associated object is required because the `UISearchBar` only holds a `weak` reference to its `delegate`.

ReactiveCocoa/UIKit/iOS/UISearchBar.swift
```diff
@@ -0,0 +1,58 @@
...
+	/// Sets the text of the search bar.
+	public var text: BindingTarget<String?> {
+		return makeBindingTarget { $0.text = $1 }
+	}
+
+	/// A signal of text values emitted by the search bar upon end of editing.
+	///
+	/// - important: Creating this `Signal` requires the reactive extension
+	///	  provider to become the `delegate` of the search bar. Setting your own 
+	///   `delegate` is not supported and will result in a runtime error.
```
mdiep: I think rewording this would make it a little clearer:

> Creating this `Signal` will make the reactive extension provider the `delegate` of the search bar. Setting your own `delegate` is not supported and will result in a runtime error.

andersio: The text could be aligned better too.

dcaunt: I'll update the comments

mdiep:
Thanks!

PR #3336, opened 2016-12-11
Remove duplicate documentation thats already in ReactiveSwift

iv-mexx:
### Duplicate Documentation

As discussed in #3320, this now removes documentation thats duplicated from ReactiveSwift and which is more ReactiveSwift specific than ReactiveCocoa anyway.

Removed are
* Framework Overview
* Basic Operators
* Design Guidelines

Originally I had intended to keep Design Guidelines here, but then realised its all about ReactiveSwift specifics and not about ReactiveCocoa as well... Anyway, I've updated it in c688036 and removed it in the following commit ea72899 so if you feel like we should keep this document here, I can just revert that one commit.

I've added references to these documents (with links to them in ReactiveSwift) to the Documentation/README.md so that its easier to find them.

### Debugging Techniques

I've updated this document, but slightly different that in https://github.com/ReactiveCocoa/ReactiveSwift/pull/114. For ReactiveCocoa, I've kept the UILabel as example. DynamicProperty is still removed since the problem that was discussed does not occur anymore.


mdiep:
Fantastic work. Thanks! 💖

PR #3342, opened 2016-12-16
Public `NSObject.makeBindingTarget`.

andersio:
It probably makes better sense to let users achieve what they need immediately with this handy shortcut, and contribute those back at a later point.

### DIFF
--- a/ReactiveCocoa/NSObject+BindingTarget.swift
+++ b/ReactiveCocoa/NSObject+BindingTarget.swift
@@ -13,7 +13,7 @@ extension Reactive where Base: NSObject {
 	///
 	/// - returns:
 	///   A binding target that holds no strong references to the object.
-	internal func makeBindingTarget<U>(_ key: StaticString = #function, action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
+	public func makeBindingTarget<U>(_ key: StaticString = #function, action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
 		return BindingTarget(on: UIScheduler(), lifetime: lifetime) { [weak base = self.base] value in
 			if let base = base {
 				action(base, value)
--- a/ReactiveCocoa/NSObject+BindingTarget.swift
+++ b/ReactiveCocoa/NSObject+BindingTarget.swift
@@ -6,15 +6,15 @@ extension Reactive where Base: NSObject {
 	/// references the object so that the supplied `action` is triggered only if
 	/// the object has not deinitialized.
 	///
-	/// - important: The binding target is bound to the main queue.
-	///
 	/// - parameters:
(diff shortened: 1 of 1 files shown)


ReactiveCocoa/NSObject+BindingTarget.swift
```diff
@@ -13,7 +13,7 @@ extension Reactive where Base: NSObject {
 	///
 	/// - returns:
 	///   A binding target that holds no strong references to the object.
-	internal func makeBindingTarget<U>(_ key: StaticString = #function, action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
+	public func makeBindingTarget<U>(_ action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
```
sharplet: What do you think about making the scheduler a parameter with a default value of `UIScheduler`, instead of hard-coding it?

andersio:
@sharplet `lifetime` is `NSObject` exclusive for the moment.

ReactiveCocoa/NSObject+BindingTarget.swift
```diff
@@ -6,15 +6,15 @@ extension Reactive where Base: NSObject {
...
 	/// - parameters:
+	///   - scheduler: An optional scheduler that the binding target uses. If it
+	///                is not specified, a UI scheduler would be used.
 	///   - action: The action to consume values from the bindings.
 	///
 	/// - returns:
 	///   A binding target that holds no strong references to the object.
-	public func makeBindingTarget<U>(_ action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
-		return BindingTarget(on: UIScheduler(), lifetime: lifetime) { [weak base = self.base] value in
+	public func makeBindingTarget<U>(on scheduler: SchedulerProtocol? = nil, _ action: @escaping (Base, U) -> Void) -> BindingTarget<U> {
```
sharplet: Can this be non-optional?

andersio: Thought we can use only literals. 🙈

mdiep:
I'd be inclined to leave this `internal`.

If you're adding binding targets for AppKit or UIKit, you should contribute them back here. If you're not, and you're writing Swift, then you probably shouldn't be using `NSObject` anyway?

andersio:
> If you're adding binding targets for AppKit or UIKit, you should contribute them back here.

What I thought of is the round-trip time of having a thing available in RAC though (or something we rejected to be in RAC).

akashivskyy:
Another argument for making this public is that developers often create or use custom third-party controls in the project and want to add reactive support for them in extensions.

In my opinion, RAC should provide a simple API to support and encourage that.

dcaunt:
Yes, I've been implementing this for some of my custom views already.

PR #3344, opened 2016-12-16
Update ReactiveSwift to 1.0.0-rc.1.

### LINKED ISSUE #3251: ReactiveCocoa 5.0 and UIKit leaks in the iOS 10.0 SDK
### What's it about?

Several components are leaking in the iOS SDK. There is no competent workaround viable at the framework level to this.
1. https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3210#issuecomment-254594506
2. https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/iOS/UISwitch%2BRx.swift#L22
3. https://github.com/ReactiveX/RxSwift/issues/842
### Known Leaking Components
1. UISwitch
#### Suspected
1. UITextField
2. UITextView
### Recommended Workaround

Whenever you use the reactive extensions in `Signal`s or `BindingTarget`s for these components, assume they would **never** deinitialize.

Limit the lifetime of your bindings using a third source, e.g. the superview, whenever:
1. the bindings must be teared down when the affected view hierarchy is removed; and
2. the source of the bindings would outlive the affected view hierarchy.
#### Example

This WOULD LEAK if the lifetime of your view model outlives view hierarchy, e.g. you reuse them.

```
mySwitch.isOn.isEnabled <~ viewModel.isEnabled
```

Therefore, the binding should be constrained with a lifetime from some others that are not affected by the issue. In the following snippet, it is the ancestor of `mySwitch` in the view hierarchy.

```
mySwitch.isOn.isEnabled <~ viewModel.isEnabled.take(during: reactive.lifetime)
```
### Note

This is an iOS specific issue. tvOS is not affected.

andersio:
Note: The iOS 10.2 SDK has fixed #3251.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode8.1
+osx_image: xcode8.2
 before_install: true
 install: true
 git:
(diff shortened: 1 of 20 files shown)


andersio:
https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3344/commits/d847c493b7665c8a026851278da274ca4f577aec uses the workaround in
https://github.com/ReactiveCocoa/ReactiveObjCBridge/pull/11/commits/1c48c759137aeaad8beda98d9fceb9dee74c8220 to the duplicated iOS 10.2 simulators.

PR #3345, opened 2016-12-16
Bound the properties of `CocoaAction` to the main thread.

### LINKED ISSUE #3315: CocoaAction may not observe changes to its underlying action on the main thread

andersio:
Fixes #3315.

### DIFF
--- a/ReactiveCocoa/CocoaAction.swift
+++ b/ReactiveCocoa/CocoaAction.swift
@@ -2,8 +2,8 @@ import Foundation
 import ReactiveSwift
 import enum Result.NoError
 
-/// CocoaAction wraps an Action for use by a GUI control (such as `NSControl` or
-/// `UIControl`), with KVO, or with Cocoa Bindings.
+/// CocoaAction wraps an Action for use by a UI control (such as `NSControl` or
+/// `UIControl`).
 public final class CocoaAction<Sender>: NSObject {
 	/// The selector for message senders.
 	public static var selector: Selector {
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/CocoaAction.swift
```diff
@@ -37,9 +35,11 @@ public final class CocoaAction<Sender>: NSObject {
 			producer.start()
 		}
 
-		isEnabled = action.isEnabled
-		isExecuting = action.isExecuting
-		
+		isEnabled = Property(initial: action.isEnabled.value,
+		                     then: action.isEnabled.producer.observe(on: UIScheduler()))
```
sharplet: Do you think it would make sense for `PropertyProtocol` to have a version of `observe(on:)`? I've been wondering if there's a potential race between reading `property.value` and starting `property.producer`.

andersio: It is safe here, as `property.producer` would always have the latest value overwritten the initial value should any race happen anyway. `PropertyProtocol.observe(on:)` makes sense too.

andersio: These can be replaced with https://github.com/ReactiveCocoa/ReactiveSwift/pull/157 once merged.

Issue #3310, opened 2016-11-23
Provide SignalProducers to observe component's values


dmcrodrigues:
I'm a big fan of the `Reactive` approach, great job with this @andersio, but I think we lost some flexibility along the way.

Let's take for example a `UITextField`.

```swift
// Rex (RAC4)
var rex_text: MutableProperty<String?>

// RAC5
var textValues: Signal<String?, NoError>
var continuousTextValues: Signal<String?, NoError>
```

Using Rex, and because it was using a `MutablePropery` to provide the binding, we have a flexibility to choose between using a `signal` if we are interested on following changes or a `producer` if we want the current value followed by any change over time. Depending on how you are using this bindings it can be quite useful to have both options.

An example can be something like a login button being enabled if we have a valid username and password or disabled otherwise.

```swift
loginButton.rex_enabled <~ 
    combineLatest(usernameTextField.rex_text.producer, passwordTextField.rex_text.producer) // I'm using `.producer` just to empathise this
        .map { $0?.characters.count ?? 0 > 0 && $1?.characters.count ?? 0 > 0 }
```

We can argue that probably we should be using an `Action` but I don't think everyone is using or is going to use them. This issue was recently reported here: https://github.com/ReactiveCocoa/ReactiveSwift/issues/99.

Another example could be using `UIDatePicker`'s `date` even when the user doesn't change the pre-selected date.

```swift
let dateProperty: MutableProperty<NSDate?> = NSMutableProperty(nil)
...
// Our property will contain always the date of the picker
dateProperty <~ datePicker.rex_date.producer
```

Using RAC5 we have signals to observe any changes which covers a lot of situations but we loose the ability to use the current value and following changes, if there are any, in an easy way. 

What I would like to propose is introduce a parallel `SignalProducer` that triggers both the current value of the control followed by any changes over time giving the ability for callers to choose which behaviour they really want without needing to find a workaround to restore the previous functionality.

andersio:
The argument that brought up by @liscio that were agreed on is that controls should not be the source of truth. Based on this principle, I am not sure how such producers would fit. One is supposed to have initialise the controls and established observations to them at the very beginning of the `UIControl`'s lifetime.

https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3229

RuiAAPeres:
I would agree with @andersio and @liscio. I see the behaviour/semantics of a UI Control as being hot. So I would say:

[...]

PR #3347, opened 2016-12-16
Will add a Property value for the isOn value

cristianames92:
This PR is related to https://github.com/ReactiveCocoa/ReactiveSwift/pull/160 and intends to make a simpler way of composing switches with the `On` values and the `Enable` State.

### DIFF
--- a/ReactiveCocoa/UIKit/iOS/UISwitch.swift
+++ b/ReactiveCocoa/UIKit/iOS/UISwitch.swift
@@ -12,4 +12,19 @@ extension Reactive where Base: UISwitch {
 	public var isOnValues: Signal<Bool, NoError> {
 		return controlEvents(.valueChanged).map { $0.isOn }
 	}
+	
+	/// A signal of the enable state in `Bool` emitted by the switch.
+	public var isEnableValues: Signal<Bool, NoError> {
+		return controlEvents(.valueChanged).map { $0.isEnabled }
+	}
+	
+	/// A property of on-off states in `Bool` emmited by the switch
+	var isOnProperty: Property<Bool> {
+		return Property(initial: base.isOn, then: self.isOnValues)
+	}
+	
+	/// A property of the enable state in `Bool` emmited by the switch
+	var isEnableProperty: Property<Bool> {
+		return Property(initial: base.isEnabled, then: self.isEnableValues)
+	}
 }


mdiep:
Thanks for the PR!

We've come to the conclusion that `Property` isn't an appropriate abstraction for reactive versions of UIKit properties. Since the underlying properties aren't KVO compatible and only send events for user changes (not programatic ones), the `Property.value` has very weird semantics.

Any UIKit properties should be exposed as a `Signal` and/or `BindingTarget`.

cristianames92:
Thanks @mdiep for the quick response, but what about the case when you try to make a composition using different Switches (In this case). You only have the `isOnValues` signal for this purpose but you will not have the very first value, so you will have to make either a producer out of it or a property. The other thing is the `isEnableValues` signal. Why this signal isn't part of the extension yet? In my case i have to make a composition using this state, something that with Rex i could do easily. I'm maybe missing some new structures for this purpose, so please tell me if you can about it.

andersio:
We are discouraging ambiguous layering and anti-patterns.

> You only have the isOnValues signal for this purpose but you will not have the very first value, so you will have to make either a producer out of it or a property.

The view itself should not be the source of truth, but the controller or the view model that bind with it.

> The other thing is the isEnableValues signal. Why this signal isn't part of the extension yet?

Whether a control is enabled or not is unanimously decided by the controller or the view model. User interactions would not affect `isEnabled` in any sense, so there is no point to have it (so as many other properties).

Related issue: #3310

cristianames92:
@andersio you are right about the responsibilities in this case of the UISwitch and the anti patterns. However i still think that even if the programmer is the one who controls the enabled state of the UISwitch, i don't think that it's a bad idea to have the property, for the cases when are combining the state of the switch with the enabled state of it. It's more clear to use this than mixing your inner state with the state of the switch for this purpose.

andersio:
> for the cases when are combining the state of the switch with the enabled state of it. It's more clear to use this than mixing your inner state with the state of the switch for this purpose.

The availability of the button is determined by your business logic, and propagates to your view. The state change of the button is originated from user interactions with your view, and propagates back to your business logic. They are in a reverse direction to to each other, and I do not see why you'd combine them together.



P.S. #3353 is related if you are interested.

cristianames92:
I see what you mean. Then i think that this pr is correctly closed.

Issue #3192, opened 2016-09-17
MutableProperty not found in RAC 5


hermannkuschke:
Hi guys

We just tried to update to RAC 5, but it seems like there is no MutableProperty to be found. All our uses of MutableProperty gives error messages. We went and checked online and it seems like the file Property.swift in https://github.com/ReactiveCocoa/ReactiveCocoa/tree/master/ReactiveCocoa/Swift cannot be found. Could you please inform me as quickly as possible as to what is going on?

Regards,

Hermann Kuschke

andersio:
`MutableProperty` is available in [ReactiveSwift](https://github.com/ReactiveCocoa/ReactiveSwift), which is now one of the dependency of ReactiveCocoa.

hermannkuschke:
Thank you for your quick reply.

andersio:
Please be noted that the Objective-C API, and the bridge between ReactiveObjC and ReactiveSwift would receive a similar deal (#3182). If your codebase depend on them, you might have to include them as separate dependencies.

justinvallely:
Hi guys,

I'm having the same issue as @hermannkuschke. Looks like ReactiveSwift is being installed with ReactiveCocoa via cocoapods. I'm still getting the build errors about MutableProperty not being found. Am I missing a step? Thank you.

hermannkuschke:
@justinvalley what we did was to not install them through cocoapods, but to instead include them as dependencies. We then manually ensured that ReactiveSwift's syntax conformed to that of Swift 3. This is just temporary though, but it works.

andersio:
The podspec would include ReactiveSwift for you. But in your code, you would need to explicitly import ReactiveSwift, since ReactiveCocoa does not reexport ReactiveSwift at this moment.

justinvallely:
That solves the mystery. Thanks gents!

Issue #3188, opened 2016-09-14
Multiple Swift 3.0 build issues


garethcurtis:
Hi there,

I've just updated to Swift 3.0 and I'm having some trouble getting ReactiveCocoa going. I was using it as a Pod but understand that's not ready yet, so have used Carthage instead. I'm getting the following errors. Any ideas please? Have I missed something?

I believe we last had it compiling with Swift 3.0 in Xcode 8.0 Beta 6.

Thanks,

Gareth.

![screen shot 2016-09-14 at 16 09 19](https://cloud.githubusercontent.com/assets/9380217/18517875/c8d6e4a6-7a95-11e6-923e-883365e17303.png)

andersio:
You would need to import `ReactiveSwift` too.

PR #3355, opened 2016-12-21
Mention the necessary import of ReactiveSwift

iv-mexx:
I see this issue coming up time and again, like

* http://stackoverflow.com/questions/41155298/use-of-unresolved-operator
* https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3192
* https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3188

When updating fro RAC 4 to RAC 5, its often not clear that its necessary to import ReactiveSwift as well. 

I just added a quick note about that in some places where I think it's most appropriate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ In version 5.0, we split ReactiveCocoa into multiple repositories for reasons ex
 
 **If you’re using both the Swift and Objective-C APIs**, you likely require both ReactiveCocoa and [ReactiveObjCBridge][], which depend on [ReactiveSwift][] and [ReactiveObjC][].
 
+**Attention:** If youre using ReactiveCocoa, you'll most likely need to import ReactiveSwift as well when using classes or operators that are implemented in ReactiveSwift.
+
 #### ReactiveCocoa
 The ReactiveCocoa library is newly focused on Swift and the UI layers of Apple’s platforms, building on the work of [Rex](https://github.com/neilpa/Rex).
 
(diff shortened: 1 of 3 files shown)


andersio:
A better solution seems like having ReactiveCocoa reexporting ReactiveSwift. IIRC @sharplet has attempted it before. Is it still a viable option?

iv-mexx:
Yeah, that would of course be a better solution. 
Somehow I thought this was not yet possible...

andersio:
Perhaps `@_exported` can be reconsidered. It is not publicly supported, but it is used in action (Foundation, overlays, etc).

sharplet:
> A better solution seems like having ReactiveCocoa reexporting ReactiveSwift. … Is it still a viable option?

It _almost_ works. Two of the issues you linked would be fixed, but not the one about `<~`, because… `@_exported import` and modular exports via the bridging header don't re-export operator declarations.

However, it may still be helpful to re-export Result and ReactiveSwift, it's just that it might become even _more_ confusing that the operators don't appear to exist without `import ReactiveSwift`. But maybe not.

I think the complete solution (workaround?) would be to somehow include the operator declarations in both ReactiveSwift and ReactiveCocoa. But I don't know how to best manage that dependency across three different package managers.

andersio:
> I think the complete solution (workaround?) would be to somehow include the operator declarations in both ReactiveSwift and ReactiveCocoa. But I don't know how to best manage that dependency across three different package managers.

I think that would be fine to redeclare `<~` in ReactiveCocoa, while reusing the precedencegroup from ReactiveSwift (if possible).

mdiep:
This seems like a great improvement to me. We can consider exporting separately.

PR #3362, opened 2017-01-01
Bypass the message forwarding for common method signatures.

andersio:
While Objective-C message forwarding is convenient as a method interception medium, it comes with a tangible overhead that might not be ideal in performance critical use sites, e.g. `prepareForReuse` and potential delegaet intercepting signals in collection views.

The PR includes concrete implementation templates for the following method signatures:

```
[0 arg]  v@:
[1 args] v@:c   v@:l   v@:q   v@:@
[2 args] v@:@q  v@:@l  v@:@d  v@:@f  v@:@@
[3 args] v@:@@q v@:@@@
```
This set of signatures covers common property setters and many critical collection delegate requirements. Note that so far it only works when `void` is the return type.

The method interception routines would use the generated implementation from these templates, instead of the ObjC forwarding chain. These templates still respect the RAC interoperability guarantees of method interception.

Delegate interception might benefit from this PR, should we reuse the method interception facility and `forwardingTarget(for:)` to implement a delegate proxy.

#### Benchmarks
InterceptingTest, 50000 iterations, Whole Module Optimization
macOS 10.12.2, i5 4258U

| Test | Substituted | RC1 | Speedup |
| ---- | ----------- | ---------- | ------- |
| Simple Signal* | 0.020 (8%) | 0.020 (8%) | - |
| No arguments (`v@:`) | 0.053 (16%) | 0.141 (6%) | 2.66x |
| RAC-KVO, 1 argument (`v@:q`) | 0.815 (4%)  | 1.017 (3%) | 1.25x |
| KVO-RAC, 1 argument (`v@:q`) | 0.810 (3%)  | 1.472 (3%) | 1.82x |

\* Emit a value to the `Signal` through a `dynamic` method.

In the two KVO interoperability tests, RAC only accounts for 5-15% of the CPU time in multiple runs. It also mitigated the regression due to the IMP cache thrashing in the KVO-RAC interoperability test.

### DIFF
--- a/ReactiveCocoa/NSObject+Intercepting.swift
+++ b/ReactiveCocoa/NSObject+Intercepting.swift
@@ -72,7 +72,6 @@ extension NSObject {
 		assert(checkTypeEncoding(typeEncoding))
 
 		return synchronized {
-			let alias = selector.alias
 			let interopAlias = selector.interopAlias
 			let stateKey = AssociationKey<InterceptingState?>(interopAlias)
 
@@ -106,7 +105,7 @@ extension NSObject {
 					subclassAssociations.setValue(selectorCache, forKey: selectorCacheKey)
 					subclassAssociations.setValue(true, forKey: interceptedKey)
 
-					enableMessageForwarding(subclass, selectorCache)
+					enableMessageForwarding(subclass, selectorCache, typeEncoding)
 					setupMethodSignatureCaching(subclass, signatureCache)
 				}
 
@@ -177,7 +176,7 @@ extension NSObject {
 ///
 /// - parameters:
 ///   - realClass: The runtime subclass to be swizzled.
-private func enableMessageForwarding(_ realClass: AnyClass, _ selectorCache: SelectorCache) {
+private func enableMessageForwarding(_ realClass: AnyClass, _ selectorCache: SelectorCache, _ typeEncoding: UnsafePointer<Int8>) {
 	let perceivedClass: AnyClass = class_getSuperclass(realClass)
 
 	typealias ForwardInvocationImpl = @convention(block) (Unmanaged<NSObject>, AnyObject) -> Void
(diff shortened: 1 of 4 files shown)


mdiep:
Unfortunately, I don't think this speedup is worth the added complexity/code—especially when it's not born out of the needs of an actual app.

We need to strike a careful balance between speed and maintainability—especially in already complex areas of the code like this. I don't think this code is slow enough in absolute numbers or is speed up enough in relative numbers, to warrant the complexity.

I have a hard enough time following this code as-is. 😅

andersio:
If it is hard to follow, it can always be improved. But I would hate to dismiss it when it could be very well be relevant in the lights of hot spots in quite a few delegate uses, and it is not hard to foresee at all.

Of course, there are alternatives in delegate interception, and we haven't ever had prototypes to compare the performances of the approaches yet.

liscio:
@andersio How do these speedups compare to those afforded by whole module optimization when Reactive{Swift,Cocoa} are compiled as source alongside the host application being profiled? Or is that what your WMO benchmark is testing? (i.e. the test is compiled alongside the RAC source in the same "module".)

There is a *huge* performance benefit that we're all going to see when those optimization limitations are lifted once we reach the stable ABI, so it's important to frame all performance work with that in mind.

andersio:
@andersio What I was benchmarking is the speedup of the PR when the intercepted methods are invoked repeatedly. This whole thing actually happens in the ObjC runtime realm, just that the implementations are written in Swift. It is also internal to RAC, so it doesn't have a real stake in cross module performance. So I don't think it would address the things you mentioned.

liscio:
I would suggest that we sit on this PR until we have some evidence of real-world usage that points to this specific code as a bottleneck. As @mdiep says, it might not be worth the added maintenance cost of the increased complexity.

Also, those speedup percentages are very difficult to interpret without a whole other application written around it. If we're talking such a low % of CPU usage when *all the test is doing is message interception* then I don't see such a high value in attacking this specific bit of code. It could be thousands of a percentage point when surrounded by other tasks.

Don't get me wrong—I love optimizations and having super fast code. But I ignore all of these jobs until Instruments tells me that there's an actual problem in a more complete application.

PR #3371, opened 2017-01-11
Add pressed signal for NSButton

ohwutup:
I thought a simple way to get this done would be to just make the `trigger` property on NSControl `internal` rather than `private` so that it could be accessed by an NSButton extension and then the signal would just return itself. If there is a way that you all would prefer to have this done, I am happy to make changes, but this seems to work nicely too! 🚀

### DIFF
--- a/ReactiveCocoa/AppKit/NSButton.swift
+++ b/ReactiveCocoa/AppKit/NSButton.swift
@@ -4,8 +4,24 @@ import AppKit
 
 extension Reactive where Base: NSButton {
 
-	public var pressed: Signal<Base, NoError> {
-		return trigger.map { [unowned base = self.base] in base }
+	internal var associatedAction: Atomic<(action: CocoaAction<Base>, disposable: Disposable?)?> {
+		return associatedValue { _ in Atomic(nil) }
 	}
 
+	public var pressed: CocoaAction<Base>? {
+		get {
+			return associatedAction.value?.action
+		}
+
+		nonmutating set {
+			base.target = newValue
+			base.action = newValue.map { _ in CocoaAction<Base>.selector }
+			associatedAction
+				.swap(newValue.map { action in
+					let disposable = isEnabled <~ action.isEnabled
+					return (action, disposable)
+				})?
+				.disposable?.dispose()
+		}
+	}
 }
--- a/ReactiveCocoa/AppKit/NSButton.swift
+++ b/ReactiveCocoa/AppKit/NSButton.swift
@@ -4,7 +4,7 @@ import AppKit
 
 extension Reactive where Base: NSButton {
 
-	internal var associatedAction: Atomic<(action: CocoaAction<Base>, disposable: Disposable?)?> {
+	internal var associatedAction: Atomic<(action: CocoaAction<Base>, disposable: CompositeDisposable)?> {
 		return associatedValue { _ in Atomic(nil) }
 	}
 
@@ -15,15 +15,17 @@ extension Reactive where Base: NSButton {
 
 		nonmutating set {
 			associatedAction
-				.swap(newValue.map { action in
-					let disposable = CompositeDisposable()
-					disposable += isEnabled <~ action.isEnabled
-					disposable += trigger.observeValues { [unowned base = self.base] in
-						action.execute(base)
+				.modify { action in
+					action?.disposable.dispose()
+					action = newValue.map { action in
+						let disposable = CompositeDisposable()
+						disposable += isEnabled <~ action.isEnabled
+						disposable += trigger.observeValues { [unowned base = self.base] in
+							action.execute(base)
+						}
+						return (action, disposable)
 					}
-					return (action, disposable)
-				})?
-				.disposable?.dispose()
+			}
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


andersio:
To be consistent, `pressed` should accept a `CocoaAction` instead.

ohwutup:
Hey @andersio! Thanks for the feedback 😄 . I am working on making this change now but I'm not totally sure the best way to copy the pattern from UIControl, any help would be greatly appreciated!

ReactiveCocoa/AppKit/NSButton.swift
```diff
@@ -4,8 +4,24 @@ import AppKit
...
+		return associatedValue { _ in Atomic(nil) }
 	}
 
+	public var pressed: CocoaAction<Base>? {
+		get {
+			return associatedAction.value?.action
+		}
+
+		nonmutating set {
+			base.target = newValue
```
andersio: Can you use `trigger` instead of assigning `CocoaAction` directly? Otherwise, this would override `trigger` when used together.

I'd expect the caveat to be solved soon, but for now let's reduce the reach of it.

ohwutup: Hey @andersio, I'm not totally sure if I understand how to use the `trigger` here instead of assigning the target/action. Any advice or clues? Thanks!

andersio: Attach an observer to `trigger` that calls `CocoaAction.execute`.

ReactiveCocoaTests/AppKit/NSButtonSpec.swift
```diff
@@ -0,0 +1,40 @@
...
+			}
+
+			pressed <~ SignalProducer(action.values)
+			button.reactive.pressed = CocoaAction(action)
+			expect(pressed.value) == false
+
+			button.performClick(nil)
+			expect(pressed.value) == true
+		}
+	}
```
mdiep: Could you also test that the `enabled` state of the button changes correctly?

ReactiveCocoa/AppKit/NSButton.swift
```diff
@@ -0,0 +1,29 @@
...
+		nonmutating set {
+			associatedAction
+				.swap(newValue.map { action in
+					let disposable = CompositeDisposable()
+					disposable += isEnabled <~ action.isEnabled
+					disposable += trigger.observeValues { [unowned base = self.base] in
+						action.execute(base)
+					}
+					return (action, disposable)
+				})?
```
mdiep: I don't think this is safe here—the binding needs to happen inside the lock from `associatedAction`. This should use `modify` instead of `swap`.

ReactiveCocoa/AppKit/NSButton.swift
```diff
@@ -0,0 +1,29 @@
+import ReactiveSwift
+import enum Result.NoError
+import AppKit
+
+extension Reactive where Base: NSButton {
+
+	internal var associatedAction: Atomic<(action: CocoaAction<Base>, disposable: Disposable?)?> {
```
mdiep: Why is `disposable` optional? Seems like just the tuple needs to be optional? Not the `Disposable` inside?

ReactiveCocoa/AppKit/NSButton.swift
```diff
@@ -0,0 +1,29 @@
...
+		}
+
+		nonmutating set {
+			associatedAction
+				.swap(newValue.map { action in
+					let disposable = CompositeDisposable()
+					disposable += isEnabled <~ action.isEnabled
+					disposable += trigger.observeValues { [unowned base = self.base] in
+						action.execute(base)
+					}
```
mdiep: I think this could be:

```swift
disposable += action <~ trigger.map { [unowned base = self.base] _ in base }
```

?

ReactiveCocoa/AppKit/NSButton.swift
andersio: It can be strong typed as `CompositeDisposable`, I suppose.

ReactiveCocoa/AppKit/NSButton.swift
andersio: Hmm, true. `UIControl` would need to be updated too.

ReactiveCocoa/AppKit/NSButton.swift
ohwutup: This isn't compiling for me, giving me an error that `Binary operator '<~' cannot be applied to operands of type 'CocoaAction<Base>' and 'Signal<Base, NoError>'` but after looking at the `<~` function definition and the BindingTargetProtocol and BindingSourceProtocol, it seems like it should work to do that binding. That being said, doing `action <~ trigger.map {...` does not return a disposable right? So I'm not sure I understand how this should be added to the CompositeDisposable instance?

ReactiveCocoaTests/AppKit/NSButtonSpec.swift
ohwutup: I'm not so sure the best way to test this. Is there an an example that you know of that I can take a look at?

PR #3385, opened 2017-01-27
Improved delegate interception for UISearchBar.

andersio:
Note: Invoking `UISearchBar.delegate.set` and `Reactive<UISearchBar>.proxy` concurrently yields undefined behaviour in a similar way to method interception.

### DIFF
--- a/ReactiveCocoa/DelegateProxy.swift
+++ b/ReactiveCocoa/DelegateProxy.swift
@@ -37,18 +37,28 @@ internal class DelegateProxy<Delegate: NSObjectProtocol>: NSObject {
 
 private let hasSwizzledKey = AssociationKey<Bool>(default: false)
 
-internal protocol DelegateProxyProtocol: class {
-	associatedtype Delegate: NSObjectProtocol
+extension DelegateProxy {
+	// FIXME: This is a workaround to a compiler issue, where any use of `Self`
+	//        through a protocol would result in the following error messages:
+	//        1. PHI node operands are not the same type as the result!
+	//        2. LLVM ERROR: Broken function found, compilation aborted!
+	internal static func proxy<P: DelegateProxy<Delegate>>(
+		for instance: NSObject,
+		setter: Selector,
+		getter: Selector,
+		_ key: StaticString = #function
+	) -> P {
+		return _proxy(for: instance, setter: setter, getter: getter, key) as! P
+	}
 
-	weak var forwardee: Delegate? { get set }
+	private static func _proxy(
+		for instance: NSObject,
+		setter: Selector,
+		getter: Selector,
+		_ key: StaticString = #function
+	) -> AnyObject {
+		let key = AssociationKey<DelegateProxy<Delegate>?>(key)
 
-	init(_ originalSetter: @escaping (AnyObject) -> Void)
-}
-
-extension DelegateProxy: DelegateProxyProtocol {}
-
-extension DelegateProxyProtocol {
-	internal static func proxy(for instance: NSObject, setter: Selector, getter: Selector, _ key: AssociationKey<Self?>) -> Self {
 		return instance.synchronized {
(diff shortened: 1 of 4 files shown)


ReactiveCocoa/UIKit/iOS/UISearchBar.swift
```diff
@@ -2,27 +2,68 @@ import ReactiveSwift
...
+				guard let base = base else { return }
+				unsafeBitCast(impl, to: OriginalImpl.self)(base, #selector(setter: base.delegate), proxy)
+			}
+
+			base.associations.setValue(proxy, forKey: proxyKey)
+
+			proxy.forwardee = base.delegate
+			unsafeBitCast(impl, to: OriginalImpl.self)(base, #selector(setter: base.delegate), proxy)
+
+			return proxy
```
mdiep: It seems like we'd want to reuse this for other controls with delegates. Maybe we should extract this somewhere? Maybe event to `DelegateProxy`?

ReactiveCocoa/UIKit/iOS/UISearchBar.swift
```diff
@@ -2,27 +2,68 @@ import ReactiveSwift
...
-		let delegate = ReactiveUISearchBarDelegate()
-		base.delegate = delegate
-		objc_setAssociatedObject(base, &delegateKey, delegate, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
-		return delegate
+					let newImpl: @convention(block) (NSObject, UISearchBarDelegate) -> Void = { object, delegate in
+						let proxy = object.associations.value(forKey: proxyKey)!
+						proxy.forwardee = delegate
+					}
+
+					class_replaceMethod(subclass,
```
mdiep: I still think it'd be nice to wrap some of the Obj-C runtime APIs to make the intent clearer at call sites like this.

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -0,0 +1,36 @@
...
+		self.originalSetter = originalSetter
+	}
+
+	override func forwardingTarget(for selector: Selector!) -> Any? {
+		return interceptedSelectors.contains(selector) ? nil : forwardee
+	}
+
+	func intercept(_ selector: Selector) -> Signal<(), NoError> {
+		interceptedSelectors.insert(selector)
+		originalSetter(self)
```
mdiep: This seems like it could happen just once—at init. (See also: my comment below about extracting the logic to set up the proxy.)

andersio: A lot of Cocoa classes cache information about delegates. Since we only intercept a subset but not all, and the user supplied delegate might be changed, we need to force the delegater to update their cache by invoking the setter again.

mdiep: Can you add this as a comment in the code?

andersio: ICYMI: Comments have been added.

mdiep:
iOS doesn't seem to be building—even with multiple runs.

ReactiveCocoa/UIKit/iOS/UISearchBar.swift
```diff
@@ -39,7 +36,7 @@ extension Reactive where Base: UISearchBar {
 	/// - note: To observe text values that change on all editing events,
```
dcaunt: The `important` comment for this signal and the `continuousTextValues` signal can be updated, now that `delegate`s are proxied.

andersio:
@mdiep Seems even Swift 3.1 beta in 8.3 beta 1 failed to build the original version. https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3385/commits/7ddfd1c6f1418076ce1c1fb9f508e1bc89b618ef worked the issue around, probably by avoiding the use of `Self` w/ NSObject subclasses.

PR #3389, opened 2017-02-01
Removed references to MapKit in iOS and macOS

duncandee:
MapKit import is breaking when running release builds on device, since its only used in tvOS I have removed it from being included in the iOS and macOS targets.

Crash: 

Termination Description: DYLD, Library not loaded: @rpath/libswiftMapKit.dylib | Referenced from: /private/var/containers/Bundle/Application/C7F32BE6-5223-4ED7-9B86-7770CE357CAB/Casino.app/Frameworks/ReactiveCocoa.framework/ReactiveCocoa | Reason: image not found


mdiep:
> since its only used in tvOS

That's not true. It's used on iOS on macOS. It doesn't have an `@available` annotation for those platforms because it's available on all supported versions of those platforms.

PR #3390, opened 2017-02-03
Add `isSecureTextEntry` binding to UITextField.

jklausa:
It's common in apps to allow users to switch the password entry field to plaintext, in order to help them spot typos they may have made while inputting them.

This PR makes it easier by exposing `isSecureTextEntry` property of `UITextField` (technically of `UITextInputTraits` protocol) in a reactive way.

### DIFF
--- a/ReactiveCocoa/UIKit/UITextField.swift
+++ b/ReactiveCocoa/UIKit/UITextField.swift
@@ -42,4 +42,9 @@ extension Reactive where Base: UITextField {
 	public var continuousAttributedTextValues: Signal<NSAttributedString?, NoError> {
 		return controlEvents(.editingChanged).map { $0.attributedText }
 	}
+
+	/// Sets the secure text entry attribute on the text field.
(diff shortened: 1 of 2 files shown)


jklausa:
I also have a broader question — how is `ReactiveCocoa`'s API surface area expected to grow? 

Is the idea to provide bindings to most commonly used properties/actions (`UILabel.text`, `UITextField.continuousTextValues`, etc.) and users can add more esoteric ones on their own?

Is there a bar of "it seems like this could be conceivably useful for multiple people"? 

I can see you probably not wanting PR's for every last property of a [UI|App]Kit control (or maybe not!), but I was just curious what the philosophical approach is here.

Thanks for the awesome work on the project so far to everyone involved!

mdiep:
> how is ReactiveCocoa's API surface area expected to grow?

That's a good question. I think we expect to accept PRs for any property that people find useful.

mdiep:
Thanks for the PR!

PR #3394, opened 2017-02-07
Hide the NSControl interception.

andersio:
Similar to #3385.

A new protocol `ActionMessageSending` is introduced. Conforming `NSObject` subclasses would automatically gain a lazily-initialised proxy via `Self.reactive.proxy`, and the instances would have their `target` and `action` setters redirected to the proxy.

Like any interception API, concurrent use has undefined behaviour, though AppKit components should be manipulated on the main thread anyway.

### DIFF
--- a/ReactiveCocoa/AppKit/ActionProxy.swift
+++ b/ReactiveCocoa/AppKit/ActionProxy.swift
@@ -19,11 +19,7 @@ internal final class ActionProxy<Owner: AnyObject> {
 	// In AppKit, action messages always have only one parameter.
 	@objc func consume(_ sender: Any?) {
 		if let action = action {
-			if let target = target {
-				_ = target.perform(action, with: sender)
-			} else {
-				NSApp.sendAction(action, to: nil, from: sender)
-			}
+			NSApp.sendAction(action, to: target, from: sender)
 		}
 
 		observer.send(value: owner)
--- a/ReactiveCocoa/AppKit/ActionProxy.swift
+++ b/ReactiveCocoa/AppKit/ActionProxy.swift
@@ -1,0 +1,104 @@
+import AppKit
+import ReactiveSwift
+import enum Result.NoError
+
+internal final class ActionProxy<Owner: AnyObject> {
+	internal weak var target: AnyObject?
+	internal var action: Selector?
+	internal let signal: Signal<Owner, NoError>
+
+	private let observer: Signal<Owner, NoError>.Observer
+	private unowned let owner: Owner
+
+	internal init(owner: Owner, lifetime: Lifetime) {
+		self.owner = owner
+		(signal, observer) = Signal<Owner, NoError>.pipe()
+		lifetime.ended.observeCompleted(observer.sendCompleted)
+	}
+
+	// In AppKit, action messages always have only one parameter.
+	@objc func consume(_ sender: Any?) {
+		if let action = action {
+			if let target = target {
+				_ = target.perform(action, with: sender)
+			} else {
+				NSApp.sendAction(action, to: nil, from: sender)
+			}
+		}
+
+		observer.send(value: owner)
+	}
+}
+
+private let hasSwizzledKey = AssociationKey<Bool>(default: false)
+
+@objc internal protocol ActionMessageSending: class {
+	weak var target: AnyObject? { get set }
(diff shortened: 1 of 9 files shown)


ReactiveCocoa/AppKit/ActionProxy.swift
```diff
@@ -0,0 +1,104 @@
...
+		lifetime.ended.observeCompleted(observer.sendCompleted)
+	}
+
+	// In AppKit, action messages always have only one parameter.
+	@objc func consume(_ sender: Any?) {
+		if let action = action {
+			if let target = target {
+				_ = target.perform(action, with: sender)
+			} else {
+				NSApp.sendAction(action, to: nil, from: sender)
```
mdiep: Can we remove the `if` and just pass `target` as `to` here?

If not, I think this deserves a comment in the code.

ReactiveCocoa/AppKit/ActionProxy.swift
```diff
@@ -0,0 +1,104 @@
...
+}
+
+private let hasSwizzledKey = AssociationKey<Bool>(default: false)
+
+@objc internal protocol ActionMessageSending: class {
+	weak var target: AnyObject? { get set }
+	var action: Selector? { get set }
+}
+
+extension Reactive where Base: NSObject, Base: ActionMessageSending {
```
mdiep: Instead of adding `ActionMessageSending`, how about just extending `NSControl` here?

That's where `action` and `target` are defined. I don't think there's any usage of that pattern on macOS outside of `NSControl`.

ReactiveCocoa/AppKit/ActionProxy.swift
```diff
@@ -0,0 +1,104 @@
...
+			// Swizzle the instance only after setting up the proxy.
+			let subclass: AnyClass = swizzleClass(base)
+
+			let targetSetter = #selector(setter: base.target)
+			let actionSetter = #selector(setter: base.action)
+
+			// Swizzle the original setters, and redirect subsequent target and action
+			// assignments to the proxy.
+			try! ReactiveCocoa.synchronized(subclass) {
+				print(subclass)
```
mdiep: `print` left in here

ReactiveCocoa/AppKit/ActionProxy.swift
```diff
@@ -0,0 +1,104 @@
...
+					let newActionSetterImpl: @convention(block) (NSObject, Selector?) -> Void = { object, selector in
+						let proxy = object.associations.value(forKey: key)!
+						proxy.action = selector
+					}
+
+					class_replaceMethod(subclass,
+					                    actionSetter,
+					                    imp_implementationWithBlock(newActionSetterImpl as Any),
+					                    actionSetterTypeEncoding)
+				}
```
mdiep: Some of this code is repeated from `DelegateProxy`. Can we share the swizzling code somewhere?

ReactiveCocoa/AppKit/ActionProxy.swift
```diff
@@ -0,0 +1,104 @@
+import AppKit
+import ReactiveSwift
+import enum Result.NoError
+
+internal final class ActionProxy<Owner: AnyObject> {
+	internal weak var target: AnyObject?
+	internal var action: Selector?
+	internal let signal: Signal<Owner, NoError>
```
mdiep: Can we give this a more descriptive name? `invoked`, maybe?

mdiep:
Sorry that I'm just now getting to this. Love the idea, but I think we should make a few improvements, as noted above.

ReactiveCocoa/AppKit/ActionProxy.swift
andersio: Yeah, I think so.

ReactiveCocoa/AppKit/ActionProxy.swift
andersio: A counter example would be `NSMenuItem`.

ReactiveCocoa/AppKit/ActionProxy.swift
andersio: Oops.

mdiep:
This is great. ✨

PR #3395, opened 2017-02-07
Add UIPickerView reactive extensions

dcaunt:
This adds support for proxying selectors with arguments in `ProxyDelegate` and uses this to provide `selections` on the `reactive` extension for `UIPickerView`.

I have added tests for the reactive extensions but not for the proxying behaviour (see [UISearchBarSpec](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/14d36c7c1bfd7e95af5862ac0328d679774d0b8e/ReactiveCocoaTests/UIKit/UISearchBarSpec.swift#L69-L107)). I can add these if necessary.

### DIFF
--- a/ReactiveCocoa/UIKit/iOS/UIPickerView.swift
+++ b/ReactiveCocoa/UIKit/iOS/UIPickerView.swift
@@ -1,0 +1,43 @@
+import ReactiveSwift
+import enum Result.NoError
+import UIKit
+
+private class PickerViewDelegateProxy: DelegateProxy<UIPickerViewDelegate>, UIPickerViewDelegate {
+	@objc func pickerView(_ pickerView: UIPickerView, didSelectRow row: Int, inComponent component: Int) {
+		forwardee?.pickerView?(pickerView, didSelectRow: row, inComponent: component)
+	}
+}
+
+extension Reactive where Base: UIPickerView {
+
+	private var proxy: PickerViewDelegateProxy {
+		return .proxy(for: base,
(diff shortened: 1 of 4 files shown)


ReactiveCocoa/UIKit/iOS/UIPickerView.swift
```diff
@@ -0,0 +1,43 @@
...
+	/// Reloads the specified component
+	public var reloadComponent: BindingTarget<Int> {
+		return makeBindingTarget { $0.reloadComponent($1) }
+	}
+
+	/// Create a signal which sends a `value` event for each row selection
+	///
+	/// - returns:
+	///   A trigger signal.
+	public var selections: Signal<(Int, Int), NoError> {
```
mdiep: Would it be better to add a `Selection` type for this?

```swift
struct UIPickerViewSelection {
    var row: Int
    var component: Int
}

public var selections: Signal<UIPickerViewSelection, NoError> { ... }
```

(That's an honest question.)

I find it a little strange that UIKit doesn't use an `NSIndexPath` for this. Maybe we should just adopt that.

dcaunt: The same had occurred to me, so I think it would – otherwise the order of row and component is ambiguous. I think a new type would be clearer than `NSIndexPath`.

andersio:
I'd recommend named tuple instead, since it is rather compact structure, unlike the keyboard change context.

mdiep:
Thanks!

Issue #3383, opened 2017-01-24
I cannot run playground 


HanXiaoTeng:
I already build result,reactiveSwift and reactiveCocoa,and i open the ReactiveCocoa-ios.playground, but it still "run Sanbox" then "objc[15129]: pthread_rwlock_wrlock failed (11)"

alper:
I did a fresh checkout and tried out the steps for the playground. I got stuck at the step:

> Build ReactiveSwift-iOS scheme

This scheme seems to be geared towards a MacOS target and not an iOS one.

sharplet:
@alper That can happen if you run checkout step while the workspace is already open in Xcode. Can you try closing Xcode and reopening it?

alper:
- I deleted everything.
- Did a new checkout.
- Closed the playground and the workspace.
- Did the git submodule init
- Opened the workspace and started building, but ReactiveSwift-iOS still targets My Mac.

alper:
I'm following the instructions from the playground.

I tried again from scratch and got the same result.

![scheme](https://cloud.githubusercontent.com/assets/72429/22744590/11743ffa-ee1e-11e6-9561-919bd09b65d2.jpg)

ikesyo:
You may need this `git submodule update --init --recursive`. The missing `--recursive` option is important and that should be documentation bug.

alper:
Ok. Now I'm further along, but now it can't find the built libraries.

(Also the playground is empty.)

```
Playground execution failed: error: Sandbox.xcplaygroundpage:9:8: error: no such module 'Result'
import Result
       ^

error: Sandbox.xcplaygroundpage:10:8: error: no such module 'ReactiveCocoa'
import ReactiveCocoa
       ^

error: Sandbox.xcplaygroundpage:11:8: error: no such module 'ReactiveSwift'
import ReactiveSwift
       ^
```

ikesyo:
Are you sure that you open (select) the playground within the workspace? Otherwise playground files can't find frameworks. That is a limitation or requirement of Xcode itself.

alper:
Yes, I am. It still can't find them.

Travis4938:
Guys, I unfortunately  trapped  into this problem although I taked the steps as the guide in the playground。 Here is the log ：
-------------------------------------------------------------------------------------------------
Playground execution failed: error: Sandbox.xcplaygroundpage:11:8: error: no such module 'ReactiveCocoa'
import ReactiveCocoa
[...]

PR #3396, opened 2017-02-08
Correctly initialize the submodules

alper:
This seems to be a bug as discussed in #3383


ikesyo:
Could you apply the change to other playgrounds in this repo as well?

alper:
Done.

PR #3397, opened 2017-02-09
Fix #selector usage in README

kean:


sharplet:
Thanks for the PR!

This is an interesting one, because this code will compile just fine if inside a view controller. I personally err on the side of brevity in situations like this. What do you think?

kean:
There is not indication that this code belongs inside the UIViewController impl. And if it was inside it would probably spell simply `reactive.trigger(for: #selector(viewWillAppear(_:)))` with an implicit `self` instead of `viewController.reactive.trigger(for: #selector(viewWillAppear(_:)))`.

And triggers are mostly used outside of the implementation of the objects for which they are created, aren't they? I'm just starting with ReactiveCocoa, and that was my initial impression.

PR #3404, opened 2017-02-28
Update ReactiveSwift to 1.1.0.

andersio:


ReactiveCocoa.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "ReactiveCocoa"
-  s.version      = "5.0.0"
+  s.version      = "5.0.1"
```
mdiep: This doesn't need to be updated here.

ikesyo:
It appears that now RAS 1.1 is not binary compatible against RAS 1.0 so we should explicitly require RAS 1.1 or higher in `Cartfile`.

ikesyo:
Addressed https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3404#issuecomment-283060819 and https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3404#issuecomment-283062963 in 41bcdf7.

PR #3410, opened 2017-03-11
add UITextField set textColor Binding Target

c12dd:


c12dd:
OK

mdiep:
Thanks for the PR!

PR #3413, opened 2017-03-15
Allow specifying KVO options for `NSObject.reactive.values(forKeyPath:)`

Burgestrand:
Original motivation for this is you don't always want the initial value.

A real-world example is from dealing with an `AVAsset` in `AVFoundation`. When you read `asset.duration` and the duration is not known yet, then `AVAsset` will busy-load asset metadata to give you a value, which can lock up the main thread. You can avoid this by reading the status of the `duration` key using `AVAsynchronousKeyValueLoading`, and then use KVO to deliver updates if the duration is not yet known.

### DIFF
--- a/ReactiveCocoa/NSObject+KeyValueObserving.swift
+++ b/ReactiveCocoa/NSObject+KeyValueObserving.swift
@@ -24,6 +24,32 @@ extension Reactive where Base: NSObject {
 			disposable += self.lifetime.ended.observeCompleted(observer.sendCompleted)
 		}
 	}
+
+	/// Create a signal all changes of the property specified by the key path.
+	///
+	/// The signal completes when the object deinitializes.
+	///
+	/// - note:
+	///	  Does not send the initial value. See `producer(forKeyPath:)`.
+	///
+	/// - parameters:
+	///   - keyPath: The key path of the property to be observed.
+	///
+	/// - returns:
+	///   A producer emitting values of the property specified by the key path.
+	public func signal(forKeyPath keyPath: String) -> Signal<Any?, NoError> {
+		return Signal { observer in
+			let disposable = CompositeDisposable()
+			disposable += KeyValueObserver.observe(
+				self.base,
+				keyPath: keyPath,
+				options: [.new],
+				action: observer.send
+			)
+			disposable += self.lifetime.ended.observeCompleted(observer.sendCompleted)
+			return disposable
+		}
+	}
 }
 
 internal final class KeyValueObserver: NSObject {
@@ -126,7 +152,7 @@ extension KeyValueObserver {
 		let observer: KeyValueObserver
 
 		if isNested {
-			observer = KeyValueObserver(observing: object, key: keyPathHead, options: options) { object in
+			observer = KeyValueObserver(observing: object, key: keyPathHead, options: options.union(.initial)) { object in
 				guard let value = object?.value(forKey: keyPathHead) as! NSObject? else {
 					action(nil)
 					return
--- a/ReactiveCocoa/NSObject+KeyValueObserving.swift
+++ b/ReactiveCocoa/NSObject+KeyValueObserving.swift
@@ -13,7 +13,7 @@ extension Reactive where Base: NSObject {
 	///
 	/// - returns:
 	///   A producer emitting values of the property specified by the key path.
-	public func values(forKeyPath keyPath: String) -> SignalProducer<Any?, NoError> {
+	public func producer(forKeyPath keyPath: String) -> SignalProducer<Any?, NoError> {
 		return SignalProducer { observer, disposable in
 			disposable += KeyValueObserver.observe(
 				self.base,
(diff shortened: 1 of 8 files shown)


ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -10,15 +10,16 @@ extension Reactive where Base: NSObject {
 	///
 	/// - parameters:
 	///   - keyPath: The key path of the property to be observed.
+	///   - options: KVO options, see `NSKeyValueObservingOptions` for more documentation.
```
sharplet: I think it would be worth documenting the default values here.

andersio:
Without an `initial` value, it should probably be a `Signal`.

mdiep:
I think it's fine to be a `SignalProducer`. There's no guarantee that a `SignalProducer` will send the current value—only that a `Signal` can't send the current value.

Burgestrand:
Technically, I guess `values(forKeyPath: …).skip(first: 1)` is _technically_ similar. Thinking about this again, is the feature creep worth it?

I also thought about doing it as a signal, and more or less came to the same conclusion as @mdiep, and because  return type ambiguity is a thing and naming is hard I used the same implementation with backwards-compatible defaults. :)

ReactiveCocoa/NSObject+KeyValueObserving.swift
Burgestrand: I agree.

I had it in at first but removed it when I saw that the generated (jazzy) documentation included the default values anyway. I'll add it later, it doesn't add much noise but probably adds some clarity, especially if working with the quick-doc.

andersio:
@mdiep Point being using the best fit construct. The situation here is no different from `Property`.

@Burgestrand `skip(first:)` would not prevent the issue of yours from happening.

Burgestrand:
Ah, very true, if it's expensive to retrieve the initial value this PR still has value.

andersio:
An alternative would be `producer(forKeyPath:)` and `signal(forKeyPath:)`, stealing the nomenclature from `Property`.

Burgestrand:
@andersio and thus also deprecate using `values(forKeyPath:)` while keeping it as an alias for `producer(forKeyPath:)`?

ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -10,15 +10,16 @@ extension Reactive where Base: NSObject {
...
 	///
 	/// - returns:
 	///   A producer emitting values of the property specified by the key path.
-	public func values(forKeyPath keyPath: String) -> SignalProducer<Any?, NoError> {
+	public func values(forKeyPath keyPath: String, options: NSKeyValueObservingOptions = [.initial, .new]) -> SignalProducer<Any?, NoError> {
```
andersio: It should not be exposed, since given the API, `old` and `prior` would not be usable at all.

andersio:
@Burgestrand Yeah.

Burgestrand:
I've done some work on doing two separate methods. Would you like me to force-push this pull-request (and thus keep the discussion), or should I open a new pull request?

mdiep:
No need to force-push or open a new PR. You don't even necessarily need a revert commit.

Burgestrand:
I already had the work in a separate branch, and merging these two was too much work. The original commit is still available in my repository if anybody cares about it; either way I pushed over it instead of creating a new pull request.

Burgestrand:
I checked the build.

The failing test case is the one I raised concerns about earlier, [a test that it should not retain a replaced value in a nested key value path](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/a475495d9a687fc6f190dd70781e1e10942604b2/ReactiveCocoaTests/KeyValueObservingSpec.swift#L147).

The test fails in master as well, assuming one adds `parentObject.reactive.values(forKeyPath: "rac_object.rac_value").start()` to the beginning of it. I don't know enough about the underlying mechanics here to know what's wrong with it, both in master and in this PR.

andersio:
> The failing test case is the one from master that I raised concerns about earlier, a test that it should not retain a replaced value in a nested key value path. It fails in master as well

It looks like a missed bit when I ported it to Swift. I will look into it. For this PR though, perhaps it should be taken out to stick to the said feature.

> I added .initial to the heads of the KVO chain in my patch, but I believe that'll cause multiple changes to be emitted for multi-nested key paths.

It doesn't seem to be the case. It would cause an initial value to be emitted, but based on the semantic of `Signal` it wouldn't be caught by the API user.

Burgestrand:
Fair enough, if the behavior is otherwise correct I'll mark the culprit test as pending.

mdiep:
This looks good to me, but needs some conflicts resolved since #3418 was merged. Would you mind resolving them?

Burgestrand:
@mdiep Sure, I'll take a swing at it.

mdiep:
Thanks for the PR!

Burgestrand:
Thanks for sticking with me through it :)

PR #3417, opened 2017-03-22
Animation for BindingTargets

startupthekid:
This PR is a proposal for animatable signals by adding a small extension on `BindingTarget`. Targets now have `animate(source:withDuration:delay:options:)` methods, one for non-optional and one for optional handling, that can be called to animate any values through the pipeline.

This does not work with `NSLayoutConstraint` however due to the nature of how it's animated using `view.layoutIfNeeded` instead of applying the value directly. I'm open to any ideas on how to go about implementing that.


mdiep:
This might be better with an API like the one that was a part of [ReactiveAnimation](https://github.com/ReactiveCocoa/ReactiveAnimation/blob/30819d390d0056649dab0414444908c6181c3b5d/ReactiveAnimation/Animation.swift). (That code is RAC 3, so it looks a little different.)

startupthekid:
@mdiep how would you suggest going about implementing that? A similar API to ReactiveAnimation or something closer to how the `reactive` struct is embedded in every object?

mdiep:
Essentially like it is in the link above. Call `observer.send(value:)` inside the animation block. The values will be processed downstream inside the animation block.

startupthekid:
Sounds great, I'll close out this PR and open a new one once I finish.

PR #3418, opened 2017-03-24
Fixed an occurrence of exception due to key path tail observers not being detached.

andersio:
A spun-off from #3413.

This happens specifically when a nested key path is being observed, for example `parent.child`, where `parent` is being replaced by a new object.

`KeyValueObserver` is not eager enough to clean up the tails, causing the then-deallocated object (the original `parent`) to have still observers attached, and eventually the KVO exception.

A new test case was also added for non-nested key path, just in case.

### DIFF
--- a/ReactiveCocoaTests/KeyValueObservingSpec.swift
+++ b/ReactiveCocoaTests/KeyValueObservingSpec.swift
@@ -144,8 +144,30 @@ class KeyValueObservingSpec: QuickSpec {
 				expect(values) == [0, 1, 0, 10]
 			}
 
+			it("should not retain the replaced value") {
+				let parentObject = NestedObservableObject()
+
+				autoreleasepool {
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/KeyValueObservingSpec.swift
```diff
@@ -144,9 +144,31 @@ class KeyValueObservingSpec: QuickSpec {
 				expect(values) == [0, 1, 0, 10]
 			}
 
+			it("should not retain the replaced value") {
+				let parentObject = NestedObservableObject()
+
+				autoreleasepool {
```
andersio: Since KVO producers retain the object being observed, they were wrapped in autoreleasepool (well, closure does the job too) so that they would deinitialize when out of scope.

PR #3419, opened 2017-03-27
Fixed Swift 3.1 compiler crashes for UIPickerView.swift and UISearchBar.swift

benasher44:
Hello! We recently tried the Swift 3.1 compiler upgrade, and we came across crashes in these 2 places. The fixes seem pretty straightforward to me, but maybe there's something I'm missing here about how this was done before, since typically I'd expect `#selector()` to take an argument in the format `<type>.<method>` insteed of `<var holding instance>.<method>` (possible I just haven't encountered the latter syntax before). Here's a sample of the crash trace I was getting before (~same for both files): https://gist.github.com/benasher44/52c3a234c8a98948cdb24c8adbca522d

Thanks for your consideration!

### DIFF
--- a/ReactiveCocoa/UIKit/iOS/UIPickerView.swift
+++ b/ReactiveCocoa/UIKit/iOS/UIPickerView.swift
@@ -12,8 +12,8 @@ extension Reactive where Base: UIPickerView {
 
 	private var proxy: PickerViewDelegateProxy {
 		return .proxy(for: base,
-		              setter: #selector(setter: base.delegate),
-		              getter: #selector(getter: base.delegate))
+		              setter: #selector(setter: Base.delegate),
+		              getter: #selector(getter: Base.delegate))
 	}
 
 	/// Sets the selected row in the specified component, without animating the
--- a/ReactiveCocoa/UIKit/iOS/UISearchBar.swift
+++ b/ReactiveCocoa/UIKit/iOS/UISearchBar.swift
@@ -15,8 +15,8 @@ private class SearchBarDelegateProxy: DelegateProxy<UISearchBarDelegate>, UISear
 extension Reactive where Base: UISearchBar {
 	private var proxy: SearchBarDelegateProxy {
 		return .proxy(for: base,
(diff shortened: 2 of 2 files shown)


NachoSoto:
Weird, I didn't run into these. 

> but maybe there's something I'm missing here about how this was done before, since typically I'd expect #selector() to take an argument in the format <type>.<method> insteed of <var holding instance>.<method>

Those represents 2 different things, `Type.method` would get the method in a type, but that'd be making an assumption on what type `base` is! Which is why `base` is more correct.

What crash are you getting? Can you file a [Jira](https://bugs.swift.org) instead?

NachoSoto:
Oh you're using a development snapshot (`swift-3.1-DEVELOPMENT-SNAPSHOT-2017`), this doesn't actually crash with the released `3.1`. So I recommend that you file this against Apple instead, since it looks like another regression on their part.

benasher44:
I'm using the March 24 snapshot, which was released on Friday. I thought this matched HEAD `swift-3.1-branch` only a few hours ago, but maybe I'm crazy. Looks like the `swift-3.1-RELEASE` release was published while I was diving into this. Will get that tag and update with once I'm done. Thanks!

benasher44:
To answer:
> What crash are you getting? Can you file a Jira instead?

See the link to the gist in the description.

andersio:
`#selector` accepts any valid expression resolvable to an ObjC method. So either instance method reference or unbound method reference works.

benasher44:
That sounds right, since in Obj-C it only matters where the selector is dispatched (i.e. the target).

That said, I'm going to close this, since this appears to work find in Xcode 8.3. I'm unsure why the [changes that occurred since the snapshot I was using](https://github.com/apple/swift/compare/swift-3.1-DEVELOPMENT-SNAPSHOT-2017-03-24-a...swift-3.1-RELEASE) would have fixed anything, since the only change appears to be how `swift` prints the version. There's probably some gap in my understanding of how toolchain releases work. Thanks again, and sorry for the trouble!

PR #3420, opened 2017-03-28
Revamped `DelegateProxy` for public API exposure.

andersio:
#### Changes
1. `DelegateProxy` is now public.

2. Replaced the `DelegateProxy.intercept` methods with overloads of `DelegateProxy.reactive.trigger(for:)` and `DelegateProxy.reactive.signal(for:)`.


andersio:
Considering rebuilding `DelegateProxy` on top of ObjC message forwarding, and this will impact the public API of `DelegateProxy`. So close for now.

PR #3426, opened 2017-03-30
Update dependencies & CI settings for Swift 3.1.

andersio:


ReactiveCocoaTests/UIKit/UIButtonSpec.swift
```diff
@@ -7,6 +7,8 @@ import enum Result.NoError
 
 class UIButtonSpec: QuickSpec {
 	override func spec() {
+		UIControl.swizzleForTesting()
```
andersio: This isn't the best solution. But I am not aware of any nice way to "inject" this at startup. 😕

PR #3428, opened 2017-04-01
Refactoring for Swift 3.1.

andersio:


mdiep:
That test doesn't want to pass on this branch. 😕

ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -8,6 +8,8 @@ import CoreGraphics
 
 class InterceptingSpec: QuickSpec {
 	override func spec() {
+		ForwardInvocationTestObject.swizzleForTesting()
```
NachoSoto: I'd put this in `beforeAll`. If these tests aren't gonna run there's no point in calling this function.

ReactiveCocoaTests/UIKit/UIButtonSpec.swift
```diff
@@ -7,6 +7,8 @@ import enum Result.NoError
 
 class UIButtonSpec: QuickSpec {
 	override func spec() {
+		UIControl.swizzleForTesting()
```
NachoSoto: Ditto

ReactiveCocoaTests/InterceptingSpec.swift
andersio: `beforeAll` is not available in Quick IIRC.

NachoSoto: Oh I think its `beforeSpec` now.

ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -824,13 +828,23 @@ class InterceptingSpec: QuickSpec {
 	}
 }
 
+extension DispatchQueue {
```
NachoSoto: This can be private

andersio:
#3446 seems to have a better approach. Closing for now.

Issue #3432, opened 2017-04-02
NSButton inside NSStackView does not emit state


ohbargain:
After upgrade to 5.0.2, NSButton embedded inside NSStackView stopped emitting state signals. 

I will make pull request with test case.

Xcode version: 8.3
macOS version: 10.12.4

andersio:
This is due to a KVO interoperability issue with the swizzling logic in `ActionProxy`. Fixed in #3435.

PR #3433, opened 2017-04-02
Add spec to demonstrate NSButton inside NSStackView failure

ohbargain:
demonstrates #3432


andersio:
I have nailed down the bug. It is a KVO interoperability issue in the swizzling logic, since NSStackView somehow uses KVO.

PR #3434, opened 2017-04-02
Fix mkmapview spec deallocation

### LINKED ISSUE #3430: MKMapView fails to deallocate
macOS: 10.12.4
Xcode: 8.3

MKMapView seems to be failing to deallocate after each test. I suspect it's not RAC's fault but reporting here to share information nonetheless.

ohbargain:
fixes #3430


ohbargain:
closing. realized this is wrong approach.

PR #3435, opened 2017-04-02
Improve the interoperability of all ObjC swizzling activities in the framework.

### LINKED ISSUE #3432: NSButton inside NSStackView does not emit state

andersio:
Fixes #3432 and #3433.

Since fixing one with new test cases leads to another, it all ends up in the same branch. 😅

#### Highlights

1. https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3435/commits/e8bfbfc40fade886f0458d07dfd7c45f7ab9cf11: Fixed an interoperability issue in the swizzling logic. The original implementation had not taken account of the existence of other swizzling users.

1. https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3435/commits/b774210cba5c0579693c301ae95f5aa2fea7bbfa: Replaced the unreliable use of `#function` as keys with the setter selector, which is constant for a given proxy.

1. https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3435/commits/d3b850e3c58302f796627d9ef38b0681e910f39b: Refined the DelegateProxy implementation, so that it always respects the latest implementation in the perceived class (in case it has been swizzled externally).

1. https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3435/commits/434227c25add109440397e8acf1442ab18d1d063: `ActionProxy` now guards for the absence of `NSApp`. Probably only relevant in tests.

### DIFF
--- a/ReactiveCocoa/AppKit/ActionProxy.swift
+++ b/ReactiveCocoa/AppKit/ActionProxy.swift
@@ -55,18 +55,23 @@ extension Reactive where Base: NSObject, Base: ActionMessageSending {
 			// The proxy must be associated after it is set as the target, since
 			// `base` may be an isa-swizzled instance that is using the injected
 			// setters below.
-			base.target = proxy
-			base.action = #selector(proxy.invoke(_:))
+			typealias TargetSetter = @convention(c) (NSObject, Selector, AnyObject?) -> Void
+			typealias ActionSetter = @convention(c) (NSObject, Selector, Selector?) -> Void
+
+			let setTargetImpl = class_getMethodImplementation(superclass, #selector(setter: base.target))
+			unsafeBitCast(setTargetImpl, to: TargetSetter.self)(base, #selector(setter: base.target), proxy)
+
+			let setActionImpl = class_getMethodImplementation(superclass, #selector(setter: base.action))
+			unsafeBitCast(setActionImpl, to: ActionSetter.self)(base, #selector(setter: base.action), #selector(proxy.invoke(_:)))
+
 			base.associations.setValue(proxy, forKey: key)
 
 			let newTargetSetterImpl: @convention(block) (NSObject, AnyObject?) -> Void = { object, target in
 				if let proxy = object.associations.value(forKey: key) {
 					proxy.target = target
 				} else {
-					typealias Setter = @convention(c) (NSObject, Selector, AnyObject?) -> Void
-					let selector = #selector(setter: ActionMessageSending.target)
-					let impl = class_getMethodImplementation(superclass, selector)
-					unsafeBitCast(impl, to: Setter.self)(object, selector, target)
+					let impl = class_getMethodImplementation(superclass, #selector(setter: self.base.target))
+					unsafeBitCast(impl, to: TargetSetter.self)(object, #selector(setter: self.base.target), target)
 				}
 			}
 
@@ -74,10 +79,8 @@ extension Reactive where Base: NSObject, Base: ActionMessageSending {
 				if let proxy = object.associations.value(forKey: key) {
 					proxy.action = action
 				} else {
-					typealias Setter = @convention(c) (NSObject, Selector, Selector?) -> Void
-					let selector = #selector(setter: ActionMessageSending.action)
-					let impl = class_getMethodImplementation(superclass, selector)
-					unsafeBitCast(impl, to: Setter.self)(object, selector, action)
+					let impl = class_getMethodImplementation(superclass, #selector(setter: self.base.action))
+					unsafeBitCast(impl, to: ActionSetter.self)(object, #selector(setter: self.base.action), action)
 				}
 			}
 
--- a/ReactiveCocoa/AppKit/ActionProxy.swift
+++ b/ReactiveCocoa/AppKit/ActionProxy.swift
@@ -43,22 +43,37 @@ extension Reactive where Base: NSObject, Base: ActionMessageSending {
 			}
 
 			let proxy = ActionProxy<Base>(owner: base, lifetime: lifetime)
-			base.associations.setValue(proxy, forKey: key)
 
 			proxy.target = base.target
 			proxy.action = base.action
 
+			// The proxy must be associated after it is set as the target, since
+			// `base` may be an isa-swizzled instance that is using the injected
+			// setters below.
 			base.target = proxy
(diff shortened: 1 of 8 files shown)


andersio:
There is still an interoperability issue with the interception API of our own.

andersio:
There is one corner case of interoperability that hasn't been fixed (https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3435/commits/c35e9bab3ffedc8f6c097e338833bdffe203c2ab), because it could require non-trivial changes to the method interception logic.

I wonder if it worths the effort, since it is about supporting the simultaneous use of KVO/method interception and the proxies on the same setter... Is this a real world use case, heh?

andersio:
Seems like the MapViewSpec keeps failing with Xcode 8.3. 🙃

ReactiveCocoa/AppKit/ActionProxy.swift
```diff
@@ -2,7 +2,7 @@ import AppKit
 import ReactiveSwift
 import enum Result.NoError
 
-internal final class ActionProxy<Owner: AnyObject> {
+internal final class ActionProxy<Owner: AnyObject>: NSObject {
```
andersio: The `NSObject` is added for KVO interoperability tests to pass. I expect this to be removed by supporting native Swift object in the KVO API.

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -53,36 +54,46 @@ extension DelegateProxy {
...
 			instance.swizzle((setter, newSetterImpl), key: hasSwizzledKey)
 
-			// Set the proxy as the delegate.
-			let realClass: AnyClass = class_getSuperclass(object_getClass(instance))
+			// Retrieve the original delegate setter.
+			//
+			// `instance.objcClass` (aka `+class`) would report the perceived
+			// class instead of the real runtime subclass generated by RAC or
+			// KVO.
+			let realClass: AnyClass = instance.objcClass
```
andersio: While in most cases the superclass of the real runtime class is the original/perceived class, it is not true if the instance is first swizzled by RAC then by KVO. It would end up with being `NSKVONotifying_*_RACSwift`, i.e. two levels of inheritance.

andersio:
(Tackled a couple of flaws and consolidated the patches.)

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -53,57 +54,64 @@ extension DelegateProxy {
...
+				let impl = class_getMethodImplementation(object.objcClass, selector)
+				unsafeBitCast(impl, to: Setter.self)(object, selector, delegate)
+			}
+
 			let newSetterImpl: @convention(block) (NSObject, AnyObject?) -> Void = { object, delegate in
-				let proxy = object.associations.value(forKey: key)!
-				proxy.forwardee = (delegate as! Delegate?)
+				let proxyMap = object.associations.value(forKey: proxyMapKey)
+
+				if let proxy = proxyMap.first(where: { $0.0 == setter })?.1 as! DelegateProxy<Delegate>? {
```
mdiep: Why is `proxyMap` an array if we're only going to handle 1 item?

andersio: In rare circumstances (e.g. collection views) there could be more than one delegate being proxied. After it being eventually public, users might also have classes that got multiple delegates. So it is made to support multiple proxies.

andersio:
P.S. I wonder if we should drop a note in README or release note to discourage a mixed use of RAC and ReactiveObjC swizzling-fuelled API. There is no test coverage in this area (yet), and while they can live together, they gonna collide when being applied to the same selector. The scope of interoperability addressed here is limited to KVO and RAC itself.

ReactiveCocoa/DelegateProxy.swift
mdiep: I'll still don't understand. How could more than one be proxied? I don't see why a collection view would have more than one delegate.

andersio: `dataSource` and `collectionViewLayout` are delegates, aren't they?

mdiep: Okay, I understand what's happening now.

It seems like we could instead create an `AssociationKey` from `setter` since it's a `Selector`.

If for some reason that doesn't work, we should be able to use `[Selector: AnyObject]` since `Selector` is `Hashable`.

andersio: These all work, but just a matter of preferences. The differences are minimal in the unit of bytes. 🙂

> It seems like we could instead create an AssociationKey from setter since it's a Selector.

It could collide with others if we use it directly. In the meantime, I'd prefer to avoid a new selector unless it is necessary.

> If for some reason that doesn't work, we should be able to use [Selector: AnyObject] since Selector is Hashable.

A dictionary is bulkier than a plain array, and the expected # of elements is low (generally 1).

mdiep: > > It seems like we could instead create an AssociationKey from setter since it's a Selector.
>
> It could collide with others if we use it directly. In the meantime, it could use a prefixed selector, but I'd prefer to avoid a new selector.

Why? This seems like a much preferable option. The selector mangling could happen in AssociationKey so that everything else gets it for free.

andersio: Changed in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3435/commits/78346b55202b4f688a0f846f4c6082912b7f23e7.

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -53,57 +53,64 @@ extension DelegateProxy {
...
 		setter: Selector,
-		getter: Selector,
-		_ key: StaticString = #function
+		getter: Selector
 	) -> P {
-		return _proxy(for: instance, setter: setter, getter: getter, key) as! P
+		return _proxy(for: instance, setter: setter, getter: getter) as! P
 	}
 
 	private static func _proxy(
```
mdiep: I think we could unify the proxy could here with the proxy code in `ActionProxy`.

PR #3436, opened 2017-04-02
Disable the MapView deallocation check for macOS.

andersio:
Apparently all the updated branches and local retries just keep failing on this. ☹️


mdiep:
😞

PR #3439, opened 2017-04-03
Support for native Swift object values in `AnyObject` properties in RAC KVO API.

andersio:
Turns out ObjC association works with Swift objects too, which enables proper handling of weak properties with direct key paths. Root and intermediate nodes in a nested key path are not covered, since native Swift object does not support ObjC KVC anyway.

This applies only to `AnyObject` properties, which may hold a native Swift object that can receive ObjC message without inheriting `NSObject`.

### DIFF
--- a/ReactiveCocoa/NSObject+Lifetime.swift
+++ b/ReactiveCocoa/NSObject+Lifetime.swift
@@ -10,6 +10,27 @@ fileprivate let lifetimeKey = AssociationKey<Lifetime?>(default: nil)
 /// Holds the `Lifetime.Token` of the object.
 fileprivate let lifetimeTokenKey = AssociationKey<Lifetime.Token?>(default: nil)
 
+internal func lifetime(of object: AnyObject) -> Lifetime {
+	if let object = object as? NSObject {
+		return object.reactive.lifetime
+	}
+
+	return synchronized(object) {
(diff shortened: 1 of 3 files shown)


andersio:
The only failing test case is fixed in #3435.

mdiep:
> Turns out ObjC association works with Swift objects too

Do we know whether this is intentional/supported? "Happens to work" always makes me nervous. I wouldn't want to be in a spot where we depend on something that no longer works.

andersio:
It is supported based on my understanding. See https://github.com/apple/swift/blob/master/stdlib/public/runtime/HeapObject.cpp#L601.

ReactiveCocoa/NSObject+Lifetime.swift
```diff
@@ -10,6 +10,27 @@ fileprivate let lifetimeKey = AssociationKey<Lifetime?>(default: nil)
 /// Holds the `Lifetime.Token` of the object.
 fileprivate let lifetimeTokenKey = AssociationKey<Lifetime.Token?>(default: nil)
 
+internal func lifetime(of object: AnyObject) -> Lifetime {
```
mdiep: Maybe we should expose this in the future. And maybe even create a shortcut version of `take(during: lifetime(of: object))`. 🤔

PR #3442, opened 2017-04-05
Continuous UITextField signals now react to all editing events.

### LINKED ISSUE #3423: `continuousTextValues` doesn't emits new value on autocomplete
I've a `gmm` snippet which expands to my email configured on iPhone keyboard settings. Now I want to implement classic login form where submit button enabled when both fields are valid. But that autocompletion doesn't emit new value on `continuousTextValues` so my submit button still disabled after I "entered" email.

andersio:
Fixes #3423.

~~Note: macOS test is expected to fail due to the merged #3433.~~

### DIFF
--- a/ReactiveCocoaTests/UIKit/UITextFieldSpec.swift
+++ b/ReactiveCocoaTests/UIKit/UITextFieldSpec.swift
@@ -37,15 +37,17 @@ class UITextFieldSpec: QuickSpec {
 		}
 
 		it("should emit user initiated changes to its text value continuously") {
-			textField.text = "Test"
-
 			var latestValue: String?
 			textField.reactive.continuousTextValues.observeValues { text in
 				latestValue = text
 			}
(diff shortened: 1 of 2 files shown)


ReactiveCocoaTests/UIKit/UITextFieldSpec.swift
```diff
@@ -44,8 +44,10 @@ class UITextFieldSpec: QuickSpec {
 				latestValue = text
 			}
 
-			textField.sendActions(for: .editingChanged)
-			expect(latestValue) == textField.text
+			for event in UIControlEvents.editingEvents {
+				textField.sendActions(for: event)
+				expect(latestValue) == textField.text
```
mdiep: Since this expectation is the same on every iteration, it doesn't seem like this change adds much.

We should probably improve the test here.

PR #3446, opened 2017-04-05
[gardening] Silence the warnings

ikesyo:
- Avoids passing optional values to string interpolation
- Avoids using `class func initialize()` which should be unavailable in future Swift version
- Replaces deprecated API usages


ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -830,36 +834,28 @@ private class ForwardInvocationTestObject: InterceptedObject {
...
-				}
-
-				let success = class_addMethod(ForwardInvocationTestObject.self,
-				                ObjCSelector.forwardInvocation,
-				                unsafeBitCast(impl, to: IMP.self),
-				                ObjCMethodEncoding.forwardInvocation)
-
-				assert(success)
-				assert(ForwardInvocationTestObject.instancesRespond(to: ObjCSelector.forwardInvocation))
+	static func _initialize() {
```
NachoSoto: This can be private

ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -830,36 +834,28 @@ private class ForwardInvocationTestObject: InterceptedObject {
...
 
-				return 0
-			}()
-		}
+		let success2 = class_addMethod(ForwardInvocationTestObject.self,
+		                               ForwardInvocationTestObject.forwardedSelector,
+		                               _rac_objc_msgForward,
+		                               ObjCMethodEncoding.forwardInvocation)
 
-		_ = Static.token
```
NachoSoto: Use precondition instead: we don't want to ignore these errors in release.

ReactiveCocoaTests/InterceptingSpec.swift
ikesyo: No, this must be called from `InterceptingSpec`'s `beforeSuite`.

ReactiveCocoaTests/InterceptingSpec.swift
ikesyo: This is test code.

ReactiveCocoaTests/InterceptingSpec.swift
ikesyo: fd14f89

ReactiveCocoaTests/UIKit/UIControl+EnableSendActionsForControlEvents.swift
```diff
@@ -1,6 +1,6 @@
 import UIKit
 
-private let rac_swizzleToken: Void = {
+private func rac_swizzle() {
```
andersio: Perhaps merge this with `UIControl._initialize`.

ReactiveCocoaTests/InterceptingSpec.swift
andersio: Assertion is enough. It is not for production, and they are not expected to fail.

.swift-version
```diff
@@ -1 +0,0 @@
-3.1
```
mdiep: We don't want to specify this anymore? 😕

ikesyo: Since we don't support SwiftPM in ReactiveCocoa 5.x (currently) so this is not needed now. This is for [swiftenv](https://github.com/kylef/swiftenv).

ReactiveCocoaTests/InterceptingSpec.swift
NachoSoto: > it is not for production 

Don't we run tests in release mode though? We used to, and if not I'd strongly recommend it cause it's how we find Swift bugs with `-O`. And then we wouldn't want `assert` getting ignored here.

> they are no expected to fail

Even more reason to make them preconditions then.

andersio: Debug mode according to the scheme.

ikesyo: We are using Release configuration in CI.

PR #3453, opened 2017-04-14
Reentrance support in `CocoaTarget.invoke`.

### LINKED ISSUE #3452: Deadlock caused by resignFirstResponder
At latest version (5.0.3). Related to this change:

> Continuous text signals of UITextField now emits the latest value for all editing events emitted. In other words, replacements due to autocompletions are now covered by the signals. (#3442 , kudos to @andersio)

Code snippet:

```
import UIKit
import ReactiveSwift
import ReactiveCocoa

internal class ViewController: UIViewController {

    @IBOutlet internal var textField: UITextField!
    
    internal var textProperty: MutableProperty<String> = MutableProperty("")
    
    internal override func viewDidLoad() -> Void {
        super.viewDidLoad()
    
        let textSignal = self
            .textField
            .reactive
            .continuousTextValues
            .skipNil()
            .filter({ (value: String) -> Bool in
                return (value.characters.count == 4)
            })
            .map { (value: String) -> String in
                self.textField.resignFirstResponder() //*** -[NSLock lock]: deadlock (<NSLock: 0x6000000d2b40> 'org.reactivecocoa.ReactiveSwift.Signal.sendLock')

                return value
        }
        
        self.textProperty <~ textSignal
    }
}
```

andersio:
Fixes #3452.

WIP:
- [x] Full buffering implementation.
- [x] ~~Test coverage for `UIGestureRecognizer` (flipping `isEnabled` in observers).~~ _Cancellation in `UIGestureRecognizer` does not result in recursion._

### DIFF
--- a/ReactiveCocoa/CocoaTarget.swift
+++ b/ReactiveCocoa/CocoaTarget.swift
@@ -4,16 +4,52 @@ import enum Result.NoError
 
 /// A target that accepts action messages.
 internal final class CocoaTarget<Value>: NSObject {
+	private enum State {
+		case idle
+		case sending
+		case invokeAfterSending
+	}
+
 	private let observer: Observer<Value, NoError>
 	private let transform: (Any?) -> Value
-	
+
+	private var state: State
+
 	internal init(_ observer: Observer<Value, NoError>, transform: @escaping (Any?) -> Value) {
 		self.observer = observer
 		self.transform = transform
+		self.state = .idle
 	}
-	
-	@objc internal func sendNext(_ receiver: Any?) {
-		observer.send(value: transform(receiver))
+
+	/// Broadcast the action message to all observers.
+	///
+	/// Reentrancy is supported, and the action message would be deferred until the
+	/// delivery of the current message has completed. Deferred messages would be
+	/// coalesced.
+	///
+	/// - note: It should only be invoked on the main queue.
+	///
+	/// - parameters:
+	///   - sender: The object which sends the action message.
+	@objc internal func invoke(_ sender: Any?) {
+		switch state {
+		case .idle:
+			state = .sending
+			observer.send(value: transform(sender))
+
+			while state == .invokeAfterSending {
+				state = .sending
+				observer.send(value: transform(sender))
+			}
+
+			state = .idle
+
+		case .sending:
+			state = .invokeAfterSending
+
+		case .invokeAfterSending:
+			break
+		}
 	}
 }
 
--- a/ReactiveCocoa/CocoaTarget.swift
+++ b/ReactiveCocoa/CocoaTarget.swift
@@ -6,8 +6,7 @@ import enum Result.NoError
 internal final class CocoaTarget<Value>: NSObject {
 	private enum State {
 		case idle
-		case sending
-		case invokeAfterSending
+		case sending(queue: [Value])
 	}
 
 	private let observer: Observer<Value, NoError>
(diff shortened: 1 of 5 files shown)


ReactiveCocoa/CocoaTarget.swift
```diff
@@ -4,16 +4,56 @@ import enum Result.NoError
...
+		switch state {
+		case .idle:
+			state = .sending
+			observer.send(value: transform(sender))
+
+			while case let .replayAfterSending(values) = state {
+				if values.count == 1 {
+					state = .sending
+				} else {
+					state = .replayAfterSending(Array(values.dropFirst()))
```
mdiep: This isn't safe.

Individual reads and writes are atomic, but this write depends on the value of the previous read. So they need to be locked together.

Otherwise this read could happen, another thread could append in `replayAfterSending`, then this would remove the first value but unintentionally drop the last value.

It might be worth trying to write a test that catches that as well.

andersio: There is no another thread, since UIKit calls back on the main queue only.

mdiep: Oh, right. 🤦‍♂️

ReactiveCocoa/CocoaTarget.swift
```diff
@@ -4,16 +4,56 @@ import enum Result.NoError
 
 /// A target that accepts action messages.
 internal final class CocoaTarget<Value>: NSObject {
+	private enum State {
+		case idle
+		case sending
+		case replayAfterSending([Value])
+	}
```
mdiep: I think it would probably simplify the code do this:

```swift
private enum State {
    case idle
    case sending(queue: [Value])
}
```

What do you think?

ReactiveCocoa/CocoaTarget.swift
```diff
@@ -4,16 +4,51 @@ import enum Result.NoError
...
-	
-	@objc internal func sendNext(_ receiver: Any?) {
-		observer.send(value: transform(receiver))
+
+	/// Broadcast the action message to all observers.
+	///
+	/// Reentrancy is supported, and the action message would be deferred until the
+	/// delivery of the current message has completed.
+	///
+	/// - note: It should only be invoked on the main queue.
```
sharplet: Do you think it would be worth guarding this with `dispatchPrecondition`?

andersio: Doesn't seem so. Target-action pattern is exclusive to UIKit and AppKit.

ReactiveCocoa/CocoaTarget.swift
```diff
@@ -4,16 +4,51 @@ import enum Result.NoError
 
 /// A target that accepts action messages.
 internal final class CocoaTarget<Value>: NSObject {
+	private enum State {
+		case idle
+		case sending(queue: [Value])
+	}
+
 	private let observer: Observer<Value, NoError>
 	private let transform: (Any?) -> Value
```
mdiep: It looks like this `transform` could be eliminated. It's only ever force-casted to `Value`.

mdiep:
Just needs the changelog conflict resolved.

ReactiveCocoa/CocoaTarget.swift
andersio: `Any?` is the sender. `Value` is not necessarily the sender, but a property of it.

mdiep: But in actual usage, this transform is always `{ $0 as! Value }`.

andersio: Please see https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3472.

PR #3457, opened 2017-04-24
Fixed a bridging issue in `KeyboardChangeContext`.

andersio:
Discovered by @gkaimakas.

Updated the test case to match the notification payload.


mdiep:
Ugh. Travis simulators are constantly changing. Can you fix that up here so the tests run?

PR #3461, opened 2017-04-27
Strip the value optionality of `DynamicProperty`.

andersio:
Based on my observation, `DynamicProperty` is primarily used establish bindings or KVO observations in place (and got thrown away immediately), and sometimes as a stored property alongside the observed object.

So should we consider replacing `weak` with `unowned`, so that we can strip the optionality from `Value`? A safe, read-only version as `Property(object:keyPath:)` is added to compensate for the need of guaranteed safe access post deinitialization.

An alternative would be folding `Property(object:keyPath:)` into `DynamicProperty`. This would however mean that every `DynamicProperty` would have to cache the value. Probably not a huge issue though, since values from the ObjC world are mostly pointer or generally not-so-large values.


mdiep:
> So should we consider replacing weak with unowned, so that we can strip the optionality from Value?

I'm strongly 👎 on this. I don't want to make it easy for people to do the wrong thing.

We could probably expose a `BindingTarget` in a different way to remove the `?` there though.

andersio:
Keeping a copy is also a way, like how `Property(_:)` works now vs `Property(capturing:)`.

mdiep:
> As mentioned, keeping a copy is also an alternative, like how Property(_:) works now vs Property(capturing:). The DynamicProperty would still work after the object deinitializes, except invoking the setter resulting in nothing.

That sounds fine to me. 👍

ReactiveCocoa/DynamicProperty.swift
```diff
@@ -2,65 +2,59 @@ import Foundation
...
 	}
 
 	/// The lifetime of the property.
 	public var lifetime: Lifetime {
 		return object?.reactive.lifetime ?? .empty
 	}
 
+	/// The binding target of the property.
+	public var bindingTarget: BindingTarget<Value> {
+		return BindingTarget(lifetime: lifetime) { [weak object, keyPath] value in
```
mdiep: This should be able to use the `BindingTarget.init(object:keyPath:)` added below? Or remove that `init`?

mdiep:
This mostly looks good to me. But I think it needs to wait for 7.0 at this point.

andersio:
TBH I'd prefer to deprecate `DynamicProperty` and encourage using binding targets and KVO producers.  🙈

andersio:
ping @mdiep

mdiep:
I wonder if we could just remove `DynamicProperty`? The separate read-only `Property` and `BindingTarget` seem preferable.

andersio:
Perhaps it should be up to the user to decide. Some may prefer to have a more compact API surface, which `DynamicProperty` could deliver.

PR #3463, opened 2017-05-20
Unified Action and Property binding API for UI controls.

andersio:
Introducing a new `<~>` bidirectional binding operator. This replaces the assignment-based `CocoaAction` action binding API, and is extended to support two-way bindings with properties.

A set of the existing control `BindingTarget`s are upgraded to be `ActionBindable` and `ValueBindable`. Both still support unidirectional bindings via `<~`.

- [x] Skeleton and API mock
- [x] Implementation
- [x] Test coverage

### Before
```swift
switch.reactive.toggled = CocoaAction(viewModel.toggle) { $0.isOn }
button.reactive.pressed = CocoaAction(viewModel.submit)
refreshControl.reactive.refresh = CocoaAction(viewModel.refresh)
```

### After
```swift
switch.reactive.isOn <~> viewModel.toggle
button.reactive.pressed <~> viewModel.submit
refreshControl.reactive.isRefreshing <~> viewModel.refresh
```

### Before
```swift
usernameField.text = viewModel.username.value
viewModel.username <~ usernameField.reactive.textValues
```

### After
```swift
viewModel.username <~> usernameField.reactive.text
```

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -151,6 +151,10 @@
 		9A54A21C1DE00D09001739B3 /* ObjC+Selector.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A54A21A1DE00D09001739B3 /* ObjC+Selector.swift */; };
 		9A54A21D1DE00D09001739B3 /* ObjC+Selector.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A54A21A1DE00D09001739B3 /* ObjC+Selector.swift */; };
 		9A54A21E1DE00D09001739B3 /* ObjC+Selector.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A54A21A1DE00D09001739B3 /* ObjC+Selector.swift */; };
+		9A6523711ED09B2E00C3BA8D /* BidirectionalBinding.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6523701ED09B2E00C3BA8D /* BidirectionalBinding.swift */; };
+		9A6523721ED09B2E00C3BA8D /* BidirectionalBinding.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6523701ED09B2E00C3BA8D /* BidirectionalBinding.swift */; };
+		9A6523731ED09B2E00C3BA8D /* BidirectionalBinding.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6523701ED09B2E00C3BA8D /* BidirectionalBinding.swift */; };
+		9A6523741ED09B2E00C3BA8D /* BidirectionalBinding.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6523701ED09B2E00C3BA8D /* BidirectionalBinding.swift */; };
 		9A6AAA0E1DB6A4CF0013AAEA /* InterceptingSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6AAA0D1DB6A4CF0013AAEA /* InterceptingSpec.swift */; };
 		9A6AAA0F1DB6A4CF0013AAEA /* InterceptingSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6AAA0D1DB6A4CF0013AAEA /* InterceptingSpec.swift */; };
 		9A6AAA101DB6A4CF0013AAEA /* InterceptingSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A6AAA0D1DB6A4CF0013AAEA /* InterceptingSpec.swift */; };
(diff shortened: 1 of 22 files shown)


liscio:
@andersio I have a question regarding the "Continuous values or committed values as default?" checkbox you have in the task list.

Are you asking about the default values to be shown when the binding is first established, when validation / formatting fails, or something else?

andersio:
@liscio Whether binding with `text` bidirectionally should be continuous by default, or be consistent with the signals, i.e. only "committed" values (`editingDidEnd`).

liscio:
@andersio In that case, my vote is to not make them continuous by default, as that is a pattern that I still feel should be discouraged in many cases. 

Granted, if the controls can be configured separately to be continuous or otherwise (as is the case on macOS), then only one endpoint is required. We could also parameterize the vendor of the binding on `UIControlEvents`, so they can be easily changed for most controls, and that might achieve a similar result.

Why do I keep singing this tune? Because one of my pet peeves is "overly reactive UI" where e.g. a text field that accepts hexadecimal colors is causing my screen to flash wildly as I enter each character of the color value. Or I want to apply transparency of 85% and as I type '85' my object on the screen first disappears (8%), then re-appears at the desired transparency (85%).

There is a place for continuously updated values (e.g. sliders, knobs, direct manipulation of the objects on the screen), but a text field is rarely the place to churn so wildly.

andersio:
@liscio 

> Granted, if the controls can be configured separately to be continuous or otherwise (as is the case on macOS), then only one endpoint is required. We could also parameterize the vendor of the binding on UIControlEvents, so they can be easily changed for most controls, and that might achieve a similar result.

I do agree this is important, especially when we start working on Swift 4 with Smart Key Paths, we might be able to invent new APIs to drop explicit binding targets altogether.

> Why do I keep singing this tune? Because one of my pet peeves is "overly reactive UI" where e.g. a text field that accepts hexadecimal colors is causing my screen to flash wildly as I enter each character of the color value. Or I want to apply transparency of 85% and as I type '85' my object on the screen first disappears (8%), then re-appears at the desired transparency (85%).

Unrelated to the question, I have thought about introducing (optional) throttling and debouncing. But these probably would require an explicit barrier notion that forces all values to be written. That said these would be addictive.

liscio:
> I do agree this is important, especially when we start working on Swift 4 with Smart Key Paths, we might be able to invent new APIs to drop explicit binding targets & signals altogether.

Yep, I am looking forward to that.

>  I have thought about introducing (optional) throttling and debouncing that applies on user input.

This is a very interesting concept, though I'm failing to see a good use case for it at first glance. Time for another proposal thread, I guess… :P

andersio:
@liscio Search query with real-time suggestion that is. Though for this particular example, one may argue that the debouncing should be applied internally to the network request in the view model.

andersio:
ping @ReactiveCocoa/reactivecocoa 
This PR covers only named UIKit writable targets.

sharplet:
I'm on vacation right now, but if it can wait, would like to review this in a couple of weeks once I'm back!

andersio:
The latest iteration has added two key path subscripts that yield a `ValueBindable` which appears when `Base` conforms to `ReactiveControlConfigurable` and `ReactiveContinuousControlConfigurable`, respectively.

```
// UIDatePicker: ReactiveControlConfigurable
datePicker.reactive[\.date]

// UITextField: ReactiveContinuousControlConfigurable
textField.reactive[continuous: \.text]
```

Note that the first subscript overrides the subscript provided in `NSObject+BindingTarget`. Since Swift picks _default implementations_ with the most fit constraints, there is no return type ambiguity. This technique is also used by the Standard Library collection protocols.

andersio:
Two subscripts yielding `ValueBindable` have been added. But they are kept internal for now, since I am not sure if these are pragmatic enough to be part of the public API.

mdiep:
Sorry that it's taken me so long to look at this.

I'd really like to split this into 2 discussions:

1. Action bindings
2. Property bindings

Because although they're alike in some ways, and treated together here, they're really 2 separate topics. I'm not sure that the concepts should be unified under a single approach. Evaluating them separately would make conceptual review and code review easier in my mind. Or maybe we should consider them together but split the code into 2 PRs.

When I look at the examples above, one difference stands out to me: under the current system, actions are always on the right side but values can be on either. Given that, I'm not sure why actions need bindings like this at all. *Can't we continue to use assignment but add extra features on top?*

I'm also curious: *why this is targeted for ReactiveCocoa instead of ReactiveSwift?* These don't seem tied to Cocoa, so they would be suitable for inclusion in ReactiveSwift.

Also, when I look at this line:

```
viewModel.username <~> usernameField.reactive.text
```

It's not totally clear to me which is the source of truth, so I'm not sure that a symmetric operator is the best idea. It'd be nice if we could make the source of truth obvious.

mdiep:
> It's not totally clear to me which is the source of truth, so I'm not sure that a symmetric operator is the best idea. It'd be nice if we could make the source of truth obvious.

Maybe something like this?

```swift
usernameField.reactive.text =~> viewModel.userName
```

iluvcapra:
I totally ran into this issue immediately after downloading the cocoapod, the first thing I tried to write was a program where two sliders are summed to set a third slider, and changing the third slider updates the value of the first two. You can make it work but it's extra lines and not as elegant as the library seems to promise.

Sometimes you just need to initialize the slider position, but sometimes you want the slider to continuously change the model, while other times you just want the slider's value to track the model while it's being changed by some third party. I'm looking at options for a project for realtime audio generation, and part of that is you tend to have a lot of processes with sliders that move under their own will continuously, except when the user clicks on them, when they transition from a "read" mode to a "write" mode and they start writing their continuous value back to the model. The model then records what values the user made with the slider and plays them back, again with the slider in a "reading" aspect.

I understand why the bidirectional operator is bad because the "truth" implicitly is the model, overruled by the user's click event, and the operator doesn't express that hierarchy, the idea that the slider (or "fader" in my use) is either the truth → to the model, or is a display element the model is using as a target of _its_ truth. Maybe what the bidirectional operator is missing is a context or profile that clearly spells out when the left-hand item is reading, and when it's the truth.

andersio:
@mdiep `=~>` doesn't seem to imply the bidirectional nature.

> I'm also curious: why this is targeted for ReactiveCocoa instead of ReactiveSwift?

It has a heavy reliance of `Lifetime`. But it could be implemented like `BindingTarget` with ReactiveCocoa supplying conveniences that supply ObjC runtime backed `Lifetime` no doubt.

PR #3466, opened 2017-05-24
Introduce `take(duringLifetimeOf:)`.

andersio:
`take(duringLifetimeOf:)` uses the internal `lifetime(of:)`, which supports deinit observation of native Swift objects. This is Cocoa specific since it requires the presence of the ObjC runtime.

Alternatives: `take(alongside:)`, `take(coexisting:)`, `take(ifAlive:)`, `take(untilDeinitOf:)`, `take(referringTo:)`, `take(associating:)`, `take(tying:)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# master
+*Please put new entries at the top.
+
+1. New operator: `take(duringLifetimeOf:)`. (#3466, kudos to @andersio)
+   It is available on `Signal` and `SignalProducer`, and supports both Objective-C and native Swift objects.
+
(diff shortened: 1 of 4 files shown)


mdiep:
`referring` definitely doesn't sound right.

This is a tricky one to name. `take(whileAlive:)`?

sharplet:
I like @jeamesbone's suggestion in Slack of `take(duringLifetimeOf:)`.

andersio:
```swift
take(untilReplacement:)
take(duringLifetimeOf:)
```

It doesn't seem too bad.

PR #3467, opened 2017-05-27
DelegateProxy revamp

andersio:


andersio:
Well now it is up and running. It is true that there are opportunities to share some part of the implementations with method interception though.

ReactiveCocoa/ObjC+MessageForwarding.swift
```diff
@@ -0,0 +1,249 @@
...
+				}
+
+				invocation.setSelector(alias)
+				invocation.invoke()
+
+				return
+			}
+		}
+
+		guard !isDelegateProxy else {
```
andersio: NOTE: Perhaps the DelegateProxy would be better to have its own simplified version of `forwardInvocation`. Much of the code is useless for the proxy, or can be assumed non-interoperable as we own the proxy type. 🤔

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -1,106 +1,269 @@
...
+		invocation.invoke(withTarget: forwardee)
 	}
 }
 
-private let hasSwizzledKey = AssociationKey<Bool>(default: false)
+internal func isDelegateProxy(_ type: AnyClass) -> Bool {
+	return type is _DelegateProxyProtocol.Type
+}
+
+extension Reactive where Base: NSObject, Base: DelegateProxyProtocol {
```
mdiep: I think these methods could go on `DelegateProxy` directly.

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -1,106 +1,269 @@
...
+	var lifetime: Lifetime { get }
+}
+
+public struct DelegateProxyConfiguration {
+	fileprivate let objcProtocol: Protocol
+	fileprivate let lifetime: Lifetime
+	fileprivate let originalSetter: (AnyObject) -> Void
+}
+
+open class DelegateProxy<Delegate: NSObjectProtocol>: NSObject, DelegateProxyProtocol, _DelegateProxyProtocol {
```
mdiep: Why make this `open`? Why do you envision subclassing it?

andersio: > Subclassing is no longer necessary, but can still be used to supply non-void-returning implementations.

Method interception only supports `<U...>(_ args: U...) -> Void`, but not `<U..., T>(_ args: U...) -> T`.

ReactiveCocoa/DelegateProxy.swift
andersio: I agree that a shorthand on `DelegateProxy` would be nice. But what the overriding is about is that `reactive.trigger(for:)` is available on all `NSObject`, thus also being available on `DelegateProxy`.

ReactiveCocoa/DelegateProxy.swift
mdiep: This seems like it gives only very small gains for quite a bit of effort.

Since it's not automatic, it doesn't seem any better than setting the `delegate` yourself and using a custom delegate alongside `DelegateProxy`. The type inference on `Reactive.proxy(forKey:)` is also a bit problem since the `DelegateProxy` subclass is only specified in the return type.

I think we should leave this `final` and remove the generics.

dmcrodrigues:
@mdiep any chance of moving this forward?

mdiep:
Somehow I missed this. I just got back from vacation, but I'll this to my list. Want to fix up the conflicts, @andersio?

andersio:
Sure. Will do it this week.

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -1,106 +1,259 @@
...
+	/// - warnings: `DelegateProxy` does not support protocols containing methods that are
+	///             non-void returning. It would trap immediately when the proxy
+	///             initializes with a protocol containing such a required method, or when
+	///             the proxy is asked to intercept such an optional method.
+	///
+	/// - parameters:
+	///   - key: The key of the property that stores the delegate reference.
+	///
+	/// - returns: The proxy.
+	public func proxy<Delegate, Proxy: DelegateProxy<Delegate>>(_: Delegate.Type = Delegate.self, forKey key: String) -> Proxy {
```
mdiep: Should we use a `KeyPath` here?

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -1,106 +1,259 @@
...
 
-	internal var interceptedSelectors: Set<Selector> = []
+	internal final let lifetime: Lifetime
+
+	private final let writeLock = NSLock()
+	private final var interceptedSelectors: Set<Selector>
+	private final let originalSetter: (AnyObject) -> Void
+	private final let objcProtocol: Protocol
+
+	public required init(config: DelegateProxyConfiguration) {
```
mdiep: Should this be `public`? Or should we only expose the `proxy` method below?

If this is `public`, should we inline the properties from `DelegateProxyConfiguration` in the `init`?

andersio: The initialiser is meant to be exposed, while the details of the configuration are meant to be hidden. This is to satisfy the compiler constraint of instantiating instances from the metatype of the non-final `DelegateProxy`, which requires `init(DelegateProxyConfiguration)` to be always present in all subclasses, so that the instantiation via the metatype always succeeds.

mdiep: Why is `DelegateProxy` non-final?

ReactiveCocoa/DelegateProxy.swift
```diff
@@ -1,106 +1,262 @@
...
+}
+
+/// An opaque delegate proxy configuration.
+public struct DelegateProxyConfiguration {
+	fileprivate let objcProtocol: Protocol
+	fileprivate let lifetime: Lifetime
+	fileprivate let originalSetter: (AnyObject) -> Void
+}
+
+public class DelegateProxy<Delegate: NSObjectProtocol>: NSObject, DelegateProxyProtocol {
```
mdiep: Why can't `DelegateProxy` be `final`?

andersio: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3467#discussion_r136692297

But if you still think it is more sensible for us to start with `final`, I am fine with that.

mdiep: But if we made it `final`, then we could make [this method private](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3467/files#r250601283)? That seems worth doing

mdiep:
Let's get this merged before doing a new release for ReactiveSwift 5.

The only outstanding question from me is [this](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3467#discussion_r257713741). If we can't do that, you're good to merge.

ReactiveCocoa/DelegateProxy.swift
andersio: Yep.

andersio: In any case I think intercepting non-void returning methods is probably very niche, and a rather dangerous zone to walk into with little return. So closing it should be fine.

mdiep:
The tests need to be updated to not inherit from `DelegateProxy` now that it's final.

mdiep:
@andersio I'd like to release ReactiveCocoa 9.0. Still want to get this in?

mdiep:
It looks like there may still be some legitimate test failures?

PR #3468, opened 2017-05-27
Interception Acceleration and improve KVO interoperability.

andersio:
Preceded by #3362, *Bypass the message forwarding for common method signatures*.

This PR resurrects #3362 from the void to replace the current IMP swapping trick used for an KVO interoperating scenario #3302.

### To-do
- [ ] Written Spec for Swizzling Interoperability.
- [ ] Wait for DelegateProxy revamp (#3467).

### The Problem
As mentioned in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3302#issuecomment-269062819:
> Limited interoperability with isa-swizzling in general: RAC would not subclass any external runtime subclass.
>
> If a selector is being implemented or overriden in the runtime subclass, and subsequently swizzled by RAC, concurrent invocation of the selector would lead to undefined behaviour.

TL;DR: It isn't safe to observe a key path, and then swizzle the setter with RAC.

This is due to [the said IMP swapping trick](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/NSObject%2BIntercepting.swift#L181), which causes the actual implementation of the selector indeterministic when it is invoked concurrently.

Note that this applies to only automatic KVO. Manual KVO doesn't use isa-swizzling, so it won't mess with RAC at all.

### The Solution
We want to guarantee that **for all automatic KVO notification compliant key paths** using RAC interception on top of KVO does not lead to undefined behaviour.

To achieve this, RAC would provide concrete implementations for a defined set of setter signatures, so that we have the compile-time information to invoke the KVO implementation directly, instead of swapping implementations and invoking through `NSInvocation`.

IOW for [all the types supported by automatic KVO](https://www.mikeash.com/pyblog/friday-qa-2009-01-23.html), we would have a concrete setter generator for it. The signature of the generator would be:
```swift
// `Any` should hold a `@convention(block) (Unmanaged<AnyObject>, U...) -> Void
(InterceptorInfo) -> Any
```

⚠️ For cases that do not fit in the defined set, we would now **trap** by default and support them case by case, since isa swizzling framework is rather rare. ⚠️

### Bonus
Message forwarding is expensive. Concrete implementations cut the overhead.

### Future Work
We may extend the set of generators beyond types supported by automatic KVO. For example, #3362 suggested to cover all commonly used signatures in delegates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# master
+*Please put new entries at the top.*
+
+1. Intercepting a selector that has been swizzled by KVO beforehand no longer causes concurrent invocations of the selector to exhibit undefined behavior. (#3467, kudos to @andersio)
+
+1. Improved method interception performance for a limited set of method signatures. (#3467)
+
 # 5.0
 
 ### Table of Contents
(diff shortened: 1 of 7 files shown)


stevegomez17:
No 6.xtag

PR #3470, opened 2017-05-31
Fix bad productName values

liscio:
This (should be) a very minimally invasive PR. 

**The background:**

In general, Xcode appears to behave very badly when using workspaces and its "Find Implicit Dependencies" feature in the scheme editor. Many folks throw up their hands and resort to manual dependency specifications.

Well I think that its terrible behavior can be attributed to a bug in Xcode that causes it to not update the `productName` value in the `.pbxproj` file when a target is renamed. This happens very commonly when creating a framework project, then duplicating the framework targets with suffixes for other platforms.

For some folks, this patch may improve their build experience in Xcode quite a bit—especially for follks like me with very large workspaces that involve a lot of subprojects.

Sorry it took so long to write this one up after making the branch—I hadn't verified that the fix actually worked until I hit the issue again today, and fixing the `productName` value resolved my dependency issue. I'll try and cook up a Radar for this now that I know the steps.

### DIFF
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -977,7 +977,7 @@
 			dependencies = (
 			);
 			name = "ReactiveCocoa-tvOS";
-			productName = ReactiveCocoa;
+			productName = "ReactiveCocoa-tvOS";
 			productReference = 57A4D2411BA13D7A00F7D4B1 /* ReactiveCocoa.framework */;
 			productType = "com.apple.product-type.framework";
 		};
@@ -1014,7 +1014,7 @@
 			dependencies = (
 			);
 			name = "ReactiveCocoa-watchOS";
-			productName = ReactiveCocoa;
+			productName = "ReactiveCocoa-watchOS";
 			productReference = A9B315541B3940610001CB9C /* ReactiveCocoa.framework */;
 			productType = "com.apple.product-type.framework";
 		};
@@ -1032,7 +1032,7 @@
 			dependencies = (
 			);
 			name = "ReactiveCocoa-macOS";
-			productName = ReactiveCocoa;
+			productName = "ReactiveCocoa-macOS";
 			productReference = D04725EA19E49ED7006002AA /* ReactiveCocoa.framework */;
 			productType = "com.apple.product-type.framework";
 		};
@@ -1050,7 +1050,7 @@
 				D04725F819E49ED7006002AA /* PBXTargetDependency */,
 			);
 			name = "ReactiveCocoa-macOSTests";
-			productName = ReactiveCocoaTests;
+			productName = "ReactiveCocoa-macOSTests";
 			productReference = D04725F519E49ED7006002AA /* ReactiveCocoaTests.xctest */;
 			productType = "com.apple.product-type.bundle.unit-test";
 		};
@@ -1068,7 +1068,7 @@
 			dependencies = (
 			);
 			name = "ReactiveCocoa-iOS";
-			productName = ReactiveCocoa;
+			productName = "ReactiveCocoa-iOS";
 			productReference = D047260C19E49F82006002AA /* ReactiveCocoa.framework */;
 			productType = "com.apple.product-type.framework";
 		};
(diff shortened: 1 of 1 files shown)


sharplet:
Is this value related to the `PRODUCT_NAME` build setting at all?

liscio:
> Is this value related to the PRODUCT_NAME build setting at all?

I don't know if it's related to anything, to be honest. Perhaps it did, once upon a time…

In Xcode, it appears to be set to the same value as `name` when the target is first created, and modifying either the `PRODUCT_NAME` or other build settings does not appear to change its value.

Imagine this sequence of events (just tested myself):

1. You create an Xcode project called `fooFramework` with an iOS framework target
2. Close Xcode, and inspect the `.pbxproj` file. Note the `name` and `productName` values are identical.
3. Change the `PRODUCT_NAME` value to `foo`.
4. Close Xcode, inspect `.pbxproj`, and `productName` still doesn't change. (`productReference`—just below it—does.)
5. Change the target name from `fooFramework` to `fooFramework-iOS`
6. Close Xcode, inspect `.pbxproj`, and `productName` *still* doesn't change.

My _guess_ is that this value in the `.pbxproj` is semantically deprecated (That is—a sensible value is now deduced in code) but someone still makes a reference to this value somewhere deep in the bowels of Xcode.

I think my guess is likely true, because if you check out the changes in https://github.com/ReactiveCocoa/ReactiveSwift/pull/420, the `productName` value was still set to `ReactiveCocoa` which doesn't relate to either the target or product name, yet that still mostly worked out.

mdiep:
Well that certainly makes sense. With WWDC a week away, maybe we should wait on this and I'll try to find someone in a lab to confirm the correct value?

liscio:
> Well that certainly makes sense. With WWDC a week away, maybe we should wait on this and I'll try to find someone in a lab to confirm the correct value?

That's definitely a worthwhile course of action either way. Don't stop visiting the Xcode labs until you find someone that _truly_ seems to understand the file format.

(Possibly helpful anecdote/tip -- Once upon a time in the Xcode lab, Chris Espinosa helped me dive into a *really* old `.pbxproj` file that was causing Xcode all kinds of grief and helped me through a similar workaround. I'm pretty sure he's no longer working on the Xcode team these days, but I'd bet he'd be able to help find you someone to speak with if you're coming up short in the proper lab.)

As for whether or not to sit on this stack of PRs, I'd say that even without a definite answer it's pretty clear that these values are not being set by Xcode after a target is initially created. None of these operations will cause the value to change:

* Duplicating the target
* Renaming it
* Changing any build settings that might possibly be related

So either the value is a vestigial artifact of the long-lived file format and has no bearing on Xcode's operation, or fixing this will fix a lot of very subtle problems for people. 🤷‍♂️

mdiep:
Well I think there's a third possiblity: it's supposed to have the same value as `PRODUCT_NAME`. Any of the 3 ways, it's a Xcode bug. But I don't want to break things more by trying to fix it.

You said it fixed the issue for you locally, which makes me feel pretty good about it. But this particular bug has always been non-deterministic IME. And since things have been like this for years, I don't think another week to wait for proper confirmation is a big deal.

I'm seriously impressed that you figured this out. This is one of the more confusing Xcode bugs that I've hit. If you file a radar, definitely drop the number in here and I'll bring it to the lab.

liscio:
> Well I think there's a third possiblity: it's supposed to have the same value as PRODUCT_NAME. Any of the 3 ways, it's a Xcode bug. But I don't want to break things more by trying to fix it.

I thought that this might be the case as well, however if that's the case then the "fix" won't help anything in practice with ReactiveSwift.

The problem I originally encountered appeared to be caused by the fact that more than one target produces the same product (e.g. 4x targets that build ReactiveSwift.framework). I noticed that my builds of a macOS app target were triggering the `ReactiveSwift-tvOS` target to build in some cases, which was bonkers.

So if we go down the path of just setting it to the value of `PRODUCT_NAME`, then we might as well leave the file broken as it is. :)

> And since things have been like this for years, I don't think another week to wait for proper confirmation is a big deal.

I agree that it's not a big deal to wait. I waited this long to even open the PRs, after all…

> If you file a radar, definitely drop the number in here and I'll bring it to the lab.

Will do.

mdiep:
Okay, I brought this to the Xcode lab at WWDC, but didn't really receive any useful information. To summarize:

1. The `xcodeproj` file is very old and hard to understand. The engineers I spoke with weren't familiar with `productName` and didn't know what it represented.

2. They seemed skeptical that this would be used for implicit dependencies. They seemed to think it would be based on the target name, SDKs, and supported architectures. (Making it extra surprising that Xcode would select a product with a different SDK.)

3. They also seemed very surprised that this would be an intermittent issue. (In my experience, it's been intermittent. I don't know about yours.)

4. While they didn't think that this would affect it, they also thought it was unlikely that it would break anything.

In light of that—especially (4)—I'd say let's go ahead and merge it. But if anyone disagrees, please speak up.

mdiep:
🚢

PR #3472, opened 2017-06-04
Introduce `mapControlEvents`.

andersio:
With reentrance support in #3453, mapping control events should be done via `CocoaTarget` so that the actual value —  rather than the control reference itself — is cached when a control event is sent recursively.

[The test case has been updated to reveal the issue.](https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3472/commits/46430da32aba3da605c829f4619ea4dbacf56072#diff-14b588001e801f83b72681a6117e4f75R144) When running the test case on `master`:
```swift
error: -[ReactiveCocoaTests.UITextFieldSpec should_not_deadlock_when_the_text_field_is_asked_to_resign_first_responder_by_any_of_its_observers]
expected to equal <[1, 2, 2]>, got <[1, 2, 3]>
```

`CocoaTarget` should have cached the value when the event is sent. But due to the current implementation of convenience signals, `CocoaTarget` captures the reference instead, and the correct value is overwritten by a subsequent mutation before `CocoaTarget` dequeues the reference and pipes it to `map`.

The transform can be replaced by smart key path in Swift 4.

TODO:
- [x] Test case
- [x] Changelog

### DIFF
--- a/ReactiveCocoa/UIKit/UIControl.swift
+++ b/ReactiveCocoa/UIKit/UIControl.swift
@@ -44,11 +44,23 @@ extension Reactive where Base: UIControl {
 	/// - parameters:
 	///   - controlEvents: The control event mask.
 	///
-	/// - returns: A signal that sends the control each time the control event 
-	///            occurs.
+	/// - returns: A signal that sends the control each time the control event occurs.
 	public func controlEvents(_ controlEvents: UIControlEvents) -> Signal<Base, NoError> {
+		return mapControlEvents(controlEvents, { $0 })
+	}
+
+	/// Create a signal which sends a `value` event for each of the specified
+	/// control events.
+	///
+	/// - parameters:
+	///   - controlEvents: The control event mask.
+	///   - transform: A transform to reduce `Base`.
+	///
+	/// - returns: A signal that sends the reduced value from the control each time the
+	///            control event occurs.
+	public func mapControlEvents<Value>(_ controlEvents: UIControlEvents, _ transform: @escaping (Base) -> Value) -> Signal<Value, NoError> {
 		return Signal { observer in
-			let receiver = CocoaTarget(observer) { $0 as! Base }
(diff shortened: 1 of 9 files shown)


ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -44,11 +44,23 @@ extension Reactive where Base: UIControl {
...
-	/// - returns: A signal that sends the control each time the control event 
-	///            occurs.
+	/// - returns: A signal that sends the control each time the control event occurs.
 	public func controlEvents(_ controlEvents: UIControlEvents) -> Signal<Base, NoError> {
+		return mapControlEvents(controlEvents, { $0 })
+	}
+
+	/// Create a signal which sends a `value` event for each of the specified
+	/// control events.
+	///
```
mdiep: Can we add a `- note` or `- important` here that explains why this is needed? (i.e. that events can be sent recursively, so you may want to transform if…)

ReactiveCocoa/UIKit/UIControl.swift
```diff
@@ -44,11 +44,23 @@ extension Reactive where Base: UIControl {
 	/// - parameters:
 	///   - controlEvents: The control event mask.
 	///
-	/// - returns: A signal that sends the control each time the control event 
-	///            occurs.
+	/// - returns: A signal that sends the control each time the control event occurs.
 	public func controlEvents(_ controlEvents: UIControlEvents) -> Signal<Base, NoError> {
```
mdiep: Should we deprecate this in favor of `mapControlEvents`?

andersio: There are control events that are not value affecting though, e.g. `editingDidBegin`.

ReactiveCocoa/UIKit/UIControl.swift
andersio: Done.

@mdiep

PR #3475, opened 2017-06-09
Update ReactiveSwift to 2.0.0 alpha 2, Result to 3.2.3, and Nimble to 7.0.1.

andersio:
Tested locally with Xcode 8.3.3 and Xcode 9.0 beta 1.

UITextField deallocation test is disabled due to the iOS 11 UIKit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ matrix:
         - JOB=CARTHAGE-watchOS
     - script:
       - gem install cocoapods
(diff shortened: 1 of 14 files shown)


mdiep:
It looks like the CocoaPods build may be broken. 😞

andersio:
Argh, forgot to update the podspec.

andersio:
Need to wait for alpha 2 to be available on Pod Trunk.

PR #3476, opened 2017-06-11
Update the release roadmap.

andersio:
@mdiep I wonder if we need to bump the major version up when we move to Swift 4.0 in the fall (after 6.0). It doesn't seem necessary since I do not expect the API to break in Swift 3.2 mode, other than associated type constraints that render existing external constraints redundant. But Swift 3.1 projects that specify at that point only `~> 6.0` but not the last minor 3.1 compliant release would break not because of the API but the tools (which is not addressed by the semver spec). 🤔 

@ReactiveCocoa/reactivecocoa

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,23 +135,20 @@ If you need any help, please visit our [GitHub issues][] or [Stack Overflow][]. 
 ## Release Roadmap
 **Current Stable Release:**<br />[![GitHub release](https://img.shields.io/github/release/ReactiveCocoa/ReactiveCocoa.svg)](https://github.com/ReactiveCocoa/ReactiveCocoa/releases)
 
+### In Development
 #### ReactiveCocoa 6.0
-It targets Swift 3.1.x and ReactiveSwift 2.0. The estimated schedule is Spring 2017.
+It targets Swift 3.1 and ReactiveSwift 2.0. The estimated schedule is July 2017.
 
-Since ReactiveSwift 2.0 would contain breaking changes, the public API of ReactiveCocoa is considered breaking too.
+ReactiveCocoa 6.0 has continued the work in polishing the APIs, preparing for the eventual API freeze for library resilience.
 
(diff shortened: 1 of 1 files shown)


liscio:
My vote is to say, "better safe than sorry," and move the major version up with Swift 4.

Maybe 6.0 is "Compiled as Swift 3.2, introduces keyPath and whatever other funky features we can adopt with minimal change to existing code."

7.0 would be "Compiled as Swift 4, contains all the intense 4.0-related cleanup, and adds whatever other new stuff we can dream up later on."

That way you can move forward on "Swift 4 things" with less baggage. Folks that are tied to Swift 3.2 can continue using the pre-6.0 release in case they can't link/compile against something for some crazy reason you didn't anticipate.

Thoughts?

andersio:
That seems a sensible approach, though with conditional conformances landing in possibly Swift 4.x it would mean yet another bump in major version in less than 12 months. So we are riding the Chrome route, aren't we?

🙈

mdiep:
It's not quite the Chrome route.

My preference is to bump major versions only when we're ready to _drop_ support for a Swift version.

- RAC 6 supports Swift 3.2 _and_ Swift 4. Major new advantages from Swift 3.2 can be adopted. Swift 4 support doesn't take advantage of new features.

- RAC 7 drops support for Swift 3.2, adopts Swift 4 features, and would support at least Swift 4.1 as well.

This makes it so our users can upgrade Swift versions ASAP. I wouldn't want to hold up Swift 4 support on significant internal refactoring, for instance.

So we'd ship RAC 6 with Swift 3.2, support Swift 4. If Swift 4 ships conditional conformances, we can adopt them on master and ship as RAC 7. If Swift 4 only had something like KeyPaths, then I'd say we should use `#if`s to support those additive features without dropping support.

andersio:
Conditional conformance is not a part of 4.0 btw, just a maybe for a 4.x.

PR #3477, opened 2017-06-11
`reactive(_:)` binding targets.

andersio:
`reactive(_:)` is available on every reference type that is `ReactiveExtensionsProvider`. By default it pipes values on `UIScheduler` since it is otherwise insane to manually overload `reactive(_:)` for all AppKit and UIKit classes.

There are ineligible binding targets though, specifically those which wrap a method call.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "ReactiveCocoa/ReactiveSwift" ~> 1.1
+github "ReactiveCocoa/ReactiveSwift" "2.0.0-alpha.2"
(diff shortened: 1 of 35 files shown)


andersio:
This is a prototype for key path support in RAC. 

Related: https://github.com/ReactiveCocoa/ReactiveSwift/issues/331
@ReactiveCocoa/reactivecocoa

mdiep:
I think I'm 👎 on this change. Specifically, I think we should maintain the named binding targets. That makes the API nicer IMO.

Maybe we could add a subscript to `Reactive` that takes a KeyPath as a first step? We could even use it to implement the named properties. But I don't think we should commit to KeyPaths as the sole API right now. Maybe in a few months we'll find that's a good idea, but I think we should be a bit more conservative here.

PR #3478, opened 2017-06-12
Make a KeyValueObservingSpec test work with any iteration count

gparker42:


andersio:
Thank you for catching it. 😀

PR #3480, opened 2017-06-16
UIButton.reactive.pressed doesn't work on tvOS

andrei-kuzma:
The reason is that .touchUpInside event is used instead of .primaryActionTriggered

### DIFF
--- a/ReactiveCocoa/UIKit/UIButton.swift
+++ b/ReactiveCocoa/UIKit/UIButton.swift
@@ -8,15 +8,23 @@ extension Reactive where Base: UIButton {
 		get {
 			return associatedAction.withValue { info in
 				return info.flatMap { info in
-					return info.controlEvents == .touchUpInside ? info.action : nil
+					return info.controlEvents == pressEvent ? info.action : nil
 				}
 			}
 		}
 
 		nonmutating set {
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/UIKit/UIButton.swift
```diff
@@ -8,16 +8,24 @@ extension Reactive where Base: UIButton {
...
 		}
 
 		nonmutating set {
-			setAction(newValue, for: .touchUpInside)
+			setAction(newValue, for: pressEvent)
 		}
 	}
 
+    private var pressEvent: UIControlEvents {
+        #if os(tvOS)
```
andersio: Could we have this being `#available(iOS 9.0, tvOS 9.0)` instead?

andersio:
Could you please also update the [UIButtonSpec](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoaTests/UIKit/UIButtonSpec.swift#L43)? The test case for `primaryActionTriggered` should be conditionally enabled for iOS 9.0+ and tvOS 9.0+. Thanks! 😸

andrei-kuzma:
@andersio 
Updated.

PR #3484, opened 2017-07-03
Replace NSObject with NSObjectProtocol in extensions

nickdomenicali:
I did this after wanting to grab the lifetime from a protocol that conforms to `NSObjectProtocol` and couldn't.

There are a couple other `NSObject` extensions this doesn't include but that's mostly because I'm new to this code base and didn't want to mess anything up without first getting feedback.

### DIFF
--- a/ReactiveCocoa/NSObject+Association.swift
+++ b/ReactiveCocoa/NSObject+Association.swift
@@ -48,7 +48,7 @@ internal struct Associations<Base: AnyObject> {
 	}
 }
 
-extension Reactive where Base: NSObject {
+extension Reactive where Base: NSObjectProtocol {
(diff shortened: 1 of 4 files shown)


ReactiveCocoa/NSObject+Association.swift
```diff
@@ -72,8 +72,8 @@ extension Reactive where Base: NSObject {
 	}
 }
 
-extension NSObject {
-	@nonobjc internal var associations: Associations<NSObject> {
+extension NSObjectProtocol {
+	@nonobjc internal var associations: Associations<NSObjectProtocol> {
```
sharplet: This should probably be `Associations<Self>`

PR #3491, opened 2017-07-09
KVO with Smart Key Path.

andersio:
Optionals are special cased, because force casting does not open a generic parameter for optional inspection. So even if `U` turns out to an optional, coercing `Any?` to `U` fails.

#### Checklist
- [x] Updated CHANGELOG.md.


ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -55,6 +55,113 @@ extension Reactive where Base: NSObject {
 			return disposable
 		}
 	}
+
+	#if swift(>=3.2)
+	private func producer<U>(
+		for keyPath: KeyPath<Base, U>,
+		action: @escaping (Signal<U, NoError>.Observer) -> (Any?) -> Void
```
mdiep: Since we're always calling `observer.send(value: $0 as X)`, could we change this to be more like a `map` instead? Transform the value (`(Any?) -> U`) and then send it on the observer?

ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -55,6 +55,105 @@ extension Reactive where Base: NSObject {
...
+
+	#if swift(>=3.2)
+	private func producer<U>(
+		for keyPath: KeyPath<Base, U>,
+		transform: @escaping (Any?) -> U
+	) -> SignalProducer<U, NoError> {
+		return SignalProducer { observer, lifetime in
+			let disposable = KeyValueObserver.observe(
+				self.base,
+				keyPath: keyPath._kvcKeyPathString!,
```
mdiep: I'm not comfortable with the use of `_kvcKeyPathString` here.

Besides the fact that it's underscore-prefixed to indicate that it's somewhat private, this is `nil` for some some `KeyPath`s—even those on `NSObject`s. The property itself needs to be `dynamic` in order for this to be non-`nil`.

We need to use the built-in `KeyPath` KVO API.

andersio: We can wrap the official closure-based KVO API, but it does not support observing weak properties (and properties of these) as the RAC version does.

andersio: Either way I don't have a strong opinion, since I haven't been using KVO for a while. Even when I am using it, I don't recall any need of observing weak properties.

liscio: FWIW, I have some observations on weak properties all over the place.

Wouldn't it make sense to add a variant that uses `ReferenceWritableKeyPath` for those that want to support weak references? That'll allow us to use the "official" API.

andersio: The actual problem is that the official one has not supported emitting `nil` when weak references are zeroed. I am not aware that the mechanics have now accommodated this, and the "new" KVO API is just a wrapper around what we have used for ages.

liscio: > The actual problem is that the official one has not supported emitting nil when weak references are zeroed.

@andersio I just verified that this is the case in a playground. Yuck!

iv-mexx: > Besides the fact that it's underscore-prefixed to indicate that it's somewhat private, this is nil for some some KeyPaths—even those on NSObjects. The property itself needs to be dynamic in order for this to be non-nil.

@mdiep I was not aware of that until I tried it yesterday, but the same is already required even in RAC with the current String based `DynamicProperty`.

Apparently KVO in Swift as a whole only works for properties that are `@dynamic`, no matter if you use String based `KeyPath`s or the new smart `KeyPaths`.

```
class NSPerson: NSObject {
    @objc dynamic var firstName: String = ""
    @objc var lastName: String = ""
}
let person = NSPerson()


person.observe(\NSPerson.firstName) { _, _ in print("KVO: \(person.firstName)")}
person.observe(\NSPerson.lastName) { _, _ in print("KVO: \(person.lastName)")}

let firstName = DynamicProperty<String>(object: person, keyPath: "firstName")
let lastName = DynamicProperty<String>(object: person, keyPath: "lastName")

firstName.signal.observeValues { print("DynamicProperty: \($0)")}
lastName.signal.observeValues { print("DynamicProperty: \($0)")}

person.firstName = "John"
person.lastName = "Doe"
```

This prints the following:

```
DynamicProperty: John
KVO: John
```

So neither plain Swift 4 Smart Key Path KVO nor the existing String based RAC DynamicProperty work for properties  that are missing the `dynamic` even if they are `@objc`

ianbytchek:
> @mdiep (But feel free to argue for it there!)

My understanding is that this concerns mostly new users who are struggling with the framework alone and might not know such caveats. Or maybe not. Anyway, this is not any different than using the existing sting KVO:

- Unless the property is `dynamic` you can observe all you want, it won't work. Users just have to know this, whether they learn it the easy or the hard way…
- To me the argument fades in the light that the current `keyPath` is a `String` type. Yes, it is intended to be used with the `#keyPath()` but that doesn't change the fact that it literally accepts anything, not only properties that are not `dynamic`, but ones that are not `@objc` and ones that don't exist at all…

My personal opinion of a fanatical ReactiveSwift & Cocoa user is divided. I always get a sense that this project is trying to be safe than sorry too much. On one hand this is the only right choice, especially for the project of that significance. On another hand it just feels punishing – I want it to embrace all the cutting edge features and not to maintain a bunch of shims within every own project.

The obvious concern with private `_`-prefixed public API is that it won't be there tomorrow. Have you tried compiling a large project that you haven't open for a year or two? 😄 At best you'll get a ton of warnings, at worst you'll need to update frameworks and refactor your codebase, and at very worst you'll have to find the old-enough system that's still capable building it. There's a great chance that the undocumented API will be replaced with a better public variant in the next version. And even if the disaster strikes, with public pre-releases and huge community the workaround will be PR-ed well before the next Swift version is out of beta… or all of us will just have to evolve our projects if we want to use the new version of Swift. We all did 4 times already.

Even Apple/Swift use tricks and hacks, and it servers everyone well. Yes, in theory things might go wrong, but in practice the benefits are worth the risk and there's nothing wrong with taking it. IMHO.

andersio:
> The obvious concern with private _-prefixed public API is that it won't be there tomorrow.

I don't think this is a high priority concern since this would become part of the Swift runtime ABI for Darwin, and would be maintained until the Foundation overlay decides to use something else. The only concern is whether the behaviour would change, but I don't see strong evidence that it would swing in a completely different direction.

ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -53,6 +53,107 @@ extension Reactive where Base: NSObject {
 			signalLifetime += lifetime.observeEnded(observer.sendCompleted)
 		}
 	}
+
+	#if swift(>=3.2)
```
mdiep: Do we still support anything before this?

ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -53,6 +53,107 @@ extension Reactive where Base: NSObject {
...
+	private func producer<U>(
+		for keyPath: KeyPath<Base, U>,
+		transform: @escaping (Any?) -> U
+	) -> SignalProducer<U, NoError> {
+		guard let kvcKeyPath = keyPath._kvcKeyPathString else {
+			fatalError("Cannot use `signal(for:)` on a non Objective-C property.")
```
mdiep: This should say `producer(for:)`.

ReactiveCocoa/NSObject+KeyValueObserving.swift
andersio: No.

ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -53,6 +53,105 @@ extension Reactive where Base: NSObject {
...
+	///	  Does not send the initial value. See `producer(forKeyPath:)`.
+	///
+	/// - parameters:
+	///   - keyPath: The key path of the property to be observed.
+	///
+	/// - returns: A producer emitting values of the property specified by the
+	///            key path.
+	public func signal<U>(for keyPath: KeyPath<Base, U?>) -> Signal<U?, NoError> {
+		return signal(for: keyPath) { $0 as! U? }
+	}
```
mdiep: Why are these overloads needed? Can't optionals use the below variants? Won't these introduce ambiguities?

mdiep:
Let's try to get this into the upcoming release as well.

ReactiveCocoa/NSObject+KeyValueObserving.swift
andersio: > Optionals are special cased, because force casting does not open a generic parameter for optional inspection. So even if U turns out to an optional, coercing Any? to U fails.

Not sure if the runtime casting machinery has changed lately.

PR #3498, opened 2017-08-09
[Xcode 9 beta 5] Fixed an issue causing infinite recursion in the Swift runtime.

andersio:
#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveCocoa/ObjC+RuntimeSubclassing.swift
+++ b/ReactiveCocoa/ObjC+RuntimeSubclassing.swift
@@ -110,7 +110,7 @@ private func subclassName(of class: AnyClass) -> String {
 ///   - class: The class to swizzle.
 ///   - perceivedClass: The class to be reported by the methods.
 private func replaceGetClass(in class: AnyClass, decoy perceivedClass: AnyClass) {
-	let getClass: @convention(block) (Any) -> AnyClass = { _ in
+	let getClass: @convention(block) (UnsafeRawPointer?) -> AnyClass = { _ in
 		return perceivedClass
 	}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ matrix:
     - script:
       - gem install cocoapods
       - pod repo update
-      - pod lib lint ReactiveCocoa.podspec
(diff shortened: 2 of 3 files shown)


ReactiveCocoa/ObjC+RuntimeSubclassing.swift
```diff
@@ -110,7 +110,7 @@ private func subclassName(of class: AnyClass) -> String {
 ///   - class: The class to swizzle.
 ///   - perceivedClass: The class to be reported by the methods.
 private func replaceGetClass(in class: AnyClass, decoy perceivedClass: AnyClass) {
-	let getClass: @convention(block) (Any) -> AnyClass = { _ in
```
mdiep: Why does this cause infinite recursion? 😕

andersio: The stack is overflown by infinite recursion of `bridgeAnyObjectToAny`. We shouldn't have used `Any` here anyway.

andersio: screen shot 2017-08-10 at 1 54 52 am

Presumably the bridging call relies on `-class` and `+class`, which in turns require bridging due to the use of `Any`.

liscio:
This one bit me pretty badly recently during development. Any idea what is causing the CI build to fail? I know nothing about CocoaPods, myself…

ikesyo:
@liscio @andersio 

```
$ pod lib lint ReactiveCocoa.podspec
 -> ReactiveCocoa (6.0.0)
    - WARN  | swift_version: The validator for Swift projects uses Swift 3.0 by default, if you are using a different version of swift you can use a `.swift-version` file to set the version for your Pod. For example to use Swift 2.3, run: 
    `echo "2.3" > .swift-version`
[!] ReactiveCocoa did not pass validation, due to 1 warning (but you can use `--allow-warnings` to ignore it).
You can use the `--no-clean` option to inspect any issue.
The command "pod lib lint ReactiveCocoa.podspec" exited with 1.
```

We should pass `--swift-version=3.0` to `pod lib lint ReactiveCocoa.podspec` to silence the warning.

PR #3504, opened 2017-08-27
Added cancelButtonClicked signal to UISearchBar

albsala:
Since there is a signal for UISearchBar.continuousTextValues which can be binded to a filter, I think there should be a signal for the cancel button so such filter can be cancelled.

This is my first contribution so I hope my request follows the way you work.

#### Checklist
- [X] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveCocoa/UIKit/iOS/UISearchBar.swift
+++ b/ReactiveCocoa/UIKit/iOS/UISearchBar.swift
@@ -48,7 +48,6 @@ extension Reactive where Base: UISearchBar {
 	/// A void signal emitted by the search bar upon any click on the cancel button
 	public var cancelButtonClicked: Signal<Void, NoError> {
 		return proxy.intercept(#selector(UISearchBarDelegate.searchBarCancelButtonClicked))
-			.map { () -> Void in }
 	}
 	
 }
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/UIKit/iOS/UISearchBar.swift
```diff
@@ -41,4 +45,10 @@ extension Reactive where Base: UISearchBar {
 			.map { [unowned base] in base.text }
 	}
 	
+	/// A void signal emitted by the search bar upon any click on the cancel button
+	public var cancelButtonClicked: Signal<Void, NoError> {
+		return proxy.intercept(#selector(UISearchBarDelegate.searchBarCancelButtonClicked))
+			.map { () -> Void in }
```
mdiep: This `map` isn't needed. `Void` and `()` are synonymous, so this isn't actually changing the value. Can you remove it?

albsala:
I was testing different things and I guess I left that unnecessary map, I'm sorry. It's removed now.

BTW quick question: I use that kind map when I need to bind two components with different value types and the value is not needed (e.g. Signal<String, Error> with an Action<Void, Void, Error>). Is there a cleaner way to do that? Thank you!

mdiep:
This is great! Thanks!

> BTW quick question: I use that kind map when I need to bind two components with different value types and the value is not needed (e.g. Signal<String, Error> with an Action<Void, Void, Error>). Is there a cleaner way to do that? Thank you!

Nope. `.map { _ in }` is the way to go.

PR #3186, opened 2016-09-14
Remove warning on XCode8 for #3183

yageek:
The option `receiver-is-weak` seems to be ignored according to the llvm
mailing list:
http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20160418/156202.html.

PR #3506, opened 2017-08-29
Remove warning on Xcode 9 (copy #3186)

vox-humana:


ikesyo:
The branch is not under maintenance anymore. Could you open a pull request at https://github.com/ReactiveCocoa/ReactiveObjC?

vox-humana:
ReactiveObjC already has such fix since Xcode 8 betas. But I'm not sure whether ReactiveObjC support iOS 7 like original ReactiveCocoa in 2.5 version? Yes, we still have to support iOS 7 😞
Thanks

PR #3509, opened 2017-09-15
iOS + tvOS: disable code coverage to fix app submission with Xcode 9

NachoSoto:
See https://github.com/Carthage/Carthage/issues/2056.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please put new entries at the top.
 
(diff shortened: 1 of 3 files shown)


andersio:
What about watchOS and macOS?

NachoSoto:
@andersio they already had it disabled. My guess is that at some point Apple changed the default value for that setting to `false`, so newer schemes were correct, but `iOS` and `Mac`, being older, had it set to `true`.

Issue #3525, opened 2017-10-03
Swift 4 migration issue: Nil is not compatible with expected argument type 'IMP' (aka 'OpaquePointer')


joelorenz20:
Hi

I am trying to migrate to swift 4 but I am caught up with these error messages.
Any idea how I can fix this?

Thanks

<img width="628" alt="screen shot 2017-10-03 at 8 42 31 pm" src="https://user-images.githubusercontent.com/11159818/31125671-a90f8c6c-a87b-11e7-9a49-e94e3f662703.png">

antonincharvat:
Solving the same issue few days ago. Apparently there are still some issues with RAC and Swift 4. We have figured out a temporary solution. Check out my thread: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3519

andersio:
ReactiveCocoa uses Swift 3.2, and apparently CocoaPods has problems in supporting mixed source use cases. As mentioned by @antonincharvat there is an intermediate solution linked in #3519.

joelorenz20:
Fix in [#3519](https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3519) seems to solve this issue. 

Thanks @antonincharvat & @andersio

alper:
Ah, Swift 4 support means that it works _with_ swift4 project, not that the project itself has migrated to swift4!? Now I get it.

andersio:
@alper Yes. Swift 3.2 frameworks work with anything that is built by the same compiler regardless of the language modes used, and the interoperability does work well for Carthage and Xcode project users.

It is just CocoaPods who hasn’t managed to support it properly.

Issue #3519, opened 2017-09-21
Can't compile RAC


antonincharvat:
Xcode 9 + Swift 4 project.
Tried all last releases of RAC, none of them works.
Can't even compile the project because of errors in RAC. Getting following errors in the pod:
![snimek obrazovky 2017-09-21 v 13 06 58](https://user-images.githubusercontent.com/19424035/30693081-56af315e-9ece-11e7-84b7-5d94920d270c.png)

andersio:
6.0.2 and 6.1.0 alpha 2 are supposed to fix this, and I heard that it works. Have you tried these versions?

stevegomez17:
NSObject+intercepting+Swift

antonincharvat:
Yes, I tried both of them, I clean the project, delete derived data, but the errors keep showing up.
![snimek obrazovky 2017-09-21 v 16 31 43](https://user-images.githubusercontent.com/19424035/30701453-98055626-9eea-11e7-90ab-cfb1c9fb00a8.png)

alexpopov:
Similar issues but I have more errors. ReactiveCocoa 6.0.2 through Cocoapods 
<img width="80" alt="Cocoapods errors" src="https://user-images.githubusercontent.com/903856/30720369-ff369260-9edb-11e7-8b4f-a61b023f76e1.png">

andersio:
I will investigate. In the meantime, you may override the Swift version for RAC in your Podfile.
https://github.com/CocoaPods/CocoaPods/issues/6791

antonincharvat:
Apparently this temporary solution with overriding Swift version to 3.2 for RAC is working, thank you. I will be regularly checking new releases of RAC to have working Swift 4 version asap.

andersio:
7.0.0 alpha 2 should now work with both Swift 3.2 and Swift 4.0. Feel free to report if you still encounter any issue.

PR #3526, opened 2017-10-04
Swift 4 support + Update the Travis CI config.

andersio:
There are a lot of force unwrapping in Objective-C runtime API call sites. A majority of them is previously IUO.

While ideally we can wipe 3.2 implementations away, #3519 and #3525 are motivations to keep them, at least until CocoaPods does their work. 

#### Checklist
- [x] Updated CHANGELOG.md.


andersio:
Ping @ReactiveCocoa/reactivecocoa

liscio:
Thanks for undertaking this, @andersio. Looks like it's not as much of a huge transition as I feared it'd be (with all the API changes on macOS, specifically.)

I'm left with a question, though—why not set SWIFT_VERSION = 4.0 in the `.xcodeproj` file along with these changes? Is it just a band-aid for supporting the language until the `${RAC_MAJOR_VERSION}` that is officially updated for Swift 4 and beyond?

andersio:
@liscio Good point. Setting it to 4.0 should do no harm.

andersio:
Argh. SWIFT_VERSION=4.0 is not working for Xcode 8.3.

liscio:
@andersio Argh. This further reinforces my frustration in Xcode's inability to override build settings at the workspace level… 😠

andersio:
@liscio We can drop Xcode 8 support though. Apparently there is no point to continue Xcode 8 support at all.

.travis.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - gem install cocoapods
       - pod repo update
       - pod lib lint ReactiveCocoa.podspec
-      - pod lib lint ReactiveMapKit.podspec
+      - pod lib lint ReactiveMapKit.podspec --swift-version=3.2
```
andersio: This can be removed once the Swift 4 code gets into a release.

PR #3542, opened 2017-11-17
Add tintColor BindingTarget for UIView and UIButton

iv-mexx:
#### Checklist
- [x] Updated CHANGELOG.md.


ReactiveCocoa/UIKit/UIButton.swift
```diff
@@ -41,6 +41,7 @@ extension Reactive where Base: UIButton {
 		return makeBindingTarget { $0.setImage($1, for: state) }
 	}
 
+	/// Sets the image of the button for the .normal state
```
iv-mexx: For a short moment I thought `tintColor` needs to be added to `UIButton` separately, and in the process I saw that this comment was missing. Hope its ok to add that in this commit as well

PR #3546, opened 2017-11-20
Implement Property + SmartKeyPath

iv-mexx:
#### Checklist
- [x] Updated CHANGELOG.md.

In #440, Smart Key Path support was added for `BindingTarget`, but it was not yet possible to use Smart Key Paths as `BindingSource`. 

This PR now adds 
* `NSObject.reactive.producer(for:keyPath:)` with `KeyPath`
* `NSObject.reactive.signal(for:keyPath:)` with `KeyPath`
* `Property(object:keyPath:)` with `KeyPath`

I've left the "old" methods with the same signature (only `String` instead of `KeyPath` for `keyPath`) alone - should we replace them? (I would say "Yes").

Also, I've not yet touched `DynamicProperty` because it would be either necessary to create a whole new class that uses `KeyPath` instead of `String`, or to change the existing one, and I again wanted to ask it we should keep the Version that takes `String` or replace it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
 # master
 *Please put new entries at the top.
 
-1. Added `tintColor` binding target to `UIView`. (#3542)
+1. Added `tintColor` binding target to `UIView`. (#3542, kudos to @iv-mexx)
+1. Smart Key Paths for `Property`. (#3546, kudos to @iv-mexx)
+	
+	Example: 
+	```swift
+	let title = Property(object: viewController, keyPath: \UIViewController.title)
+	```
+
 
 # 7.0.0
 1. Update ReactiveSwift to 3.0.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,14 @@
 # master
 *Please put new entries at the top.
 
-1. Added `tintColor` binding target to `UIView`. (#3542)
+1. Added `tintColor` binding target to `UIView`. (#3542, kudos to @iv-mexx)
```
iv-mexx: I did not attribute myself in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3542 and added this here

andersio:
#3491 did the work. The question here is that we have been supporting weak key path observation since the ObjC era, and the “new” strong typed API (still) doesn’t according to @liscio.

iv-mexx:
Oh no, somehow I missed that open PR

PR #3548, opened 2017-11-21
DynamicProperty of optional property

### LINKED ISSUE #3547: DynamicProperty of optional property: Fatal error on force unwrap
Consider the following snippet:

```
class MyObject: NSObject {
    @objc dynamic var name: String?

    init(name: String) {
        self.name = name
    }
}

let object = MyObject(name: "Joe")

let dynamicProperty = DynamicProperty<String?>(object: object, keyPath: "name")
dynamicProperty.producer.startWithValues { print("Dynamic: \($0)")}
object.name = "Jack"
object.name = nil
```

When setting the property to nil, the following crash occurs 
> Fatal error: Unexpectedly found nil while unwrapping an Optional value

in [NSObject+KeyValueObserving.swift:79](https://github.com/ReactiveCocoa/ReactiveCocoa/blob/master/ReactiveCocoa/NSObject%2BKeyValueObserving.swift#L79)

```
init(object: NSObject, keyPath: String) {
  self.producer = object.reactive.producer(forKeyPath: keyPath).map { $0 as! Value }
}
```

This was introduced in #3461 that was intended to strip the optionality of DynamicProperty - but what about DynamicProperties that are Optional to begin with?

iv-mexx:
#### Checklist
- [x] Updated CHANGELOG.md.

Adds necessary overloads to `DynamicProperty` and `Property` to enable the usage with optional `Value`. 

Refs #3547.

### DIFF
--- a/ReactiveCocoa/DynamicProperty.swift
+++ b/ReactiveCocoa/DynamicProperty.swift
@@ -11,12 +11,13 @@ public final class DynamicProperty<Value>: MutablePropertyProtocol {
 	private weak var object: NSObject?
 	private let keyPath: String
 	private let cache: Property<Value>
+	private let transform: (Value) -> Any?
 
 	/// The current value of the property, as read and written using Key-Value
 	/// Coding.
 	public var value: Value {
 		get { return cache.value }
-		set { object?.setValue(newValue, forKeyPath: keyPath) }
+		set { object?.setValue(transform(newValue), forKeyPath: keyPath) }
 	}
 
 	/// The lifetime of the property.
@@ -45,10 +46,11 @@ public final class DynamicProperty<Value>: MutablePropertyProtocol {
 		return cache.signal
 	}
 
-	internal init(object: NSObject, keyPath: String, cache: Property<Value>) {
+	internal init(object: NSObject, keyPath: String, cache: Property<Value>, transform: @escaping (Value) -> Any?) {
 		self.object = object
 		self.keyPath = keyPath
 		self.cache = cache
+		self.transform = transform
 	}
 	
 	/// Create a typed mutable view to the given key path of the given Objective-C object.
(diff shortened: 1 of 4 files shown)


ReactiveCocoa/DynamicProperty.swift
```diff
@@ -45,16 +45,42 @@ public final class DynamicProperty<Value>: MutablePropertyProtocol {
...
+	public convenience init(object: NSObject, keyPath: String) {
+		self.init(object: object, keyPath: keyPath, cache: Property(object: object, keyPath: keyPath))
+	}
+}
+
+extension DynamicProperty where Value: OptionalProtocol {
+	
+	/// The current value of the property, as read and written using Key-Value
+	/// Coding.
+	public var value: Value {
```
andersio: Not quite sure why this needs to be overriden.

ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -55,6 +55,34 @@ extension Reactive where Base: NSObject {
...
+	///   - keyPath: The key path to observe.
+	public convenience init(object: NSObject, keyPath: String) {
+		// `Property(_:)` caches the latest value of the `DynamicProperty`, so it is
+		// saved to be used even after `object` deinitializes.
+		self.init(UnsafeOptionalKVOProperty(object: object, keyPath: keyPath))
+	}
+	
+	// `Property(unsafeProducer:)` is private to ReactiveSwift. So the fact that
+	// `Property(_:)` uses only the producer is explioted here to achieve the same effect.
+	private final class UnsafeOptionalKVOProperty: PropertyProtocol {
```
andersio: `UnsafeKVOProperty` can be shared, since the only thing different is the transform.

ReactiveCocoa/DynamicProperty.swift
iv-mexx: I don't know why it happens, but otherwise the `should write the underlying object` Testcase fails with:

> expected to be nil, got <<null>>

andersio: I am a bit skeptical if this overload is called in a generic context (`MutablePropertyProtocol` specifically) though.

ReactiveCocoa/NSObject+KeyValueObserving.swift
iv-mexx: Alright, I've changed that

ReactiveCocoa/DynamicProperty.swift
iv-mexx: Is ae79efb what you meant?

andersio: Yep. Looks nice.

PR #3557, opened 2017-12-28
Allow NSObject.reactive.signal to also intercept and return the 'return' value of the executed selector.

mishagray:
Added the ```includeReturnValue``` to reactive.signal(for selector: Selector)

NSObject.reactive.signal(for selector: Selector, includeReturnValue: Bool = false)

Will add a return value to the end of the array, for each value of the Signal.
If the selector does not return a value, than the flag is ignored and no value is appended to the array.

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveCocoa/NSObject+Intercepting.swift
+++ b/ReactiveCocoa/NSObject+Intercepting.swift
@@ -33,7 +33,7 @@ extension Reactive where Base: NSObject {
 
 	/// Create a signal which sends a `next` event, containing an array of 
 	/// bridged arguments, at the end of every invocation of `selector` on the 
-	/// object.
+	/// object.  if includeReturnValue == true, then the last element of the array will include any return value of the invocation, if the 'selector' returns a value.  No value is appended if the method returns 'Void'.
 	///
 	/// It completes when the object deinitializes.
 	///
@@ -42,10 +42,12 @@ extension Reactive where Base: NSObject {
 	///
 	/// - parameters:
 	///   - selector: The selector to observe.
-	///
+	///   - includeReturnValue:  the array will include the value returned by the invocation.  Ignored if the selector returns Void.
 	/// - returns: A signal that sends an array of bridged arguments.
-	public func signal(for selector: Selector) -> Signal<[Any?], NoError> {
-		return base.intercept(selector).map(unpackInvocation)
+	public func signal(for selector: Selector, includeReturnValue: Bool = false) -> Signal<[Any?], NoError> {
+		return base.intercept(selector).map {
+			unpackInvocation($0, includeReturnValue: includeReturnValue)
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


andersio:
It looks great with a quick skim. I will take a second look ASAP.

In the meantime, could you please clean up the indentation, the excess blank lines & break the long lines? The GitHub shows quite a few nits.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -42,10 +45,12 @@ extension Reactive where Base: NSObject {
 	///
 	/// - parameters:
 	///   - selector: The selector to observe.
-	///
+	///   - includeReturnValue:  the array will include the value returned by the invocation.  Ignored if the selector returns Void.
 	/// - returns: A signal that sends an array of bridged arguments.
-	public func signal(for selector: Selector) -> Signal<[Any?], NoError> {
-		return base.intercept(selector).map(unpackInvocation)
+	public func signal(for selector: Selector, includeReturnValue: Bool = false) -> Signal<[Any?], NoError> {
```
ikesyo: This is API-compatible but binary-incompatible change. So the original API should be kept as is and the new API should be added as a separete one.

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -45,7 +45,31 @@ extension Reactive where Base: NSObject {
...
+	/// It completes when the object deinitializes.
+	///
+	/// - note: Observers to the resulting signal should not call the method
+	///         specified by the selector.
+	///
+	/// - parameters:
+	///   - selector: The selector to observe.
+	///   - includeReturnValue:  the array will include the value returned by the invocation.  Ignored if the selector returns Void.
+	/// - returns: A signal that sends an array of bridged arguments.
+	public func signal(for selector: Selector, includeReturnValue: Bool) -> Signal<[Any?], NoError> {
```
mdiep: This seems to me the wrong way to do this. Instead, it would be better to have a `struct` that tied together the parameters and return value.

```swift
struct Invocation {
    var parameters: [Any?]
    var returnValue: Any?
}
```

ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -456,5 +480,73 @@ private func unpackInvocation(_ invocation: AnyObject) -> [Any?] {
...
+			value = extractRet((Selector?).self)
+		case .undefined:
+			var size = 0, alignment = 0
+			NSGetSizeAndAlignment(retRawEncoding, &size, &alignment)
+			let buffer = UnsafeMutableRawPointer.allocate(bytes: size, alignedTo: alignment)
+			defer { buffer.deallocate(bytes: size, alignedTo: alignment) }
+			
+			invocation.copyReturnValue(to: buffer)
+			value = NSValue(bytes: buffer, objCType: retRawEncoding)
+		}
```
mdiep: There's code duplication here between this and the parameters. I think they could be combined so that all this logic doesn't need to be repeated.

ReactiveCocoaTests/InterceptingSpec.swift
```diff
@@ -930,4 +990,178 @@ private class InterceptedObject: NSObject {
...
+	var selector: Selector {
+		switch I.self {
+		case is CChar.Type:
+			return #selector(InterceptedObject.testReturnValuesC(arg:))
+		case is CShort.Type:
+			return #selector(InterceptedObject.testReturnValuesS(arg:))
+		case is CInt.Type:
+			return #selector(InterceptedObject.testReturnValuesI(arg:))
+		case is CLong.Type:
+			return #selector(InterceptedObject.testReturnValuesL(arg:))
```
mdiep: I think this logic unnecessarily complicates the test. It would be better to pass in the `Selector` directly if possible.

PR #3565, opened 2018-01-24
Add showsCancelButton, textDidBeginEditing and ...EndEditing to UISearchBar

banjun:
I want to show/hide the cancel button of a UISearchBar when its focus changed, with mapping with some other app specific conditions.
With the missing `textDidBeginEditing` reactive extension, I can use ReactiveCocoa to bind all I need instead of using its delegate.

As UISearchBar is not a UITextField nor UIControl, other common extensions such as `mapControlEvents` cannot be used for that.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please put new entries at the top.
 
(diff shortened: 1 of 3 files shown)


mdiep:
This looks great! Well done! Thank you for writing tests. ✨

PR #3566, opened 2018-01-30
More useful system keyboard event notification observing

ra1028:
This proposal has no breaking changes.

I've wanted to observe the keyboard changes by the event types. like below.

```swift
NotificationCenter.default.reactive.keyboard(.willShow, .willHide)
    .take(duringLifetimeOf: self)
    .observeValues { [unowned self] context in
        //  Do something...

        if context.event == .willShow {
            self.tableView.scrollToNearestSelectedRow(at: .none, animated: true)
        }
}
```

To be honest, I wanted to implement as follows, but it has breaking changes.
What do you think of this?

```swift
public enum KeyboardEvent {
    case .willShow
    ...

    struct Context {
        private let base: [AnyHashable: Any] ...
	public let event: KeyboardEvent ...
	public var beginFrame: CGRect ...
	public var endFrame: CGRect ...
    	public var animationCurve: UIViewAnimationCurve ...
    	public var animationDuration: Double ...
    	@available(iOS 9.0, *) ...
    	public var isLocal: Bool ...
    	fileprivate init(userInfo: [AnyHashable: Any], event: KeyboardEvent) ...
    }
}

extension Reactive where Base: NotificationCenter {
    public func keyboard(_ events: KeyboardEvent...) -> Signal<KeyboardEvent.Context, NoError> {
        return .merge(
            events.map { event in
                notifications(forName: event.notificationName)
                    .map { notification in KeyboardChangeContext(userInfo: notification.userInfo!, event: event) }
            }
        )
    }
}

extension Reactive where Base: NotificationCenter {
    @available(*, deprecated, message="Use .keyboard(.willChangeFrame) instead")
    public var keyboardChange: Signal<KeyboardEvent.Context, NoError> {
        return keyboard(.willChangeFrame)
    }
}
```

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
+++ b/ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
@@ -80,12 +80,7 @@ extension Reactive where Base: NotificationCenter {
 	/// - returns: A `Signal` that emits the context of system keyboard events.
 	public func keyboard(_ first: KeyboardEvent, _ tail: KeyboardEvent...) -> Signal<KeyboardChangeContext, NoError> {
 		let events = [first] + tail
-		return .merge(
-			events.map { event in
-				notifications(forName: event.notificationName)
-					.map { notification in KeyboardChangeContext(userInfo: notification.userInfo!, event: event) }
-			}
-		)
+		return .merge(events.map(_keyboard))
 	}
 	
 	/// Create a `Signal` that notifies whenever the system keyboard announces an
@@ -94,6 +89,11 @@ extension Reactive where Base: NotificationCenter {
 	/// - returns: A `Signal` that emits the context of every change in the
 	///            system keyboard's frame.
 	public var keyboardChange: Signal<KeyboardChangeContext, NoError> {
-		return keyboard(.willChangeFrame)
+		return _keyboard(.willChangeFrame)
+	}
+	
+	private func _keyboard(_ event: KeyboardEvent) -> Signal<KeyboardChangeContext, NoError> {
+		return notifications(forName: event.notificationName)
+			.map { notification in KeyboardChangeContext(userInfo: notification.userInfo!, event: event) }
 	}
 }
--- a/ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
+++ b/ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
@@ -12,7 +12,7 @@ public enum KeyboardEvent {
 	case didChangeFrame
 	
 	/// The name of the notification to observe system keyboard events.
-	var notificationName: Notification.Name {
+	fileprivate var notificationName: Notification.Name {
 		switch self {
 		case .willShow:
 			return .UIKeyboardWillShow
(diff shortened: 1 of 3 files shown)


ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
```diff
@@ -37,19 +68,31 @@ public struct KeyboardChangeContext {
...
+	}
+	
 	/// Create a `Signal` that notifies whenever the system keyboard announces an
 	/// upcoming change in its frame.
 	///
 	/// - returns: A `Signal` that emits the context of every change in the
 	///            system keyboard's frame.
 	public var keyboardChange: Signal<KeyboardChangeContext, NoError> {
-		return notifications(forName: .UIKeyboardWillChangeFrame)
-			.map { notification in KeyboardChangeContext(notification.userInfo!) }
```
mdiep: I think this should continue to use the old code just because it'd likely be more performant.

ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
```diff
@@ -37,19 +68,31 @@ public struct KeyboardChangeContext {
...
 		base = userInfo
+		self.event = event
 	}
 }
 
 extension Reactive where Base: NotificationCenter {
+	/// Create a `Signal` that notifies whenever the system keyboard announces specified events.
+	///
+	/// - returns: A `Signal` that emits the context of system keyboard events.
+	public func keyboard(_ events: KeyboardEvent...) -> Signal<KeyboardChangeContext, NoError> {
```
mdiep: You should probably be required to pass at least one event.

We can do that like so:

```swift
func keyboard(_ first: KeyboardEvent, _ rest: KeyboardEvent…) -> Signal<KeyboardChangeContext, NoError> {
```

mdiep:
This makes sense to me as an additive change. 👍

mdiep:
I'd say lets open an issue for the breaking change that we can file against 8.0.

ra1028:
@mdiep 
I fixed them.

> I'd say lets open an issue for the breaking change that we can file against 8.0.

If this proposal accepted, so I'll consider to open an issue.
Thanks!

ReactiveCocoa/UIKit/iOS/UIKeyboard.swift
```diff
@@ -37,19 +68,43 @@ public struct KeyboardChangeContext {
...
+	
+	/// Create a `Signal` that notifies whenever the system keyboard announces specified events.
+	///
+	/// - parameters:
+	///   - first: First type of system keyboard event to observe.
+	///   - second: Second type of system keyboard event to observe.
+	///   - tail: Rest of the types of system keyboard events to observe.
+	///
+	/// - returns: A `Signal` that emits the context of system keyboard events.
+	public func keyboard(_ first: KeyboardEvent, _ second: KeyboardEvent, _ tail: KeyboardEvent...) -> Signal<KeyboardChangeContext, NoError> {
```
andersio: Not sure why we would need two code paths. Variadic supports 1 argument.

ra1028: @andersio
Because it was commented as above,  I thought that it's better not to `Signal.merge` when only 1 argument.
Two code paths are techniques to keep arguments from being empty.
Variadic allows the codes such as below.
```swift
NotificationCenter.default.reactive.keyboard()
    .observeValues {...}
```
If better to allow them, I'll fix it.

mdiep:
Thanks @ra1028! ✨

PR #3570, opened 2018-02-15
Handle forwarded selectors without any implementation.

andersio:
#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveCocoa/NSObject+Intercepting.swift
+++ b/ReactiveCocoa/NSObject+Intercepting.swift
@@ -155,8 +155,18 @@ private func enableMessageForwarding(_ realClass: AnyClass, _ selectorCache: Sel
 			}
 		}
 
-		let method = class_getInstanceMethod(perceivedClass, selector)!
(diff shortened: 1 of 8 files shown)


ReactiveCocoa/NSObject+Intercepting.swift
```diff
@@ -155,8 +155,18 @@ private func enableMessageForwarding(_ realClass: AnyClass, _ selectorCache: Sel
 			}
 		}
 
-		let method = class_getInstanceMethod(perceivedClass, selector)!
```
andersio: We currently assume that, if the intercepted class has declared its support of a selector via `-respondsToSelector:`, there _must_ be an implementation registered with the Objective-C runtime.

However in practice this is not often the case. For example, we have encountered a crash when intercepting `JSQMessageCollectionCellView` from the JSQMessageViewController framework, whose message forwarding based handling of menu actions clashes with the said assumption.

mdiep:
Nice. 👏

PR #3572, opened 2018-02-25
 Make CocoaTarget to public 

ra1028:
### Motivation
I need this when introduce extended operators in my app.
e.g
```swift
public extension Reactive where Base: CADisplayLink {
    static func link(to runloop: RunLoop, forMode mode: RunLoopMode, framesPerSecond: Int = 60) -> SignalProducer<CADisplayLink, NoError> {
        return .init { observer, lifetime in
            let target = CocoaTarget(observer)
            var displayLink: CADisplayLink? = .init(target: target, selector: target.selector)
            displayLink?.preferredFramesPerSecond = framesPerSecond
            displayLink?.add(to: runloop, forMode: mode)
            
            lifetime.observeEnded {
                displayLink?.invalidate()
                displayLink = nil
            }
        }
    }
}
```

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # master
 *Please put new entries at the top.
 
-1. Add `showsCancelButton`, `textDidBeginEditing` and `textDidEndEditing` extensions to `UISearchBar` (#3566)
+1. Add `showsCancelButton`, `textDidBeginEditing` and `textDidEndEditing` extensions to `UISearchBar` (#3565)
 1. `NotificationCenter.reactive.keyboard(_:)` for system keyboard notification by the event types. (#3566, kudos to @ra1028)
 1. `CocoaTarget` now public (#3572, kudos to @ra1028)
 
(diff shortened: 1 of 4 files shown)


ra1028:
I amended wrong PR-number(#3565) of CHANGELOG as well.

mdiep:
I think our preference would be to include APIs that wrap these things if possible.

ra1028:
@mdiep 
What kind of interface are you intending?

mdiep:
If you're wanting this because you want a reactive `CADisplayLink` API, then let's add one.

ra1028:
@mdiep 
I don't mean what I want reactive CADisplayLink API only.
It's just one example.

To extend all APIs with ReactiveCocoa is not practical.
I hope that it can be extended in each user's codebase as needed.

mdiep:
> To extend all APIs with ReactiveCocoa is not practical.
> I hope that it can be extended in each user's codebase as needed.

Our preference is that users would contribute those APIs back to ReactiveCocoa so that everyone doesn't need to implement everything themselves.

ra1028:
hmm, I see.

PR #3573, opened 2018-03-11
RFC: Use Sourcery to generate BindingTargets for UIKit

iv-mexx:
#### Checklist
- [ ] Updated CHANGELOG.md.

_This is not yet completed but I wanted to put this up for discussion before continuing._

The code for providing UI Bindings (I'm talking mainly about `BindingTarget` right now) is always the same and has to be copy&pasted for each new attributed to be added. 

Also, currently they are not complete. Every now and the I find myself in need of a less common used attribute thats still missing. E.g. right now I was about to add `borderStyle` to `UITextfield` and `accessoryType` to `UITableViewCell`.

So I played around a bit with [Sourcery](https://github.com/krzysztofzablocki/Sourcery) to see how it would work to generate the Boilerplate for these Bindings automatically.

*Note:* I did not yet integrate Sourcery in the Build Process or anything, I've just installed it via homebrew and ran it in the root directory of Reactive Cocoa. Currently there's a .sourcery.yml Configuration file with very basic configuration.

## Code Generation

I've only looked at `BindingTarget` right now, and I've ignored some special cases. Adding special cases manually is not a problem, just do the same as before.

The Template is setup to look for Protocols named `ReactiveUI<Foo>` and for all variables declared in the Protocol.

It will then generate an Extension: 

```swift
extension Reactive where Base: <Foo> {
  var <VariableName>: <VariableType> { 
    makeBindingTarget { $0.<VariableName> = $1 }
  }
}
```

This is an actual example how the Protocol has to be defined:

```swift
protocol ReactiveUINavigationItem {
  var title: BindingTarget<String?> { get }
}
```

## Generate other Bindings

I have not yet looked into the other kinds of bindings and if/how they can be generated as well

## Still to figure out

The current Source Files are linked to varying targets (macOS, iOS, watchOS, tvOS). I think it will thus be necessary to run Sourcery once for each platform and always include only the Source Files that apply to that platform.

Right now I've ignored this and only converted some UIKit Extensions without regard to the platform

### DIFF
--- a/.sourcery.yml
+++ b/.sourcery.yml
@@ -1,0 +1,6 @@
+sources:
+  - ReactiveCocoa/UIKit
+templates:
+  - SourceryTemplates/
+output:
+  ReactiveCocoa/UIKit
--- a/ReactiveCocoa.xcodeproj/project.pbxproj
+++ b/ReactiveCocoa.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		004FD0071E26CDB300A03A82 /* NSButtonSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 004FD0061E26CDB300A03A82 /* NSButtonSpec.swift */; };
 		006518761E26865800C3139A /* NSButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = 006518751E26865800C3139A /* NSButton.swift */; };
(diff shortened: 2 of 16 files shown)


mdiep:
My preference would be to use something like Sourcery to generate these bindings once and then maintain them. (Or maybe keep a script around that can generate them, but don't make it part of the normal build process.)

I dislike generated source code on the whole. It adds a layer of indirection that I found cumbersome.

kaybutter:
Do we really need the explicit Binding-Targets? The KeyPath-Subscript-API has been working just fine for me, or am I missing some limitation?

I'm referring to this syntax:
``` swift
navigationItem.reactive[\.title] <~ someTitleProducer
```

PR #3583, opened 2018-04-07
Mitigate a regression in type inference.

andersio:
https://bugs.swift.org/browse/SR-7346


mdiep:
This is fine, but we should maybe wait to hear back after your last comment in the ticket? Or clarify whether we should work around it since it appears to be a legitimate swift bug.

andersio:
It breaks builds with Xcode 9.3 which should probably be patched ASAP.

mdiep:
I'm good to ship then.

mdiep:
I'm good to ship then.

PR #3589, opened 2018-04-25
UIApplication Reactive Extensions

cocoahero:
#### Checklist
- [x] Updated CHANGELOG.md.


cocoahero:
If anyone is able to help write a spec for this, it would be appreciated. The current spec that is written crashes, most likely because `UIApplication.shared` isn't available.

andersio:
@mdiep Mocking `UIApplication` seems impossible through normal means. Should we drop the unit test, or introduce some sort of `UIApplicationProtocol` to replace the direct extension constraint to `UIApplication`?

mdiep:
I think you can use `UIApplication.shared` if you add a test host app.

But I'd be okay dropping the unit test in this case.

cocoahero:
@andersio @mdiep I have removed the unit test case. In terms of resolving the project conflict, would you prefer a rebase or a merge?

mdiep:
Either is fine 🙂

cocoahero:
This should be ready to go now. 🙂

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
 2. Added `title` binding target to `UIViewController` (#3588, kudos to @cocoahero).
 3. Added several trigger signals for view lifecycle events to `UIViewController` (#3588, kudos to @cocoahero).
 
+# 7.X.0
+1. Added `applicationIconBadgeNumber` binding target to `UIApplication` (#3589, kudos to @cocoahero).
```
mdiep: This needs to go at the top under `master`. 🙈

cocoahero: 😅 Fixed!

mdiep:
Awesome, thanks!

PR #3590, opened 2018-04-27
support Cocoapods 1.5.0 static framework

mishagray:
#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # master
 *Please put new entries at the top.
 
+1. Add support for Cocoapods 1.5.0 static frameworks (#3590, kudos to @mishagray)
+
 1. Add `becomeFirstResponder` and `resignFirstResponder` extensions to `UIResponder`. (#3585, kudos to @Marcocanc)
 
 # 7.2.0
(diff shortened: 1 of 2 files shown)


mdiep:
Seems fine to me. @ikesyo or @andersio?

mdiep:
Let's go for it. 😄

andersio:
`ObjCRuntimeAliases.h` is meant to be private header. This change makes stuff in this header become part of the public API of ReactiveCocoa built with CocoaPods.

mishagray:
@andersio yeah... the public/private issue is more about using getting cocoa pods to generate a functional modulemap for both static and dynamic frameworks.   If somebody smarter than me can figure out how to fix the Cocoapods.podspec to get cocoa pods to generate the write modulemap for both dynamic static frameworks with the header being 'private' that would be better.   

Alternatively... is for somebody to create a custom modulemap that Cocoapods 'likes' for both static/dynamic frameworks.  Either strategy may work.   Is there a Cocoapods podspec expert out there that can figure this out?

But with a lot of developers (myself included) trying to move from dynamic frameworks to static frameworks ... (smaller code! faster load time!) this feels like an ok workaround for now.   This was the only pod I found out of dozens that didn't port very smoothly.

PR #3591, opened 2018-05-02
Make UITextField and UITextView text and attributedText signals non-optional.

Marcocanc:
It looks like (at least on iOS 11) it is not possible for `UITextField` and `UITextView` to return `nil` on their `text` and `attributedText` values, even when setting any of them to nil (a simple Playground will prove this). It seems that Apple has [changed this over time](https://stackoverflow.com/questions/18741335/textfield-is-returning-nil-when-it-is-empty/18741469), but I think it would make sense (at least for `text`) to not be optional.

I also noticed that the `UITextView` signals were already force-unwrapping `text` and `attributedText`.

#### Checklist
- [x] Updated CHANGELOG.md.


mdiep:
This seems fine to me. @andersio?

mdiep:
@Marcocanc I think if you update the CHANGELOG, we can go ahead and merge this.

Marcocanc:
@mdiep Alright, will do. There's one more thing though: `attributedTextValues` and `continuousAttributedTextValues` on UITextField. Do you think it's fine to use `?? NSAttributedText()` there?

andersio:
@mdiep I believe it was modelled as optional originally so that the API matches that of UITextField. I don't see any technical problem for the change though, since Apple sticks with it being IUO for multiple SDK iterations now.

mdiep:
I'm okay to use `??`. 👍

mdiep:
@Marcocanc Do you want to merge master into this? It's not passing CI (and the CHANGELOG needs an update). We'll be releasing a new major version soon, and I'd like to get this in.

mdiep:
This has another conflict. 😓 Sorry!

mdiep:
Thanks for all your work on this!

Issue #3602, opened 2018-06-12
Fails to build for watchOS using Xcode 10 from Carthage


NachoSoto:
**Cartfile**:
```
github "ReactiveCocoa/ReactiveCocoa" "xcode-10"
```

**Command:**
`$ carthage update ReactiveCocoa --platform watchos`

-------

This results in this command:
```
/usr/bin/xcrun xcodebuild -workspace /project/Carthage/Checkouts/ReactiveCocoa/ReactiveCocoa.xcworkspace -scheme ReactiveMapKit -configuration Release -derivedDataPath /Users/nachosoto/Library/Caches/org.carthage.CarthageKit/DerivedData/10.0_10L176w/ReactiveCocoa/36f0383edcc2de9431baaf4f75a1f8be201176cb -sdk watchsimulator -destination platform=watchOS\ Simulator,id=F192C9E9-FAA3-4FBD-B755-8BE25DCA115B -destination-timeout 3 ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES build
```

And that fails with a bunch of errors like this because of missing `UIKit` classes (which shouldn't be compiled on the watchOS framework):
```
/project/Carthage/Checkouts/ReactiveCocoa/ReactiveCocoa/UIKit/ReusableComponents.swift:15:11: error: use of undeclared type 'UITableViewCell'
extension UITableViewCell: Reusable {}
          ^~~~~~~~~~~~~~~
```

Not sure what's going on honestly, because `ReactiveMapKit` builds correctly from Xcode.

mdiep:
We should probably switch to the `canImport` checks when using Swift 4.2. That might clear these up?

ikesyo:
I assume this is due to the fact that `ReactiveMapKit` is a universal-framework target. Since MapKit is not available on watchOS, removing `watchos` and `watchsimulator` from `ReactiveMapKit`'s `SUPPORTED_PLATFORMS` build setting would fix the issue.

ikesyo:
Should be addressed by #3604.

PR #3604, opened 2018-06-12
 MapKit is not available on watchOS

ikesyo:
So building ReactiveMapKit for watchOS does not make sense.

Ref: #3602

#### Checklist
- [ ] Updated CHANGELOG.md.


ReactiveCocoa.xcodeproj/project.pbxproj
```diff
@@ -2020,13 +2024,18 @@
...
+				CLANG_WARN_RANGE_LOOP_ANALYSIS = YES;
+				CLANG_WARN_STRICT_PROTOTYPES = YES;
 				CODE_SIGNING_REQUIRED = NO;
 				CURRENT_PROJECT_VERSION = 1;
 				GCC_OPTIMIZATION_LEVEL = 0;
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
 				PRODUCT_BUNDLE_IDENTIFIER = "org.reactivecocoa.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = "$(PROJECT_NAME)";
+				SWIFT_COMPILATION_MODE = wholemodule;
```
mdiep: We probably shouldn’t be turning this on now

ReactiveCocoa.xcodeproj/project.pbxproj
```diff
@@ -1999,6 +1999,10 @@
 			buildSettings = {
 				APPLETVOS_DEPLOYMENT_TARGET = 9.0;
 				BITCODE_GENERATION_MODE = bitcode;
+				CLANG_WARN_BLOCK_CAPTURE_AUTORELEASING = YES;
+				CLANG_WARN_COMMA = YES;
+				CLANG_WARN_RANGE_LOOP_ANALYSIS = YES;
+				CLANG_WARN_STRICT_PROTOTYPES = YES;
```
mdiep: Do we want these settings in the project file?

ikesyo: Actually those are at project level. The four duplicated-like diffs are for each configurations (Debug, Profile, Release, Test).

ReactiveCocoa.xcodeproj/project.pbxproj
ikesyo: I don't remember why optimizations are disabled for Release configuration but I follow that for now.

ReactiveCocoa.xcodeproj/project.pbxproj
ikesyo: Those are automatically added by Xcode 9.4 but should be removed?

mdiep: That's what we've typically done, because we try to have them in the `.xcconfig`s instead. But I don't personally feel strongly about this.

ikesyo: Once https://github.com/jspahrsummers/xcconfigs is updated this could be removed.

ikesyo: Filed an issue: https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3605

PR #3606, opened 2018-06-13
 Fix Xcode 10 compilation issue on optional IMP

hkellaway:
Im running the latest Xcode 10 Beta (10L176w) and encountering the following error message:

```.../Pods/ReactiveCocoa/ReactiveCocoa/NSObject+Intercepting.swift:218:62: Cannot convert value of type 'OpaquePointer?' to specified type 'IMP' (aka 'OpaquePointer')```

Here is the least to do to address it (treat the `IMP` instance as an Optional - but sort of changes logic in that there's no handling for when `IMP` comes out `nil`.

I'm not familiar enough with the internals of ReactiveCocoa to understand if this impacts other areas. 

#### Checklist
- [ ] Updated CHANGELOG.md.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "ReactiveCocoa/ReactiveSwift" "4.0.0-rc.1"
+github "ReactiveCocoa/ReactiveSwift" "xcode-10"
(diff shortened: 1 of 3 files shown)


MartinPrusa:
still works with `if let impl = impl, impl != _rac_objc_msgForward {`

andersio:
This is odd. Using the same build (10L176w) doesn't result in the said error for me. 🤔

PR #3609, opened 2018-07-17
Conform `Observer` to `BindingTargetProvider`

olejnjak:
#### Checklist
- [x] Updated CHANGELOG.md.

I was wondering whether there is a reason that `Signal.Observer` doesn't conform to `BindingTargetProvider`? 

Sometimes I need to chain some signals so this would come in handy. And I think that code 
```swift
observer <~ action.completed
```

is much more readable than

```swift
action.completed.take(duringLifetimeOf: self).observeValues { [weak self] in self?.observer.send(value: ()) }
```

If there's a reason this not possible, of course close the PR 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 *Please put new entries at the top.
 
 # 8.0.0-rc.1
-1. Add support for Cocoapods 1.5.0 static frameworks (#3590, kudos to @mishagray)
-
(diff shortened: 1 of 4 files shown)


ikesyo:
This is a duplicate of https://github.com/ReactiveCocoa/ReactiveSwift/pull/635 (should be done in ReactiveSwift).

PR #3622, opened 2018-10-04
Add swift version to podspecs

olejnjak:
Now with the latest Xcode release it is unable to integrate ReactiveCocoa without further changes like adding post install hook into Podfile. For libraries that depend on ReactiveCocoa it is even more complicated.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please put new entries at the top.
 
+1. Add `swift_version` to podspecs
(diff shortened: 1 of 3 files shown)


mdiep:
Thanks for the PR!

I'm worried this might be overly specific. @andersio @ikesyo Are you knowledgable about this?

andersio:
@mdiep This LGTM. Apparently CocoaPods does not respect the `.swift-version` anymore, and we need to tell CocoaPods what the language mode the pod is meant to be built with.

PR #3623, opened 2018-10-05
Change 'impl' to optional

khanirteza:
I have changed the `impl` variable to optional. Because, it failed to compile in Xcode 10 due to the following error,
`Cannot convert value of type 'OpaquePointer?' to specified type 'IMP' (aka 'OpaquePointer')`

### DIFF
--- a/ReactiveCocoa/NSObject+Intercepting.swift
+++ b/ReactiveCocoa/NSObject+Intercepting.swift
@@ -215,7 +215,7 @@ private func enableMessageForwarding(_ realClass: AnyClass, _ selectorCache: Sel
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


andersio:
Hmm. Do you happen to use CocoaPods? I assume #3622 should help resolve this then.

andersio:
I've tested building ReactiveCocoa 8.0.1 via the Xcode project and via CocoaPods (with host projects of 4.1 mode and 4.2 mode) using Xcode 10 GM, and this issues can no longer be reproduced. I would close this for now, but please do get in touch if ReactiveCocoa 8.0.1 does not help resolve the issue.

PR #3646, opened 2019-03-15
Update ReactiveSwift to 5.0

Marcocanc:
#### Checklist
- [x] Updated CHANGELOG.md.

PR #3647, opened 2019-03-15
Fix warnings for deprecated methods, and pick up latest Quick & Nimble dependencies

liscio:
These warnings have been driving me nuts for far too long. Since Swift 3.x support was dropped in Xcode, it's safe to say that the new methods are going to be there.

#### Checklist
- [x] Updated CHANGELOG.md.

PR #3648, opened 2019-03-15
Update podspec to ReactiveSwift 5.0

Marcocanc:
Forgot to update the podspec in #3646

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode10.1
+osx_image: xcode10.2
 before_install: true
 install: true
 branches:
@@ -14,7 +14,7 @@ cache:
 jobs:
   include:
     - stage: unit tests
-      osx_image: xcode10.1
+      osx_image: xcode10.2
       script:
         - XCODE_SCHEME=ReactiveCocoa-macOS
           XCODE_SDK=macosx
--- a/ReactiveCocoa.podspec
+++ b/ReactiveCocoa.podspec
@@ -24,7 +24,7 @@ Pod::Spec.new do |s|
   s.watchos.source_files = "ReactiveCocoa/WatchKit/*.{swift}"
   s.module_name = 'ReactiveCocoa'
 
-  s.dependency 'ReactiveSwift', '~> 4.0'
+  s.dependency 'ReactiveSwift', '~> 5.0'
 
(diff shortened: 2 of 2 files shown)


ReactiveCocoa.podspec
```diff
@@ -24,7 +24,7 @@ Pod::Spec.new do |s|
   s.watchos.source_files = "ReactiveCocoa/WatchKit/*.{swift}"
   s.module_name = 'ReactiveCocoa'
 
-  s.dependency 'ReactiveSwift', '~> 4.0'
+  s.dependency 'ReactiveSwift', '~> 5.0'
 
   s.pod_target_xcconfig = { "OTHER_SWIFT_FLAGS[config=Release]" => "$(inherited) -suppress-warnings" }
   s.swift_version = '4.1.2'
```
ikesyo: This should be `4.2`.

ReactiveCocoa.podspec
```diff
@@ -24,7 +24,7 @@ Pod::Spec.new do |s|
   s.watchos.source_files = "ReactiveCocoa/WatchKit/*.{swift}"
   s.module_name = 'ReactiveCocoa'
 
-  s.dependency 'ReactiveSwift', '~> 4.0'
+  s.dependency 'ReactiveSwift', '~> 5.0'
```
ikesyo: FYI ReactiveSwift 5.0 is not published to CocoaPods yet. Please wait for a while.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode10.1
+osx_image: xcode10.2
```
ikesyo: We need https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3647 before using Xcode 10.2 image.

ReactiveCocoa.podspec
ikesyo: That has been published.

.travis.yml
mdiep: It has been merged now. ✨

ReactiveCocoa.podspec
Marcocanc: Should we merge one of the Swift 4.2 branches first then?

ikesyo: Ah okay this is not required, I'm sorry for confusing you!

(Both Swift 4.2 compiler and Swift 5 compiler support Swift 4 mode, so this is still valid)

PR #3656, opened 2019-03-27
9.0.0

mdiep:
I don't think this needs an RC; I think we can go right to a release. cc @ReactiveCocoa/reactivecocoa

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # master
 *Please put new entries at the top.
+
+# 9.0.0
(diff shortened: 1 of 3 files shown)


NachoSoto:
It's only appropriate then that the Swift 5.0 release is RAS 10.0 😗

mdiep:
I don't think there's any way to make the podspec lint in this PR since ReactiveMapKit needs the ReactiveCocoa release to validate.

ikesyo:
https://github.com/CocoaPods/CocoaPods/releases/tag/1.7.0.beta.3

> Adds support for referring to other podspecs during validation

Looks like this is what we want now.

Issue #3660, opened 2019-04-11
Crashed on Smart Key Path KVO when values are enums


wordlessj:
Current implementations of `producer(for:)` and `signal(for:)` use old Objective-C API, which get values from `value(forKey:)`. It works for most types except enums, `value(forKey:)` returns an `NSNumber`, then force-casted to the original enum type, which crashes. Maybe we should use Swift KVO API instead?

ikesyo:
A good news is that is exactly https://github.com/ReactiveCocoa/ReactiveCocoa/pull/3491.

wordlessj:
I may not have made It clear, this PR is the origin of the crash. It gets values from `value(forKey:)` which returns `NSNumber` for enums, then `producer(for:)` and `signal(for:)` force cast `NSNumber` to the specific enum type which will crash. If we use the Swift KVO api, values passed to the closure will be the correct type.

ikesyo:
#3667 should had fixed this.

PR #3667, opened 2019-04-20
Allow for observable enum values

gfontenot:
Previously, when using the `NSObject.producer/signal(for:)` syntax with
Swift 4 KeyPaths, we were force-casting the resulting value to the expected
type. This works for native Objective-C types, but breaks for Objective-C
enums, which are sent as their raw Integer types. This means the force-cast
would result in a crash if you tried to observe an enum value.

To fix this, we can add a special override for the KeyPath observation API 
that handles the RawRepresentable conversion automatically. This will then
be used if the user specifies a KeyPath that points to a RawRepresentable
type.

I ran into this need when trying to observe `\AVPlayer.timeControlStatus`,
which is an enum value. The runtime crash was unexpected and hard to
understand at first. This issue is solvable in the consuming codebase by
mapping the value over the enum initializer manually, but it would be _really_
nice to not have this edge case in the first place.

The added tests will result in a crash without the new code. I can add more
tests around the specific optional behavior if needed.

### DIFF
--- a/ReactiveCocoa/NSObject+KeyValueObserving.swift
+++ b/ReactiveCocoa/NSObject+KeyValueObserving.swift
@@ -106,6 +106,22 @@ extension Reactive where Base: NSObject {
 		return producer(for: keyPath) { $0 as! U? }
 	}
 
+	/// Create a producer which sends the current value and all the subsequent
+	/// changes of the property specified by the key path.
+	///
+	/// The producer completes when the object deinitializes.
+	///
+	/// - parameters:
+	///   - keyPath: The key path of the property to be observed.
+	///
+	/// - returns: A producer emitting values of the property specified by the
+	///            key path.
+	public func producer<U>(for keyPath: KeyPath<Base, U?>) -> SignalProducer<U?, Never> where U: RawRepresentable {
+		return producer(for: keyPath) {
+			$0.flatMap { value in U(rawValue: value as! U.RawValue)! }
+		}
+	}
+
 	/// Create a signal all changes of the property specified by the key path.
 	///
 	/// The signal completes when the object deinitializes.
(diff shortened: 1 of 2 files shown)


ReactiveCocoa/NSObject+KeyValueObserving.swift
```diff
@@ -106,6 +106,22 @@ extension Reactive where Base: NSObject {
...
+	/// The producer completes when the object deinitializes.
+	///
+	/// - parameters:
+	///   - keyPath: The key path of the property to be observed.
+	///
+	/// - returns: A producer emitting values of the property specified by the
+	///            key path.
+	public func producer<U>(for keyPath: KeyPath<Base, U?>) -> SignalProducer<U?, Never> where U: RawRepresentable {
+		return producer(for: keyPath) {
+			$0.flatMap { value in U(rawValue: value as! U.RawValue) }
```
gfontenot: I'm not sure if this should force-unwrap the enum value or not. Strictly speaking, if we have a value, we should be able to create a valid enum instance from it. But there's also no "harm" in returning the optional here since we have to return an optional anyway.

gfontenot:
I'm pretty sure this is going to fail CI (for a reason beyond the test failure that seems to be on master inside `ReactiveMapKit`), but I'm not sure what the actual failure is. Any guidance here would be appreciated.

ikesyo:
Looks like this fixes #3660.

ReactiveCocoa/NSObject+KeyValueObserving.swift
andersio: This should trap in my opinion. If a type is imported as a Swift enum equivalent, it should be exactly convertible in both directions. Most importantly, given that the property is optional, `nil` has its own contextual semantic, and we should not silently corrupt that semantic with the scenario of "failed conversion that is supposed to succeed".

andersio:
LGTM otherwise.

PR #3670, opened 2019-05-10
WatchKit additional reactive bindings

tdimeco:
Hey,

I have added some new WatchKit reactive bindings for standard interface objects. I leave some questions/remarks inside the code below.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/ReactiveCocoa/WatchKit/WKInterfaceActivityRing.swift
+++ b/ReactiveCocoa/WatchKit/WKInterfaceActivityRing.swift
@@ -1,0 +1,13 @@
+import ReactiveSwift
+import WatchKit
+import HealthKit
+
+@available(watchOSApplicationExtension 2.2, *)
+extension Reactive where Base: WKInterfaceActivityRing {
+	/// Sets the summary of the activity ring.
+	///
+	/// - Parameter animated: Whether updates are animated.
+	public func activitySummary(animated: Bool) -> BindingTarget<HKActivitySummary?> {
+		return makeBindingTarget { $0.setActivitySummary($1, animated: animated) }
+	}
+}
--- a/ReactiveCocoa/WatchKit/WKInterfaceButton.swift
+++ b/ReactiveCocoa/WatchKit/WKInterfaceButton.swift
@@ -3,7 +3,7 @@ import WatchKit
 
 extension Reactive where Base: WKInterfaceButton {
 	/// Sets the title of the button.
-	public var text: BindingTarget<String?> {
+	public var title: BindingTarget<String?> {
 		return makeBindingTarget { $0.setTitle($1) }
 	}
 	
--- a/ReactiveCocoa/WatchKit/WKInterfaceInlineMovie.swift
+++ b/ReactiveCocoa/WatchKit/WKInterfaceInlineMovie.swift
@@ -1,0 +1,30 @@
+import ReactiveSwift
+import WatchKit
+
+@available(watchOSApplicationExtension 3.0, *)
+extension Reactive where Base: WKInterfaceInlineMovie {
+	/// Sets the url of the movie.
+	public var movieURL: BindingTarget<URL> {
+		return makeBindingTarget { $0.setMovieURL($1) }
+	}
+	
+	/// Sets the video gravity of the movie.
+	public var videoGravity: BindingTarget<WKVideoGravity> {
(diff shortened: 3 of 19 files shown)


ReactiveCocoa/WatchKit/WKInterfaceButton.swift
```diff
@@ -3,7 +3,7 @@ import WatchKit
 
 extension Reactive where Base: WKInterfaceButton {
 	/// Sets the title of the button.
-	public var text: BindingTarget<String?> {
+	public var title: BindingTarget<String?> {
```
tdimeco: I have deprecated the `text` attribute in favour of `title` (to match the WatchKit name), is that ok?

ReactiveCocoa/WatchKit/WKInterfaceLabel.swift
```diff
@@ -13,7 +13,7 @@ extension Reactive where Base: WKInterfaceLabel {
 	}
 	
 	/// Sets the color of the text of the label.
-	public var textColor: BindingTarget<UIColor> {
+	public var textColor: BindingTarget<UIColor?> {
```
tdimeco: I have changed the type to optional, to match the WatchKit original type. Is that ok to do it that way, without some kind of deprecation?

ReactiveCocoa/WatchKit/WKInterfaceActivityRing.swift
```diff
@@ -0,0 +1,13 @@
+import ReactiveSwift
+import WatchKit
+import HealthKit
+
+@available(watchOSApplicationExtension 2.2, *)
+extension Reactive where Base: WKInterfaceActivityRing {
+	/// Sets the summary of the activity ring.
+	///
+	/// - Parameter animated: Whether updates are animated.
+	public func activitySummary(animated: Bool) -> BindingTarget<HKActivitySummary?> {
```
tdimeco: Is it ok to use a computed property instead of a function, in order to allow users to customise the animation behaviour of the binding?

ReactiveCocoa/WatchKit/WKInterfaceInlineMovie.swift
```diff
@@ -0,0 +1,30 @@
...
+		return makeBindingTarget { $0.setVideoGravity($1) }
+	}
+	
+	/// Sets the poster image of the movie.
+	public var posterImage: BindingTarget<WKImage?> {
+		return makeBindingTarget { $0.setPosterImage($1) }
+	}
+	
+	/// Whether the movie loops.
+	public var loops: BindingTarget<Bool> {
```
tdimeco: Do you have any suggestion for the name of this property? `isLoops` seems weird, so I use the original name.

ReactiveCocoa/WatchKit/WKInterfaceInlineMovie.swift
```diff
@@ -0,0 +1,30 @@
...
+		return makeBindingTarget { $0.setLoops($1) }
+	}
+	
+	/// Whether the movie autoplays.
+	public var autoplays: BindingTarget<Bool> {
```
tdimeco: Do you have any suggestion for the name of this property? `isAutoplays` seems weird, so I use the original name.

ReactiveCocoa/WatchKit/WKInterfaceActivityRing.swift
mdiep: Yeah, that makes sense. 👍 

We have a few existing `func`s that return `BindingTarget`s.

ReactiveCocoa/WatchKit/WKInterfaceButton.swift
mdiep: Great 👍

ReactiveCocoa/WatchKit/WKInterfaceLabel.swift
mdiep: Not really, but there's also not really a way to deprecate it. I think we'll just have to go with it.

ReactiveCocoa/WatchKit/WKInterfaceInlineMovie.swift
mdiep: I think `loops` makes sense. 👍

mdiep:
This is great, thanks! 👏🏻

PR #3675, opened 2019-06-15
Add a binding target for the barTintColor of UINavigationBar

rehatkathuria:
Exposes binding sugar for the `barTintColor` on `UINavigationBar`.


mdiep:
Looks great! Thanks!

PR #3677, opened 2019-08-02
Update dependencies for Xcode 11

olejnjak:
Hi,

not sure why, but I am unable to compile ReactiveCocoa when used as a dependency. Updating dependencies seems to fix the issue.

I assume that when Carthage checkouts ReactiveCocoa it checkouts ReactiveSwift 6.0.0 as stated in Cartfile.resolved. Then if my Cartfile requires ReactiveSwift 6.1.0 it is then overwritten (only a guess as I am unable to compile dependencies and calling `carthage update` in _Carthage/Checkouts/ReactiveCococa_ and then calling `carthage build` in my project succeeds.

#### Checklist
- [x] Updated CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # master
 *Please put new entries at the top.
 1. Add reactive extensions for standard WatchKit interface objects.
+1. Update dependencies so ReactiveCocoa can be used with Xcode 11 (#3677, kudos to @olejnjak)
 
 # 10.0.0
 1. Update ReactiveSwift to 6.0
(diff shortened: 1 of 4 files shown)


olejnjak:
Not sure why the single test is failing, every time I do clean checkout it works just fine 😔🤷‍♂️

mluisbrown:
When I make the same changes in this PR locally all tests pass (using Xcode 10.2, iPhone 6s simulator as used by Travis). So really not sure what the problem is here. Would be great to get this merged so that we can start building our apps with Xcode 11 beta.

olejnjak:
I am not sure if this is not Carthage issue because if I require _ReactiveSwift ~> 6.1_ in my Cartfile it still fails to compile.

I submitted new [issue on Carthage](https://github.com/Carthage/Carthage/issues/2844)

mdiep:
The tests pass locally for me too. 🤷‍♂

PR #3693, opened 2019-11-21
Add support for Swift Package Manager

simba909:
This PR adds support for Swift Package Manager to the project. I should point out that this is based off of the great stuff already done by @fabio-cerdeiral-ck in #3692, and by @sharplet in #3676. Big thanks to those guys for putting in the effort 👏 

Please let me know if there's anything that needs fixing!

#### Checklist
- [x] Updated CHANGELOG.md.

Issue #3690, opened 2019-10-06
Crash during runloop's autorelease pool drain on macOS 10.15 Catalina


iby:
Experiencing a naughty crash with buttons and all inheriting classes inside NSStackView with enabled `Detaches Hidden Views` when the window gets closed and deallocated in the latest Catalina build:

```
2019-10-06 15:04:27.548688+0100 App[4474:58829] [general] Caught exception during runloop's autorelease pool drain of client objects NSRangeException: Cannot remove an observer <NSStackView 0x10461aca0> for the key path "hidden" from <NSButton 0x10461c7f0> because it is not registered as an observer. userInfo: (null)
2019-10-06 15:04:27.550252+0100 App[4474:58829] [General] An uncaught exception was raised
2019-10-06 15:04:27.550313+0100 App[4474:58829] [General] Cannot remove an observer <NSStackView 0x10461aca0> for the key path "hidden" from <NSButton 0x10461c7f0> because it is not registered as an observer.
	0   CoreFoundation                      0x00007fff36166d63 __exceptionPreprocess + 250
	1   libobjc.A.dylib                     0x00007fff6bf56bd4 objc_exception_throw + 48
	2   Foundation                          0x00007fff387a7395 -[NSObject(NSKeyValueObserverRegistration) _removeObserver:forProperty:] + 578
	3   Foundation                          0x00007fff387a7106 -[NSObject(NSKeyValueObserverRegistration) removeObserver:forKeyPath:] + 74
	4   Foundation                          0x00007fff387bf963 -[NSObject(NSKeyValueObserverRegistration) removeObserver:forKeyPath:context:] + 190
	5   AppKit                              0x00007fff334e0cc9 -[NSStackView dealloc] + 233
	6   CoreFoundation                      0x00007fff360c5176 -[__NSArrayI dealloc] + 73
	7   libobjc.A.dylib                     0x00007fff6bf6653a _ZN19AutoreleasePoolPage12releaseUntilEPP11objc_object + 134
	8   libobjc.A.dylib                     0x00007fff6bf4cc70 objc_autoreleasePoolPop + 175
	9   CoreFoundation                      0x00007fff3608767e _CFAutoreleasePoolPop + 22
	10  CoreFoundation                      0x00007fff360b59fc __CFRunLoopRun + 2404
	11  CoreFoundation                      0x00007fff360b4e13 CFRunLoopRunSpecific + 499
[...]

PR #3694, opened 2019-12-01
Work around an issue where NSButtonSpec test fails

simba909:
This works around the failing test that is blocking #3693. 

I've done some research and the failing test seems to be the same issue reported in #3690. Setting `stackView.detachesHiddenViews = false` "solves" the issue, but it's probably not the proper solution. Investigating this further however requires far higher knowledge about Obj-C class/method swizzling (and how RAC does this) than what I currently possess.

Thought I would set up a PR anyway to work around the issue for now, so other PR's could get merged before a more proper fix is in place. Please feel free to close this PR if a proper fix is required.

#### Checklist
- [ ] Updated CHANGELOG.md.


andersio:
If I read the test results correctly, CI is failing in #3693 due to the recurring issue of MKMapView being somehow internally retained.

simba909:
@andersio Ah, my bad. Closing this for now then.

PR #3698, opened 2020-01-08
Add support for macOS Catalyst

### LINKED ISSUE #3697: Unable to compile targeting macOS Catalyst using SwiftPM (fix exists)
I'm unable to compile the the SwiftPM package when targeting macOS Catalyst, because NSSlider and other types aren't available yet. 
I have fixed this locally in [a fork](https://github.com/nkristek/ReactiveCocoa/tree/hotfix/catalyst) in [this commit](https://github.com/nkristek/ReactiveCocoa/commit/8c34e20ab3d8708ee2a7f2be18bf3f490126740d), but those additional `#if !targetEnvironment(macCatalyst)` checks have to be removed once the types become available for that target.
Should I make a PR with this fix?

nkristek:
This PR adds support for macOS Catalyst by adding `#if !targetEnvironment(macCatalyst)` checks for types which aren't available in macOS Catalyst. This is needed, since when targeting macOS Catalyst you can import both UIKit and AppKit, but you can't use classes like `NSSlider`. 

**Please note:** These checks probably need to be removed once the types eventually become available. For now, it is required to use this library at all.

#### Checklist
- [x] Updated CHANGELOG.md.

#### Additional notes

- closes #3697

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # master
-*Please put new entries at the top.
+1. Don't include code which uses unavailable classes (like `NSSlider`) when targeting macOS Catalyst (#3698, kudos to @nkristek)
 
 # 10.2.0
 1. Update ReactiveSwift to 6.2.
--- a/ReactiveCocoa/AppKit/ActionProxy.swift
+++ b/ReactiveCocoa/AppKit/ActionProxy.swift
@@ -1,4 +1,4 @@
-#if canImport(AppKit)
+#if canImport(AppKit) && !targetEnvironment(macCatalyst)
 import AppKit
 import ReactiveSwift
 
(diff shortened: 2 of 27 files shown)


andersio:
Why should we prefer this over `@available(macCatalyst, unavailable)`?

nkristek:
Thanks for your feedback! 
I sticked with adding a condition to the already existing `#if` checks, since in some cases, multiple `@available(macCatalyst, unavailable)` will be needed per file while only one `#if` would have been needed to modify, all while not increasing the line count.
The final deciding factor was, that for example in `ReactiveCocoa/Shared/NSLayoutConstraint` it is actually needed to modify the `#if` statement as I have done in this PR (otherwise you get the error: `Use of undeclared type 'NSLayoutConstraint'`). To switch between both ways of marking the code as unavailable would introduce additional complexity I wanted to avoid (since you probably also need to remember to remove these checks once the classes become available in macCatalyst, depending on what Apple plans going forward).
If you still prefer using `@available(macCatalyst, unavailable)` when applicable, I'm happy to alter this PR accordingly as I just want to be able to use this library in a macCatalyst application.

andersio:
That's fair. 👍

PR #3703, opened 2020-05-01
Fix watchOS build error

JaviSoto:
`WKInterfaceButton.title` is gone.


ReactiveCocoa/Deprecations+Removals.swift
```diff
@@ -18,7 +18,7 @@ import WatchKit
 extension Reactive where Base: WKInterfaceButton {
 	@available(*, deprecated, renamed: "title")
```
andersio, line 19:
```suggestion
	@available(*, unavailable, renamed: "title")
```

JaviSoto: 👍

PR #3710, opened 2020-06-17
Bump SPM swift-tools-version to 5.2

mluisbrown:
For the same reason as this was done in ReactiveSwift, which is so that test dependencies aren't exposed to clients of this package.


andersio:
Could you update `.swift-version` too?

mluisbrown:
@andersio  Done 👍

mluisbrown:
This test is failing on macCatalyst:

```
it("should accept changes from bindings to its zoom enabled state")
```

No idea why 🤷

andersio:
MapKit tests sometimes fail because MapKit intermittently retains the views a little longer.

